PR #75, opened 2023-01-31
issue-40 - game support cyberpunk 2077

halgari:
Implements support for Cyberpunk 2077

- Can detect Cyberpunk from EGS, Steam, and GoG
- Can start Cyberpunk and deploy mods via redmod
- Tested with several of the most popular mods
- Implemented the `stress-test` verb which downloads the mods uploaded in the past day and tries to install them

As of right now modding for Cyberpunk is dead thanks to an update today and the fact that a key mod hasn't yet been updated. I want to merge this code in so we can continue to work on other games, but I'll loop back around to this game before closing out the ticket.


halgari:
Standard tests pass, I need to rework the configured env tests

PR #77, opened 2023-02-02
fixed ReplaceExtension function cropping filenames

TanninOne:


halgari:
Thanks for this, I've seen this bug crop up in some places and hadn't sat down to fix it yet. 

Test configuration is broken here for some reason, I'll debug it and merge this when it's fixed.

halgari:
fixes #78

PR #90, opened 2023-02-06
Cleanup: Benchmark Projects (Fixes #88)

### LINKED ISSUE #88: Consider Refactoring Benchmark Code
# Improvement

## Currently 

As is; every time you want to run a new benchmark you have to:
- Comment all Other Lines  
- Uncomment the Benchmark you want to Run  

In some cases; we make temporary changes to the benchmark project to profile our code and this can creep into the individual git logs/commits; for example: [#86cd831](https://github.com/Nexus-Mods/NexusMods.App/commit/86cd831599684cd404191a98e950d70714f5acec#diff-9b05a9c709a222c0f685f35e5e7fe19a6b4a11e81fe59e1b44c4e118fa5126a3) removes such case.

## We want
A way to run benchmarks without having to modify the source code between successive runs.

## Design

A simple design that uses reflection; such as the one below to automatically autofill selectable benchmarks to a menu:  

```csharp
/// <summary>
/// Common base objects for benchmarks.
/// </summary>
public interface IBenchmark { }
```

```csharp
var benchmarks = Assembly.GetExecutingAssembly()
    .GetTypes()
    .Where(x => x.IsAssignableTo(typeof(IBenchmark)) && x != typeof(IBenchmark))
    .ToArray();
```
```csharp
public class SomeBenchmark : IBenchmark
```

Sewer56:
Fixes #88


Sewer56:
image

Ships with simple menu; benchmarks are automatically populated using reflection.

PR #108, opened 2023-02-10
20 flexible downloader

TanninOne:


Sewer56:
General feedback unrelated to any code:

- We use [Nullable Reference Types](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references) i.e. Compile Time null checking in our projects. 

The base class libraries are annotated with nullables; as such; Roslyn (the C# compiler) can inform you when a null check is necessary or not e.g. it's redundant here
image

e.g. While missing here
image

[Note: This gets called from `DownloaderTask` for first chunk; without any exception handling for that call; thus may potentially throw `NullReferenceException`. If there is no content; consider at least throwing a custom exception.]

I'll leave code specific stuff directly as comments.

src/Networking/NexusMods.Networking.HttpDownloader/AdvancedHttpDownloader.cs
```diff
@@ -0,0 +1,734 @@
...
+
+        private bool IsDownloadSlow(ChunkState chunk, double referenceSpeed)
+        {
+            var now = DateTime.Now;
+            if ((now - chunk.Started).TotalMilliseconds < MinCancelAge)
+            {
+                // don't cancel downloads that were only just started
+                return false;
+            }
+            if ((chunk.Read / chunk.Size) > 0.8)
```
Sewer56, line 288: Here `chunk.Read` and `chunk.Size` are both integer types (longs); thus cannot produce fractional output after division.  
This will always return either 1 [post completion] or 0.

You'll need to cast to a float here.

halgari:
We may loop around in the future and improve this, but this is certainly a good starting point and I want to hook it in as the default downloader, so I'm merging this.

Issue #110, opened 2023-02-13
Advanced downloader breaks with files smaller than the chunk size


halgari:
# Bug Report

Making the new AdvancedHttpDownloader the default in the app results in error such as this, and all downloads fail. It looks like the initial request to get the chunks returns the entirety of the file and so no future requests need to be made, but the app then assumes this is a partial download and starts requesting after the end of the content. 

Mod being downloaded: https://www.nexusmods.com/cyberpunk2077/mods/5802?tab=files

And the HTTP log:
[14_Full.txt](https://github.com/Nexus-Mods/NexusMods.App/files/10722776/14_Full.txt)

PR #118, opened 2023-02-14
20 flexible downloader

TanninOne:
addresses #110


src/Networking/NexusMods.Networking.HttpDownloader/AdvancedHttpDownloader.cs
```diff
@@ -414,6 +411,7 @@ private async Task ReadStreamToEnd(Stream stream, ChunkState chunk, ChannelWrite
             {
                 byte[] buffer = _bufferPool.Rent(ReadBlockSize);
                 lastRead = await stream.ReadAsync(buffer, 0, ReadBlockSize, cancel);
+                job.ReportNoWait(Size.From(lastRead));
```
halgari, line 414: I think this needs to be the wait version so we respect bandwidth limiting, but I want to do some testing with that and make sure it doesn't mess up the rest of the code.

PR #130, opened 2023-02-20
Game: Bannerlord Support

Aragas:
Implemented:
- [x] Steam
- [x] GOG
- [x] Epic
- [x] Windows Game Store
- [x] Game Version Resolution
- [x] Save Location
- [x] Testing Mods
- [x] Installing Mods
- [x] Diagnosing Mods
- [x] Sorting Mods


Aragas:
Feedback:
* There's a need for a conditional `ITool`. For example, tool should only be available if a specific mod is installed.
* Metadata for `Mod`. You can add some metadata for a concrete file within a mod, but no for the mod.
This can be worked around by using a concrete mod file to store the data.

Aragas:
Needs #265

Aragas:
Needs #292 for proper multimod support

Aragas:
Needs #312

Aragas:
We also should be able to use #327, as BLSE has similar diagnostics capabilities

Aragas:
I believe the MVP is finished, but I still need to test sorting!
From what I found, the app only uses it currently with the flatten-list command, which seem to not work currently.
I did consider the tests, found `ALoadoutSynrchonizerTest`, but the tests are quite volatile at this stage, so I'm not sure how to approach it currently. The best option might be to wait for the sorting to be integrated to the UI

Aragas:
Added Diagnotic support as part of #327

Aragas:
As mentioned in #700, we also need the ability to discover already installed (manually or the official built-in) mods

erri120:
> As mentioned in #700, we also need the ability to discover already installed (manually or the official built-in) mods

@nepellion

When you add a game, we expect the installation to be "clean". Adding a game with mods already installed is not supported. @halgari

Aragas:
> When you add a game, we expect the installation to be "clean". Adding a game with mods already installed is not supported. @halgari

That might be a big issue, since the game itself distributes at least 5 mods, 4 of which can be enabled/disabled

nepellion:
> When you add a game, we expect the installation to be "clean". Adding a game with mods already installed is not supported.

This might be the case, however some games allow installing mods from within/while running. (I'm not 100% sure if currently, these would be tracked as "Generated files", ignored or deleted based on tracked game files state)

I imagine this would be best handled on a per-game basis as an optional implementation anyway, this would mostly/only affect games with their own mod management, which each have their own way of doing things.

e.g. In #700 (Blade & Sorcery), the mods could be read and displayed without any problems from the mod files, as they reflect 1:1 with the contents of the mod archives from Nexus Mods, however, they can't be easily managed, as mods installed in-game are managed through the integrated mod.io account, very similar to how Steam Workshop would keep tracking and managing it's mods, therefore disabling them in manager wouldn't have effect, since they'd just sync again at runtime.

Aragas:
I agree that we need some game level optional layer that could parse the present folders as archives and do a "virtual" installation

halgari:
I've let this sit out way too long, I apoligize for that. Does this PR mostly work aside from mod sorting? Should we merge it then add sort support, or what do you think the best course of action is here @Aragas?

Aragas:
> I've let this sit out way too long, I apoligize for that. Does this PR mostly work aside from mod sorting? Should we merge it then add sort support, or what do you think the best course of action is here @Aragas?

Thant's okay! The Bannerlord support also brings up the BUTR Bannerlord management library that was designed for Native Launcher/Vortex in terms of the API, so I was waiting for the NexusMods App API to stabilize, so you won't have the burden of adapting those API's.

In it's current state, it should support every feature NexusMods.App has, but I wasn't able to test sorting on the UI.  
And we don't have yet the ability to find any existing mods within the game's installation, but it can be resolved when the feature arrives. Note tho, because of that you won't be able to correctly launch the game unless we hardcode the pre-existing official mods.

I'm fine with this PR's being in a draft for a while longer to reduce the mainenance burden, up to you to decide @halgari. I expect to be able to support with any issues within 1-2 days max

Aragas:
Sorting is implemented and tested

PR #131, opened 2023-02-20
Fixed: Missing virtual keyword

Aragas:
Without the `virtual` keyword, the overridden `GetGameFiles` method won't be used when the game is used as `AGame` type


halgari:
Yep, that likely saved me some hours of debugging

PR #133, opened 2023-02-21
Fixed: Added GameStore To Differentiate Installations

Aragas:
I found out while testing with the CLI that if you have multiple game installations with the same version, the discovery mechanism will break at those lines.
https://github.com/Nexus-Mods/NexusMods.App/blob/44194f9f2518b58d38f914a6113a43f17d47feea/src/NexusMods.DataModel/JsonConverters/GameInstallationConverter.cs#L13-L14
There's not enough 'uniqueness' when using only a (Slug, Version) tuple.
This could be fixed by adding some unique GUID for the concrete installation, but if I understand correctly, we have a constraint for any store that you can have only one (Game, Version) tuple installed at any given time.
This means that a store identifier should be enough for solving the 'uniqueness' issue, (Slug, Store, Version) tuple basically.

This is an example of how this could be fixed

### DIFF
--- a/src/NexusMods.CLI/Verbs/ListGames.cs
+++ b/src/NexusMods.CLI/Verbs/ListGames.cs
@@ -27,8 +27,8 @@ public class ListGames : AVerb
             from install in game.Installations.OrderBy(g => g.Version)
             select install;
         
-        await _renderer.Render(new Table(new[] { "Game", "Version", "Path" },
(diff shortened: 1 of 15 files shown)


Aragas:
This implementation also won't handle cases when you have multiple game installations of the save version that don't have a Store (Unknown).
Storing the absolute path will fix that

erri120:
Replaced with #265, thanks for the PR.

PR #134, opened 2023-02-22
oauth authentication

TanninOne:
This code still needs some cleanup and unit tests but please review if fundamentally the code structure is to your liking, in particular around the way platform dependent code is handled.

### DIFF
--- a/src/Networking/NexusMods.Networking.NexusWebApi/OAuth.cs
+++ b/src/Networking/NexusMods.Networking.NexusWebApi/OAuth.cs
@@ -1,0 +1,270 @@
+﻿using DynamicData;
+using GameFinder.Common;
+using Microsoft.Extensions.Logging;
+using NexusMods.Common;
+using NexusMods.DataModel.Abstractions;
+using NexusMods.DataModel.Interprocess;
+using NexusMods.Networking.NexusWebApi.DTOs;
+using NexusMods.Networking.NexusWebApi.Types;
+using System;
+using System.Buffers.Binary;
+using System.Collections.Generic;
+using System.CommandLine.Parsing;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Threading.Tasks;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+
+namespace NexusMods.Networking.NexusWebApi;
+
+/// <summary>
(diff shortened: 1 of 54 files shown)


erri120:
Protocol registration on Linux can be done via [XDG Desktop Entries](https://wiki.archlinux.org/title/Desktop_entries) ([Specs](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html)). As an example, here is part of the desktop entry of [Steam](https://archlinux.org/packages/multilib/x86_64/steam/) from my machine:

```ini
[Desktop Entry]
Name=Steam (Runtime)
Comment=Application for managing and playing games on Steam
Exec=/usr/bin/steam-runtime %U
MimeType=x-scheme-handler/steam;x-scheme-handler/steamlink;
```

The important bit is the `MimeType` field ([Specs](https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html#idm45530160413952)).

src/Networking/NexusMods.Networking.NexusWebApi/OAuth.cs
```diff
@@ -0,0 +1,270 @@
...
+            }
+        }
+    }
+
+    private string MakeUUID()
+    {
+        return Guid.NewGuid().ToString();
+    }
+
+    private string ToBase64(byte[] input)
```
halgari, line 227: These would be good candidates for [extension methods](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods). We do Base64 conversion enough it would be nice to have them in a more central project and usable by components.

tests/Networking/NexusMods.Networking.NexusWebApi.Tests/Types/NXMModUrlTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    public void ToleratesUnknownParameters()
+    {
+        var parsed = NXMUrl.Parse("nxm://skyrim/mods/123/files/456?whatdis=value");
+        Assert.NotNull(parsed);
+    }
+
+    [Fact()]
+    public void RejectsInvalidUrls()
+    {
+        Assert.Throws<ArgumentException>(() => NXMUrl.Parse("notnxm://skyrim/mods/123/files/456"));
```
halgari, line 67: I need to go through and update the rest of the codebase, but using Fluent Assertions results not only in slightly more readable code, but also nicer error messages such as `expected 42 but got 43, a difference of 1`. 

https://fluentassertions.com/exceptions/

halgari (changes requested):
Aside from these two comments, it looks good. I love how componentized it is, looks great.

PR #145, opened 2023-03-01
Add `.gitattributes` file

erri120:
The [`.gitattributes`](https://www.git-scm.com/docs/gitattributes) file can be used to tell git how to handle certain files. It can also be used to influence the [GitHub language detection](https://github.com/open-papyrus/docs/blob/master/.gitattributes).

The important changes here are the LF normalization and EOL specification for certain text files to prevent any cross-platform issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,61 @@
+# Auto detect text files and perform LF normalization
+* text=auto
+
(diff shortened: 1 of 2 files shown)


Sewer56:
LGTM

Issue #99, opened 2023-02-08
Optimise: Free 1-line Performance Wins Throughout the Codebase


Sewer56:
# Improvement

## Currently 
There are some free performance wins such as:  
- Stack & Heap allocations where arrays are zero'd and then immediately discarded.  
- Heap Allocations for temporary small values <64 bytes.  

## We want
Change a few lines in our code and get free performance 👍.

Sewer56:
Superseded by #214

PR #94, opened 2023-02-07
Cleanup .csproj by using nested shared Directory.Build.props files.

Sewer56:
fixes #73

PR #146, opened 2023-03-01
Add `.editorconfig` file

erri120:
- forces UTF-8 (not utf-8-bom, which VS likes to use)
- forces final new line and trims trailing whitespace for better git compatibility and less diff clutter
- changes warning severities:
  - [CS1591](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591): "Missing XML comment for publicly visible type or member" from _warning_ to _suggestion_. This reduces clutter in the build log while keeping the suggestions visible in the IDE.
  - [CS4014](https://learn.microsoft.com/en-US/dotnet/csharp/language-reference/compiler-messages/cs4014): "Task not awaited" from _warning_ to **error**. This is just a precaution as nothing in the codebase currently triggers this warning. I changed this to be an error, as an issue related to a not awaited task is very hard to debug.
- contains a couple debatable style choices. These don't really matter, but it's best to have a unified code base and not 20 different coding styles.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,68 @@
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+max_line_length = 80
+
+# Microsoft .NET properties
+csharp_new_line_before_members_in_object_initializers = false
+csharp_preferred_modifier_order = public, private, protected, internal, file, new, static, abstract, virtual, sealed, readonly, override, extern, unsafe, volatile, async, required:suggestion
+csharp_style_prefer_utf8_string_literals = true:suggestion
+csharp_style_var_elsewhere = true:warning
+csharp_style_var_for_built_in_types = true:warning
+csharp_style_var_when_type_is_apparent = true:warning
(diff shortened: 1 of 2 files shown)


Sewer56:
Note for [CS1591](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591):

It was introduced as part of https://github.com/Nexus-Mods/NexusMods.App/pull/94 , when I consolidated the projects to use `Directory.Build.props` and `NuGet.Build.props`. 

Anything we could potentially publish to NuGet [and thus using `NuGet.Build.props`]; had the XML documentation flag set; thus leading to the situation we're in; as a good chunk of this codebase has been originally ripped out from Wabbajack.

While I don't mind suppressing it for now; it's my personal goal to get this number down to 0 instead as part of the refactoring work I'm doing throughout the codebase and also looking for free wins https://github.com/Nexus-Mods/NexusMods.App/issues/99

halgari:
Merging this for now, we can backout the warnings change once we decide to fix a few more of them.

PR #147, opened 2023-03-01
Small codecov fixes

erri120:
- adds a [validation](https://api.codecov.io/validate) workflow for `codecov.yaml`
- updates `codecov.yaml` to only post a comment after 2 builds. This prevents codecov from posting a comment when only 1 build finishes, as the comment often contains incomplete information. This value should be updated when more builds are added.


Sewer56:
@halgari random CI failure unrelated to PR. Not sure if it's a strange one off; an issue related to test parallelism.

PR #160, opened 2023-03-02
Remove RocksDB and LMDB links in the datamodel project

halgari:
We no longer use these libraries so let's remove them


Sewer56:
test comment [testing slack integration]

PR #173, opened 2023-03-07
Nullable warnings as errors

erri120:
This adds `NullableErrors.Build.props` which changes the nullable warnings to error:

```xml
<WarningsAsErrors>Nullable</WarningsAsErrors>
```

I imported this file in almost all projects, the only projects that are not using this are:

- all test projects
- `NexusMods.App.UI`

Working with nullability in UI projects is very tedious and would require a lot of changes, which I consider outside the scope of this PR.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -66,3 +66,6 @@ dotnet_diagnostic.cs4014.severity = error
 [{*.yaml,*.yml}]
 indent_style = space
 indent_size = 2
+
+[*.csproj]
+indent_size = 4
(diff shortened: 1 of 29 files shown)


Sewer56 (approved):
Actual contents of the commits seem good to me. LGTM

One thing to note, this PR makes the indents inconsistent with other parts of the `.csproj` files, as the existing files use the old VS default of 2 spaces per level of nesting, while new lines use 4 spaces. That should probably be fixed before merging.

it would be nice of we could set some sort of rule for that, to standardise it.

Note: Checked this from my phone.

Sewer56:
In that case we should probably standardise it soon; even if we have to run through the lines manually. It's good to merge though.

PR #176, opened 2023-03-08
File system abstraction

### LINKED ISSUE #168: File system mocking
Currently, we use the real file system everywhere. While this makes it easier to implement features, it also makes it harder to write meaningful unit tests. Here are a couple options:

- Use a virtual file system for testing like [System.IO.Abstractions](https://github.com/TestableIO/System.IO.Abstractions). Since we have custom implementations for various IO related tasks, like dealing with paths, the library might not be a good, and we'd have to use a custom implementation or something else entirely.
- Keep using the real file system, but provide utility functions and wrappers for test runners to easily create, modify, verify and delete files.

## Design

* Implement an interface for mocking filesystem access. Should accept `AbsolutePath` types and return our app value types, for example, return `Size` instead of `ulong`, 
* Create a default implementation of this `IFileSystem` interface
* Add `IFileSystem` as a parameter type to `AbsolutePath`. 
* Update `AbsolutePath` to access the filesystem via `IFileSystem`. Most of the time we shouldn't need to add a `IFilesystem` parameter in the rest of the app. If we have an `AbsolutePath` in a given context we can use the `IFileSystem` from that path, instead of requiring it as an input to the user-code function. 
* For the remaining usecases we'll need to add `IFileSystem` as an additional parameter/component
* Be sure to register the default `IFileSystem` in the DI framework so we can import it into any component

erri120:
**Still work-in-progress and a draft PR for now.** Resolves #168 

Methods available in `IFileSystem` so far:

- `IFileSystem CreateOverlayFileSystem`
- `AbsolutePath FromFullPath` (WIP, should have better name)
- `AbsolutePath FromDirectoryAndFileName` (WIP, should have better name)
- `IFileEntry GetFileEntry`
- `IDirectoryEntry GetDirectoryEntry`
- `Stream OpenFile`
- `Stream ReadFile`
- `Stream WriteFile`
- `Stream CreateFile`
- `void CreateDirectory`
- `bool DirectoryExists`
- `void DeleteDirectory`
- `bool FileExists`
- `void DeleteFile`
- `void MoveFile`
- `Task<byte[]> ReadAllBytesAsync`
- `Task<string> ReadAllTextAsync`
- `Task WriteAllBytesAsync`
- `Task WriteAllTextAsync`
- `Task WriteAllLinesAsync`


erri120:
A quick rundown on my progress:

- created `IFileSystem` interface with some basic functions like `OpenFile`
- changed the internal constructor for `AbsolutePath` to **require** a `IFileSystem` implementation. All public static methods of `AbsolutePath` like `FromFullPath` **optionally** accept an implementation. These methods will use a static instance of `FileSystem` if they don't receive an implementation. This basically prevents every code from breaking, while still adding the new file system abstraction support.
- added an abstract class `BaseFileSystem` which implements some `IFileSystem` methods like `OpenFile` and forwards these calls to `OpenFileInternally`, which is implemented by its inheritors.
- added `FileSystem` and `InMemoryFileSystem` which inherit from `BaseFileSystem`. The first one uses the actual file system, while the latter is completely in-memory and can be used for testing.

As discussed in #168, to have decent Wine support without having to rewrite everything, I've added `IFileSystem CreateOverlayFileSystem(Dictionary<AbsolutePath, AbsolutePath> pathMappings)` to `IFileSystem` for easy creation of an "overlay" file system.

These path mappings are used in `BaseFileSystem`, so that the actual implementations in `FileSystem` and `InMemoryFileSystem` don't notice this:

```mermaid
flowchart TD
    A[IFileSystem.OpenFile] -- path = /foo/bar --> B[BaseFileSystem.OpenFile]
    B -- path = /foo/bar --> C[BaseFileSystem.GetMappedPath]
    C -- path = /foo/baz --> D[FileSystem.InternalOpenFile]
```

This level of abstraction should make it easy for existing code to "just work", without having to deal with path mappings.

erri120:
As part of this PR, I've also added `AutoFileSystem` using AutoFixture:

https://github.com/Nexus-Mods/NexusMods.App/blob/efcd44f2f2f58844b50d3fad6282474408074bcd/tests/NexusMods.Paths.Tests/AutoData/AutoFileSystemAttribute.cs

This makes writing tests very easy, without having to deal with hardcoded values:

https://github.com/Nexus-Mods/NexusMods.App/blob/efcd44f2f2f58844b50d3fad6282474408074bcd/tests/NexusMods.Paths.Tests/FileSystem/InMemoryFileSystemTests.cs#L65-L71

erri120:
Setting this PR ready for review before it becomes too large. Some stuff that I've not yet implemented:

- `IDirectoryEntry` as an abstraction for `DirectoryInfo`
- all of the custom file/directory enumerations
- `AbsolutePath.DeleteDirectory` which uses `DirectoryInfo` and the file/directory enumeration helpers
- `InMemoryFileSystem` does not support any `FileMode` other than `FileMode.Open`

The Windows test is expected to fail, since our path code still contains bugs.

erri120:
Tests fail due to #186.

src/NexusMods.Paths/FileSystemAbstraction/IFileSystem.cs
```diff
@@ -0,0 +1,162 @@
...
+public interface IFileSystem
+{
+    /// <summary>
+    /// Creates a new <see cref="FileSystem"/> that allows for path mapping.
+    /// </summary>
+    /// <param name="pathMappings">Path mappings</param>
+    /// <returns></returns>
+    IFileSystem CreateOverlayFileSystem(Dictionary<AbsolutePath, AbsolutePath> pathMappings);
+
+    AbsolutePath FromFullPath(string fullPath);
```
Sewer56, line 18: Missing XML docs here; and namespace doesn't match up.
If the latter is intended, suppress it; otherwise make it match up the directory structure [preferred].

src/NexusMods.Paths/FileSystemAbstraction/FileVersionInfo.cs
```diff
@@ -0,0 +1,9 @@
+namespace NexusMods.Paths;
+
+public record FileVersionInfo(Version ProductVersion)
```
Sewer56, line 3: Please use `record struct` here, as this element composes of only 1 item [which itself is heap allocated] and does not need to be heap allocated.

Don't forget the docs 😉

src/NexusMods.Paths/AbsolutePath.cs
```diff
@@ -15,6 +15,11 @@ public partial struct AbsolutePath : IEquatable<AbsolutePath>, IPath
 
     private static readonly string DirectorySeparatorCharStr = Path.DirectorySeparatorChar.ToString();
 
+    /// <summary>
+    /// File system implementation.
+    /// </summary>
+    public IFileSystem FileSystem { get; set; }
```
Sewer56, line 21: This increases item size by `sizeof(nint)`; which puts the struct at `3x(sizeof(nint))` size total; basically the boundary between when it's worth to still use a struct vs. a class.

Given in the actual app we're most likely going to use `FileSystem` directly [as `InMemoryFileSystem` is mostly a testing/debugging tool]; we'll end up copying the same instance across each path item; which is a bit unoptimal. 

For now I think this is okay. Down the road though, I think ideally we'd just have extension methods on `IFileSystem` or on `AbsolutePath` which perform an operation using the path; such that memory wise `IFileSystem` can be decoupled from `AbsolutePath`.

erri120: This is mostly just a stopgap measure to prevent this PR from breaking everything. Once newer code uses `IFileSystem` and we update the older code, we can remove this property from `AbsolutePath`, as well as all the utility methods in `AbsolutePath` which use `IFileSystem` (eg: `AbsolutePath.ReadAllTextAsync` just redirects to `FileSystem.ReadAllTextAsync`).

Sewer56: Yeah; just remember, the later it's done the harder it'll be to migrate 😉

Sewer56 (approved):
LGTM; just make sure to fix up [or suppress] the namespaces before merging; and add the missing docs [if they're still missing].

PR #178, opened 2023-03-08
Explicit Persistence and removing the Datastore on entities

halgari:
This refactors the Entities so they no longer implicitly persist themselves to the IDataStore. As a side-effect we then have no reason for them to contain a `IDataStore` so I yanked that as well. Collections will still auto-persist any object inserted into them, but that feels natural as we're calling things like `.Add`. If we don't like these semantics we might be able to yank that as well, but for now, having collections write-on-add is nice for code simplicity.

### DIFF
--- a/src/NexusMods.DataModel/Abstractions/Entity.cs
+++ b/src/NexusMods.DataModel/Abstractions/Entity.cs
@@ -82,9 +82,12 @@ public abstract record Entity : IWalkable<Entity>
     [JsonIgnore]
     public IId DataStoreId
     {
-        get => _id ?? throw new UnpersistedEntity(this);
+        get => _id ?? throw new UnpersistedEntity();
         set => _id = value;
     }
+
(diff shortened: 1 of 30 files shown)


src/NexusMods.DataModel/Abstractions/Entity.cs
```diff
@@ -91,10 +82,13 @@ public void EnsureStored()
     [JsonIgnore]
     public IId DataStoreId
     {
-        get => _id ??= Persist();
+        get => _id ?? throw new UnpersistedEntity();
```
Sewer56, line 85: Please make a `ThrowHelper` method for this exception, to allow this getter to be inlined.

Throwing exceptions prevents inlining of code, which means that every access to this field suddenly has to go through a method call to a method which is only few instructions at best.

Sewer56:
LGTM

Sewer56 (approved):

PR #179, opened 2023-03-09
Fix Linux path issues

erri120:
- `/` is now a valid directory
- fixes `FromFullPath` to use `/` as a directory instead of `""`
- fixes the constructor not to remove `/` from the directory, if the directory is `/`
- fixes `GetFullPath` to not return `//foo` instead of `/foo`


Sewer56 (commented):
Verified works as intended; and looks good to me. 
Just the two elements I commented on earlier.

Sewer56:
LGTM

Sewer56 (approved):

PR #181, opened 2023-03-09
More AbsolutePath fixes for Linux

erri120:


Sewer56 (approved):

Sewer56:
This PR introduces platform specific test logic as opposed to normalising the path(s) for the given platforms as done before. 

No issues there; the only caveat (CC. @halgari ), we'll need to merge code coverage results together between individual job runners at some point. [This will be needed anyway but good time for reminder]

Sewer56 (approved):

PR #182, opened 2023-03-09
Rework the IPC system

halgari:
This reworks the IPC code for the App. 

### Old Code:
 * Used a 3rd party socket/shared memory system
 * Project hasn't been updated in a few months
 * Has some rather concerning GitHub issues about leaking resources
 * Is a bit overkill for our project, we're likely to only ever insert a message every few seconds, not thousands per second
 
### New Code:
  * Uses Sqlite as the log store
  * Data is stored in a separate file from the main DataStore so we don't fragment the old data store
  * Messages are deleted after 10 seconds
  * Every ~10 minutes a running process will run a "delete old messages" command
  * The size/last modified file stats for the .sqlite file are used to trigger a new polling
  * Every 10 seconds the file is polled by doing a `SELECT WHERE` just in case some race condition blocked up the cheaper approach.

### DIFF
--- a/src/NexusMods.DataModel/Interprocess/InterprocessConsumer.cs
+++ b/src/NexusMods.DataModel/Interprocess/InterprocessConsumer.cs
@@ -1,51 +1,20 @@
-using System.Reactive.Subjects;
-using Cloudtoid.Interprocess;
-using Microsoft.Extensions.Logging;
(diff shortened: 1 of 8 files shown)


erri120 (approved):
LGTM and works on Linux.

PR #185, opened 2023-03-11
Use icon rendered from the Next site's SVG

BellCubeDev:
The colors now match the official Next color scheme and there is no longer an ugly border along the sides as this image was rendered directly from the SVG.


halgari (approved):

Sewer56:
LGTM

Issue #186, opened 2023-03-13
Make tests more stable


erri120:
CI keeps failing "randomly" on various tests and need a re-run to be fixed. Our tests should not arbitrarily fail.

The tests that fail randomly are the following:

```log
2023-03-13T11:29:39.2661094Z   Failed NexusMods.CLI.Tests.VerbTests.AnalyzeArchive.CanAnalyzeArchives [235 ms]
2023-03-13T11:29:39.2661734Z   Error Message:
2023-03-13T11:29:39.2662664Z    Expected LastTable.Rows to be a collection with 3 item(s), but found an empty collection.
```

```log
2023-03-13T11:36:36.1457435Z   Failed NexusMods.DataModel.Tests.InterprocessTests.CanSendAndReceiveMessages [7 s]
2023-03-13T11:36:36.1458168Z   Error Message:
2023-03-13T11:36:36.1458842Z    Expected dest to be a collection with 128 item(s), but found an empty collection.
```

Sewer56:
This is a concurrency issue, we can sometimes run parallel tests on the same DataModel, leading to behaviour that is non-deterministic.

I was considering bringing it up on Slack at some point.

It's probably mostly fixable with the inclusion of `xunit.runner.json` (set to copy to output dir).

```json
{
  "parallelizeAssembly": false,
  "parallelizeTestCollections": false,
  "maxParallelThreads": 1 
}
```

erri120:
`AutoFixture` is probably what we want: https://github.com/Nexus-Mods/NexusMods.App/pull/176#issuecomment-1462091681

It's similar to DI and basically completely removes the need for hardcoded variables.

PR #187, opened 2023-03-14
More path fixes

erri120:
This PR is a continuation of #179 and fixes multiple path issues on both Linux and Windows. I've updated and added more tests to ensure we actually test on methods on multiple paths.


erri120:
Test fail due to #186

halgari:
I reran the tests and they pass now, LGTM

Sewer56:
I'm reading through the code too as we speak

halgari (approved):
Looks good to me

src/NexusMods.Paths/AbsolutePath.cs
```diff
@@ -84,19 +85,34 @@ internal AbsolutePath(string? directory, string fileName, IFileSystem fileSystem
...
-        var index = fullPath.LastIndexOf(Path.DirectorySeparatorChar);
-        if (index < 0)
-        {
-            // No directory
+        var span = fullPath.AsSpan();
+
+        // path is not rooted
+        var rootLength = GetRootLength(span);
+        if (rootLength == 0)
             return new AbsolutePath(null, fullPath, fileSystem ?? Paths.FileSystem.Shared);
```
Sewer56, line 93: I have a question about this: Would a non-rooted path not be considered a non-full path; and thus be invalid use of the API?

Should we either throw [throwhelper] or use `Path.GetFullPath()` here?

src/NexusMods.Paths/AbsolutePath.cs
```diff
@@ -109,6 +125,73 @@ public static AbsolutePath FromFullPath(string fullPath, IFileSystem? fileSystem
...
+
+    /// <summary>
+    /// Gets the length of the root of the path.
+    /// </summary>
+    /// <param name="path"></param>
+    /// <returns></returns>
+    /// <exception cref="PlatformNotSupportedException">
+    /// The current platform is not supported.
+    /// </exception>
+    private static int GetRootLength(ReadOnlySpan<char> path)
```
Sewer56, line 151: This method [as well as `IsValidDriveChar`] would benefit from `[MethodImpl(MethodImplOptions.AggressiveInlining)]`, as they have very short bodies. 

It seems the JIT doesn't want to inline them, presumably due to multiple uses; but with the attributes they are successfully inlined.

src/NexusMods.Paths/AbsolutePath.cs
erri120: I think that's part of a bigger design issue with `AbsolutePath`: `Directory` can be null and `FileName` can be empty. There is also the interface `IPath` which also has a `FileName` property, which is implemented as `Path.GetFileName(FileName)`.

We should make a separate issue for this.

src/NexusMods.Paths/AbsolutePath.cs
erri120: It's probably not inlining the function due to the exception being thrown.

src/NexusMods.Paths/AbsolutePath.cs
```diff
@@ -84,19 +85,34 @@ internal AbsolutePath(string? directory, string fileName, IFileSystem fileSystem
...
+        // path is not rooted
+        var rootLength = GetRootLength(span);
+        if (rootLength == 0)
             return new AbsolutePath(null, fullPath, fileSystem ?? Paths.FileSystem.Shared);
+
+        // path is only the root directory
+        if (span.Length == rootLength)
+            return new AbsolutePath(fullPath, "", fileSystem ?? Paths.FileSystem.Shared);
+
+        var slice = span[rootLength..];
```
Sewer56, line 99: Use `SliceFast` here to elide bounds checks. 
We know the root here isn't out of bounds because it is either 0 or found; but the JIT isn't smart enough to pick that up.

For the lines below, it seems a bounds check is still emitted as JIT can't tell that slice should be length > 1 [this should be guaranteed by above checks].

So for `slice[^1]` use the following method to index without a bounds check [taken from `CommunityToolkit.HighPerformance`]

```csharp
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

/// <summary>
/// Returns a reference to an element at a specified index within a given <see cref="Span{T}"/>, with no bounds checks.
/// </summary>
/// <typeparam name="T">The type of elements in the input <see cref="Span{T}"/> instance.</typeparam>
/// <param name="span">The input <see cref="Span{T}"/> instance.</param>
/// <param name="i">The index of the element to retrieve within <paramref name="span"/>.</param>
/// <returns>A reference to the element within <paramref name="span"/> at the index specified by <paramref name="i"/>.</returns>
/// <remarks>This method doesn't do any bounds checks, therefore it is responsibility of the caller to ensure the <paramref name="i"/> parameter is valid.</remarks>
[MethodImpl(MethodImplOptions.AggressiveInlining)]
public static ref T DangerousGetReferenceAt<T>(this Span<T> span, int i)
{
    ref T r0 = ref MemoryMarshal.GetReference(span);
    ref T ri = ref Unsafe.Add(ref r0, (nint)(uint)i);

    return ref ri;
}
```

src/NexusMods.Paths/AbsolutePath.cs
```diff
@@ -84,19 +85,34 @@ internal AbsolutePath(string? directory, string fileName, IFileSystem fileSystem
...
+        var span = fullPath.AsSpan();
+
+        // path is not rooted
+        var rootLength = GetRootLength(span);
+        if (rootLength == 0)
             return new AbsolutePath(null, fullPath, fileSystem ?? Paths.FileSystem.Shared);
+
+        // path is only the root directory
+        if (span.Length == rootLength)
+            return new AbsolutePath(fullPath, "", fileSystem ?? Paths.FileSystem.Shared);
```
Sewer56, line 97: Can you hoist `fileSystem ?? Paths.FileSystem.Shared` to a variable; i.e. `fileSystem ??= Paths.FileSystem.Shared;`, and then use `fileSystem` below. This should save a bit of runtime by making the branches over the constructors faster due to less instructions total generated.

src/NexusMods.Paths/AbsolutePath.cs
```diff
@@ -84,19 +85,34 @@ internal AbsolutePath(string? directory, string fileName, IFileSystem fileSystem
...
+        {
+            // root directory (eg: "/" or "C:\\") is the directory
+            return new AbsolutePath(span[..rootLength].ToString(), slice.ToString(), fileSystem ?? Paths.FileSystem.Shared);
         }
 
-        // Windows: "C:\foo", directory should be "C:"
-        // Linux: "/foo", directory should be "/"
-        var directory = index == 0 ? DirectorySeparatorCharStr : fullPath[..index];
+        // everything before the separator
+        var directorySpan = span[..(rootLength + separatorIndex)];
```
Sewer56, line 111: Make sure to `SliceFast` here too. It should be safe as we handle the case of trailing slash above; but the JIT isn't smart enough to know that and inserts unnecessary bounds check.

src/NexusMods.Paths/AbsolutePath.cs
```diff
@@ -109,6 +125,75 @@ public static AbsolutePath FromFullPath(string fullPath, IFileSystem? fileSystem
...
+        return (uint)((value | 0x20) - 'a') <= 'z' - 'a';
+    }
+
+    /// <summary>
+    /// Joins two path components together.
+    /// </summary>
+    /// <param name="left"></param>
+    /// <param name="right"></param>
+    /// <returns></returns>
+    private static string JoinPathComponents(string left, string right)
```
Sewer56, line 189: Please use span [at least for `right`] and aggressive inline this.

Previous string joins in this PR were replaced with this method; and all spans were `ToString()'d` when we have no use for the new string, leading to allocations.

Normally not a big deal but this also affects `CombineUnchecked` and `GetFullPath` which I consider high performance APIs.

src/NexusMods.Paths/AbsolutePath.cs
```diff
@@ -148,17 +234,17 @@ public void GetFullPath(Span<char> buffer)
 
         Directory.CopyTo(buffer);
 
+        // directory without file name
         if (string.IsNullOrEmpty(FileName)) return;
```
Sewer56, line 238: We can probably also just check if length == 0, avoiding 1 extra compare and branch; since we know FileName is not null based on nullability.

Sewer56:
Anyway, nothing else apart from perf related comments. 
All in all, good stuff 👍🤞

src/NexusMods.Paths/AbsolutePath.cs
```diff
@@ -148,17 +234,17 @@ public void GetFullPath(Span<char> buffer)
 
         Directory.CopyTo(buffer);
 
-        if (string.IsNullOrEmpty(FileName)) return;
+        // directory without file name
+        if (FileName.Length == 0) return;
 
-        // on Linux: Directory="/", FileName="foo" should return "/foo" and not "//foo"
-        if (OperatingSystem.IsWindows() || Directory != DirectorySeparatorCharStr)
+        if (IsRootDirectory(Directory))
```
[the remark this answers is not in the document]
erri120: I chose to use `IsRootDirectory` instead of `Directory.EndsWith`, as it better conveys, what is being checked. We handle root directories in a special way, since they end with a directory separator character. However, if we change the code to just check for that separator character, it's not really clear that this very specific condition, is for root directories.

You can argue that this would be solved by adding a comment, but then you have two sources of truth, and we might end up in a similar situation as before, where the handling of root directories is kinda broken.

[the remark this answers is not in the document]
Sewer56: I was just curious from a performance standpoint really. Ideally the tests should be comprehensive enough to catch that kind of error should it ever occur.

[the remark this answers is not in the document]
Sewer56: Thinking about it, `GetFullPathWithSeparator()` should probably be deprecated; as `GetFullPath()` was reworked to guarantee a separator for root paths [not having it was originally by design, as perf. measure].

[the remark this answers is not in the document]
Sewer56: > https://github.com/Nexus-Mods/NexusMods.App/pull/187#discussion_r1135805350

I'd prefer standardising to the approach that is faster at runtime since `GetFullPath()` is a high activity API.

And yeah, `GetFullPathWithSeparator` can now just be replaced with `GetFullPath` since we now guarantee a separator at the end of root path.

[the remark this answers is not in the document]
erri120: Moved to issue https://github.com/Nexus-Mods/NexusMods.App/issues/189

src/NexusMods.Paths/AbsolutePath.cs
erri120: Moves to issue https://github.com/Nexus-Mods/NexusMods.App/issues/189

PR #191, opened 2023-03-14
Manage game UI 4

halgari:
Building on the Spine->Left Menu->Right Content flow. Implements the "Manage found games", "manage game" user flow:

NexusMods App_mK2Z75G3wj


Sewer56:
Some things to note.

Before reviewing; if you haven't touched the UI code, you should read [UI Coding Conventions](https://github.com/Nexus-Mods/NexusMods.App/blob/main/docs/UICodingConventions.md).

And you might need some IDE Setup; here is [guidance for Rider](https://docs.avaloniaui.net/docs/getting-started/ide-support/jetbrains-rider-setup) <= this should be added to UI Coding Conventions and/or other setup docs.

To open Preview [in Rider], click one of the buttons top right of the editor 
image

[This isn't explained anywhere either, not even on plugin page, so I might aswell mention.]

-----------

In any case; most of this looks fine to me; but as Avalonia is something I haven't yet played around with [although am familiar with WPF which is very, very similar]; I can't call myself fully qualified for Avalonia related matters; I need more experience here. 

Overall project setup looks relatively clean. We should probably have an extension method called `AddViewAndViewModel` as most `ViewModel`(s) and Views have 1:1 mapping.

My only request for this PR is to clean up redundant code marked by Rider's `Problems` window. I'd like this number to hit 0 by the end of the week. As part of my docs and cleanup; we've already went from 1500 [including missing docs] to ~200 [excluding missing docs] now, so that's some progress.

image

Sewer56 (approved):

erri120:
Is this view supposed to not have a scrollbar?

erri120 (changes requested):

halgari:
> Is this view supposed to not have a scrollbar?
> 
> 

Thanks, I forgot to add that in. It's not specified in the designs anywhere but I added it anyways

erri120 (approved):

PR #195, opened 2023-03-15
Cache found games

erri120:
All GameFinder methods are pure, meaning they don't change the internal state of the handler. Repeatedly calling `FindOneGameById` for every game will reparse all files on every invocation.

This PR adds an internal cache to `AGameLocator` to improve performance.

### DIFF
--- a/src/NexusMods.StandardGameLocators/AGameLocator.cs
+++ b/src/NexusMods.StandardGameLocators/AGameLocator.cs
@@ -6,12 +6,13 @@ using NexusMods.Paths;
 namespace NexusMods.StandardGameLocators;
 
 public abstract class AGameLocator<TStore, TRecord, TId, TGame> : IGameLocator
-where TStore : AHandler<TRecord, TId>
-where TRecord : class
-where TGame : IGame
+    where TStore : AHandler<TRecord, TId>
+    where TRecord : class
+    where TGame : IGame
 {
     private readonly ILogger _logger;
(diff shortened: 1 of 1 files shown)


src/NexusMods.StandardGameLocators/AGameLocator.cs
```diff
@@ -21,21 +22,22 @@ protected AGameLocator(ILogger logger, AHandler<TRecord, TId> handler)
...
-            else
-            {
-                yield return new GameLocatorResult(Path(found));
+                    _logger.LogError("While looking for games: {Error}", error);
             }
         }
+
+        return Ids(tg)
+            .Where(id => _cachedGames.ContainsKey(id))
+            .Select(id => _cachedGames[id])
```
Sewer56, line 39: PR LGTM.

Just note the double dictionary lookup, a non-LINQ solution would save you a line [and some perf] here, even if it's not as pleasing to the eye.

```csharp
foreach (var id in Ids(tg))
    if (_cachedGames.TryGetValue(id, out var found))
        yield return new GameLocatorResult(Path(found));
```

Sewer56 (approved):
Just the one note above

src/NexusMods.StandardGameLocators/AGameLocator.cs
erri120: Yeah, `TryGet` pattern is not very LINQ friendly.

PR #196, opened 2023-03-15
Use CliWrap for OpenURL on Windows

### LINKED ISSUE #165: OpenUrl broken on Windows
# Bug Report

## Summary
We can't log into the app currently due to `explorer.exe <url>` opening a copy of Windows Explorer instead of the default app. Not sure if this is a problem with our login URL or the way we're calling explorer. 

## Steps to reproduce
Try and login via the app, a file manager window opens instead of the default browser.

A browser opens.

erri120:
Resolves #165

### DIFF
--- a/src/NexusMods.Common/OSInterop/OSInteropWindows.cs
+++ b/src/NexusMods.Common/OSInterop/OSInteropWindows.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics;
+using CliWrap;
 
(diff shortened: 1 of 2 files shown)


erri120:
Should be tested by someone on Windows.

halgari (approved):
If this is confirmed to work on Windows I'm all for it

Sewer56 (approved):

PR #202, opened 2023-03-16
`AbsolutePath` API Improvement

### LINKED ISSUE #189: `AbsolutePath` API Improvements
From previous discussions:
- https://github.com/Nexus-Mods/NexusMods.App/pull/187#discussion_r1135610423
- https://github.com/Nexus-Mods/NexusMods.App/pull/187#discussion_r1135778351

`AbsolutePath.Directory` is currently nullable, and the only place where it's set to `null`, is in `FromFullPath` when the input path isn't rooted. This raises the question, if a path that is not rooted, meaning it doesn't have a root directory (eg: `foo`) is considered an "absolute path".

`AbsolutePath.GetFullPath` has two versions: `string GetFullPath()` and `void GetFullPath(Span<char>)`. Both methods create a full path, except one writes it to a span, while the other simply returns a heap-allocated string. Internally, they work differently and it might be worth consolidating the internals. `string GetFullPath` could call `GetFullpath(Span<char>)` instead of doing something else internally.

`GetFullPathWithSeparator` should be phased out now that #187 and #176 have been merged, however the custom file/directory enumeration methods still need to be converted to use `IFileSystem` from #176.

erri120:
Resolves #189

Changes:
- `AbsolutePath` is now marked as `readonly`
- `AbsolutePath` requires a rooted path and doesn't accept relative paths
- `AbsolutePath.Directory` is guaranteed to not be null
- `AbsolutePath.GetFullPath` and all overloads use the same function internally
- removed `AbsolutePath.GetFullPathWithSeparator`
- added `AddFileSystem` which adds a singleton implementation for `IFileSystem`
- marked methods which redirect to `_fileSystem` as obsolete, new code should use `IFileSystem` directly
- removed `FileEntry` and replaced all usages with `IFileEntry`
- added and implemented file and directory enumerator methods to `IFileSystem`
- more tests


halgari (approved):

Sewer56:
Sorry for the delay in reviewing this PR, I was investigating.

There's a performance regression in this PR in the case of memory efficiency; compare the old and new benchmarks:

New:
```
|                   Method |    Mean |    Error |   StdDev |      Gen0 |      Gen1 | Allocated |
|------------------------- |--------:|---------:|---------:|----------:|----------:|----------:|
|       EnumerateFiles_New | 3.718 s | 0.0208 s | 0.0194 s | 3000.0000 | 1000.0000 |  59.31 MB |
|       EnumerateFiles_Old | 3.714 s | 0.0299 s | 0.0280 s | 5000.0000 | 1000.0000 |  81.39 MB |
| EnumerateDirectories_New | 3.740 s | 0.0221 s | 0.0196 s | 2000.0000 | 1000.0000 |  44.65 MB |
| EnumerateDirectories_Old | 3.737 s | 0.0227 s | 0.0212 s | 2000.0000 | 1000.0000 |  33.44 MB |
| EnumerateFileEntries_New | 3.745 s | 0.0262 s | 0.0245 s | 4000.0000 | 1000.0000 |  66.45 MB |
```

Old:
```
|                   Method |     Mean |    Error |   StdDev |       Gen0 |      Gen1 | Allocated |
|------------------------- |---------:|---------:|---------:|-----------:|----------:|----------:|
|       EnumerateFiles_New |  3.700 s | 0.0204 s | 0.0181 s |  2000.0000 | 1000.0000 |  39.52 MB |
|       EnumerateFiles_Old |  3.734 s | 0.0164 s | 0.0137 s |  5000.0000 | 1000.0000 |  81.39 MB |
| EnumerateDirectories_New |  3.728 s | 0.0146 s | 0.0137 s |  1000.0000 |         - |   29.5 MB |
| EnumerateDirectories_Old |  3.714 s | 0.0256 s | 0.0240 s |  2000.0000 | 1000.0000 |  33.44 MB |
| EnumerateFileEntries_New |  3.737 s | 0.0321 s | 0.0301 s |  2000.0000 | 1000.0000 |  48.69 MB |
```

In these benches pretty much everything (~99%) allocated should be used up/returned to user so this basically represents actual memory use.

Memory profiler didn't point me to the exact location new allocations were happening [but it was close, maybe code was inlined]. That said, I figured what the deal is.

----------

# Quick Explanation

See previous Enumerator implementation(s)

image

The method `TransformEntry` gets called for each file present in a directory. If directory is not set, it is set, guaranteeing it's never null for the API consumer.

The enumerator walks directory by directory [it doesn't multithread or interleave them etc.]. When it finishes walking a directory, the `CurrentDirectory` is set to `null` so it will be updated when the next file is enumerated via `TransformEntry`.

[`FileSystemEnumerator` lacks an `OnDirectoryStarted` API, so we're forced to do a check on every new file]

What this effectively means that `CurrentDirectory` string is deduplicated in memory; and is the reason why `AbsolutePath` stores `Directory` and `FileName` as separate components. I wanted to be able to get similar memory savings to the older approach used in Wabbajack [original paths code was WJ code], without sacrificing much performance compared to simply just using a single string.

Consider reverting the changes in the enumerators and documenting them [if desired] for future reference.

----------

By the way, I'm completely happy with this PR; everything else is 👍👍.

Sewer56 (changes requested):
see above comment

Sewer56 (approved):

PR #203, opened 2023-03-16
Added: Remaining Documentation, Warning Resolutions & PR (closes #103)

### LINKED ISSUE #103: Improvement: Reduce amount of warnings produced by project.
# Improvement

## Currently 

Currently, in classic .NET Developer fashion we have a nice amount of 500+ warnings throughout our codebase.  
It might not be a bad idea to clean those up.  

Main source of warnings include:  
- Undocumented Public APIs.  [mostly this]
- No matching function parameters.  
- Nullable related errors (e.g. possible null reference assignment).  
- Async methods without any asynchronous code.  

## We want
As some warnings out there are legitimately useful; we should clear up all of the warnings which belong 

## Supporting Information
Not the most exciting task but it's a very nice way of getting more familiar with the codebase 👍 

Note: Normally I would assign this a `small` label; however, this is 500+ warnings we're dealing with, so this might realistically take up to 3 days.

Sewer56:
# About

This PR completes the work on https://github.com/Nexus-Mods/NexusMods.App/issues/103 , adding remaining TODOs, trivial cleanups and documentation for our projects all throughout.

As of the merging of this PR, the following should hold true:
- 100% Public API Documentation Coverage for all Library Projects.
- 0 Warnings [down from ~200 before this sprint]
- 0 'Problems' in Jetbrains [down from 1500 before this sprint]

After merging this PR, I'll start putting up issues for some things I found during this documentation & refactoring process [marked by TODOs in code]. 

It wouldn't make it for this week's sprint planning; but should still be okay.

# Additional Changes

Things worth noting.

- 156ded40b4861b1551ee195035e74c78a57f7faa : Missing XML Documentation now once again results in compile warnings.  
- 5fa66960bdf8ddccd013e9e0fe43ec750fe880bd : Makes versions of `Microsoft.Extensions.*` dependencies more consistent between projects.


erri120:
Should probably merge #202 before this, both PRs are in conflict with each other.

Sewer56:
Sorry for the delay in reviewing this PR, I was investigating.

There's a performance regression in this PR in the case of memory efficiency; compare the old and new benchmarks:

New:
```
|                   Method |    Mean |    Error |   StdDev |      Gen0 |      Gen1 | Allocated |
|------------------------- |--------:|---------:|---------:|----------:|----------:|----------:|
|       EnumerateFiles_New | 3.718 s | 0.0208 s | 0.0194 s | 3000.0000 | 1000.0000 |  59.31 MB |
|       EnumerateFiles_Old | 3.714 s | 0.0299 s | 0.0280 s | 5000.0000 | 1000.0000 |  81.39 MB |
| EnumerateDirectories_New | 3.740 s | 0.0221 s | 0.0196 s | 2000.0000 | 1000.0000 |  44.65 MB |
| EnumerateDirectories_Old | 3.737 s | 0.0227 s | 0.0212 s | 2000.0000 | 1000.0000 |  33.44 MB |
| EnumerateFileEntries_New | 3.745 s | 0.0262 s | 0.0245 s | 4000.0000 | 1000.0000 |  66.45 MB |
```

Old:
```
|                   Method |     Mean |    Error |   StdDev |       Gen0 |      Gen1 | Allocated |
|------------------------- |---------:|---------:|---------:|-----------:|----------:|----------:|
|       EnumerateFiles_New |  3.700 s | 0.0204 s | 0.0181 s |  2000.0000 | 1000.0000 |  39.52 MB |
|       EnumerateFiles_Old |  3.734 s | 0.0164 s | 0.0137 s |  5000.0000 | 1000.0000 |  81.39 MB |
| EnumerateDirectories_New |  3.728 s | 0.0146 s | 0.0137 s |  1000.0000 |         - |   29.5 MB |
| EnumerateDirectories_Old |  3.714 s | 0.0256 s | 0.0240 s |  2000.0000 | 1000.0000 |  33.44 MB |
| EnumerateFileEntries_New |  3.737 s | 0.0321 s | 0.0301 s |  2000.0000 | 1000.0000 |  48.69 MB |
```

In these benches pretty much everything (~99%) allocated should be used up/returned to user so this basically represents actual memory use.

Memory profiler didn't point me to the exact location new allocations were happening [but it was close, maybe code was inlined]. That said, I figured what the deal is.

----------

# Quick Explanation

See previous Enumerator implementation(s)

image

The method `TransformEntry` gets called for each file present in a directory. If directory is not set, it is set, guaranteeing it's never null for the API consumer.

The enumerator walks directory by directory [it doesn't multithread or interleave them etc.]. When it finishes walking a directory, the `CurrentDirectory` is set to `null` so it will be updated when the next file is enumerated via `TransformEntry`.

[`FileSystemEnumerator` lacks an `OnDirectoryStarted` API, so we're forced to do a check on every new file]

What this effectively means that `CurrentDirectory` string is deduplicated in memory; and is the reason why `AbsolutePath` stores `Directory` and `FileName` as separate components. I wanted to be able to get similar memory savings to the older approach used in Wabbajack [original paths code was WJ code], without sacrificing much performance compared to simply just using a single string.

----------

By the way, I'm completely happy with this PR; everything else is 👍👍.

PR #219, opened 2023-03-21
Added: Faster Image Blur without Re-Encode (fixes #194)

### LINKED ISSUE #194: Rework image blur code
# Bug Report

The code I wrote in the [GameWidget](https://github.com/Nexus-Mods/NexusMods.App/blob/61c39a627deda2277e9b339104d19721707a879c/src/NexusMods.App.UI/Controls/GameWidget/GameWidget.axaml.cs#L33-L62) is an offense against nature, and should really be replaced.

## Summary
The issue is that we want to modify the image using Skia's routines, but the image in the start and end of the transformation should be an Avalonia bitmap. Avalonia uses Skia for drawing, but I couldn't find any sort of good translation at the time. We should figure out how to convert these bitmaps and remove the ugly code that saves and loads the images from PNGs.

That the transformation would be done without any sort of encoding/decoding/saving of the image data.

Sewer56:
# About

This PR addresses the issue in the original `GameWidget` code where the image transformation was inefficient and relied on saving and loading PNG images. The following changes have been made to improve performance and remove the need for re-encoding:

- Moved the blurring of the image to an extension method located in `NexusMods.App.UI.Extensions.SkiaExtensions`.
- Implemented custom conversion between Skia and Avalonia bitmaps in `NexusMods.App.UI.Extensions.SkiaAvaloniaExtensions`.

I had to figure out the conversion between Skia and Avalonia bitmaps from scratch, as there was no existing documentation or examples I could find anywhere.

## Benchmark Results

The following benchmarks were taken within the live app in Release Mode, with no specific benchmark code written.  
The tests were performed on a 5900X with DDR4 3000MHz CL16 memory.  

- Converting from Avalonia to Skia: ~0.0254ms
- Blurring the image: ~5.7181ms
- Converting from Skia to Avalonia: ~0.4908ms (further optimization is not possible)

No encoding, transcoding or decoding is performed in any of these steps.


Sewer56:
I changed this to draft because I noticed I found out it's possible to extract data from `Bitmap` too; so I'll add another special case for that tomorrow.

halgari (approved):

Sewer56:
Should be good to merge

PR #222, opened 2023-03-22
Implement Configuration and Fix File Logging Issues (fixes #218 , fixes #126)

### LINKED ISSUE #126: Move hardcoded settings into a DI injected settings object(s)
Currently the settings for the downloader, datastore, temp storage, etc, are either injected as bare values directly into the project, or hardcoded as constants in the component. This makes construction of these objects overly verbose and doesn't allow for tweaking of the values without a recompile. 

An overview is available here: https://benfoster.io/blog/net-core-configuration-legacy-projects

The gist is that we want to move the app settings into a .json file, that is mapped to a strongly typed object. 

# Requirements

* Build strongly typed objects for components that need configuration
* Combine these objects into a `AppConfig` Object
* DI the sub objects into the required components
* Hook all this up to the .json file in `NexusMods.App`

### LINKED ISSUE #218: NLog: Logging does not use Platform Appropriate Path
# Bug Report

## Summary
The current file logging configuration in the C# code using `NLog` does not work on other platforms due to the use of hardcoded Windows-style path. This leads to undefined behaviour on other platforms.

See the relevant code snippet below:

```csharp
var fileTarget = new FileTarget("file")
{
    FileName = @"c:\tmp\logs\nexusmods.app.current.log",
    ArchiveFileName = @"c:\tmp\logs\nexusmods.app.{##}.log",
    ArchiveOldFileOnStartup = true,
    MaxArchiveFiles = 10,
    Layout = "${processtime} [${level:uppercase=true}] (${logger}) ${message:withexception=true}",
    Header = "############ Nexus Mods App log file - ${longdate} ############"
};
```

## Steps to reproduce
1. Run the application on a Linux or OSX environment.
2. Perform any action that triggers a log event.
3. Well, try to find the log file. Where is it?

The file logging configuration should work on all platforms, including Windows, Linux, and OSX. 

## Other information

I will fix this as part of https://github.com/Nexus-Mods/NexusMods.App/issues/126 , and simply use a subdirectory of program directory.

Sewer56:
# About

The changes introduced in this PR resolve the following configuration & logging related issues:

- #126 
- #218

# Changes

The following changes have been made...

## Configuration

- Implemented strongly typed objects for components requiring configuration.  
  - Each library has its own object, e.g.  `NexusMods.FileExtractor` has `IFileExtractorSettings`.  
  - When settings are not provided; sensible defaults are applied.  
  - Configuration components are directly injected via DI where appropriate.  
  
- Main App Configuration is exposed as `AppConfig.cs` in `NexusMods.App`.  
  - This configuration combines all of the individual library configurations into one.  
  - `appsettings.json` is provided; with values matching our previous defaults out of the box.  

All potentially useful previously hardcoded constants are now available through the config. If there's anything else that might be useful, let me know.

## Default Value Rules

These rules are explained in the source code, `AppConfig.json`

- Zero & Negative values are treated as default values if they are invalid for a specific field.  
- Arrays must not contain duplicates [ensured at boot time].  
- Relative Paths are expanded to Absolute Paths; and allow for path expansion.

I added a `Sanitize` function to individual settings objects to ensure these rules are enforced.

## Path Expansion

- Added a new API, `KnownFolders.ExpandPath`, to expand paths using known monikers for various folders, `{EntryFolder}`, `{CurrentDirectory}`, `{HomeFolder}` and `{MyGames}`.
  - This API is used to ensure that the configuration paths are cross-platform compatible.
  - Provided sample unit tests for the new `KnownFolders.ExpandPath` API.

## File Logging

- Updated the NLog file logging configuration to use settings specified from our `AppConfig`.
- Saving file logs should now specify a correct path on all platforms.


erri120:
A `JsonConverter<AbsolutePath>` can be implemented like so:
- write a string `path.GetFullPath()`
- read a string `AbsolutePath.FromFullPath(fullPath)`

Sewer56:
> A `JsonConverter<AbsolutePath>` can be implemented like so:
> 
> * write a string `path.GetFullPath()`
> * read a string `AbsolutePath.FromFullPath(fullPath)`

I'm not sure if this would be necessary in this case as paths can be specified absolute or relative.
If I did implement the conversion however; the following would hold true:

- The converter would call `AbsolutePath.FromFullPath(fullPath)` to transform string from config to Absolute Path.
- Sanitize function would then convert back to string with `AbsolutePath.GetFullPath()`; expand the paths.
- And then Sanitize would have to then convert back using `AbsolutePath.FromFullPath(fullPath)` again.

Not really sure what the benefit would be here; it would just amount to extra code with no real gain. 

The purpose of `AbsolutePath` has always really been for paths sourced from the OS as per [Case Sensitivity Design Doc](https://github.com/Nexus-Mods/NexusMods.App/blob/main/docs/decisions/backend/0003-case-sensitivity.md); but we're sourcing these paths from the users; and they can be relative paths too, so <derpy_face_emoji<.

That said, if desired, I do see potential benefit in custom `JsonConverter` that would expand the string as part of the deserialize operation; I don't mind implementing that.

halgari:
> That said, if desired, I do see potential benefit in custom `JsonConverter` that would expand the string as part of the deserialize operation; I don't mind implementing that.

It would be nice to have, just for the reason that we're likely to add paths in the future, and also likely to forget to sanitize these extra paths.

Sewer56:
# About

I made some changes in the following commit; however after looking at the problem a bit further; I noticed there is no 'perfect solution'; one way or another tradeoffs have to be made.

After some thinking; I think I found a decent-ish compromise; but I don't mind making further changes, or reverting to older version.

# Summary of Changes

In this update, two significant changes have been made:

1. **Added `ConfigurationPath` struct**: This new struct introduces the `RawPath` property, which stores the path as set by the user. 
    - The method `GetFullPath()` is used to expand the path.
    - This will enable preserving path expansion monikers like `{EntryFolder}` in future config editing in the UI/CLI.
    - Explicit sanitization of paths in `Sanitize()` no longer required.

2. **Switched from `Microsoft.Extensions.Configuration.Json` to `System.Text.Json`**. 
    - We lose the ability for automatic configuration from environment variables & cmd parameters.  
    - But we can now support custom serialization; and co. So we have guarantee nobody forgot to sanitize a value.
        - We could add additional types like `ConfigurationPath` to auto sanitize on deserialize if desired.

## Problem 

In the previous version, the code relied on `Microsoft.Extensions.Configuration.Json` as requested in original issue. However, this library does not allow customization of the (de)serialization process using `JsonConverter`(s) or similar. 

This is because `Microsoft.Extensions.Configuration` is more of a 'key-value' store for strings, and values can come from various sources, such as command-line arguments and environment variables. This makes the customizing the JSON (de)serialization process impossible; because under the hood library just stores strings.

PR #224, opened 2023-03-23
More FileSystem changes for Wine

erri120:
`BaseFileSystem` will can now convert a Windows path to a Linux path when configured to do so. This is only useful for Wine mappings and is disabled by default.


halgari (approved):
Looks simple and concise

Sewer56 (approved):
LGTM

PR #226, opened 2023-03-25
Convert "Maximize" button to a toggle

BellCubeDev:
# About
While doing a basic compile & run test, I noticed that the Maximize button was one-way. We can't have that!

# Changes
I've renamed `MaximizeCommand` to `ToggleMaximizeCommand` and implemented a basic ternary conditional to toggle the window state.


Sewer56:
IIRC there was a change introduced at some point recently-ish where the title bar got borked. 
The title bar can no longer be dragged either [at least on my machine].

Sewer56 (approved):
Anyway, looks good to me. Thanks!

BellCubeDev:
> The title bar can no longer be dragged either [at least on my machine].

I noticed the *no dragging* thing, too. I'm not experienced enough with C# to go and find it, though.

PR #227, opened 2023-03-26
Merge `main` into `87-fomod-xml`

BellCubeDev:
Conflict resolution for the changes made to `main` this month and the `87-fomod-xml` branch. Tested by compiling, running, and exploring the CLI options a little bit.


BellCubeDev:
Rendered obsolete by #237

Sewer56:
Please do keep this branch around; some of the tests here will still be useful in the future 👍
I wasn't able to consume them within the timeframe of this week/sprint, because we haven't yet implemented plugin/file dependency checks.

[NMA determines which files it will push out ahead of time, so we need to query NMA which files it'll expect in output directory; and that's not implemented yet]

PR #232, opened 2023-03-28
Datamodel refactor

halgari:
I realized while working on the code, that the overall structure of the data model could use some work. Several parts of the datamodel had overlapping concerns (multiple ways of doing things), and objects like `Root` encouraged a strange programming model where we could run into issues down the line if the UI used multiple roots and they updated inconsistently. This PR condenses all of these methods into a few self-contained classes, that we can more easily reason about.

Instead of going over the old structure in detail, this will be a overview of how the code works now: 

## Loadouts 
Are stored as immutable entities in the `EntityCategory.Loadouts` table. The most up-to-date revision of every loadout is found in the `EntityCategory.LoadoutRoots` table. Each loadout the user has is given a unique `LoadoutId`, this id does not change as the list is updated. Instead each revision of the list updates the `DataStoreId` (as it does for all Entity classes). This means that we always want to refer to a loadout via a LoadoutId, which using a `LoadoutRegistry` can be resolved to the current `Loadout` instance.

## LoadoutRegistry
A singleton class that provides routines for looking up loadouts by LoadoutId, as well as methods for updating loadouts, and mods based on Ids. This has replaced the concept of the "Root" in the old model. Root objects provided a unneeded level of indirection. Now the top level Ids for Loadouts are stored in their own table. 

This class should never contain any sort of business level logic, such as working with tools, games, or adding mods. The sole purpose of this class is to be the CRUD methods for working with Loadouts

## LoadoutManager
This is the "business logic" of the app. It deals with list application, managing games, etc. All the updates this class performs to a loadout should be through the LoadoutRegistry. This is also a singleton class and involves pulling in a lot of DI components. We may break this class apart in the future, as so far it it's starting to turn into a bit of a god object. 

## LoadoutMarker
A shorthand way of passing around a LoadoutId, and a LoadoutManager together. In the future parts of this class may be moved into LoadoutManager or a separate class.

# Usage of classes

Use: 
  * Loadouts - if you need read-only access to a loadout. Start by using this class and use others as needed
  * LoadoutManager - if you need to add mods, manage games, etc. 
  * LoadoutRegistry - if you need IObservable views of loadouts, and want to be notified when loadouts change
  * LoadoutMarker - if you need read-write access to a loadout and don't want to pass the LoadoutManager/LoadoutId pair through a bunch of routines


src/NexusMods.DataModel/Cursors.cs
```diff
@@ -0,0 +1,11 @@
+﻿using NexusMods.DataModel.Loadouts;
+
+namespace NexusMods.DataModel;
+
+/// <summary>
+/// A logical
+/// </summary>
+/// <param name="LoadoutId"></param>
+public record struct LoadoutCursor(LoadoutId LoadoutId);
```
Sewer56, line 9: Don't forget the docs here :wink:

halgari: It has docs...just not a full sentence :P  . And actually that file should be deleted, not sure why it got left in the PR.

Sewer56 (approved):
Looks good to me, a logical cleanup that makes things simpler.
We should probably produce a doc on how this is all wired up at some point.

PR #234, opened 2023-03-29
File system bug fixes

erri120:
- fixes path mappings with `GetKnownPath`
- adds `IFileSystem.EnumerateRootDirectories`

### DIFF
--- a/src/NexusMods.Paths/FileSystemAbstraction/BaseFileSystem.cs
+++ b/src/NexusMods.Paths/FileSystemAbstraction/BaseFileSystem.cs
@@ -211,6 +211,30 @@ public abstract class BaseFileSystem : IFileSystem
         => InternalGetDirectoryEntry(GetMappedPath(path));
 
     /// <inheritdoc/>
+    public IEnumerable<AbsolutePath> EnumerateRootDirectories()
+    {
+        if (!OperatingSystem.IsLinux() && !OperatingSystem.IsWindows())
+            throw new PlatformNotSupportedException();
+
+        if (OperatingSystem.IsLinux())
+            yield return FromFullPath("/");
+
+        // go through the valid drive letters on Windows
+        // or on Linux with cross platform path conversion enabled, in case
+        // the current file system is an overlay file system for Wine
+        if (OperatingSystem.IsWindows() || (OperatingSystem.IsLinux() && _convertCrossPlatformPaths))
+        {
+            for (var i = (uint)'A'; i <= 'Z'; i++)
+            {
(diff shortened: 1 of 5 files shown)


Sewer56 (commented):
Is it possible to use an array for `knownPathMappings` here?

`KnownPath` is an enum with linear values [from 0] and known fixed size, so you could probably make a quick struct with a Dictionary-like interface in 10 mins and it'd perform better and without over-allocation.

Here's some dummy code to get you started.
```csharp
/// <summary>
/// Quick map for known paths.
/// </summary>
public struct KnownPathMap
{
    private AbsolutePath?[] _values;

    public KnownPathMap()
    {
        _values = new AbsolutePath?[Enum.GetValues<KnownPath>().Length];
    }

    public void Clear() => Array.Fill(_values, default);

    public bool ContainsKey(KnownPath key) => _values[(int)key].HasValue;

    public bool Remove(KnownPath key, out AbsolutePath value)
    {
        var item = _values[(int)key];
        value = item.GetValueOrDefault();
        _values[(int)key] = null;
        return item.HasValue;
    }

    public bool TryGetValue(KnownPath key, out AbsolutePath value)
    {
        var item = _values[(int)key];
        value = item.GetValueOrDefault();
        return item.HasValue;
    }

    public AbsolutePath? this[KnownPath key]
    {
        get => _values[(int)key];
        set => _values[(int)key] = value;
    }

    public IEnumerator<AbsolutePath> Values()
    {
        for (var x = 0; x < _values.Length; x++)
        {
            if (_values[x].HasValue)
                yield return _values[x].Value;
        }
    }
}
```

erri120:
> Is it possible to use an array for `knownPathMappings` here?
> 
> `KnownPath` is an enum with linear values [from 0] and known fixed size, so you could probably make a quick struct that wraps an array with a Dictionary-like interface in 10 mins and there wouldn't be any collisions
> 

`BaseFileSystem` and `InMemoryFileSystem` are not optimized at all. Since the new APIs have only existed for a couple weeks, I want to make those stable before dealing with performance. Converting GameFinder to use our custom `IFileSystem` instead of `System.IO.Abstractions.IFileSystem` was a good experiment to point out APIs that can be improved or added (#224, #231, #233, #234).

I'd move performance optimizations to a separate issue and deal with them after the API is more stable and we use `IFileSystem` in our code more. With this approach, we can just focus on performance, instead of dealing with bug fixes, and can write benchmarks.

Sewer56:
Alright, works for me.

Sewer56 (approved):

PR #237, opened 2023-03-31
Miscellaneous FOMOD Improvements (fixes #198)

### LINKED ISSUE #198: Cleanly Integrate FOMOD XML support
Based on Tannin's work, I think we're ready to move forward on integrating FOMOD support into the app. 

Use https://github.com/Nexus-Mods/NexusMods.App/tree/87-fomod-xml for reference. 

I'm not too keen on the idea of using git submodules here because they tend to be a rather brittle and we already have solid package management via nuget. 

Something we can fix in this prototype, is the converting of IModInstaller to async. Ideally mod installers should always be side-effect free. Any code that needs to perform IO should be moved into a IFileAnalyzer so that mod installation is a purely functional operation: you feed it mod info, it returns install info, with no side effects to the disk.  

Objectives:

* [x] Release the FOMOD XML library via Nuget 
* [x] Move the [analysis code](https://github.com/Nexus-Mods/NexusMods.App/compare/main...87-fomod-xml#diff-aa580f1cb2562c35f061663ab1b33aa224464a51bda267e2147c16a622ea533fR44-R63) into a `IFileAnalyzer` and cache the XML so we don't have to run extractors during the mod install
* [x] Remove the modifications to the API that were required by making IModInstaller async

For now, let's keep the FOMOD code as a black box, we shouldn't need to modify it to accomplish these goals, just write a package/publish script for the library code, and then import it into the app via Nuget.

Sewer56:
# Changes

- XML installer tests now run on actual stubbed `.fomod` files.
  - And there's tests for `.7z` and `.rar` files.  
- Added additional test cases for FOMODs.  
- FOMODs that specify an Asset/Image URL should no longer break CLI.  
- Added: Functionality to add names to tables.
  - Tables for groups now have titles.
- FOMOD tests now double as integration tests; some code was unstubbed, and we now also test communication between `IFileAnalyzer` and `IModInstaller`.
- FOMODs are now entirely loaded from memory, with no extraction from disk.  
- FOMODs with broken schema should no longer throw an exception; only log an error.  
- Images are now cached as part of analysis data [with the XML]. However, improvements can still be made, as the JSON gets serialized as Base64, increasing space use for images by 25%.  
- Cleaned up some of the code around FOMODs.  

## Remaining tasks

- Plugin/File Dependency Handling is not Yet Implemented.
  - Need to implement `IPluginDelegates` and look up the current loadout for files to be deployed.
  - Figure out all the files once and just look up in a dictionary.
  - It's unclear if it's possible to do this efficiently.
- Saving and Restoring User Preferences Across Re-Deployment.


Sewer56:
This seems to be failing because after update of DataModel to main, `.sqlite` database handle isn't released after disposing the object pool in `SqliteDataStore`; I'm not sure why.

I'm doubling the unit test for installer as integration test, and thus use a real temporary datastore to ensure stuff transfers correctly from the analyzer to installer.

halgari (approved):
Lotta code here, but it's clean, and does what we need it to do

Sewer56:
I think I know why;
https://github.com/Nexus-Mods/NexusMods.App/blob/15609d01e3f85f331770ba57b855ab8bc9805163/src/NexusMods.DataModel/ConnectionPoolPolicy.cs#L28

We open connections here but they never get closed down the road.

Sewer56:
Okay, I figured out a fix; so I guess I fixed a bug on main too 😅

erri120:
I'm gonna look at this over the weekend and on Monday.

src/NexusMods.Paths/TemporaryFileManager.cs
```diff
@@ -11,6 +11,8 @@ public class TemporaryFileManager : IDisposable, IAsyncDisposable
     private readonly AbsolutePath _basePath;
     private readonly bool _deleteOnDispose;
 
+    // TODO: For unit tests we should inject IFileSystem here.
```
erri120, line 14: Will look into that (#239).

erri120 (changes requested):
The tests should ideally use `InMemoryFileSystem` as the `IFileSystem` implementation. Since you have "real" files that you want to test, adding a method `AddExistingFile` to `InMemoryFileSystem` and calling that as part of the test setup phase, should be enough to get this to work.

As part of #239, I will look into updating `TemporaryFileManager`.

Sewer56:
@erri120 I am aware the tests currently run on the FileSystem. I did originally try to write the code to be entirely in memory; though I felt that the improvements needed to make that possible warrant separate issues/tickets. ~~I was also running out of time in the Sprint~~

More specifically, we need:  
- An In-Memory `IDataStore` .  
- Unpacking of `.7z` and `.rar` in memory (In-Memory `FileExtractor`).
  - We currently use binary `7z`/`7zz` for unpacking; thus can't use in-memory FileSystem.
  - or we just drop non-zip tests entirely; that's probably ok too.
- `TemporaryFileManager` needs `IFileSystem` support as noted in TODO; this is now tracked in https://github.com/Nexus-Mods/NexusMods.App/issues/239.

I felt like some of these were more appropriate to be done at the same time we transition the entire codebase to using `IFileSystem`; allowing for other existing tests to also be done in-memory.

erri120 (approved):

PR #241, opened 2023-04-03
PR for ARD-0005

halgari:


erri120 (changes requested):
Just for completion, the file format section should mention which endianness is being used. The [BinaryReader](https://learn.microsoft.com/en-us/dotnet/api/system.io.binaryreader.readuint32?view=net-8.0#remarks) and [BinaryWriter](https://learn.microsoft.com/en-us/dotnet/api/system.io.binarywriter.write?view=net-8.0#system-io-binarywriter-write(system-uint32)) use the little-endian format.

Sewer56:
Since I got an OK to poke around with this; here's an alternative proposal for format I spent today working on. It's a draft; but it should be fairly complete. 

I feel like all this time reverse engineering file formats and experimenting with compression over the years is coming in handy here; I have a lot of experience in this domain. Did go a bit overboard here; and spent a bit longer than expected, but hey, this was fun. 

No major revisions until I'm done with sprint stuff though :p.

# Format Summary

This file format is a semi-SOLID archive format for storing game mod content; intended to double up as a packaging format for uploading mods. We use SOLID compression to bundle up small files together, while keeping the large files as separate compressed blobs. Therefore the large files can be completely skipped during the extract operation if a small file is all that is needed.

In other words; it has the following properties:
- Files under Block Size are SOLID Compressed.  
- Files above Block Size are non-SOLID Compressed.  
- Variable Block Size.  
- Stores File Hashes Within.  
- TOC in-front.  

In some cases below, numbers may be quoted. These numbers are based on a dataset of 132 Reloaded-II mods that span 7 games. Suggestions for larger datasets are welcome!

# Overall file format

| Header | TOC | Block 1 | Block 2 | ... | Block N |

All sections [including block data] are 4096 aligned to match physical sector size of modern drives and page granularity.

Field sizes used below are similar to Rust notation; with some custom types i.e.
- `u8`: Unsigned 8 bits.
- `i8`: Signed 8 bits.
- `u4`: 4 bits
- `u32/u64`: 4 Bytes or 8 Bytes (depending on variant)

Assume any bit packed values are sequential, i.e. if `u4` then `u4` is specified, first `u4` is the upper 4 bits.

All packed fields are `little-endian`; and written out when total number of bits aligns with a power of 2.
- `u6` + `u12` is 2 bytes `little-endian`
- `u15` + `u17` is 4 bytes `little-endian`
- `u26` + `u22` + `u16` is 8 bytes `little-endian`
- `u6` + `u11` + `u17` ***is 4 bytes*** `little-endian`, ***not 2+2***

## Header (8 Bytes)
- `u8[4]` Magic (`"NXUS"`)
- `u4` BlockSize
- `u4` Version/Variant
- `u8` TocCompressedPageCount
- `u8` TocPoolDecompressedPageCount
- `u8` Reserved [Feature Flags]

### Block Size

Stored so the decompressor knows how big each block is.

Size: `4 bits`, but restricted (1-12).  
Parsed as `32768 << blockSize`.  

i.e. BlockSize = 12 is `67108864` (64MiB, i.e. 2^26). 

[Note: This limited is deliberate for bit packing purposes down the road.]

### Version/Variant

Size: `4 bits` (0-15)

- `0`: 
  - Most common variant covering 99.99% of cases.
  - 20 byte FileEntry w/ `u32` Size
  - Up to 4GB (2^32) per file and 4 million files.

- `1`:
  - Variant for archives with large files >= 4GB size.
  - 32 byte per entry w/ `u64` Size
  - 2^64 bytes per file and 4 million files.

Can add additional variants or updates ontop of this scheme.

### TocPageCount

Number of 4K memory pages required to store the compressed and decompressed Table of Contents.  

- `TocPageCount` represents number of pages needed to store the entire ToC. 
- `TocPoolDecompressedPageCount` represents a buffer size large enough to store decompressed `StringPool` (see below). 

These values are encoded as `numPages - 1` (so 0 == 1) and converted to bytes the following way:
 
```csharp
if (tocPageCount >= 248) 
    return (1024 * 1024) << tocPageCount - 8; // 256MiB max.

return 4096 * tocPageCount; // just under 1MiB max.
```

As a rough reference, StringPool [main memory consumer] of (150+) games (180k files) on my PC used up 3.3MB uncompressed, 660K compressed.  

## Table of Contents (TOC)

- `u32`: FileCount  
- `u16`: BlockCount  
- `u16`: Reserved/Padding  
- `FileEntry[FileCount]`  
  - `u64`: FileHash (xxHash64)  
  - `u32/u64`: CompressedSize  
  - `u32/u64`: DecompressedSize  
  - `u26`: DecompressedBlockOffset (note: Tied to max block size)
  - `u22`: FilePathIndex (in StringPool)
  - `u16`: BlockIndex  
- Blocks[BlockCount]  
  - `u32/u64` CompressedBlockSize
- BlockCompressions
  - `u2` Compression
- StringPool (see below)  
  - `RawCompressedData...`  

This header+TOC design aims to fit under 4096 bytes when possible; >=90% of mods out there will fit (based on my dataset). This is to take advantage of read granularity; more specifically:

- **Page File Granularity**

For our use case where we memory map the file. Memory maps are always aligned to the page size, this is 4KiB on Windows and Linux (by default). Therefore, a 5 KiB file will allocate 8 KiB and thus 3 KiB are wasted.

- **Unbuffered Disk Read**

If you have storage manufactured in the last 10 years, you probably have a physical sector size of 4096 bytes.

```pwsh
fsutil fsinfo ntfsinfo c:
# Bytes Per Physical Sector: 4096
```

a.k.a. ['Advanced Format'](https://learn.microsoft.com/en-us/windows/win32/fileio/file-buffering#alignment-and-file-access-requirements). This is very convenient (especially since it matches page granularity); as when we open a mapped file (or even just read unbuffered), we can read the exact amount of bytes to get header.

### File Entries

Use known fixed size and are 4 byte aligned to improve parsing speed; size 24-32 bytes per item depending on variant.

### Blocks

Each entry contains raw size of the block; this avoids us having to have an offset for each block.

### BlockCompressions

Size: `2 bits` (0-3)

- `0`: Copy
- `1`: ZStandard
- `2`: LZ4
- `3`: Reserved

This section is padded to next multiple of 4.
If we ever need to add more formats; we still got reserved space for flags in header; and a flag could extend this to 4 bits (16 values).

### String Pool

Raw buffer of UTF-8 deduplicated strings of file paths. Each string is null terminated.
The strings in this pool are first lexicographically sorted (to group similar paths together); and then compressed using ZStd.
As for decompression, size of this pool is unknown until after decompression is done; file header should specify sufficient buffer size.

For example a valid (decompressed) pool might look like this:  
`data/textures/cat.png\0data/textures/dog.png`

String length is determined by searching null terminators. We will determine lengths of all strings ahead of time by scanning for (`0x00`) using SIMD. No edge cases; `0x00` is guaranteed null terminator due to nature of UTF-8 encoding.

See UTF-8 encoding table:
```
| First code point | Last code point |  Byte 1  |  Byte 2  |  Byte 3  |  Byte 4  | Code points |
|:----------------:|:---------------:|:--------:|:--------:|:--------:|:--------:|:-----------:|
|           U+0000 |          U+007F | 0xxxxxxx |          |          |          |         128 |
|           U+0080 |          U+07FF | 110xxxxx | 10xxxxxx |          |          |        1920 |
|           U+0800 |          U+FFFF | 1110xxxx | 10xxxxxx | 10xxxxxx |          |       61440 |
|          U+10000 |        U+10FFFF | 11110xxx | 10xxxxxx | 10xxxxxx | 10xxxxxx |     1048576 |
```

When parsing the archive; we decode the StringPool into an array of strings.

Spike: Can we make a dictionary for ZStd that compresses this more efficiently? Build StringPool from base game folder; use that during compression.

## Algorithm Implementation

### Using Memory Maps

When possible, we will work with memory mapped files for two reasons.  

Firstly, a system call is orders of magnitude slower than change to a program's local memory. Secondly; it can avoid data copies, as it can map to kernel's page (file) cache; meaning copies of data isn't unnecessarily made in user space.

### Parallel Decompression

Files are decompressed in parallel by decompressing multiple blocks in separate threads/tasks.

The decoding logic assigns each block to a given thread; and then each thread performs the following action(s):
- Allocate and `mmap` output files in advance.
- Decompress relevant data into output files.
- Flush to disk every file where all data has been written. (Check once all work per thread done.)

No GC work is needed; as most compression algorithms accept `Span<byte>`; and in cases where we might need to do native interop; we of course have pointers.

### Parallel Compression

First we must decide how to group files into blocks.

To do this, we will perform the following steps:
- Files are sorted by size. (optimize for blocks)
- Files are then grouped by extension. (optimize for ratio)
- These groups are chunked into blocks. Files whose size exceeds `BlockSize / 2` are allocated their own block.
- The groups are then re-combined in order of `total size of group data` ascending.
- Groups are assigned to threads, e.g. For 8 groups, 4 threads you assign thread(s) `1,2,3,4` + `1,2,3,4`. This handles load balancing.

Note: Grouping files by extension (equivalent to 7zip `qs` parameter) improves compression ratio, as data between different files of same type is likely to be similar (e.g. two text files).

### Choosing Compression Implementation

- For recompressing downloaded files from the web; prefer faster compression approaches.
  - Use `zstd -11` or `zstd -16` (depending on CPU, estimate power by thread count).
  - Can't have the user wait too long... 
  - We only use highest levels e.g. `zstd -22` when 'publishing' (uploading) to web.  

- Files which we know are not compressible are assigned `copy` as compression strategy in their blocks.
  - We should have some sort of mechanism to feed this info from outside the library.

- Extraordinarily large files should be handled with LZ4. 
  - To avoid case of lots of small files, and one huge file.
  - Files 1GB and beyond. Needs benchmarking.

Prefer ZStd for large files.

### Repacking/Appending Files

Speed of this operation depends on SOLID block size, but in most cases should be reasonably fast. This is because non-SOLID blocks used for big files can be copied verbatim.

Updating the ToC is inexpensive.

### Use as Packaging Format

Our inclusion of hash for each file; yields some nice benefits:  
- We don't need to hash downloaded files for NMA; hashes are known up front.  
- Can do partial download to upgrade from older version of mod.
  - We can download TOC-only, compare hashes.
  - Then only download the chunks we need to decompress our needed data.
  - Inspired by MSIX and certain Linux package formats.

## Appendix

<details>
  <summary>Sample program that estimates StringPool size for each subdirectory under `YOUR_PATH_HERE`.  
Modify to your need.</summary>

  ```csharp
  using System.Runtime.InteropServices;
  using System.Text;
  using SharpZstd.Interop;

  // ZStd
  var allDirectories = new HashSet<string>();
  var allFileNames   = new HashSet<string>();
  int numFolders = 0;

  foreach (var folder in Directory.GetDirectories(@"YOUR_PATH_HERE")) 
  {
      var directories = new HashSet<string>();
      var fileNames   = new HashSet<string>();
      numFolders++;
      
      foreach (var inner in Directory.GetDirectories(folder, "*.*", SearchOption.AllDirectories))
      {
          directories.Add(inner);
          allDirectories.Add(inner);
      }

      foreach (var inner in Directory.GetFiles(folder, "*.*", SearchOption.AllDirectories))
      {
          var name = Path.GetFileName(inner);
          fileNames.Add(name);
          allFileNames.Add(name);
      }

      var serialize = SerializeAll(directories, fileNames);
      Console.WriteLine($"[{Path.GetDirectoryName(folder)}] Number of Bytes: {serialize.Count:000000} (Compressed {Compress(serialize).Length:000000}), File Count: {fileNames.Count:0000}, Dir Count: {directories.Count:0000} | {folder}");
  }

  var serialized = SerializeAll(allDirectories, allFileNames);
  Console.WriteLine($"TOTAL Number of Bytes: {serialized.Count} (Compressed {Compress(serialized).Length:000000}), Avg: {serialized.Count / numFolders}, File Count: {allFileNames.Count}, Dir Count: {allDirectories.Count}");

  static List<byte> SerializeAll(HashSet<string> dirs, HashSet<string> files)
  {
      var result = new List<byte>();
      foreach (var res in dirs.Concat(files).Order())
          result.AddRange(Encoding.UTF8.GetBytes(res));

      return result;
  }

  // using SharpZstd.Interop
  // These are the only? up to date bindings. No wrapper, C API only.
  static unsafe Span<byte> Compress(List<byte> input)
  {
      var result = new byte[Zstd.ZSTD_compressBound((nuint)input.Count)];
      fixed (byte* inputPtr = CollectionsMarshal.AsSpan(input))
      fixed (byte* resultPtr = result)
      {
          return result.AsSpan(0, (int)Zstd.ZSTD_compress(resultPtr, (nuint)result.Length, inputPtr, (nuint)input.Count, 22));
      }
  }
  ```
</details>

<details>
  <summary>Sample program that gets file and directory count for each subfolder.</summary>

  ```csharp
  var results = new List<(string folderName, int folderCount, int fileCount)>();
  var folders = Directory.GetDirectories(@"YOUR_PATH_HERE");

  foreach (var folder in folders) 
  {
      var innerFolders = Directory.GetDirectories(folder, "*.*", SearchOption.AllDirectories);
      var files = Directory.GetFiles(folder, "*.*", SearchOption.AllDirectories);
      results.Add((Path.GetFileName(folder), innerFolders.Length, files.Length)!);
  }

  results.Sort((a, b) => a.folderCount.CompareTo(b.folderCount));
  Console.WriteLine(results.Count);
  foreach (var result in results)
      Console.WriteLine($"Folder: {result.folderName}, Folders: {result.folderCount}, Files: {result.fileCount}, Ratio {result.fileCount / (float)result.folderCount}");

  Console.WriteLine("Stats");
  var lowerQuart = results.Count / 4 * 1;
  var upperQuart = results.Count / 4 * 3;

  int totalFolders = 0;
  int totalFiles = 0;

  for (int x = lowerQuart; x < upperQuart; x++)
  {
      totalFolders = results[x].folderCount;
      totalFiles = results[x].fileCount;
  }

  Console.WriteLine($"Interquartile File/Folder Ratio: {totalFiles / (float)totalFolders}");
```
</details>

Sewer56:
Going to address some feedback while I take a break.

> Much of the byte shaving here is likely a overkill. Let's say we reduce the size of a 32bit int to 24bits, for a given chunk, if chunk sizes are 1MB in size, and we have a 6GB archive (very large for the data sizes we're discussing here), then we're talking about a 6KB savings over all, not something anyone is going to care about for files that are in an archival folder somewhere.

Even less than that.

This format is non-SOLID with SOLID blocks for grouping small files; all files above `BlockSize/2` get their own block; files under that are merged into blocks. The 6KB figure assumes all files are > `BlockSize/2`, which generally won't happen in practice.

The motivation for aggressively packing the ToC is twofold:
- Improve metadata parsing efficiency.
- For use as a packaging format.

For the first one; it's pretty trivial. Any inefficiency introduced unpacking bits will be more than offset by avoiding a read of another >= 4096 bytes from storage. (4096 is physical sector size of modern drives) 

From experience (across multiple games). most mod archives tend include less than 50 files in practice. I wanted to make an archive format that can ideally fit the entire ToC in a single 4K page for 95%+ of mods out there.

----

For the second one; I'd like to make a format people could upload directly to file hosting services (e.g. Nexus/GitHub/GB). 

Minimizing the size of the TOC allows us to be able to pull this information from the web faster; reducing server and client load.  
The idea is we can initially ask for a small download range of like first 8192 bytes (2 pages) and chances are that for 99% of the mods; we then will have the entire TOC. 

What's the point? Well, down the road I'd like to be able to support delta updates for downloads.
The idea is we parse the TOC, compare hashes against files we already have and then ***only*** download files we don't have and those with a different hash; and dynamically build the new archive from the older and current version.

Something we've talked about a while back is wanting to get involved with new games as they release and their communities. To achieve this; becoming part of the modding process itself would be very beneficial; I feel like that should be one of NMA's goals.

Some form of delta updates is something the modding community would greatly benefit from; usually most files in major mods between updates remain unchanged. 

This also ties with our longer term as down the road we want to be able to directly upload mods from NMA.

-------

In general, with each new project I take on, I always try to innovate and push the envelope forward. For example, with Reloaded in 2018-19, I added a dependency system with automatic dependency resolution and downloading, ability to load mods in real time and a modular design that can support any game out of the box; people haven't really been doing these things before in the modding space. 

Last year I added the ability for people to create delta compressed updates for their mods using a fork of VCDiff I optimized. The issue there however is the user uploading the file has to explicitly create the delta from the previous version to the next version; which involves downloading the last version. This can be annoying (and for some people slow) to do; not to mention some users are simply also just not aware of the feature. Being able to have a second delta system in place that is fully transparent from the user is handy.

> Likewise I'm hesitant to do any sort of SOLID compression as this complicates the design and requires reader code to bounce between SOLID and non-SOLID approaches to read a given file. Simplicity over file savings in other words.

With this format; the read logic is actually the same regardless of whether a file is contained within a block, split between two blocks or self contained in its own block.
  
It breaks down to the following:

- Create memory mapped files for each item extract.
- Determine blocks associated with files to extract. 
- Process each block in each thread.

Each thread that processes a block is passed the following context `List<(DecompressedRange, PointerToWriteTo)>`. 

All the thread has to do is decompress the block, writing certain ranges of the decompressed data to their designated pointer (memory mapped file address). Compression is much the same; after some pre-processing (detailed in proposal above); it just boils down to 'compress' these range(s) of these files in given order.

> The other thing to point out here is that we can change this format anytime before the official launch of the application, and even afterwards as long as we version the files. So in that case I'm much more likely to lean on the ideas of YAGNI and say build the simplest thing, then iterate on the design in the future should it prove insufficient.

Well, I got reserved space for these additional things in the header 😉 if needed down the road. Always have to account for that in binary formats.

I enjoy this and get a little passionate I suppose 😅.
Back to sprint stuff for me; happy to address any technical questions; might take a while to respond though, depending on how sprint stuff goes, I need to get that out of the way first.

halgari:
So I've done a lot of thinking about his over the past week. I do think @Sewer56's design is a bit over complicated for the simplest use cases, and I'm not sure how much we gain by building this close to the machine (4K page sizes and the like). But I do recognize that the final implementation will be rather small, and no more complex than BA2/BSA or other archive formats. I think it has the performance features we're looking for, so I say let's go for it. 

@Sewer56 feel free to edit this PR with your design doc, I'm scheduling this work for Iteration 6

Sewer56:
This will be implemented starting next week in [NexusMods.Archives.Nx](https://github.com/Nexus-Mods/NexusMods.Archives.Nx)

halgari:
Closing as this work will be done elsewhere

PR #259, opened 2023-04-07
Refactor HashRelativePath for Simplification and Efficiency (fixes #210)

### LINKED ISSUE #210: Rewrite HashRelativePath Struct to Use Single RelativePath
# Improvement

## Currently

In our C# codebase, we have a struct called `HashRelativePath`, which stores the hash of a given file and an array of `RelativePath` structs. These `RelativePath `structs are meant to be combined (joined) to store the complete relative path tied to the struct. 

i.e. The `HashRelativePath` struct is currently defined as:

```csharp
// Simplified for clarity
public struct HashRelativePath
{
    public Hash Hash;
    public RelativePath[] Parts;
}
```

However, after examining the codebase, as far as I can tell, all current uses of `HashRelativePath` only require one `RelativePath` struct, and the structs are never joined or combined; therefore making the array redundant. In addition, the relative paths can simply be joined up front, rather than being stored as an array, improving overall efficiency.

## We Want

We want to simplify the `HashRelativePath` struct by using only a single `RelativePath` struct instead of an array. This would make the struct more efficient and easier to work with. The updated `HashRelativePath` struct should look more like:

```csharp
public struct HashRelativePath
{
    public Hash Hash;
    public RelativePath Path;
}
```

# Other Info 

This code is based on the old Paths library/system. When rewriting the `Paths` library, it seems I missed this code because this code resides in the `NexusMods.Hashing.xxHash64` library when I was working on `NexusMods.Paths`.

Sewer56:
# Description

This Pull Request addresses the concern raised in issue #210. 

Originally, the design of the `HashRelativePath` struct was intended to store items in the format `DDSHASH | /foo/bar.zip | baz.bsa | file.dds`. 

However, after a thorough examination of the codebase, it appears format/storage method is not being utilized at all (and is thus not currently necessary). In practice, the codebase would serialize the provided example as: `ZIPHASH | /foo/bar.zip`, `BSAHASH | baz.bsa`, and `DDSHASH | file.dds`.

As a result of this observation, I have simplified the `HashRelativePath` struct by removing the unnecessary `RelativePath[] Parts` and replacing it with a single `RelativePath RelativePath`. This change will make the struct more efficient and easier to work with.

## Changes

- Replaced `RelativePath[] Parts` with a single `RelativePath RelativePath` in the `HashRelativePath` struct.
- Modified all instances of `HashRelativePath` that previously specified no path to now specify a default empty path.

## Breaking Changes

The changes made affect the serialization of the `HashRelativePath` struct. As a result, it is strongly advised to wipe the `bin/build` folder (or at least your DataStore) after merging this PR to avoid potential issues.

## Misc

I considered removing the `IComparable` interface from the `HashRelativePath` struct, since no sorting has been taking place (outside of tests) and the sorting method is undefined (it was previously sorting by hash unless the hashes were equal). I left it in for now though.


halgari:
@Sewer56 feel free to merge this

Issue #74, opened 2023-01-31
Auto-invalidate archive content caches when new analyzers are added/updated


halgari:
Currently when analyzing an archive we assume that the archive contents never change and hence the results of the analyzers need never be updated. However during development (and in future versions of the application) we may need to re-analyze archives to get improved analysis data. So we need some sort of versioning system for analysis data. 

* IFileAnalyzer should include some sort of version or hash for determining when the analyzer changes
* All analyzers that match a given file type should be calculated and stored with the analyzed archive in the datastore
  * Consider using a value hash, or some sort of content hash to condense all the analyzers and versions into a single small value (like a long) instead of a list of strings that must be included on every file analyzed 
* When the cached data is loaded the analyzer fingerprint should be compared against the current fingerprint and if they differ, the cached copy should be discarded.

PR #264, opened 2023-04-11
Archive Content Cache Invalidation

halgari:
Adds a Id and Revision pair to every `IFileAnalyzer`, changing the Revision number of an Analyzer will trigger a re-analysis of a file once it opens the next time. 

Unfortunately, because files can contain other files, we have to invalidate everything in the cache whenever a single IFileAnalyzer is updated. This shouldn't be too bad though as right now analysis data is only ever used during mod installation. More complex code could likely improve the user experience here, but at least for now we don't get bad analysis data whenever the code updates.


halgari:
fixes #74

erri120 (approved):

Issue #42, opened 2023-01-26
Game Support - Stardew Valley


Greg-Nexus:
## Outline

The desired outcome of this issue is that Stardew Valley:

- is automatically detected; AND
- game version is detected; AND
- can be managed by Nexus Mods App; AND
- the most common mods can be installed; AND
- the game starts with mods installed (and the mods work).

...for all applicable game stores (see below).

## Supporting Information

- Stores
  - [Steam](https://store.steampowered.com/app/413150/Stardew_Valley/)
  - [GOG](https://www.gog.com/game/stardew_valley)
  - [Xbox Game Pass](https://www.xbox.com/en-US/games/store/p/9MWR1NC6VQ6L)
- Automatic Game Detection
  - [ ] Steam
  - [ ] GOG
  - [ ] Windows Game Store
- Game Version Detection
  - [ ] Steam
  - [ ] GOG
  - [ ] Windows Game Store
- Game Managed
  - [ ] Steam
  - [ ] GOG
  - [ ] Windows Game Store

LetsTussleBoiz:
As a fervent modder and mod maker for Stardew, it's also vital that the SMAPI installation be automated as well. Potentially even Content Patcher, but that one is less obvious.

PR #265, opened 2023-04-12
Add `GameStore` to `GameInstallation`

erri120:
Replaces #133

Adds a `GameStore` value to `GameInstallation`, which is needed for #42 and other game support code that requires different functionality depending on the store.

Since this PR changes how `GameInstallation` is serialized, all tests that use pre-exported loadouts will fail. #253 added helpers for testing, which should be used instead of so pre-exported loadouts.


Sewer56:
Oh? An edge case?

Alright; let me throw in a completely unrelated edge case of my own.
There are some games where the expected ***Current Working Directory*** is ***NOT*** the EXE/main binary folder.

Will make issue post sprint; I've a tiny backlog of issues to write.

erri120:
> Another important note, `AGame.PrimaryFile` can't be store agnostic in some cases. For example, Bannerlord differs in the binary path (`bin/Win64_Shipping_Client` on Steam/GOG/Epic vs `bin/Gaming.Desktop.x64_Shipping_Client` on Xbox), so it needs to be removed or replaced with something like
> 
> ```cs
> public abstract GamePath GetPrimaryFile(GameInstallation installation);
> ```

This is also the case for Stardew Valley (#42), the primary file changes depending on operating system **and** store.

Aragas:
As a side note related to the UI, showing the store in the game list would also be helpful

halgari:
This might be fixed after merging #273

erri120:
`NexusMods.Games.BethesdaGameStudio.Tests` and `NexusMods.UI.Tests` are still failing because they use pre-exported loadouts.

erri120:
After #273, #279 and some manual changes, all pre-exported loadouts have been removed, and all tests pass "on my machine".

halgari (approved):

PR #272, opened 2023-04-12
UI consolidation work

halgari:
Consolidates a lot of the styles into organized style sheets. Cropped the logo image so it can be spaced properly. Most of the top bar is now closely tied to the designs and uses the same form of layout.


Sewer56 (approved):
Looks fine to me; I haven't looked over the FIGMA stuff but I assume it's accurate with how it's defined there.

Issue #271, opened 2023-04-12
Deploy is broken on master


halgari:
# Bug Report

## Summary
During my testing recently I modded Cyberpunk, and deployed and ran the game. It deployed, but created the following folders in the game folder:

* `in`
* `rchive`
* `ed4ext`

Calling `Flatten List` on the CLI results in the correct paths of `{Game}\bin\...`, so I'm pretty sure this a deploy issue or perhaps a problem with our paths/filesystem code

halgari:
Well that bug report is false, looking at the installed mod I see this

```
│ Cyberpunk 2077 Files                                                     │ {Game}\archive\pc\content\basegame_4_appearance.archive                 │
│ Thompson 4K-7956-1-0-1680371473.zip                                      │ {Game}\rchive\pc\mod\NPCs-Thompson.archive                              │
│ Cyberpunk 2077 HD Reworked Project Ultra Quality-7652-1-0-fix-1678972038 │ {Game}\rchive\pc\mod\HD Reworked Project.archive                        │
│ (1).zip                                                                  │                                                                         │
│ Dum Dum 4K-8049-1-0-1681170737.zip                                       │ {Game}\rchive\pc\mod\NPCs-DumDum.archive                                │
│ RED4ext-2380-1-11-1-1681244248.zip                                       │ {Game}\in\x64\winmm.dll                                                 │
│ RED4ext-2380-1-11-1-1681244248.zip                                       │ {Game}\ed4ext\RED4ext.dll                                               │
│ Game version 1.62-107-1-24-0-1681322656.zip                              │ {Game}\in\x64\plugins\cyber_engine_tweaks\fonts\NotoSans-Regular.ttf    │
│ Game version 1.62-107-1-24-0-1681322656.zip                              │ {Game}\in\x64\plugins\cyber_engine_tweaks\fonts\NotoSansThai-Regular.tt │
│                                                                          │ f                                                                       │
│ Game version 1.62-107-1-24-0-1681322656.zip                              │ {Game}\in\x64\plugins\cyber_engine_tweaks\fonts\NotoSansMono-Regular.tt │
│                                                                          │ f                                                                       │
│ Game version 1.62-107-1-24-0-1681322656.zip                              │ {Game}\in\x64\plugins\cyber_engine_tweaks\fonts\NotoSansKR-Regular.otf  │
[...]

Issue #266, opened 2023-04-12
Multiple mods in one archive


erri120:
Most of our mod installation code assumes that there is one mod in one archive. The concept of a "mod" is defined by the game support code. I will use #42 as an example:

A SMAPI mod for Stardew Valley is defined by having a `manifest.json` file which points to either an entry DLL, if it's a C# mod, or references some content pack framework, if it's a content pack mod. Regardless of the type of mod, they all have the `manifest.json` file in common. Most archives will have a single manifest file: [Example](https://www.nexusmods.com/stardewvalley/mods/239?tab=files), which the installer will find and correctly process.

It is however possible that a mod author puts multiple SMAPI mods, within one archive: [Example](https://www.nexusmods.com/stardewvalley/mods/3753?tab=files). In this case, the _Nexus Mod_ [Stardew Valley Expanded](https://www.nexusmods.com/stardewvalley/mods/3753) comes with multiple _SMAPI Mods_, a single C# mod and some content pack mods.

[Stardew Valley Expanded](https://www.nexusmods.com/stardewvalley/mods/3753) **requires** all of those _SMAPI Mods_, however it is possible that some _Nexus Mod_ has multiple **optional** _Game Mods_. [Maestros of Synth](https://www.nexusmods.com/cyberpunk2077/mods/3776) by @halgari contains multiple _Cyberpunk Mods_ within one archive that the user can mix and match.

We have to consider the following:

- One archive can contain multiple _Game Mods_.
- If multiple _Game Mods_ are found within one archive, the user should have an option to choose which mods they want to install. It's impossible for the app to know, since installation instructions for these type of mods are usually found on the mod page, instead of having machine-readable instructions like FOMOD.

Instead of having an implementation of `IModInstaller` which looks at the entire archive and returns a list of files to be extracted, we could return a list of _Game Mods_ and their associated files:

```csharp
interface IModInstaller<TGameMod> where TGameMod : IGameMod
{
    IEnumerable<TGameMod> GetGameMods(Archive archive);
}
```

```csharp
interface IGameMod
{
    AModFile[] ModFiles { get; }
}
```

```csharp
record SMAPIMod(AModFile[] ModFiles, SMAPIManifest Manifest) : IGameMod;
```

```csharp
class SMAPIModInstaller : IModInstaller<SMAPIMod>
{
    IEnumerable<SMAPIMod> GetGameMods(Archive archive)
    {
        // finds all manifest.json files
[...]

PR #263, opened 2023-04-11
Stardew Valley support

erri120:
Resolves #42

- adds basic support for Stardew Valley
- mod support:
  - SMAPI mods: working installer and analyzer
  - SMAPI itself: blocked, mainly by #244. I've put TODOs in the code for missing features.
- this PR also updates and changes some misc code:
  - `IEnumerable<T>.TryGetFirst` because `.FirstOrDefault` doesn't work well for value types
  - `RelativePath.DropFirst` now does nothing if called with `n = 0`
  - `UniversalStubbedGameLocator` now accepts a dictionary of "game files" which are put inside the temporary folder to support accessing game files within tests

PR #273, opened 2023-04-12
Rework Cyberpunk tests

halgari:
Reworks the cyberpunk tests so that they use the new testing framework. The deploy bug #271 is due to a bug in `.DropFirst` that @erri120 already found in the SDV code. But I took the time to make these tests more robust by checking the actual path names and not just a count of the files installed.


erri120 (changes requested):
The file `Resources\cyberpunk2077.1.61.zip` should be deleted, since the tests stopped using it. `ModInstallerTests` tests every `IModInstaller` for Cyberpunk, however, we should unit test each installer separately using `AModInstallerTest<TGame, TModInstaller>`.

erri120 (approved):

src/NexusMods.DataModel/ArchiveManager.cs
```diff
@@ -92,6 +92,20 @@ public bool HaveArchive(Hash hash)
...
+    /// <param name="hash"></param>
+    /// <param name="path"></param>
+    /// <returns></returns>
+    public bool TryGetPathFor(Hash hash, out AbsolutePath path)
+    {
+        var rel = NameForHash(hash);
+        path = _locations.Select(r => r.CombineUnchecked(rel))
+            .FirstOrDefault(r => r.FileExists);
+        return path != null;
+    }
```
erri120, line 108: This was an oversight on my end while reviewing the code. `.FirstOrDefault` is notorious for breaking code when the item type of the collection is a value type, because a value type will return `default` instead of `null`. This means that the check `path != null` will **always** be true.

In #263 I added the enumerable extension method `TryGetFirst`, which should be used instead of `FirstOrDefault` for enumerable of collections with value type items. I'll fix this code in my upcoming PR for #266.

PR #280, opened 2023-04-13
Fix trace logging

erri120:
- fixes NLog overwriting log filters
- sets minimum level depending on compile constants:
  - `DEBUG` will enable debug logging
  - `TRACE` will enable trace logging (only for debug builds)
  - else: informational logging


Sewer56 (approved):
LGTM.

One thing to note, we might want to allow the user to override this setting in the future via the config file.

PR #289, opened 2023-04-18
Filesystem everywhere

halgari:
Rework how we use IFileSystem and AbsolutePaths. Due to the way several components work I don't think it's possible to keep both IFileSystem and Absolute paths separate, instead those places where we convert strings to AbsolutePaths are now *required* to supply an IFileSystem. I explained this in the attached ADR. 

I also removed KnownFolders, reworking all the code to go through `IFileSystem.GetKnownFolder`


src/NexusMods.Paths/FileSystemAbstraction/IFileSystem.cs
```diff
@@ -222,6 +222,15 @@ public interface IFileSystem
     /// <returns></returns>
     Task WriteAllTextAsync(AbsolutePath path, string text, CancellationToken cancellationToken = default);
 
+    /// <summary>
+    /// Writes all text to a file using UTF-8 encoding.
+    /// </summary>
+    /// <param name="path">Path to the file.</param>
+    /// <param name="text">Text to write to the file.</param>
+    /// <param name="cancellationToken">Optional <see cref="CancellationToken"/>.</param>
```
erri120, line 230: Should be removed.

tests/NexusMods.Paths.Tests/PathExpansionTests.cs
```diff
@@ -1,15 +1,23 @@
...
+    }
+    
     [Fact]
     public void CanExpandEntryFolder()
     {
         var inputPath = "{EntryFolder}/FluffyKitten.png";
-        var expandedPath = KnownFolders.ExpandPath(inputPath);
-        var expectedPath = Path.Combine(KnownFolders.EntryFolder.GetFullPath(), "FluffyKitten.png");
+        var expandedPath = _fileSystem.ExpandKnownFoldersPath(inputPath);
+        var expectedPath = Path.Combine(_fileSystem.GetKnownPath(KnownPath.EntryDirectory).GetFullPath(), "FluffyKitten.png");
```
erri120, line 20: Tests like these should use `[Theory, AutoFileSystem]` using AutoFixture:

```csharp
[Theory, AutoFileSystem]
public void CanExpandEntryFolder(IFileSystem fs)

// or

[Theory, AutoFileSystem]
public void CanExpandEntryFolder(InMemoryFileSystem fs)
```

erri120 (changes requested):
If we want to keep `IFileSystem` as a field of `AbsolutePath`, _every_ method of `IFileSystem` that operates on a single path, like `OpenFile`, should also have a method in `AbsolutePath` which redirects to `IFileSystem`:

```csharp
interface IFileSystem
{
    //...
    Stream WriteFile(AbsolutePath path);
}

struct AbsolutePath
{
    //...
    public Stream WriteFile() => FileSystem.WriteFile(this);
}
```

erri120:
ARD `0005` is also already taken by #241.

src/NexusMods.Paths/FileSystemAbstraction/IFileSystem.cs
halgari: Why?

halgari: Without some amount of sync files we cause deadlocks in the some small situations where we need to not bounce to another thread while generating a file. Removing this and making `EnsureFiles` use `.Wait()` in "StubbedGame" causes a deadlock on systems (like CI) where we have 2 cores.

erri120: This was regarding the `cacnellationToken` parameter, which is unused for the sync methods. GitHub doesn't like it when someone just selects 1 line and shows multiple, so the context for this was likely weird.

tests/NexusMods.Paths.Tests/PathExpansionTests.cs
halgari: Updated to use Theory and InlineData, good catch

src/NexusMods.Paths/FileSystemAbstraction/IFileSystem.cs
halgari: Ah, makes sense 😆

halgari:
> If we want to keep `IFileSystem` as a field of `AbsolutePath`, _every_ method of `IFileSystem` that operates on a single path, like `OpenFile`, should also have a method in `AbsolutePath` which redirects to `IFileSystem`:
> 
> ```cs
> interface IFileSystem
> {
>     //...
>     Stream WriteFile(AbsolutePath path);
> }
> 
> struct AbsolutePath
> {
>     //...
>     public Stream WriteFile() => FileSystem.WriteFile(this);
> }
> ```

This may be the case in the long run, and we may want to move some helper methods out of FileSystem and into AbsolutePath_IO.cs. That may keep BaseFilesystem a bit cleaner

erri120 (approved):

PR #292, opened 2023-04-20
Add support for multiple mods in one archive

### LINKED ISSUE #266: Multiple mods in one archive

erri120:
Fixes #266.

Changes:

- Renamed `IModInstaller.Priority` to `IModInstaller.GetPriority`.
- Replaced `IModInstaller.GetFilesToExtractAsync` with `IModInstaller.GetModsAsync`. Instead of returning `IEnumerable<AModFile>`, the return type is now `IEnumerable<Mod>`.
- Updated `LoadoutManager` and related to support this change.
- Adds `AModMetadata` abstract class and `GroupMetadata` implementation, which is set by the `LoadoutManager`.
- Updated the SMAPI mod installer to support returning multiple mods.

https://user-images.githubusercontent.com/16577545/233375529-2821dc56-7aef-49ec-9bab-227433462c6e.mp4

The changed made to `LoadoutManager` added some complexity to account for different possibilities. The indent implementation of `IModInstaller.GetModsAsync` can be found in `StubbedGameInstaller`:

https://github.com/erri120/NexusMods.App/blob/f1595db7b9d5a493788cc365512f89c786d55d2b/tests/NexusMods.StandardGameLocators.TestHelpers/StubbedGames/StubbedGameInstaller.cs#L29-L63

The method `GetModsAsync` has a `Mod baseMod` parameter which should be copied. In this simple example, where only one mod is returned, the `LoadoutManager` sees that the mod identifier didn't change, and alters the registry by updating the existing mod in the loadout.

However, if multiple mods are returned, the installer has to assign new identifiers as seen in the `SMAPIModInstaller`:

https://github.com/erri120/NexusMods.App/blob/f1595db7b9d5a493788cc365512f89c786d55d2b/src/Games/NexusMods.Games.StardewValley/Installers/SMAPIModInstaller.cs#L103-L109

If multiple mods are returned, the original base mod, assigned to the archive, must be removed. The video shows this happening in real time:

1) the user clicks `Add Mod` and selects an archive
2) `LoadoutManager` adds a new `Mod` with `id=foo` and `status=Installing` to the loadout
3) the UI reflects this loadout change by displaying a new entry in the table and an installation spinner
4) `SMAPIModInstaller` returns three new instances of `Mod`, all with new ids
5) `LoadoutManager` adds those new mods to the loadout
6) `LoadoutManager` removes `Mod` with `id=foo` because it wasn't returned by the installer
7) the UI updates and shows the three new mods

### DIFF
--- a/src/Games/NexusMods.Games.BethesdaGameStudios/Installers/SkyrimInstaller.cs
+++ b/src/Games/NexusMods.Games.BethesdaGameStudios/Installers/SkyrimInstaller.cs
@@ -48,7 +48,7 @@ public class SkyrimInstaller : IModInstaller
     public Priority GetPriority(GameInstallation installation, EntityDictionary<RelativePath, AnalyzedFile> archiveFiles)
     {
         if (installation.Game is not SkyrimSpecialEdition)
-            return Common.Priority.None;
+            return Priority.None;
 
         // Determine one of the following:
         // - There is a 'meshes' folder with NIF file.
@@ -65,14 +65,14 @@ public class SkyrimInstaller : IModInstaller
             {
                 var subDirectory = rawPath.AsSpan(separatorIndex + 1);
                 if (AssertPathForPriority(subDirectory))
-                    return Common.Priority.Normal;
+                    return Priority.Normal;
             }
 
             if (AssertPathForPriority(rawPath))
-                return Common.Priority.Normal;
+                return Priority.Normal;
         }
 
-        return Common.Priority.None;
+        return Priority.None;
     }
 
     public ValueTask<IEnumerable<Mod>> GetModsAsync(
--- a/src/Games/NexusMods.Games.BethesdaGameStudios/Installers/SkyrimInstaller.cs
+++ b/src/Games/NexusMods.Games.BethesdaGameStudios/Installers/SkyrimInstaller.cs
@@ -45,7 +45,7 @@ public class SkyrimInstaller : IModInstaller
         _dataStore = dataStore;
     }
 
-    public Priority Priority(GameInstallation installation, EntityDictionary<RelativePath, AnalyzedFile> files)
+    public Priority GetPriority(GameInstallation installation, EntityDictionary<RelativePath, AnalyzedFile> archiveFiles)
     {
         if (installation.Game is not SkyrimSpecialEdition)
             return Common.Priority.None;
@@ -54,7 +54,7 @@ public class SkyrimInstaller : IModInstaller
         // - There is a 'meshes' folder with NIF file.
         // - There is a 'textures' folder with DDS file.
         // - There is a folder (max 1 level deep) with either of the following cases.
-        foreach (var file in files)
+        foreach (var file in archiveFiles)
         {
             var path = file.Key;
             var rawPath = file.Key.Path;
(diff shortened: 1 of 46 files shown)


halgari (approved):

Aragas:
I'm experimenting with the PR, there's a possible behavior issue. Because the extensions now control the `Mod` record, we could potentially set the `Metadata` field manually, only for it to be overriden by `LoadoutManager.InstallModsFromArchiveAsync`

erri120:
> I'm experimenting with the PR, there's a possible behavior issue. Because the extensions now control the `Mod` record, we could potentially set the `Metadata` property manually, only for it to be overriden by `LoadoutManager.InstallModsFromArchiveAsync` UPDATE: There are also other properties that could be set, like `Status`

Yup, that's another issue this solution has. I think it actually might be better to return a different type. Instead of returning `Mod`, we can return `ModToInstall`, which just has a few properties that the mod installer can set. `LoadoutManager` will then just map `ModToInstall` to a `Mod`.

This solution is much cleaner for mod installer implementations, @halgari.

erri120:
Updated `IModInstaller` to return `ModInstallerResult`.

halgari (approved):

Issue #258, opened 2023-04-06
Display current/peak download speed


Greg-Nexus:
# Story

## User Story

As a user, I want to view the current and recent speed of my downloads

## Outline
Add the following functionality to the 'In Progress' downloads page:

Display a 'Speed' section, above the 'In progress' section, with the following data:

- Recent download speed, displayed as a line graph
- Recent peak download speed (as MB/s)
- Current download speed (as MB/s)

The 'Speed' section should be collapsible/expandable.

This should not cover (these will be done iteratively in future tasks):

- Building the 'Completed' page
- Building the 'Download history' page

## Supporting Information
See the epic for designs and context: https://github.com/Nexus-Mods/NexusMods.App/issues/180

PR #293, opened 2023-04-20
Structure and layout of the download screen

halgari:
Implements the structure and layout of the downloads pages. Adds a spine button (on the far left) for a download page, but the button is purely a placeholder as that will implemented as part of #258 

Toolbars are implemented as a small amount of XAML structure, combined with styles in `ToolbarStyles.xaml`

Reworked the toolbar (unified the code with the other toolbars) in the Loadout screen deleting the files we no longer needed. 

Implemented tests for all the new VMs and views

NexusMods App_VMH473A9dL


src/NexusMods.App.UI/ViewModels/Helpers/ViewModelSelector/ViewModelDesignSelector.cs
```diff
@@ -0,0 +1,33 @@
...
+public class ViewModelDesignSelector<TEnum, TVmType, TBase> : 
+    AViewModelSelector<TEnum, TVmType, TBase>
+    where TVmType : class, IViewModelInterface
+    where TEnum : struct, Enum {
+    private static readonly Dictionary<TEnum,Type> Mappings;
+
+    static ViewModelDesignSelector()
+    {
+        Mappings = AViewModelAttribute.GetAttributes<TEnum>();
+    }
```
erri120, line 23: The base class `AViewModelSelector` already has a protected field called `Mappings`. These lines (L18-L23) can be removed.

erri120 (approved):

Issue #291, opened 2023-04-19
Incosistent updates when using .Alter


halgari:
# Bug Report

## Summary
The whole idea of the immutable IDataStore is dependent on having atomic updates to roots. This model is heavily based on the datamodel found in [Clojure](https://clojure.org). And a example of this is the .Alter method in LoadoutRegistry:

```
registry.Alter(loadoutId, loadout => loadout with {Name = "New Name"});
```

This function maintains consistency by having "at least once" run semantics. The function passed to `.Alter` may be run one or more times depending if there is a conflict. The issue is that currently in the code, the alter function never returns, because `IDataStore.PutRoot` always returns successful, even if the supplied IDs don't match. 

To solve this we should update `.PutRoot` to perform its work with [CAS](https://en.wikipedia.org/wiki/Compare-and-swap)-like semantics. If the ID used to generate the new loadout doesn't match what's in the datastore, the put should fail, and the update should be rolled back.

## Steps to reproduce
Open the UI, create a loadout, add multiple mods at once, watch as everything becomes broken and inconsistent

PR #294, opened 2023-04-20
Implement CAS support for the Datastore in LoadoutRegistry

halgari:
This implementes #291 and a working CAS on the datastore via SQLite Transactions


tests/NexusMods.DataModel.Tests/DataStoreTests.cs
```diff
@@ -33,13 +33,50 @@ public void CanGetAndSetHashedValues()
...
             To = new GamePath(GameFolderType.Game, "test.foo")
         }.WithPersist(DataStore);
         foo.DataStoreId.ToString().Should().NotBeEmpty();
         DataStore.Get<FromArchive>(foo.DataStoreId).Should().NotBeNull();
     }
 
+    [Fact]
+    public void CompareAndSwapIsAtomic()
+    {
+        const int numThreads = 10;
```
erri120, line 46: We should limit the number of threads using `Environment.ProcessorCount`. Hardcoding the number of threads might produce weird results in CI, which is often very limited.

erri120 (approved):

tests/NexusMods.DataModel.Tests/DataStoreTests.cs
Sewer56: Something small to add here.
`Environment.ProcessorCount` is ideal, but some lower bound should probably also be established. The test would be moot if by some chance it ran in a single threaded environment.

IIRC GitHub Actions has 2 threads on Windows/Linux runners and 3 on OSX.

halgari: The point here is to slam the method so we get wildly inconsistent values if the function isn't atomic, but we can probably get away with `Math.Min(4, Environment.ProcessorCount * 2)`

PR #296, opened 2023-04-24
High performance logging

erri120:
### Problem

We're using `Microsoft.Extensions.Abstractions.Logging` for all of our logging. We're also producing a lot of _trace_ logs in hot paths of our app. I fixed the issue of always logging traces in #280, which indirectly helped with performance (because the code no longer writes to console), however there is still performance to be gained.

The `ILogger` extension methods `LogTrace`, `LogDebug` and so on, all use `param object[]`, which **always** allocates a new array, even if we disabled trace logging. Furthermore, value types, which we use everywhere, get boxed to `object`.

### Solution

Microsoft provides [guidelines](https://learn.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging) for high performance logging using the [`LoggerMessage`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggermessage) class, which exposes functionality to create cacheable delegates that require fewer object allocations and reduced computational overhead compared:

```csharp
private static readonly Action<ILogger, WorkItem, Exception> s_processingPriorityItem = LoggerMessage.Define<WorkItem>(
    LogLevel.Information,
    new EventId(1, nameof(PriorityItemProcessed)),
    "Processing priority item: {Item}");

public static void PriorityItemProcessed(this ILogger logger, WorkItem workItem)
{
    if (logger.IsEnabled(LogLevel.Information)
    {
        s_processingPriorityItem(logger, workItem, default!)
    }
}
```

This prevents an array allocation for `params object[]`, removes boxing allocation since it's using generics, and doesn't do additional work, if the log level isn't even enabled.

Instead of writing this by hand, we can also make use of [source generators](https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator):

```csharp
    [LoggerMessage(
        EventId = 0,
        Level = LogLevel.Trace,
        Message = "Job {jobId} progress is {percent}")]
    public static partial void JobProgress(
        this ILogger logger,
        JobId jobId,
        Percent percent);
```

This generates the same code as before:

```csharp
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "7.0.7.1805")]
        private static readonly global::System.Action<global::Microsoft.Extensions.Logging.ILogger, global::NexusMods.DataModel.Interprocess.Jobs.JobId, global::NexusMods.DataModel.RateLimiting.Percent, global::System.Exception?> __JobProgressCallback =
            global::Microsoft.Extensions.Logging.LoggerMessage.Define<global::NexusMods.DataModel.Interprocess.Jobs.JobId, global::NexusMods.DataModel.RateLimiting.Percent>(global::Microsoft.Extensions.Logging.LogLevel.Trace, new global::Microsoft.Extensions.Logging.EventId(0, nameof(JobProgress)), "Job {jobId} progress is {percent}", new global::Microsoft.Extensions.Logging.LogDefineOptions() { SkipEnabledCheck = true }); 

        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "7.0.7.1805")]
        public static partial void JobProgress(this global::Microsoft.Extensions.Logging.ILogger logger, global::NexusMods.DataModel.Interprocess.Jobs.JobId jobId, global::NexusMods.DataModel.RateLimiting.Percent percent)
        {
            if (logger.IsEnabled(global::Microsoft.Extensions.Logging.LogLevel.Trace))
            {
                __JobProgressCallback(logger, jobId, percent, null);
            }
        }
```

### Benchmarks

I've written a benchmark to showcase the difference in performance:

|                       Method | MinimumLogLevel |      Mean |     Error |    StdDev |   Gen0 | Allocated |
|----------------------------- |---------------- |----------:|----------:|----------:|-------:|----------:|
|       Default_ReferenceTypes |           Trace | 43.057 ns | 0.0488 ns | 0.0408 ns | 0.0005 |      40 B |
|           Default_ValueTypes |           Trace | 49.025 ns | 0.1098 ns | 0.0973 ns | 0.0012 |     104 B |
| LoggerMessage_ReferenceTypes |           Trace | 11.080 ns | 0.1052 ns | 0.0984 ns |      - |         - |
|     LoggerMessage_ValueTypes |           Trace | 11.493 ns | 0.0898 ns | 0.0840 ns |      - |         - |
|       Default_ReferenceTypes |     Information | 42.167 ns | 0.1096 ns | 0.1025 ns | 0.0005 |      40 B |
|           Default_ValueTypes |     Information | 48.945 ns | 0.0987 ns | 0.0923 ns | 0.0012 |     104 B |
| LoggerMessage_ReferenceTypes |     Information |  1.728 ns | 0.0025 ns | 0.0021 ns |      - |         - |
|     LoggerMessage_ValueTypes |     Information |  1.562 ns | 0.0150 ns | 0.0140 ns |      - |         - |

The `Default_*` benchmarks used the `LogTrace` extension methods, while the `LoggerMessage_*` benchmarks use the compiled custom logging methods. The minimum logging level was set to `Trace` for the first run, and `Information` in the second run.

The table shows a ~4x speed improvement and no memory allocations, if the message should be logged. If the minimum logging level is higher than `Trace`, the compiled logging methods don't do anything and exit early, resulting in a ~43x speed improvement.

### Conclusion

Compiled logging methods using `LoggerMessage` should be used in critical hot paths, where we do a lot of trace and debug logging. I've updated the `InjectedViewLocator`, `SqliteIPC` and `SqliteDataStore` classes to use this kind of logging, because they produce 99% of all logs.

### Problem

On the topic of logging, `SqliteIPC` logs massive amounts of `IId` values. The `ToString` implementation of `AId` does the following:

```csharp
Span<byte> span = stackalloc byte[SpanSize];
ToSpan(span);
return $"{Category}-{Convert.ToHexString(span)}";
```

The `ToSpan` implementations are already optimized, however the string creation is not. C# enums are very weird sometimes and have massive performance issues when calling stringifying them. They are using reflection behind the scenes and have to account for "missing" enum values. In our case, we know the values and we know the string values.

### Solution

[`NetEscapades.EnumGenerators`](https://github.com/andrewlock/NetEscapades.EnumGenerators) is a source generator that produces `ToStringFast` extension methods for enums with the `EnumExtension` attribute:

```csharp
        public static string ToStringFast(this NexusMods.DataModel.Abstractions.EntityCategory value)
            => value switch
            {
                NexusMods.DataModel.Abstractions.EntityCategory.Loadouts => nameof(NexusMods.DataModel.Abstractions.EntityCategory.Loadouts),
                NexusMods.DataModel.Abstractions.EntityCategory.FileAnalysis => nameof(NexusMods.DataModel.Abstractions.EntityCategory.FileAnalysis),
                NexusMods.DataModel.Abstractions.EntityCategory.LoadoutRoots => nameof(NexusMods.DataModel.Abstractions.EntityCategory.LoadoutRoots),
                NexusMods.DataModel.Abstractions.EntityCategory.FileHashes => nameof(NexusMods.DataModel.Abstractions.EntityCategory.FileHashes),
                NexusMods.DataModel.Abstractions.EntityCategory.FileContainedIn => nameof(NexusMods.DataModel.Abstractions.EntityCategory.FileContainedIn),
                NexusMods.DataModel.Abstractions.EntityCategory.AuthData => nameof(NexusMods.DataModel.Abstractions.EntityCategory.AuthData),
                NexusMods.DataModel.Abstractions.EntityCategory.TestData => nameof(NexusMods.DataModel.Abstractions.EntityCategory.TestData),
                _ => value.ToString(),
            };
```

### Benchmarks

|                 Method |     Mean |    Error |   StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
|----------------------- |---------:|---------:|---------:|------:|-------:|----------:|------------:|
|                    Old | 76.17 ns | 0.132 ns | 0.117 ns |  1.00 | 0.0018 |     152 B |        1.00 |
| New_WithEnumExtensions | 32.81 ns | 0.575 ns | 0.538 ns |  0.43 | 0.0015 |     128 B |        0.84 |

The `New_WithEnumExtensions` benchmark uses `Category.ToStringFast` instead of `Category.ToString`. This single change basically doubles the performance and also reduces memory allocations.

### Conclusion

The default implementation of `ToString` for enums is horrible and should not be used in hot paths. I believe we can further improve the `ToString` method of `AId` by using `string.Create` or changing how we represent the ids in general.

### DIFF
--- a/benchmarks/NexusMods.Benchmarks/Benchmarks/HighPerformanceLogging.cs
+++ b/benchmarks/NexusMods.Benchmarks/Benchmarks/HighPerformanceLogging.cs
@@ -1,0 +1,98 @@
+using BenchmarkDotNet.Attributes;
+using Microsoft.Extensions.Logging;
+using NexusMods.Benchmarks.Interfaces;
+
+namespace NexusMods.Benchmarks.Benchmarks;
+
+[MemoryDiagnoser]
+[BenchmarkInfo("High Performance Logging", "Compares the different methods of logging")]
+public class HighPerformanceLogging : IBenchmark
+{
+    public const string LoggingMessageWithReferenceTypes = "This uses reference types: {a} and {b}";
+    public const string LoggingMessageWithValueTypes = "This uses value types: {a} and {b}";
+
+    [Params(LogLevel.Trace, LogLevel.Information)]
+    public LogLevel MinimumLogLevel { get; set; }
+
+    private ILogger _logger;
+    private string _referenceA;
+    private string _referenceB;
+    private Guid _valueA;
+    private Guid _valueB;
+
+    [GlobalSetup]
+    public void Setup()
+    {
+        _logger = new DummyLogger(MinimumLogLevel);
+        _valueA = Guid.NewGuid();
+        _valueB = Guid.NewGuid();
+        _referenceA = _valueA.ToString();
+        _referenceB = _valueB.ToString();
+    }
+
+    [Benchmark]
+    public void Default_ReferenceTypes()
+    {
+        _logger.LogTrace(LoggingMessageWithReferenceTypes, _referenceA, _referenceB);
+    }
+
+    [Benchmark]
+    public void Default_ValueTypes()
+    {
+        _logger.LogTrace(LoggingMessageWithValueTypes, _valueA, _valueB);
+    }
+
+    [Benchmark]
+    public void LoggerMessage_ReferenceTypes()
+    {
+        _logger.WithReferenceTypes(_referenceA, _referenceB);
+    }
+
+    [Benchmark]
+    public void LoggerMessage_ValueTypes()
+    {
+        _logger.WithValueTypes(_valueA, _valueB);
+    }
+
+    private class DummyLogger : ILogger
+    {
+        private readonly LogLevel _minLogLevel;
+
+        public DummyLogger(LogLevel minLogLevel)
+        {
+            _minLogLevel = minLogLevel;
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state,
+            Exception? exception, Func<TState, Exception?, string> formatter) { }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return _minLogLevel <= logLevel;
+        }
+
+        public IDisposable? BeginScope<TState>(TState state)
+            where TState : notnull => null;
+    }
+}
+
+internal static partial class LoggingExtensions
+{
+    [LoggerMessage(
+        EventId = 0,
+        Level = LogLevel.Trace,
+        Message = HighPerformanceLogging.LoggingMessageWithReferenceTypes)]
+    public static partial void WithReferenceTypes(
+        this ILogger logger,
+        string a,
+        string b);
+
+    [LoggerMessage(
+        EventId = 0,
+        Level = LogLevel.Trace,
+        Message = HighPerformanceLogging.LoggingMessageWithValueTypes)]
+    public static partial void WithValueTypes(
+        this ILogger logger,
+        Guid a,
+        Guid b);
+}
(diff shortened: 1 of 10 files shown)


Sewer56 (approved):
Looks good to me; this is pretty much how I'd do it in order to get a good tradeoff between simplicity and performance.

I like the use of source generated enum names; this is a trick I've myself previously employed


The main takeaway here is actually removed allocations; in the long run allocations leading to gen0 GC hurt much more than the minimal nanoseconds we shave.

Good stuff.

PR #302, opened 2023-04-26
Launch button fixes

halgari:
* Removes some unused button styles, and moves the ones we do use into a separate stylesheet
* Reworks IPC Jobs so that they're all based on Entity and support polymorphism, so we can filter jobs by C# type
* Reworks the Launch button to work with the new styles and IPC Jobs
* Writes up a simple "launch game" logic to the button
* Implements unit tests on the LaunchButton classes


halgari:
Got the tests to pass by disabling parallel execution, doesn't look like it made much of a impact on test run times, so we'll keep that disabled for now and see what happens long-term. It's the only thing I could think of as it runs fine in Rider, which has different parallelization semantics.

erri120 (approved):

PR #304, opened 2023-05-02
Add `IOSInformation`

### LINKED ISSUE #287: OS information abstraction
Similar to #176, certain components have different implementations depending on the operating system. This is currently handled via [`RuntimeInformation.IsOSPlatform`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.isosplatform) or [`OperatingSystem.IsLinux`](https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem.islinux) and other functions of the [`OperatingSystem`](https://learn.microsoft.com/en-us/dotnet/api/system.operatingsystem) class.

Game support components, like Stardew Valley, use a different primary file, depending on the operating system:

https://github.com/Nexus-Mods/NexusMods.App/blob/797c34479617e5b9b9d27ecd35a0d9a9c392749f/src/Games/NexusMods.Games.StardewValley/StardewValley.cs#L33-L52

This implementation will break Wine installation, since `RuntimeInformation.IsOSPlatform(OSPlatform.Linux)` will return `true` on Linux, even if you wanted to run the game inside a Wine prefix.

Instead of relying on information provided by the .NET runtime, having an abstraction over the operating system related information would be better for Wine support and maybe even enable testing Windows-only code on Linux:

```csharp
interface IOSInformation
{
    bool IsWindows { get; }

    bool IsLinux { get; }

    bool IsOSX { get; }

    bool IsWine { get; }
}
```

```csharp
private readonly IOSInformation _osInfo;

public override GamePath PrimaryFile
{
    get
    {
        if (_osInfo.IsLinux)
        {
            //...
        }

        if (_osInfo.IsWindows)
        {
            //...
        }

        if (_osInfo.IsOSX)
        {
            //...
        }

        throw new PlatformNotSupportedException();
}
```

Injecting the correct instance into the components that need it, is a similar problem to injecting the right file system (#270).

erri120:
Resolves #287.

This adds a new interface `IOSInformation` and the implementation `OSInformation` to `NexusMods.Common`. It is similar to `IFileSystem` from #176, in that it abstracts information about the host.

`IOSInformation` provides the following properties and methods:

- `OSPlatform Platform`: returns the current platform using the [OSPlatform](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.osplatform) struct
- `bool IsWindows`/`isLinux`/`isOSX`: check the current platform
- `bool IsPlatformSupported`: checks if the platform is supported
- `void PlatformSupportedGuard`: guard clause that throws an exception when the platform is not supported

The most interesting methods are `MatchPlatform<TOut>` and `SwitchPlatform`, which provide consumers an easy way to "match" or "switch" based on the current platform. `MatchPlatform` is used if you want to return something, and `SwitchPlatform` is used if you don't need a return value:

```csharp
return _osInformation.MatchPlatform(
    state: ref store,
    onWindows: (ref GameStore gameStore) =>
        gameStore == GameStore.XboxGamePass
            ? new GamePath(GameFolderType.Game, "Stardew Valley.exe")
            : new GamePath(GameFolderType.Game, "StardewModdingAPI.exe"),
    onLinux: (ref GameStore _) => new GamePath(GameFolderType.Game, "StardewValley"),
    onOSX: (ref GameStore _) => new GamePath(GameFolderType.Game, "Contents/MacOS/StardewValley")
);
```

I've also added an overload to both of those methods that accepts a `TState` parameter to prevent delegate allocations. The cost of these allocations can be seen in this benchmark I've added:

|                     Method |      Mean |     Error |    StdDev |   Gen0 | Allocated |
|--------------------------- |----------:|----------:|----------:|-------:|----------:|
|      NoState_ReferenceType |  9.446 ns | 0.1225 ns | 0.1086 ns | 0.0008 |      64 B |
|          NoState_ValueType | 10.237 ns | 0.2211 ns | 0.2952 ns | 0.0008 |      64 B |
| WithRefState_ReferenceType |  7.125 ns | 0.2163 ns | 0.3170 ns |      - |         - |
|     WithRefState_ValueType |  1.499 ns | 0.0031 ns | 0.0027 ns |      - |         - |

Matching or switching the current platform is currently not used in any critical paths, so these improvements are insignificant at best. However, going forward we should design APIs with this in mind and optionally provide an overload that accepts a "state" parameter.


Sewer56 (approved):
LGTM

PR #305, opened 2023-05-03
Use VSTest integration instead of MSBuild

erri120:
VSTest integration is [preferred](https://github.com/coverlet-coverage/coverlet#vstest-integration-preferred-due-to-known-issue) over MSBuild. This might help stabilize the CI tests a bit.

### DIFF
--- a/.github/workflows/clean_environment_tests.yaml
+++ b/.github/workflows/clean_environment_tests.yaml
@@ -34,8 +34,8 @@ jobs:
       #  run: dotnet clean
       
       - name: Tests
-        run: dotnet test --filter "RequiresNetworking!=True" /p:CollectCoverage=true /p:CoverletOutputFormat=opencover
-      
+        run: dotnet test --filter "RequiresNetworking!=True" --collect:"XPlat Code Coverage;Format=opencover"
+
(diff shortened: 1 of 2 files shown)


Sewer56 (approved):

Sewer56:
Giving it the ok, since I originally reported this in the internal Slack channel.

Sewer56:
Before merging just doubly make sure Codecov can pick up the newly made coverage files from CI logs. 

This approach will actually yeet them in a subfolder with a random GUID and I've never used Codecov without explicitly setting the XML paths (our config uses default setting of auto detection), so I don't know if it recursively scans directories.

erri120:
> Before merging just doubly manually make sure Codecov can pick up the newly made coverage files from CI logs. Seems like it did though.
> 
> This approach will actually yeet them in a subfolder with a random GUID and I've never used Codecov without explicitly setting the XML paths (our config uses default setting of auto detection), so I don't know if it recursively scans directories.

https://github.com/Nexus-Mods/NexusMods.App/pull/305#issuecomment-1532728063 seems to indicate that it picked up everything correctly.

Issue #299, opened 2023-04-25
As an extension developer, I want to set sorting rules


Aragas:
# User story

**As an** extension developer

**I want to** set sorting rules

**So that** the mod list is automatically sorted

The `Mod` class exposes a `SortRules` property that can't be accessed currently.  
One of the more logical methods where the sorting rules could be set is `ValueTask<IEnumerable<ModInstallerResult>> GetModsAsync`, but as a PoC I already tried it before #292 was merged and `SortingRules` could be set, but I had an unresolved issue.  
I need to be able to alias a `ModId` to some arbitrary mod id that the game uses.  
For example, in Bannerlord a mod can have it's own internal Module Id if a `SubModule.xml` (a mod manifest) is present. If it's present, I want to somehow alias this Module Id to a ModId, so I can resolve it from NMA data storage when I parse the manifest file and create the sorting rules. There also is another issue. if, for some reason, the alias is missing (because the mod is not yet parsed by NMA), I won't be able to add the sorting rule. This means that there might be needed a two pass system, when first all aliases are added to the data store, then the sorting rules could be applied.

erri120:
`Mod.SortRules` is not really useable right now. The Bethesda support code is the only place where we employ using, however, this is done to generate the `plugins.txt` file and doesn't use `ModId`.

For now, I'd recommend implementing an `IFileAnalyzer` and a custom `IFileAnalysisData` type that contains dependency information.

Aragas:
> For now, I'd recommend implementing an IFileAnalyzer and a custom IFileAnalysisData type that contains dependency information.

Already done! Waiting then for the abstraction to be ready for use

Al12rs:
Would this be exclusively for game extensions? I'm still not sure about what kind of extensions NMA will support in the future.

How to best handle a case where multiple extensions want to define the order of the mods?
In MO2, extensions can manipulate the priority of mods through an interface, but not to the scope of the ESP list (bethesda plugins), where the game extension has the control and the responsibility of defining the starting order. The ESPList is still exposed to other extensions through an interface, though the initial order is defined by the game extension.

Since the EspList would be a game specific feature it makes sense that the game extension would have full control over it, while mods are a common concept for all games. Mod order might not even be a valid concept for some games. 
[...]

PR #306, opened 2023-05-04
Transformer hook system for the backend

halgari:
This is somewhat like Event Hooks or MediatR, but different in that the result of each hook is fed into the next hook. 

The process is fairly simple, you register for one of the transformer hook, in the DI system and then the given method is called at certain parts of the backend process. We will extend this in the future as various games need hooks for parts of the apply process. For now we can insert logic into the various parts of the sorting system which should solve #299

### DIFF
--- a/src/NexusMods.DataModel/TransformerHooks/IBeforeMakeApplyPlan.cs
+++ b/src/NexusMods.DataModel/TransformerHooks/IBeforeMakeApplyPlan.cs
@@ -1,0 +1,23 @@
+﻿using NexusMods.DataModel.Loadouts;
+using NexusMods.Paths;
+
+namespace NexusMods.DataModel.TransformerHooks;
+
+/// <summary>
+/// Called after the flattening of mods into a single collection but before the
+/// creation of the apply plan.
+/// </summary>
+public interface IBeforeMakeApplyPlan
+{
+    /// <summary>
+    /// This method is executed after the flattening of mods into a single collection but before the
+    /// creation of the apply plan.
+    /// </summary>
+    /// <param name="files"></param>
+    /// <param name="mods"></param>
+    /// <param name="loadout"></param>
+    /// <param name="ct"></param>
+    /// <returns></returns>
+    public ValueTask<Dictionary<GamePath, (AModFile File, Mod Mod)>> BeforeMakeApplyPlan(Dictionary<GamePath, (AModFile File, Mod Mod)> files, Loadout loadout, CancellationToken ct);
+    
+}
--- a/src/NexusMods.CLI/Verbs/FlattenList.cs
+++ b/src/NexusMods.CLI/Verbs/FlattenList.cs
@@ -18,7 +18,7 @@ public class FlattenList : AVerb<LoadoutMarker>
     public async Task<int> Run(LoadoutMarker loadout, CancellationToken token)
     {
         var rows = new List<object[]>();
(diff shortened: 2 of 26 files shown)


src/NexusMods.DataModel/TransformerHooks/IBeforeMakeApplyPlan.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// <summary>
+    /// This method is executed after the flattening of mods into a single collection but before the
+    /// creation of the apply plan.
+    /// </summary>
+    /// <param name="files"></param>
+    /// <param name="mods"></param>
+    /// <param name="loadout"></param>
+    /// <param name="ct"></param>
+    /// <returns></returns>
+    public ValueTask<Dictionary<GamePath, (AModFile File, Mod Mod)>> BeforeMakeApplyPlan(Dictionary<GamePath, (AModFile File, Mod Mod)> files, Loadout loadout, CancellationToken ct);
```
erri120, line 21: Inconsistent naming, should be `BeforeMakeApplyPlanAsync`.

erri120 (changes requested):
The only issue I have with this design is that we need to properly document every hook. Instead of guessing which hook you'd want for your specific purpose, we should have some docs that list all hooks and explains what they do, maybe even when you should use them.

The API of these hooks could also be improved. Ideally, the methods on the interface should provide a read-only view of the data. This is already done with the `IAfterSort.AfterSortAsync` and `IBeforeSort.BeforeSortAsync` methods, which accept and return an `IEnumerable<Mod>`.

`IAfterFlatten.AfterFlattenAsync`, `IBeforeMakeApplyPlan.BeforeMakeApplyPlan.` instead provide and return a dictionary, which can be changed by the implementation at will. `GenerateFiles` actually showcases this and is a good example for the issues this API has:

- The implementation has to create a new `List<(AModFile, Mod)>` for files that get generated during this step.
- The provided dictionary gets updated while it's being iterated on, requiring the use of `.ToList()` to prevent any iteration issues, this is slow and allocates more memory than it should.
- The implementation wants to add files and has to manually call the `LoadoutManager` to add these files to the loadout.

`GenerateFiles` is supposed to "generate[s] files for mods that have a zero hash", and as such its only concern should be iterating over the existing files, finding the files with a zero hash, and ideally returning updated files.

The complexity should not be in the implementations of these hooks, but in the `TransformerRegistry`. Ideally, the APIs accept a read-only view over the data and only return the changes (added, updated, removed). The `TransformerRegistry` should use these changes to properly update the loadout.

These issues are similar to the ones encountered while working on #292, where complexity was added to the implementations instead of the manager, however I changed this to move the complexity to the manager, leaving the implementations simpler.

halgari:
These are good points, and something I'll consider over the weekend. My plan here was to be as open-ended in the design as possible. Returning a collection allows the callee to change the input collection in any way they want. But that does make it easier to construct buggy transformers. 

Improving these interfaces is fairly simple in several cases such as the file replacement stage. But in other case like the sorting I need to think about how to craft the API. In those cases we probably want the transformers to return ISortRules instead of allowing full access to change the mod ordering however they see fit.

In addition, for performance reasons we probably want each transformer to declare their game and file filters, instead of doing a `O(transformers * files)` scan on every use.

halgari:
Closing as this work is being done in another PR/branch

Issue #221, opened 2023-03-22
Identify solution for e2e tests on the frontend


Greg-Nexus:
## We want
We should make sure that the frontend of the app is tested automatically by e2e tests. This will reduce the need for manual testing of features or improvements on the app.

## Supporting Information
Investigate some possible solutions, on the website we use Playwright.

halgari:
Closely related to, but perhaps not a duplicate of #200

PR #307, opened 2023-05-04
End-to-End testing framework

halgari:
For #221, implements a full-app testing system for end to end testing. The created window uses a real MainWindow, real MainWindowViewModel and all the associated services. Routes the datamodel folders to a temporary location so tests start clean every time, and also implements a way to save screenshots of the headless renderer the window runs in. 

In the future we may want to upload these screenshots as part of the tests so we can verify that the code looks as expected. But this is beyond the scope of this ticket.

### DIFF
--- a/tests/NexusMods.UI.Tests/AEndToEndTest.cs
+++ b/tests/NexusMods.UI.Tests/AEndToEndTest.cs
@@ -1,0 +1,29 @@
+﻿using NexusMods.UI.Tests.Framework;
+
+namespace NexusMods.UI.Tests;
+
+/// <summary>
+/// A base test class for a test that needs the entire app, the MainWindow and a full View Model
+/// </summary>
+public class AEndToEndTest : IAsyncLifetime
+{
+    private readonly AvaloniaApp _app;
+    
+    protected WindowHost? Host { get; private set; }
+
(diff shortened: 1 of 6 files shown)


tests/NexusMods.UI.Tests/AEndToEndTest.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace NexusMods.UI.Tests;
+
+/// <summary>
+/// A base test class for a test that needs the entire app, the MainWindow and a full View Model
+/// </summary>
+public class AEndToEndTest : IAsyncLifetime
+{
+    private readonly AvaloniaApp _app;
+    
+    protected WindowHost? Host { get; private set; }
```
erri120, line 12: Since we're setting this value in `IAsyncLifetime.InitializeAsync`, we can remove nullability from this property:

```csharp
private WindowHost? _host;

// either get fancy with Debug.Assert
public WindowHost Host
{
    get
    {
        Debug.Assert(_host is not null);
        return _host;
    }
}

// or just throw an exception
public WindowHost Host => _host ?? throw new UnreachableException();
```

erri120 (changes requested):

erri120 (approved):

Issue #252, opened 2023-04-06
Component - Sidebar button, "Download" variant


captainsandypants:
### Name
Sidebar button - "Download speed" variant
Sidebar button - "Time remaining" variant

### Purpose

- Button used to navigate to the download section.  
- Button displays the current download speed as well as a total download progress ring in the style of a border.  
- When user is on the download section and so the download button is active , it can be clicked again to change from displaying speed to total download time remaining.  

### Notes
The switch from speed to time remaining will be a right click option in the long term, as well as adjustable in the settings.
More variants will be created for this button in the future i.e. a variant to show when collection download is paused until an ad is watched.

## Related tickets
https://github.com/Nexus-Mods/NexusMods.App/issues/180 - As a user, I want to download a mod

## Design

https://www.figma.com/file/8pjtQeNggvVi7RWoLNGV80/%F0%9F%A7%B0-Nexus-Mods-Design-System?node-id=6146%3A21810&t=hdwqCzigG5cLUUnV-1

![image](https://user-images.githubusercontent.com/118466382/230358742-a7b5f87f-7156-4bec-bea5-f5b1a9e4c84c.png)

PR #311, opened 2023-05-09
Spine Download Button

halgari:
Implements the spine download button as per #252 

rider64_5sXIPnPuSO

The main view model doesn't have any testable code at the moment, but the features of the button are showcased in the design view model, which does have several tests. 

Hooked it into the app, replacing the previous placeholder button:

NexusMods App_0E01E2iLgQ

### DIFF
--- a/src/NexusMods.App.UI/Controls/Spine/Buttons/Download/DownloadButtonDesignerViewModel.cs
+++ b/src/NexusMods.App.UI/Controls/Spine/Buttons/Download/DownloadButtonDesignerViewModel.cs
@@ -1,15 +1,26 @@
 ﻿using System.Windows.Input;
 using NexusMods.DataModel.RateLimiting;
 using ReactiveUI;
+using ReactiveUI.Fody.Helpers;
 
 namespace NexusMods.App.UI.Controls.Spine.Buttons.Download;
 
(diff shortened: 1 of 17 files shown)


Aragas:
> Since we're going to be dealing with download speeds and possible other units, something like [UnitsNet](https://github.com/angularsen/UnitsNet) might be useful. This project has strongly typed quantities and conversions for basically everything, including [BitRate](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/BitRate.json) ([Generated Code](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/GeneratedCode/Quantities/BitRate.g.cs)):
> 
> ```cs
> var downloadSpeed = new BitRate(3.12, KibibytePerSecond);
> var abbreviation = BitRate.GetAbbreviation(downloadSpeed); // "KiB/s"
> var inMBs = downloadSpeed.MegabytesPerSecond;
> ```
> 
> The only downsides are this project isn't trimmable, even though this is perfect for trimming, and they use `decimal` behind the scenes, which is much slower than `float` or `double`.

The library seem done in it's current state. A fork with some changes might be viable :joy:

halgari:
Seems like a bit overkill, but something we can look into the future. For now the VM/View are unit/data agnostic. In some cases they may show time remaining, other times download speed.

erri120 (approved):

Issue #283, opened 2023-04-14
Implement In-memory IDataStore for unit testing


Sewer56:
# Currently

We have an `IDataStore` (database) that is backed by SQLite3. This component is needed for some unit tests. 

However, using the SQLite-backed `IDataStore` for unit testing presents a problem, as it requires we write data to disk. This can slow down the tests and risk leaving junk files behind if the user either stops debugging during a debugging session [which kills application] or does not properly clean up files from their test code. 

Currently, writing completely in-memory unit tests in some cases is not possible because we lack an implementation of an `IDataStore` that stores all of its information in memory.

# We want

We want to create an in-memory `IDataStore` implementation to to enable for cleaner unit testing. This new implementation should store all its information in memory and not write any files to disk, making it more suitable for unit testing purposes.

# Supporting Information
N/A

# Design
The design of this improvement should include:

1. A new class that implements the `IDataStore` interface (e.g. `InMemoryDataStore`), using an in-memory data structure (such as nested Dictionary) to store data instead of writing it to disk.

## Further Work

Convert existing tests relying on `SqliteDataStore` to use `InMemoryDataStore`. Open another issue for this once this is complete.

Al12rs:
As an alternative, from my knowledge SQLite3 can be configured to work in memory, which should offer the performance improvements and avoid the disk writes. This would need to be better investigated to be sure.

One advantage with this approach could be that the tests can also cover any regressions which are related to the SQLite3 database interactions.

Don't know if simply changing it to in memory satisfies the requirements here or if the intentions where to obtain an even leaner and simpler way to test

halgari:
Yeah, the easiest way to do this would probably to use a in-memory instance

PR #341, opened 2023-05-31
In memory datastore

halgari:
Adds `UseInMemoryDataModel` to `IDataModelSettings` which will switch the datamodel over to using a in-memory version of Sqlite, and a single process sync array. 

There's two major issues with this approach: normally every connection to a in-memory database gets a new database, and we need multithreaded access to the DB, so we have to generate a unique name for the store. In addition the DB is deleted once the last connection closes, so we now open a connection to the DB and keep it live for the lifetime of the app. This should also stop someone from deleting/moving the db in the middle of the app running.

In order to get the tests to pass I also had to fix a few issues in the existing tests. Not sure how the download url tests ever passed as the files were getting truncated on every test run. In addition the FOMOD tests were using a version of the datastore that didn't come from the DI system, so I fixed that as well.

### DIFF
--- a/src/Networking/NexusMods.Networking.HttpDownloader/LocalFileDownloader.cs
+++ b/src/Networking/NexusMods.Networking.HttpDownloader/LocalFileDownloader.cs
@@ -29,7 +29,7 @@ internal static class LocalFileDownloader
         var filePath = Uri.UnescapeDataString(uri.AbsolutePath);
         var fullPath = Path.GetFullPath(filePath);
 
-        await using var stream = fullPath.ToAbsolutePath(FileSystem.Shared).Create();
+        await using var stream = fullPath.ToAbsolutePath(FileSystem.Shared).Read();
         job.Size = Size.FromLong(stream.Length);
         await using var file = destination.Create();
         return await stream.HashingCopyAsync(file, default, job);
--- a/src/NexusMods.CLI/Types/DownloadHandlers/NxmDownloadProtocolHandler.cs
+++ b/src/NexusMods.CLI/Types/DownloadHandlers/NxmDownloadProtocolHandler.cs
@@ -1,4 +1,5 @@
 ﻿using NexusMods.DataModel;
+using NexusMods.DataModel.Abstractions;
 using NexusMods.DataModel.Loadouts.Markers;
 using NexusMods.Networking.HttpDownloader;
 using NexusMods.Networking.NexusWebApi;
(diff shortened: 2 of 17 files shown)


src/Networking/NexusMods.Networking.HttpDownloader/LocalFileDownloader.cs
```diff
@@ -29,7 +29,7 @@ internal static class LocalFileDownloader
         var filePath = Uri.UnescapeDataString(uri.AbsolutePath);
         var fullPath = Path.GetFullPath(filePath);
 
-        await using var stream = fullPath.ToAbsolutePath(FileSystem.Shared).Create();
+        await using var stream = fullPath.ToAbsolutePath(FileSystem.Shared).Read();
```
halgari, line 32: I have no clue how this ever worked

src/NexusMods.CLI/Types/DownloadHandlers/NxmDownloadProtocolHandler.cs
```diff
@@ -16,13 +17,13 @@ public class NxmDownloadProtocolHandler : IDownloadProtocolHandler
     private readonly Client _client;
     private readonly IHttpDownloader _downloader;
     private readonly TemporaryFileManager _temp;
-    private readonly ArchiveAnalyzer _archiveAnalyzer;
-    private readonly ArchiveInstaller _archiveInstaller;
+    private readonly IArchiveAnalyzer _archiveAnalyzer;
```
halgari, line 20: Somehow missed this in previous PRs, they should have always been interfaces

halgari:
fixes #283

erri120 (approved):
Using a new DataStore means we loose cached values from previous runs. On one hand, this means our tests will always start from a "clean slate", which is useful for tests of _changed_ code, but on the other hand, tests of _unchanged_ code can't make use of cached values. Our game tests are impacted the most by this as they need to download files from Nexus Mods, extract the archives and analyze the files.

This is mostly just a minor inconvenience for developers who like to run all tests in the solution.

halgari:
> This is mostly just a minor inconvenience for developers who like to run all tests in the solution.
This is already the case as we used to use a new datastore on every invocation of some of our tests, by creating a new directory for the datastore on every invocation:

ADataModelTest
```
        var prefix = FileSystem.Shared.GetKnownPath(KnownPath.TempDirectory)
            .CombineUnchecked(typeof(Startup).FullName ?? "NexusMods.DataModel.Tests")
            .CombineUnchecked(Guid.NewGuid().ToString());

        container
            .AddLogging(builder => builder.SetMinimumLevel(LogLevel.Trace))
            .AddFileSystem()
            .AddSingleton<TemporaryFileManager>()
            .AddDataModel(new DataModelSettings(prefix))
            .AddStandardGameLocators(false)
```

But I don't understand how this makes it harder for devs wanting to run all the tests, I re-run all the tests on most PRs I make

PR #347, opened 2023-06-07
Upgrade To Avalonia 11-rc1, Fix Datagrid not updating

### LINKED ISSUE #335: Bug: (Sometimes) the UI of the Mod List is not Immediately Updated
# Bug Report

## Summary
In NexusMods.App, we support downloading and installing of mods in the background.  
This may be initiated by Inter Process Communication (IPC), or a similar mechanism.

In the case that a user is looking at a given game's list of mods; sometimes changes in this list are not consistently reflected in the UI. The updated list sometimes doesn't show the installed mod, or the installed mod replaces the first entry in the list. The expected behaviour is that the newly installed mod should be added to the list.

## Steps to reproduce

1. Download a mod in the background with mod list open.
2. Check the mod list in the UI.

After a mod is installed, the mod list in the UI should be updated to include the newly installed mod.

## Other information

This inconsistent behaviour has been observed randomly. In some cases, the UI updates as expected, but in others, the bug described above is encountered.

One example run is in the video attached:  

https://github.com/Nexus-Mods/NexusMods.App/assets/6697380/57603e3b-83cb-4673-a050-9c72e155145d

Notice the `Launch` button reacting to the install as expected, and page refreshing; yet no mod is seen. SkyUI only shows after a page re-open.

This is probably a trivial fix, chances are.

### LINKED ISSUE #343: Upgrade to Avalonia 11 RC1
https://github.com/AvaloniaUI/Avalonia/wiki/Avalonia-11-Porting-Guide

https://github.com/AvaloniaUI/Avalonia/releases/tag/11.0.0-rc1.1

halgari:
Avalonia 11-rc1 changes a lot of things, including many interfaces that are now `internal`. This means I had to refactor quote a few things:

* Updated to Avalonia 11-rc1 (fixes #343)
* Disabled NX support switching to ZIP files for internal archives, this removes a lot of test inconsistency
* Implemented `Eventually` and `EventuallyOnUI` which are helpers to retry UI assertions until a given time window has expired
* Screenshotting in tests is deleted, we can research how to add this again but the `IHeadless` interfaces are now internal
* Converting between Avalonia and Skia bitmaps had to be reworked as some interfaces are now internal
* Reworked many of the tests so that they use Abstract test classes instead of helper tests, and standardized how we access helper methods
* Put a `Gird` panel around the `LoadoutDataGrid` instead of a `StackPanel`, open-ended sizing like that breaks the virtualization in the Datagrid (fixes #335)
* Update several dependency libraries that were out of date
* Avalonia 11-rc1 now has both `Items` and `ItemsSource` on collections and we were using the wrong ones in the past
  * ItemsSource - the data for each item in a list
  * Items - the actual views (controls) for each item in the list


src/NexusMods.App.UI/Controls/Spine/Buttons/Icon/IconButton.axaml
```diff
@@ -42,7 +42,7 @@
             <Setter Property="BorderBrush" Value="{DynamicResource ContentSecondary}" />
             <Setter Property="Background" Value="#01FFFFFF" />
             <Setter Property="BorderThickness" Value="1" />
-            <Setter Property="BorderDashArray" Value="2, 2" />
+            <!--<Setter Property="BorderDashArray" Value="2, 2" />-->
```
halgari, line 45: These are not supported in Avalonia 11, will re-enable them later

src/NexusMods.App.UI/Initializers.cs
```diff
@@ -18,7 +18,7 @@ namespace NexusMods.App.UI;
 public static class Initializers
 {
     public static readonly ICommand ICommand = ReactiveCommand.Create(() => { });
-    public static readonly IImage IImage = new WriteableBitmap(new PixelSize(16, 16), new Vector(96, 96), PixelFormat.Rgba8888, AlphaFormat.Opaque);
+    public static IImage IImage => new WriteableBitmap(new PixelSize(16, 16), new Vector(96, 96), PixelFormat.Rgba8888, AlphaFormat.Opaque);
```
halgari, line 21: This can't run until Avalonia's DI is setup, so moving it to a computed property fixes that race condition

src/NexusMods.App.UI/RightContent/LoadoutGrid/LoadoutGridView.axaml
```diff
@@ -14,9 +14,8 @@
     <Design.DataContext>
         <loadoutGrid:LoadoutGridDesignViewModel />
     </Design.DataContext>
-    <StackPanel>
-        <Line/>
-        <Border Classes="Toolbar">
+    <Grid RowDefinitions="Auto, *">
```
halgari, line 17: Must be in a grid to allow the DataGrid to work with item virtualization

erri120 (changes requested):

erri120 (approved):

PR #348, opened 2023-06-07
macOS support

halgari:
Mostly just gets the tests to pass. The UI updates just fine, bit we still lack OAuth support and the ability to login to the site. That requires protocol registration which I'll write in a future PR. This requires #347 before merging as that has a lot of path fixes. And we may need to rewrite parts of this after merging #345.


erri120:
Will review after I finished working on #345, and it's merged. I think Steam also runs on macOS, however, GameFinder doesn't support it at all.

erri120:
As a side note: I think our codebase uses the terms "OSX" and "macOS" interchangeably. Not sure if we want to make it consistent or not, it's probably clear to everyone that both terms refer to the same OS, and even the BCL is sometimes inconsistent with this.

Sewer56:
No closing reason was provided, so what's the status on this?

Is it Indefinitely Postponed or Cancelled?

Am just curious.

PR #352, opened 2023-06-13
Manually added games

halgari:
Fixes #115. 

In this PR:
* Allow for games to be manually added
  * Creates a new game store that contains arbitrary absolute paths and versions for games the user has added by hand
* The "stress test" github action (download new mods for each game, attempt to install them and report) now:
  * Manually adds a game for the game it is about to test, pointing to a blank folder.
  * This allows us to test games that are not installed on the test system
* If the NEXUS_API_KEY environment variable is set, the DI system will use that instead of trying to use OAuth (useful for testing)


erri120:
For the stress test, we can print the results using `GITHUB_STEP_SUMMARY` ([GitHub Docs](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/)), similar to #322.

src/NexusMods.Paths/RelativePath.cs
```diff
@@ -178,7 +178,7 @@ public override int GetHashCode()
     public static implicit operator ReadOnlySpan<char>(RelativePath d) => d.Path;
 
     /// <summary/>
-    public static implicit operator RelativePath(string b) => new(b);
+    public static implicit operator RelativePath(string b) => new(PathHelpers.Sanitize(b, OS));
```
erri120, line 181: I considered doing this in my original PR #345, however, I decided against it. `PathHelpers.Sanitize` is an expensive operation and should only be called on inputs that come from outside the application like parsing CLI arguments, reading a file, and so on.

Automatically sanitizing all relative paths in the implicit converter, and the `.ToRelativePath` extensions method, is essentially disabling the "is sanitized"-assertion completely for relative paths. The assertions exist to enforce correct path handling in our code, and are only useful if we only sanitize our input paths. It's our internally created paths like `static readonly RelativePath Something = "foo\bar"` that need to be checked for correctness.

src/NexusMods.Paths/Utilities/PathHelpers.cs
```diff
@@ -119,7 +119,10 @@ public static string Sanitize(ReadOnlySpan<char> path, IOSInformation os)
 
         // Paths on Unix-based systems and Windows paths without backslashes
         // only need to be checked for trailing directory separators.
-        if (os.IsUnix() || path.Count('\\') == 0) return RemoveTrailingDirectorySeparator(path).ToString();
+        if (os.IsUnix() || path.Count('\\') == 0)
+        {
+            path = RemoveTrailingDirectorySeparator(path).ToString();
+        }
```
erri120, line 125: This should be the following:

```csharp
if (path.Count('\\') == 0) return RemoveTrailingDirectorySeparator(path).ToString();
```

erri120 (changes requested):

erri120 (approved):

Issue #255, opened 2023-04-06
Allow users to pause downloads


Greg-Nexus:
# Story

## User Story

As a user, I want to pause in progress downloads

## Outline
Add the following functionality to the 'In Progress' downloads page:

- Pausing individual mod downloads that are in progress
- Resuming individual mod downloads that have been paused
- Pause all mod downloads
- Resume all mod downloads

This should not cover (these will be done iteratively in future tasks):

- Adding 'Cancel' UI
- Adding the 'Total' progress bar above the mods table
- Adding the 'Speed' section in the main content
- Building the 'Completed' page
- Building the 'Download history' page

## Supporting Information
Figma designs: https://www.figma.com/file/Ntr9PttR5VrqpTRwwZHzLu/Mod-list?node-id=252-24430&t=DXyQTx6LsPZC5Kpy-0

PR #360, opened 2023-06-26
Implement Download Service/Tracking, Progress UI, Download Cancel UI & Misc. UI Fixes

### LINKED ISSUE #256: Allow users to cancel downloads
# Story

## User Story

As a user, I want to cancel in progress downloads

## Outline
Add the following functionality to the 'In Progress' downloads page:

- Cancel the current mod download

This should not cover (these will be done iteratively in future tasks):

- Adding the 'Total' progress bar above the mods table
- Adding the 'Speed' section in the main content
- Building the 'Completed' page
- Building the 'Download history' page

## Supporting Information
Figma designs: https://www.figma.com/file/Ntr9PttR5VrqpTRwwZHzLu/Mod-list?node-id=252-24430&t=DXyQTx6LsPZC5Kpy-0

### LINKED ISSUE #257: Display total download progress
# Story

## User Story

As a user, I want to view the total progress of all my queued downloads

## Outline
Add the following functionality to the 'In Progress' downloads page:

- Display a 'Total' progress bar above the queued downloads, displaying:
  - Total % complete
  - File size downloaded
  - File size total
  - Time remaining

This should not cover (these will be done iteratively in future tasks):

- Adding the 'Speed' section in the main content
- Building the 'Completed' page
- Building the 'Download history' page

## Supporting Information
Figma designs: https://www.figma.com/file/Ntr9PttR5VrqpTRwwZHzLu/Mod-list?node-id=252-24430&t=DXyQTx6LsPZC5Kpy-0

### LINKED ISSUE #356: Theme Bug: Swapped Hover and Pressed Styles
# Bug Report

## Summary

On FIGMA, under *Nexus Design System* we have the following button styles:

image

The existing theme implementation in NMA, uses the `focus` style for mouse hover, and `hover` as the style when user clicks the button. This is incorrect and does not match [next.nexusmods.com](next.nexusmods.com).

## Steps to reproduce

- Open `NexusMods.App`.  
- Hover over `Log-In` button.  

The `focus` and `hover` styles are swapped.

## Other information

Already fixed on branch #cancel-downloads-new in 293fbcdd640afb080b9287af852a23afecd5cc68; and confirmed as bug with design team.
This is just for tracking fixed issues

### LINKED ISSUE #357: Theme: Incorrect Value for StructuralBorderColor
# Bug Report

## Summary

In the theme, `StructuralBorderColor` is incorrectly defined as `#2DFFFFFF`, rather than `#303236`; which makes borders subtly different to their FIGMA designs. This difference is very subtle in most of the UI as the borders .

Current/App:

image

Intended/FIGMA:

image

## Other information

Already fixed as of 95d5b58964e3ee6f6bbb2e0f52c96b1b2b83de62. in branch cancel-downloads-new, this is just for progress tracking.

### LINKED ISSUE #358: FoundGamesView is causing performance issues in switching application pages.
# Improvement

## Currently 

Switching into the 'My Games' and 'Found Games' views are extremely slow, even if there are no actual found games.

## Supporting Information
Running in Debug mode on a laptop with a 3550H; 

```
00:01:54.253 [DEBUG] Finding View for NexusMods.App.UI.RightContent.FoundGamesViewModel
00:01:54.254 [INFO] Creating View
00:01:54.730 [INFO] View Created!!
```

## Cause

The `FoundGamesView` control has the following code:

```xaml
<DataContext>
    <rightContent:FoundGamesDesignViewModel />
</DataContext>
```

Constructor of `FoundGamesDesignViewModel` has the following:

```csharp
var games = Enumerable.Range(0, 10).Select(_ => new GameWidgetDesignViewModel());
```

And `GameWidgetDesignViewModel` has this:

```csharp
Image = new Bitmap(AssetLoader.Open(new Uri("avares://NexusMods.App.UI/Assets/DesignTime/cyberpunk_game.png")));
```

The image is not cached; so this image is decoded 10 times every time `FoundGamesDesignViewModel` is created; which in the case of `My Games` view, twice (total 20 decodes).

## Solution

Only use `DataContext` in design mode.

```xaml
<Design.DataContext>
    <rightContent:FoundGamesDesignViewModel />
</Design.DataContext>
```

## Other Notes

Fixed in 6cc444b070e206a74bf51c6e9f337423df0c6fb6 on branch cancel-downloads-new.

### LINKED ISSUE #359: Theme: Incorrect Text Colour on Body2RobotoRegular
# Bug Report

## Summary

`Body2RobotoRegular` style uses text colour FontDarkSecondaryColor (#aaaaaa) instead of Brand White (#ffffffff).
This affects styling of some controls; here's an example...

Incorrect (App):

image

Correct (Figma):

image

## Other information

This probably happened when styles were copy-pasted when the values from Figma were being translated over.
Fixed in 8f6a626daa285747dea3b55bff1f315a68fedb9a on branch cancel-downloads-new.

All usages of style were checked/consolidated to ensure they match with Figma.

Sewer56:
fixes #356 fixes #357 fixes #358 fixes #359 fixes #256 fixes #257

# Summary

This PR implements mainly issues #256 (Allow user to Cancel Downloads) and #257 (Display Total Download Progress), along with several minor theme/UI adjustments that surfaced during the development process.

## Note

The complexity of implementing issues #256 and #257 was considerably higher than initially anticipated. These tasks required significant back-end plumbing which only became apparent during development. Consequently, the scope of this PR has expanded.

To prevent such situations in the future, we may need to discuss strategies for handling unexpectedly large tasks. One option could be to break down these tasks into smaller, more manageable PRs. This way, we can tackle the issues incrementally, potentially treating larger issues as epics or blocked issues. This approach requires further discussion.

On a brighter note, the work completed in this PR lays the groundwork for issues #255 and #258. The majority of the remaining work for these issues will be selecting and theming a graph library; not a huge amount of work.

## Changelog

Excluding main tasks #256 (Allow user to Cancel Downloads) and #257 (Display Total Download Progress); below is a short summary of the remainder of *other* changes done.

### Rate Limiting/Job Extensions

The Rate Limiting and Job System have been expanded to allow for pausing of jobs (additional state), and calculating job throughput (e.g. Download Speed). 

Usage Examples:

```csharp
var timeProvider = new DateTimeProvider();

// Get Throughput of Single Job
var resource = new Resource<JobThroughputTests, Size>("Test Resource", int.MaxValue, Size.FromLong(long.MaxValue), timeProvider);
var job = await resource.BeginAsync("Some Job", Size.FromLong(100), CancellationToken.None);
await job.ReportAsync(Size.FromLong(5), CancellationToken.None);

Thread.Sleep(1000); // simulate wait
var jobThroughput = job.GetThroughput(); // ~5 per second
```

```csharp
// Get Total Throughput of multiple jobs
// jobs is `IEnumerable<IJob<TSize>>`
var throughput = jobs.GetTotalThroughput(new DateTimeProvider());
```

The logic used for calculating throughput, takes into account pausing of jobs and calculates stats based since the last time the job was resumed/started running. e.g. If a download is paused for 10 hours; then the user resumes it; they expect to see e.g. 10MB/s, not 0.0001 because the job was paused.

In addition, there is also the `GetTotalCompletion` API, which calculates the total completion of a group of jobs. e.g. The number of downloaded bytes in all downloads.

### Refactor Download Code & Add `DownloadService`

The `DownloadService` class has been added with the purpose of handling and observing different states of various different download tasks which have now also been refactored into their own classes (e.g. NXM Downloads, HTTP Downloads, Local Downloads etc.). 

The class maintains lists of ongoing download tasks and uses Reactive Extensions (Rx) to provide observables of started, completed, cancelled, paused, and resumed tasks.

```csharp
// Creating the service (usually done by dependency injection)
IServiceProvider provider = ... // get your IServiceProvider
var downloadService = new DownloadService(provider);

// Adding a new NXM download task
downloadService.AddNxmTask(nxmUrl); // nxmUrl is an instance of NXMUrl

// Adding a new HTTP download task
downloadService.AddHttpTask(url, loadout); // url is a string and loadout is an instance of Loadout

// Adding a generic download task
downloadService.AddTask(task); // task is an instance of IDownloadTask

// Observing started tasks
downloadService.StartedTasks.Subscribe(task => Console.WriteLine($"Task {task.Id} started."));

// Observing completed tasks
downloadService.CompletedTasks.Subscribe(task => Console.WriteLine($"Task {task.Id} completed."));

// Getting total download throughput
Size throughput = downloadService.GetThroughput();
```

etc.

You can find the relevant implementation in: `NexusMods.Networking.Downloaders.DownloadService` and accompanying tests.

### Reorganize LoadoutGrid Namespaces

LoadoutGrid has over 60 classes!! Which all previously shared a namespace; this made working with the grid very difficult as ViewModels, Interfaces, Columns, plumbing code etc. were all mixed together and required excessive scrolling when names weren't obvious.

This was reorganised to group the relevant elements together by namespace:

image

All the back-end plumbing code is now grouped together, and anything related to a single LoadoutGrid column is now neatly tied up in its own namespace.

### Added Modular Base Overlay Controls

image

There are multiple dialog boxes that can 'pop up' throughout the application; we don't yet have a 'standard' way of showing or building these yet. This PR takes some steps towards achieving that; through implementing lightweight, modular controls for various possible current and future dialog boxes.

In the above image, there are 3 controls involved:
- Background Pill (`MessageBoxBackground`)
- Base Ok/Cancel Dialog (`MessageBoxOkCancelView`)
- Cancel Download View (`CancelDownloadOverlayView`)

They all wrap each other, so `CancelDownloadOverlayView` contains `MessageBoxOkCancelView` which in turn contains MessageBoxBackground`.

This setup makes it that `CancelDownloadOverlayView` can simply be defined as:

```xml
<okCancel:MessageBoxOkCancelView MinWidth="300" MaxWidth="450" 
                                 Title="Cancel and delete downloads?"
                                 x:Name="OkCancelView" />
```

while the `MessageBoxOkCancelView` provides the general layout for controls with 'Yes/No' dialogs.

In this case, `CancelDownloadOverlayView` is simply a wrapper around `MessageBoxOkCancelView` that sets the description of the dialogue to include name of the mod to be cancelled. Everything else is included in `MessageBoxOkCancelView` itself.

In addition to all of this, overlays can now be dynamically shown via `MainWindowViewModel.SetOverlayContent`; rather than requiring specific single instance per overlay in `MainWindowViewModel`. In the future, this new API might be updated to become a stack of overlays, such that multiple can be shown one by one in succession.

### UI Fixes

(Details in issues)

Fixed: *Incorrect Button Hover/Pressed Styles* https://github.com/Nexus-Mods/NexusMods.App/issues/356 
Fixed: *Incorrect Value for StructuralBorderColor* https://github.com/Nexus-Mods/NexusMods.App/issues/357
Fixed: *FoundGamesView is causing performance issues in switching application pages* https://github.com/Nexus-Mods/NexusMods.App/issues/358
Fixed: *Incorrect Text Colour on Body2RobotoRegular* https://github.com/Nexus-Mods/NexusMods.App/issues/359

# Merge Status: On Hold

This PR is not ready to be merged. I'm currently encountering an issue with `MainWindowViewModel.SetOverlayContent`. Despite the value being set and changing successfully, the ReactiveUI bindings for `MainWindowViewModel.OverlayContent` are not being triggered, even when `PropertyChanged` is explicitly called.

Apart from this issue, the PR is generally ready for review. I've been trying to resolve this problem for over an hour without success. Any insights would be appreciated.

There might be room to add a couple more VM tests to improve our coverage.


Sewer56:
Undrafting for CI Run, seems it didn't go off last time; I'll re-draft once Codecov gets back to me.

Sewer56:
Okay, now it builds. I'll make another commit or two, now that I got more up to date diff on coverage, and should be good.

Sewer56:
Should be good for review/merge process.

src/Networking/NexusMods.Networking.Downloaders/Interfaces/IDownloadService.cs
```diff
@@ -0,0 +1,95 @@
...
+/// <summary>
+/// This is a 'download manager' of sorts.
+/// This service contains all of the downloads which have begun, or have already started.
+/// </summary>
+public interface IDownloadService
+{
+    /// <summary>
+    /// Contains all downloads managed by the application.
+    /// </summary>
+    List<IDownloadTask> Downloads { get; }
```
halgari, line 17: This should be a ReadOnlyObservableCollection, so that we can see the small delta changes, and not have to deal with large lists of items once the number of downloads reach into the thousands.

src/Networking/NexusMods.Networking.Downloaders/Interfaces/IDownloadService.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// <summary>
+    /// This gets fired whenever a download-and-install task is started.
+    /// </summary>
+    IObservable<IDownloadTask> StartedTasks { get; }
```
halgari, line 22: Likewise these can be ReadOnlyObservableCollections and/or `IObservable<IChangeSet<..>>`

src/NexusMods.App.UI/Helpers/StringFormatters.cs
```diff
@@ -0,0 +1,43 @@
...
+/// </summary>
+public static class StringFormatters
+{
+    /// <summary>
+    /// Formats the file size as a string.
+    /// </summary>
+    /// <param name="currentBytes">The currently downloaded number of bytes.</param>
+    /// <param name="totalBytes">The total number of downloaded bytes.</param>
+    /// <remarks>Complete number of downloaded bytes.</remarks>
+    public static string ToGB(long currentBytes, long totalBytes)
```
halgari, line 14: We already use the `Humanizer` library for this

src/NexusMods.App.UI/RightContent/Downloads/InProgressViewModel.cs
```diff
@@ -1,6 +1,89 @@
...
+                if (SelectedTask == null)
+                    return;
+                
+                var result = await overlayController.ShowCancelDownloadOverlay(SelectedTask);
+                if (result) 
+                    CancelSelectedTask();
+            });
+
+            // Subscribe to started tasks
+            downloadService.StartedTasks
```
halgari, line 45: This will probably be much simplified with the switch to `ReadOnlyObservableCollection`

src/NexusMods.App.UI/RightContent/Downloads/ViewModels/DownloadTaskViewModel.cs
```diff
@@ -0,0 +1,124 @@
...
+    // live-updating ViewModel.
+    private string _previousName = string.Empty;
+    private string _previousVersion = string.Empty;
+    private string _previousGame = string.Empty;
+    private DownloadTaskStatus _previousStatus = DownloadTaskStatus.Idle;
+    private long _previousDownloadedBytes = 0;
+    private long _previousSizeBytes = 0;
+    private long _previousThroughput = 0;
+    
+    public void Poll()
```
halgari, line 80: Not sure that we need this we should be able to do this via `[Reactive]` or observable chains

src/NexusMods.App.UI/RightContent/LoadoutGrid/ColumnType.cs
```diff
@@ -10,5 +10,12 @@ public enum ColumnType
     Version,
     Category,
     Installed,
-    Enabled
+    Enabled,
+    
+    // Download
+    DownloadName,
```
halgari, line 16: These aren't Loadout Columns, so they don't belon in `/LoadoutGrid/` We should either subclass some new grid, or make `LoadoutGrid` more generic.

halgari (commented):
We can go over DynamicData (ReadOnlyObservableCollection) when you have a moment. It's a fairly complex library, but what's awesome about it, is it syncs collections of data including those in UI widgets. But does it via changesets, so that only the deltas are ever emitted as events, instead of the entire collection. Even with this it still supports sorting, transforms, filters, projections, etc. So if you sort, and only two items swap places, then the delta is just the movement of those two records and not a re-creation of the entire list.

halgari (changes requested):
Forgot to add this as a "request changes" and not comment, see the previous review

src/NexusMods.App.UI/RightContent/Downloads/ViewModels/DownloadTaskViewModel.cs
Sewer56: I'm not a fan of it either, but there are specific reasons I wrote it in this way. 

1. Some metrics (e.g. throughput/download speed); inherently require polling.

Download speed is defined as the amount of bytes transferred within the last period of <x> time; and must be polled. In this case, we need some sort of poll loop anyway. 

While in theory we could technically add polling to the HTTP Downloaders themselves to calculate this info, you would still have to poll inside the HTTP Downloaders themselves. In that case, we'd end up having to write more code because both HTTP downloaders would need this mechanism for parity.

2. Some third party APIs don't expose the data we need.

Not 100% sure if we'll run into this, but I've also been burned in the past with implementing downloads from certain services, where the necessary APIs didn't expose events to obtain the data we need (e.g. Updates/Events on Download Progress). 

One notorious case in the past where I've been burned really hard was NuGet, abstracting away access to even the HTTP/Source URL from which the contents are being fetched; and not exposing any info at all. This was a bit problematic... as I couldn't handle the download myself either.

I've used self-hosted NuGet as a form of mod distribution in the past; and I had to use reflection in production to access an internal field to get HTTP URL :sweat_smile:. Not the proudest thing, but I had control over the dependency version so it was my only option short of reimplementing the client myself (download source was deliberately abstracted, so no possible PR).

----------

In any case, this was mostly a code saving measure, a polling mechanism of some sort is necessary anyway, (at least for calculating download speed). 

I figured a single polling loop would have been better than risking having multiple independent polling loops down the road (e.g. in HTTP downloaders, or specific service download implementation).

src/NexusMods.App.UI/RightContent/LoadoutGrid/ColumnType.cs
Sewer56: Fixed in 39f3c78ec5953ac03080eb47fad59a139fa9b251

src/NexusMods.App.UI/Helpers/StringFormatters.cs
Sewer56: Working as intended. 

`Humanizer` doesn't have an extension for formatting progress in format `current / total`. There is one for throughput, e.g. `1MB/s`, but not for progress.

The format shown in UI designs is `0.00 GB / 0.00 GB`. 
(Although it doesn't explicitly state whether it's `0.##` or `0.00`)

Humanizer formats as `0.##` (not customizable), so `500MB / 1000MB` would be formatted as `0.5 GB / 1 GB`.
(using `$"{currentBytes.Bytes().Humanize("GB")} GB / {totalBytes.Bytes().Humanize("GB")}"`)

This is undesirable, as the width of the download progress text would change as digits get added/removed in the rare random cases where progress represents a nice, round number for 1 second. After that 1 second, text would return to previous size, making the experience awkward.

If you instead use format `$"{currentBytes.Bytes().Humanize("0.00")} GB / {totalBytes.Bytes().Humanize("0.00")}"`, Humanizer will round to integers, in which case you will get inaccurate values (i.e. 0GB when 400+MB downloaded).

The current code ensures the download progress is formatted consistently.

src/NexusMods.App.UI/RightContent/Downloads/ViewModels/DownloadTaskViewModel.cs
halgari: I'd rather this information be in some sort of behind the scenes downloader service, otherwise people will lose the download throughput history everytime they navigate away from a view model (and the VM is destroyed)
 
 >>Some third party APIs don't expose the data we need.
 
 Not sure how this applies to the task at hand, we always have access to the download rate and although we may not have access to the full download size, those sorts of HTTP services are fairly rare.

src/NexusMods.App.UI/Helpers/StringFormatters.cs
halgari: We have quite a bit of flexability on code like this, there's nothing that says we have to match the mockups 100%

Sewer56: Yeah, though in this situation it would just be awkward if all the test is aligned 99% of the time and 1% of the time it becomes not aligned for 1 second. That's why I decided to go with this approach, it would be unsightly for the eyes.

This PR has actually already done this sort of thing with merging the pause/resume and progress bar column together, as we discussed and decided it would lead to better UX. (Happened when you were vacationing)

src/NexusMods.App.UI/RightContent/Downloads/ViewModels/DownloadTaskViewModel.cs
Sewer56: > I'd rather this information be in some sort of behind the scenes downloader service, otherwise people will lose the download throughput history everytime they navigate away from a view model (and the VM is destroyed)

Good point, is it fine if I handle this when we get around to adding the graph to the view?
At that point, keeping this history will be necessary. 

Incremental steps; don't want to hold this PR back longer than it needs to be. 

> those sorts of HTTP services are fairly rare.

Totally agree, but I had to include the example; as it was something I've ran into in the past. While extremely unlikely, and quite honestly I doubt we'll run into it, it is possible.

Writing the code this way was once again, just a cost saving measure as for throughput (etc.), we have to poll anyway.

halgari: If the tests pass at this point I'll merge this. In the future I think we should probably stop work on a ticket like this and deal with it in smaller PRs, as that will help us avoid situations where we're forced to either do good code reivews or not have something merged for awhile. That and maybe I should put a limit on the number of files someone can edit in a single PR :P

src/NexusMods.App.UI/Helpers/StringFormatters.cs
Sewer56: Can I request that https://github.com/Nexus-Mods/NexusMods.App/pull/360/commits/d1222c0edb79e13a4c0e0b841687fb8b5691d737 gets feedback from an extra set of eyeballs?

The change in this commit will lead to strings such as `677.11 MB / 953.67 MB` to be printed.

The spacing in the DataGrid for InProgress downloads is already *very tight* as is. There is not a lot of horizontal space available, and as such, this change to formatting will lead to excessive text clipping; within that menu. That's another factor that has to be considered here.

We can save some space by trimming the `Version` column [down to around 80], as it's probably the thing that end users will care about the least; and anything beyond a standard semver length string is redundant.

Extra Note: No progress will display as `0 b` in Humanizer, which means you can get a string such as `0 b / 2.58 GB`. Might want to special case this one.

src/NexusMods.App.UI/RightContent/Downloads/ViewModels/DownloadTaskViewModel.cs
Sewer56: Yeah; honestly; this one spun out of control :P

:sweat_smile: 
gosh...

PR #362, opened 2023-06-27
Basic Loadout Diagnostics

### LINKED ISSUE #327: Epic: Analyze loadout and create diagnostics
## Introduction

Similar to a compiler, warnings, and errors refer to issues with the created content, be it code or, in our case, a loadout. These _diagnostics_ can vary in complexity:

For TES and Fallout games, we [already](https://github.com/Nexus-Mods/NexusMods.App/blob/3ae7ede4dd8e54db9a6e9f9dac18f98c7954b6c7/src/Games/NexusMods.Games.BethesdaGameStudios/PluginAnalyzer.cs#L27) analyze the Plugin files (`.esp`, `.esm`, `.esl`) using [Mutagen](https://github.com/Mutagen-Modding/Mutagen) and can create a "warning" or "error" about missing dependencies. Similarly, for Stardew Valley, we also [already](https://github.com/Nexus-Mods/NexusMods.App/blob/3ae7ede4dd8e54db9a6e9f9dac18f98c7954b6c7/src/Games/NexusMods.Games.StardewValley/Analyzers/SMAPIManifestAnalyzer.cs#L68) analyze the manifest files and have a list of known dependencies.

Diagnostics can also be expanded with more complex analyzers. The [Mutagen.Bethesda.Analyzer](https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers) project sets out to provide diagnostics for a variety of issues, ranging from [Missing Assets](https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/issues/24) to [Circular Level Lists](https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/issues/8) and [Engine limit of four shadowcasting lights](https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/issues/34). Darkest Dungeon also has [engine limits](https://docs.google.com/document/d/1RbmVk8NmKI92NPCG2wLDxILMmtgyh_dFMY58N13Bxe8/edit), which could be found using a static analyzer to prevent unexpected crashes for the user.

### Diagnostics

Diagnostics should contain a short message, a detailed description, a level and optionally a reference to one or multiple mods relevant to this diagnostic. The diagnostic levels should be chosen carefully and must be game independent.

- Critical: Will cause the game to crash.
- Error: Will cause bugs and unexpected behavior, but won't necessarily crash the game.

These two are example levels that could be used, however, a formal definition of available levels should be made.

### UI/UX

The UI has to visualize the emitted diagnostics and allow the user to "dismiss" or "ignore" one, some or all of them. Furthermore, the user should be able to pick a minimum diagnostic level. The diagnostic level and short message should be easily viewable, while the detailed description could be in another menu to reduce cluttering. The list of referenced mods could be intractable and allow the user to quickly "navigate" to, or "filter" out, the relevant mods.

### Implementation

Our backend already supports `IFileAnalyzer` implementations, however, the results are only consumed by an `IModInstaller` implementation. We'd need a new mechanics, similar to that of `IFileAnalyzer`, which runs _after_ a mod has been installed, uninstalled or otherwise modified. It will analyze that single mod and emit data relevant to it. After every mod has been analyzed, the data of every mod will be used to emit diagnostics:

```mermaid
flowchart TD
    Loadout
    Loadout -- Mod A --> ModAnalyzer
    Loadout -- Mod B --> ModAnalyzer
    Loadout -- Mod C --> ModAnalyzer
    ModAnalyzer[Mod Analayzer]
    ModAnalyzer -- Data A --> DataAggregator
    ModAnalyzer -- Data B --> DataAggregator
    ModAnalyzer -- Data C --> DataAggregator
    DataAggregator[Data Aggregator]
    DataAggregator --> DataAnalyzer
    DataAnalyzer[Data Analyzer]
    DataAnalyzer -- Diagnostics --> DiagnosticEmitter
    DiagnosticEmitter[Diagnostic Emitter]
```

The "Mod Analyzer" would take a mod and statically analyze its content. The output data of this analyzer will be aggregated and send to a "Data Analyzer", which takes in all the mod analyzer outputs and emits diagnostics. Using already existing code, we can cache the mod analyzer outputs and run the data analyzer only when necessary.

erri120:
Part of #327

## Changes

- Adds `Diagnostic`, a new Entity that represents a diagnostic.
- Adds "diagnostic emitters", which are interfaces you can implement to create diagnostics.
- Adds `IDiagnosticManager` and the internal `DiagnosticManager` implementation.
- Adds an example diagnostic for Stardew Valley that checks for missing dependencies.
- Fixes various Sqlite bugs.
- **BREAKING CHANGE**: adds new and changes existing entries in `EntityCategory`

## Diagnostics

At the core of all of this is a `Diagnostic`. It has an identifier, a severity, a message, and data references. The identifier, `DiagnosticId` is a composite between a `Source` and a `Number`, where `Source` is the assembly name that produces the diagnostic and `Number` is a unique monotonic increasing number. This is similar to C# diagnostics (`CS4014`), however, instead of using two or three letters to define a namespace, we're using the entire assembly name instead to prevent collisions.

The message has a custom value object `DiagnosticMessage` and will, in a later PR, a format string. It's currently just a simple string, but will later be parsed and properly displayed in the UI with links to the referenced data.

The data references implement `IDataReference` and there are currently three implementations: `LoadoutReference`, `ModReference`, `ModFileReference`.

### Severity of Diagnostics

Probably the most important part of this PR. I've created a formal definition for all levels of severity diagnostic can have. They are based on my previous experience when creating severity levels for [Mutagen.Bethesda.Analyzers](https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers) and I've also looked at the levels created by C# code analyzers ([`DiagnosticSeverity`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.diagnosticseverity)) and the [CVE severity scale](https://www.imperva.com/learn/application-security/cve-cvss-vulnerability/) for reference.

I believe that the following levels of severity are general enough to be applicable for any game, while still offering enough difference between levels. These differences are significant for both users and developers that write analyzers and create diagnostics.

A user must quickly be able to identify which diagnostics impact their experience the most. Similarly, a developer should be able to choose the most fitting level, without thinking too much about it. A simple summary, an extended description and some examples should help with that process.

#### Severity: Critical

**Summary**: Something that will make the game unplayable.

**Description:** Unplayable, in this context, means that the user is unable to interact with the game. The most common effect of such problems is a crash. The frequency or likelihood of such causes or conditions are irrelevant. A crash that happens when the user starts the game, and a crash that happens when the user is in a very specific situation, are equal in severity.

**Examples:**

- A crash.
- A prolonged or indefinite loading screen.
- A prolonged or indefinite freeze.

#### Severity: Warning

**Summary**: Something that has an unintended adverse effect on any part of the game.

**Description:** This severity encompasses unintended problems that negatively impact the game in any of its aspects. This includes the visuals, the performance, and even the gameplay.

**Examples:**

- Missing assets that don't crash the game but manifest themselves as a noticeable abnormality. As an example, missing textures will get replaced with a default duo color checker-pattern, or a single color texture.

#### Severity: Suggestion

**Summary**: Something that doesn't indicate a problem, and offers improvements to the user.

**Description:** This severity can be used to provide helpful advises to the user. Applying these suggestions MUST NOT introduce further diagnostics of a higher severity. Furthermore, analysis and deduction MUST BE objective and based on facts. Suggestions have to be documented, and their improvements must be justifiable. Don't offer improvements if you aren't sure that the user benefits from them.

**Examples:**

- After analysis of the installed mods and their files, packing them into an archive, instead of having them as loose files, will improve load times.

## Diagnostic Emitters

An emitter takes in some data and returns zero-or-more diagnostics. There are three interfaces you can implement: `ILoadoutDiagnosticEmitter`, `IModDiagnosticEmitter`, and `IModFileDiagnosticEmitter`.

## Diagnostic Manager

The manager is the core of everything diagnostics related. It keeps track of all active diagnostics, runs the emitters if the loadout changes, and removes outdated diagnostics. It also has options that the user can later configure, once we have the UI for it.


erri120:
Our current "analyzer-hell":

[`IArchiveAnalyzer`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/Abstractions/IArchiveAnalyzer.cs): only implemented by [`ArchiveAnalyzer`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/ArchiveAnalyzer.cs) in the Data Model. This is the service that will be called when a user wants to install a mod.

[`IFileAnalyzer`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/Abstractions/IFileAnalyzer.cs): part of the mod installation process. [`ArchiveAnalyzer.AnalyzeFileAsync`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/ArchiveAnalyzer.cs#L81) calls [`AnalyzeFileInnerAsync`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/ArchiveAnalyzer.cs#L138), which hashes the input, does some recursion if the input is an archive, and calls [`IFileAnalyzer.AnalyzeAsync`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/Abstractions/IFileAnalyzer.cs#L34) on each `IFileAnalyzer` that has a matching signature.

`IFileAnalyzer.AnalyzeAsync` returns an async-enumerable of type [`IFileAnalysisData`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/Abstractions/IFileAnalysisData.cs). The [`PluginAnalyzer`](https://github.com/Nexus-Mods/NexusMods.App/blob/819887b5c2b575f395d50db60fe576f024c56092/src/Games/NexusMods.Games.BethesdaGameStudios/PluginAnalyzer.cs) implementation of `IFileAnalyzer` returns [`PluginAnalysisData`](https://github.com/Nexus-Mods/NexusMods.App/blob/819887b5c2b575f395d50db60fe576f024c56092/src/Games/NexusMods.Games.BethesdaGameStudios/PluginAnalysisData.cs), which contains some information, like a list of Masters, for the current Bethesda plugin file (`.esp`). All of this data gets saved to [`AnalyzedFile.AnalysisData`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/ArchiveContents/AnalyzedFile.cs#L54).

`AnalyzedFile.AnalysisData` is used by some implementations of `IModInstaller`, like the [`SMAPIModInstaller`](https://github.com/Nexus-Mods/NexusMods.App/blob/819887b5c2b575f395d50db60fe576f024c56092/src/Games/NexusMods.Games.StardewValley/Installers/SMAPIModInstaller.cs#L32). It's also used by the [`LoadoutManager`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/Loadouts/LoadoutManager.cs) in [`LoadoutManager.IndexAndAddGameFiles`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/Loadouts/LoadoutManager.cs#L182). This method will call `IArchiveAnalyzer.AnalyzeFileAsync` on all files inside the game directory. The resulting `AnalyzeFile` objects, will then be used in [`LoadoutManager.GetMetadata`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/Loadouts/LoadoutManager.cs#L440).

```mermaid
flowchart TD
    subgraph IFileAnalyzer
        IFileAnalyzer.AnalyzeAsync[AnalyzeAsync]
    end

    subgraph ArchiveAnalyzer
        ArchiveAnalyzer.AnalyzeFileAsync[AnalyzeFileAsync]
        ArchiveAnalyzer.AnalyzeFileInnerAsync[AnalyzeFileInnerAsync]
        ArchiveAnalyzer.AnalyzeArchiveInnerAsync[AnalyzeArchiveInnerAsync]
        ArchiveAnalyzer.AnalyzeFileAsync --> ArchiveAnalyzer.AnalyzeFileInnerAsync
    end

    subgraph LoadoutManager
        LoadoutManager.InstallModsFromArchiveAsyn…ArchiveAsync: Step 2]
        LoadoutManager.InstallModsFromArchiveAsyncStep3[InstallModsFromArchiveAsync: Step 3]
        LoadoutManager.InstallModsFromArchiveAsyncStep1 --> ArchiveAnalyzer.AnalyzeFileAsync
        ArchiveAnalyzer.AnalyzeFileInnerAsync.return -- AnalyzedArchive --> LoadoutManager.InstallModsFromArchiveAsyncStep2
        LoadoutManager.InstallModsFromArchiveAsyncStep2 --> IModInstaller.GetModsAsync -- ModInstallerResult --> LoadoutManager.InstallModsFromArchiveAsyncStep3
    end
```

(if this doesn't render, check [this link](https://mermaid.live/edit#pako:eNqdld9vgjAQx_-Vps9C4nwjyxYy50Iy9zD3sogPNziEDFrTFjen_u9rBRV_A7zQXq-fu_a-B0sa8BCpQ6OU_wQxCEU--j4j-pH511TALCbeIEnRZZAu_lAUa-Y5MNvlwJULFoyrk0mxA1nosyOwK4I4mZ9BHy1s4ZuA1QA7w6TWXo8xFCeAvfU2pTSfgk4WJvVPQyzroV7ae-YTsOdfJSBQSz0k5fhx3fAeyL1lVQopE9kHBZuELpe3aQxD2-d77gwbl9UnylXti7-MeeOreonZAlUu2Lh4kdInNE4tlKCDk4gLgnMUCxJpCEkYgcKtaYnPdYs35KHHpII0PWjDqtl-QaXnstBmdXKlDV85hDxXQ2AwrZIP7XYZxBAHgmfb0xj4SOGsO77i4BDjQrqTtvS7GvS71vReDXqvNb17q_ptuqoUr1HdVrhuRWsNb7d1WYpPxUUJmvSqi-8o81S1yLC3ky_t0AxFBkmo_1pLY_apijFDnzp6GIL49qnP1toPcsVHGkAdJXLs0HwWgsJ-Alr3GXUiSCWu_wEzHXfx)).

`LoadoutManager.GetMetadata` is perhaps the weirdest thing in all of this. This method is only for files from the **game directory**, yet it returns [`IModFileMetadata`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/Loadouts/IModFileMetadata.cs). `LoadoutManager.GetMetadata` calls all registered implementations of [`IFileMetadataSource`](https://github.com/Nexus-Mods/NexusMods.App/blob/63463e0aad834fdec11d62f75907455c07b5e248/src/NexusMods.DataModel/Abstractions/IFileMetadataSource.cs), which has a `GetMetadataAsync` method that takes the `AnalyzedFile` and returns an async-enumerable over `IModFileMetadata`. The only implementation of this is [`AnalysisMetaDataSource`](https://github.com/Nexus-Mods/NexusMods.App/blob/819887b5c2b575f395d50db60fe576f024c56092/src/Games/NexusMods.Games.BethesdaGameStudios/AnalysisMetaDataSource.cs).

halgari:
I'm not sure that this qualifies as "hell" but it is true that "analyzer" is a rather generic term, so it needs an adjective attached to it to give it context. 

- IArchiveAnalyzer - One aspect of the confusion here may be that it's common practice to abstract dependency injected classes behind interfaces for easier testing. In several of the tests, pre-calculated data is injected into the tests via a custom implementation of the `IArchiveAnalyzer` interface. The same is true of other interfaces in the system.

- IFileAnalyzer - This is designed to provide a caching layer between the file analyzers (which could take a long time to execute) and anything that wants to use that data. For Example:
  - BSA and .archive files (cyberpunk) will want to have their subfile names returned from this interface so we can quickly determine file override behavior. 
  - ESP files will need to return their masters like (they already do)
  - DDS files may return their compression format
  - INI files return a hashet of their sections and keys, so we can quickly determine if the INI is likely a mod configuration file, render settings, etc. 

Now for the big one: 

- IModFileMetadata - Perhaps we can rename or abstract this a bit. Really the idea here is to provide metadata (extra information, about a file inside a Loadout. So this could be sorting information, user-defined sort rules, etc. Since we can't modify or add extra properties to the C# classes, and since we can't really polymorphically extend records (`can't to X with {Member = Y}` where `X` is a abstract class, at lest I don't think we can). We need something to attach extra data to. Something that could work here, so to make it a generic `IMetadata` collection, then put that directly on the Entity class. I think in many situations we want a place to stick extra metadata, so perhaps that would work. At least with this approach we no longer use "ModFile" in the name.

erri120:
Part of the issue with `IModFileMetadata` is that `IFileMetadataSource` implementations are only used by `LoadoutManager.IndexAndAddGameFiles`. Looking at the doc string of `IFileMetadataSource`, I think this was supposed to be used in different places, but we forgot?

> Defines a source of file metadata, will be called whenever new files with the given filters are added to a loadout, or when they are modified.

Even if we used this in the "intended" way, it feels like we're doing the same thing twice. At least that's the case with the only implementation of `IFileMetadataSource`, [`AnalysisMetaDataSource`](https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/Games/NexusMods.Games.BethesdaGameStudios/AnalysisMetaDataSource.cs), which just takes the result of the [`PluginAnalyzer`](https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/Games/NexusMods.Games.BethesdaGameStudios/PluginAnalyzer.cs) and "adapts" the result by mapping the `IFileAnalysisData` implementation [`PluginAnalysisData`](https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/Games/NexusMods.Games.BethesdaGameStudios/PluginAnalysisData.cs) to the `IModFileMetadata` implementation [`AnalysisSortData`](https://github.com/Nexus-Mods/NexusMods.App/blob/774bf9b21ddaac70cf1dac8b09768d213d98cac2/src/Games/NexusMods.Games.BethesdaGameStudios/AnalysisSortData.cs).

Setting the metadata could also be done directly in the `IModInstaller` implementation, which is the place where the `AModFile` instances are created: https://github.com/Nexus-Mods/NexusMods.App/blob/dc12a50f98e8ce2bb1f78934abdbed7c6e589339/src/Games/NexusMods.Games.BethesdaGameStudios/Installers/SkyrimInstaller.cs#L88-L94

However, since the original game files don't go through one of the installers, this would mean that the game files won't have metadata if the installers set it.

I'm not sure what's the best solution for this. `IFileAnalysisData` and `IModFileMetadata` are technically the same thing. The former is part of `AnalyzedFile` while the latter is part of `AModFile`.

The reason I'm drilling into this, is because diagnostics would also make use of the resulting data/metadata. As an example, if we want to create diagnostics for Skyrim to warn the user about missing masters, we want a "diagnostic producer" that looks at the entire loadout and the metadata of each mod to figure out if there are missing masters. If the loadout is modified, the producer would run again with the new loadout version.

SMAPI has a similar situation where mods have a manifest that tells SMAPI which dependencies are required. Currently, the `SMAPIManifest` is an implementation of `IFileAnalysisData`, so if I want to write a "diagnostic producer", I'd have to create a new implementation of `IFileMetadataSource` which will just return the same `SMAPIManifest`.

I think we should consolidate `IFileAnalysisData` and `IModFileMetadata` into something, or at least make one implement the other, so we don't have unnecessary `IFileMetadataSource` implementations that just maps from `IFileAnalysisData` to `IModFileMetadata`.

erri120:
I've restructured our file analysis a bit and removed `IModFileMetadata` and `IFileMetadataSource`:

We now have an `IMetadata` interface for everything metadata related. `AModFile` has a property `ImmutableList<IMetadata> Metadata` and `AnalyzedFile` has a property `ImmutableList<IFileAnalysisData> AnalysisData`.

`IFileAnalysisData` now implements `IMetadata`, and I've created extension methods to easily map from `AnalyzedFile` to one of the `AModFile` inheritors (`FromArchive` and `GameFile`).

Now, installers can simply call `analyzedFile.ToFromArchive(somePath)` and all properties from `AnalyzedFile`, including `AnalysisData`, will be mapped to `AModFile`. This means that `IFileMetadatSource` is now unnecessary, which is why I've removed it.

halgari (approved):

PR #365, opened 2023-06-30
Windows and Wine path fixes

erri120:
Some small bugs I found while integrated the latest version into GameFinder:

- Paths with double backslashes weren't sanitized properly: `C:\\foo` would turn into `C://foo` instead of `C:/foo`.
- `BaseFileSystem.ConvertCrossPlatformPath` was kinda broken with Windows paths from Wine on Linux due to sanitization being in the context of the current OS. Meaning a Windows path `C:\foo` that we get from Wine on Linux has to be sanitized using the Windows specific sanitization methods, so that `C:/foo` can properly be mapped to `/c/foo`.
- Adds a `Debug.Assert` to `GetKnownPath` to make sure that no `default` values are returned. This is the case when the caller forgot to set up known path mappings, as some known paths are OS-specific.
- Adds `HasKnownPath` that returns whether the provided known path exists on the current platform or has a known path mapping.


Sewer56 (approved):
Looks good to me, no complaints.
Fairly simple, and still sufficiently efficient for the traffic this method gets.

--------

I can probably SIMD this if needed; (here you'd just load the `\` char into each ushort of a SIMD register, do equal and `movemask` to accumulate number of bytes until next `\` char, then seek to non `\` char). 

If this ever becomes a high enough traffic method, do consider it.

PR #416, opened 2023-07-06
Fix crash when adding mods from cli

### LINKED ISSUE #413: Installing mod from CLI fails if archive data isn't already in dataStore
When trying to install some medium sized mods from CLI the installation fails with `InvalidOperationException`
https://github.com/Nexus-Mods/NexusMods.App/blob/464cf7026d5034880a68e4ee0f14eea86b5b9560/src/NexusMods.DataModel/ArchiveInstaller.cs#L63-L70

Tested mods: 
[RaceMenu](https://www.nexusmods.com/skyrimspecialedition/mods/19080?tab=files)
[RealisticWaterTwo](https://www.nexusmods.com/skyrimspecialedition/mods/2182?tab=files)

Installing Racemenu from UI was successful and after that it was possible to install racemenu from CLI as well.
RealisticWaters can't be installed from UI since it is a Fomod and so it currently can't be installed.

Al12rs:
fix #413 
The FomodAnalyzer was closing a stream that was still in use causing an exception later.

### DIFF
--- a/src/Games/NexusMods.Games.FOMOD/FomodAnalyzer.cs
+++ b/src/Games/NexusMods.Games.FOMOD/FomodAnalyzer.cs
@@ -49,7 +49,7 @@ public class FomodAnalyzer : IFileAnalyzer
 
         try
         {
-            using var streamReader = new StreamReader(info.Stream);
+            using var streamReader = new StreamReader(info.Stream, leaveOpen:true);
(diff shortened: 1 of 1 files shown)


Sewer56:
lgtm

erri120 (approved):

Issue #370, opened 2023-07-03
Epic: Breakup mono repo into multiple standalone repositories


erri120:
Tasks:
- [ ] Create a meta-repository
- [ ] Move libraries (and their tests) into their own repos:
  - [ ] `NexusMods.Paths` (+ the testing helper lib)
  - [ ] `NexusMods.Hashing.xxHash64`
  - [ ] `NexusMods.FileExtractor`
  - [ ] `NexusMods.DataModel.RateLimiting`
- [ ] Convert project references into package references

The meta-repository will be included by every other repository as a submodule, and the included files will be symlinked. It should include files like `.editorconfig`, `.gitignore`, `.gitattributes`, `LICENSE`, and pre-setup `.github/workflows`.

Sewer56:
Potential candidate for closing in the future.
I checked off the other projects that got moved away from main in order to be more 'up to date'.

Another potential candidate for this is 'Workspaces'. I also considered 'Diagnostics', however that's so intrinsically tied to our DataModel that it perhaps is better left as part of the App itself.

erri120:
This can probably be closed.

PR #419, opened 2023-07-10
Reuse workflows

erri120:
Part of #370.


erri120:
The `delete_merged_branches` workflow is obsolete, GitHub supports auto-deleting branches ([Docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches)):

erri120:
I updated the workflows, the repository settings have to be updated as well since the old workflow jobs don't exist anymore @halgari.

Sewer56:
Looks good to me.
Only one thing to note. 

It might be a good idea to pin our referenced workflows to a tag or a commit to avoid potentially breaking changes.

For example, when adding a new runner/platform to a workflow we might run into a situation where a project referencing it does not support that target platform (like currently we do with OSX).

I've already been burned in the past with the likes of [CsWin32](https://github.com/microsoft/CsWin32) changing schema, which in turn made my old project uncompileable. I'd rather avoid that sort of situation on our DevOps end when dealing with other repo references.

Issue #423, opened 2023-07-10
PluginFile is incorrectly being used for .esp files and Plugins.txt file.


Al12rs:
While debugging SkyrimSpecialEditionTest.cs for #411 I found that esm and esp files in were being represented with `PluginFile` in the dataModel, which is incorrect as `PluginFile` is meant to represent the Generated Loadorder file (Plugins.txt) not the plugin files themselves.

- [x] Fix incorrect usages of `PluginFile`
- [x] Rename `PluginFile` to `LoadOrderFile` to avoid confusing terminology overloading.

Al12rs:
I can no longer reproduce the bug where esp files show up as PluginFiles. 
Not sure if it was caused by some other change laying around or if I'm simply not looking at the right moment.

Will still proceed with the renaming to avoid confusion

Al12rs:
Closing this since I can no longer reproduce and the renaming should avoid future confusion.

PR #426, opened 2023-07-11
Rename PluginFile to PluginOrderFile to avoid confusion

Al12rs:
Could no longer reproduce #423


src/Games/NexusMods.Games.BethesdaGameStudios/PluginOrderFile.cs
```diff
@@ -16,7 +16,7 @@
 namespace NexusMods.Games.BethesdaGameStudios;
 
 [JsonName("BethesdaGameStudios.PluginFile")]
```
Aragas, line 18: I think we can rename the Json name also, since we're still at the development phase

Al12rs: Updated it, as long as the others don't disapprove

erri120 (approved):

PR #430, opened 2023-07-13
Fomod nested folder support and other fixes

### LINKED ISSUE #414: Fomod installer doesn't support archives with deeper /fomod directory
Examples:
https://www.nexusmods.com/skyrimspecialedition/mods/173?tab=files
https://www.nexusmods.com/skyrim/mods/17802?tab=files

When the mod archive contains one or more top level folders before `/fomod/ModuleConfig.xml` the fomod installer fails to recognize it as a valid fomod archive and returns Priority.None causing the mod to fail to install. 

The installer should be able to recognize these mods regardless of depth of the /fomod folder as a lot of very popular SSE and SLE mods have such a nested structure.

### LINKED ISSUE #428: Fomod installer fails on some archives due to paths not being sanitized
Reproduce with main Vivid Weathers file https://www.nexusmods.com/skyrimspecialedition/mods/2187?tab=files
Selecting the Summer Season option during installation.

Requires RelativePath sanitization and RelativePaths comparisons (for case insensitivity).

Al12rs:
This adds support to Fomod archives that contain one or more subfolders at the top level.
A few tests were added to cover these cases.

This also fixes a few bugs related to unsensitized paths being used by the fomod installer and analyzer (added a test here as well).

For these purposes a couple of new methods were added to `RelativePath`, with tests as well:

- `EndsWith()`
- `FromUnsanitizedInput()`
- fix #414 
- fix #428


tests/Games/NexusMods.Games.FOMOD.Tests/NexusMods.Games.FOMOD.Tests.csproj
```diff
@@ -18,5 +18,21 @@
...
+          <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        </None>
+
+        <None Update="TestCases\NestedWithImages\random subfolder name\fomod\ModuleConfig.xml">
+          <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        </None>
+
+        <None Update="TestCases\NestedWithImages\random subfolder name\fomod\moduleTitle.png">
+          <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+        </None>
```
erri120, line 36: Similar to `TestCasesPacked`, you can use a glob for this:

```xml
<None Update="TestCases\**">
    <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
</None>
```

src/Games/NexusMods.Games.FOMOD/FomodXmlInstaller.cs
```diff
@@ -48,7 +48,13 @@ public Priority GetPriority(GameInstallation installation, EntityDictionary<Rela
         // we only intend to support xml scripted fomods, this should be good enough
         var stopPattern = new List<string> { "fomod" };
 
-        if (!archiveFiles.TryGetValue(FomodConstants.XmlConfigRelativePath, out var analyzedFile))
+        var xmlFileList = archiveFiles.Keys.Where(x => x.EndsWith(FomodConstants.XmlConfigRelativePath));
```
erri120, line 51: This `IEnumerable` will be enumerated multiple times. Use `.ToArray` instead to prevent multiple enumerations.

tests/Games/NexusMods.Games.FOMOD.Tests/FomodXmlInstallerTests.cs
```diff
@@ -137,6 +137,44 @@ public async Task InstallsFilesSimple_Using7z()
...
     }
+    
+    [Fact]
+    public async Task InstallFilesNestedWithImages()
+    {
+        using var testData = await SetupTestFromDirectoryAsync("NestedWithImages.zip");
+        var installedFiles = (await testData.GetFilesToExtractAsync()).ToArray();
+
+        installedFiles
+            .Cast<IToFile>()
```
erri120, line 148: You should verify that the items are of type `IToFile`:

```csharp
installedFiles
  .Should()
  .AllBeOfType<IToFile>()
  .Which
  .Should()
  //... continue with have count
```

erri120 (changes requested):

tests/Games/NexusMods.Games.FOMOD.Tests/NexusMods.Games.FOMOD.Tests.csproj
Al12rs: I think these might be leftovers that were done automatically from when I created the new test archives. I'll remove it.

src/Games/NexusMods.Games.FOMOD/FomodXmlInstaller.cs
Al12rs: I'll change it use TryGetFirst instead of Where so that I get what I'm looking for directly.

tests/Games/NexusMods.Games.FOMOD.Tests/FomodXmlInstallerTests.cs
Al12rs: `AllBeOfType<IToFile>()` fails the assertion saying that the found type was `ModFiles.FromArchive`, which in turn implements the `IToFile` interface. It doesn't make much sense

erri120: Right, looking at the docs, `AllBeOfType` checks for the **exact** type, which isn't very helpful in this scenario.

Al12rs: `AllBeAssignableTo<>()` works

erri120 (approved):

PR #434, opened 2023-07-17
Fix crash on 7z file extraction to long paths on windows

### LINKED ISSUE #429: "Could not find analysis data for archive" exception on specific mods install
# Bug Report

## Summary
At mod install (tested from CLI) of specific mods archives, the application crashes with the following exception:

```
Unhandled exception: System.InvalidOperationException: Could not find analysis data for archive
   at NexusMods.DataModel.ArchiveInstaller.AddMods(LoadoutId loadoutId, Hash archiveHash, String defaultModName, CancellationToken token) in C:\Dev\Repos\NexusMods.App\src\NexusMods.DataModel\ArchiveInstaller.cs:line 56
   at NexusMods.CLI.Verbs.InstallMod.<>c__DisplayClass6_0.<<Run>b__0>d.MoveNext() in C:\Dev\Repos\NexusMods.App\src\NexusMods.CLI\Verbs\InstallMod.cs:line 44
```

## Steps to reproduce
Clear the databases of the app to make sure that the archive isn't already indexed.
Run the install-mod command with one of the following files with which I was able to reproduce:
- LeanWolfs main file: https://www.nexusmods.com/skyrimspecialedition/mods/2017?tab=files
- XP32 Maximum skeleton main file: https://www.nexusmods.com/skyrimspecialedition/mods/1988?tab=files
The app will crash, but not with the above exception, rather it seems that the 7z process used to extract the archive crashed and that causes the app to crash in turn.

Following attempts to install the archive will trigger the exception above.

### Further details:
Debugging shows that on following attempts at installation of the same file, the failure seems to come from the fact that the application does have an entry in the database for the analysis data for the archive, but as an AnalyzedFile and not an AnalyzedArchive, presumably because file extraction failed on the first run, only leaving the initial file analysis data.

Interestingly they both happen to be Fomods, could be related or not.

Al12rs:
fixes #429 

This implements a workaround for a 7z.exe problem on windows in cases where the path is very long and uses forward slashes.
We normally always use forward slashes internally as windows file systems should support it, but apparently in this particular case 7z requires backwards slashes on windows.

To obtain the proper slashes, the ToNativeSeparators functions were added to the paths library.

### DIFF
--- a/src/NexusMods.Paths/Utilities/PathHelpers.cs
+++ b/src/NexusMods.Paths/Utilities/PathHelpers.cs
@@ -145,6 +145,45 @@ public static class PathHelpers
             ? slice.ToString()
             : RemoveTrailingDirectorySeparator(slice).ToString();
     }
+    
+    /// <summary>
+    /// Replaces all directory separator characters with the
+    /// native directory separator character of the passed OS.
+    /// <remarks>
+    /// Uses `/` on Unix-based systems and `\` on Windows.
+    /// </remarks>
+    /// </summary>
+    public static string ToNativeSeparators(ReadOnlySpan<char> path, IOSInformation os)
+    {
+        if (path.IsEmpty) return string.Empty;
+        if (os.IsWindows)
+        {
+            var buffer = path.Length > 512
+                ? GC.AllocateUninitializedArray<char>(path.Length)
+                : stackalloc char[path.Length]; 
+            path.CopyTo(buffer);
+            return buffer.Replace('/', '\\', buffer).ToString();
+        }
+        else
+        {
+            // Paths with backslashes instead of forward slashes need to be fixed.
+            var buffer = path.Length > 512
+                ? GC.AllocateUninitializedArray<char>(path.Length)
+                : stackalloc char[path.Length];
+
+            var bufferIndex = 0;
+            var previous = '\0';
+
+            for (var pathIndex = 0; pathIndex < path.Length; pathIndex++)
+            {
+                var current = path.DangerousGetReferenceAt(pathIndex);
+                if (previous == '\\' && current == '\\') continue;
+                buffer[bufferIndex++] = current == '\\' ? '/' : current;
+                previous = current;
+            }
+            return buffer.SliceFast(0, bufferIndex).ToString();
+        }
+    }
 
     /// <summary>
     /// Checks for equality between two paths.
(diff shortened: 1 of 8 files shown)


src/NexusMods.Paths/Utilities/PathHelpers.cs
```diff
@@ -145,6 +145,45 @@ public static string Sanitize(ReadOnlySpan<char> path, IOSInformation os)
...
+            var previous = '\0';
+
+            for (var pathIndex = 0; pathIndex < path.Length; pathIndex++)
+            {
+                var current = path.DangerousGetReferenceAt(pathIndex);
+                if (previous == '\\' && current == '\\') continue;
+                buffer[bufferIndex++] = current == '\\' ? '/' : current;
+                previous = current;
+            }
+            return buffer.SliceFast(0, bufferIndex).ToString();
```
erri120, line 184: All inputs this method gets should be assumed sanitized, which means that we only need to do something for Windows:

```csharp
public static string ToNativeSeparators(ReadOnlySpan<char> path, IOSInformation os)
{
    DebugAssertIsSanitized(path, os);
    if (path.IsEmpty) return string.Empty;
    if (os.IsUnix()) return path.ToString();

    var buffer = path.Length > 512
        ? GC.AllocateUninitializedArray<char>(path.Length)
        : stackalloc char[path.Length];
    path.CopyTo(buffer);
    return buffer.Replace('/', '\\', buffer).ToString();
}
```

tests/NexusMods.Paths.Tests/PathHelperTests.cs
```diff
@@ -74,6 +74,28 @@ public void Test_Sanitize(bool isUnix, string input, string expectedOutput)
...
+    [InlineData(true, "/foo/", "/foo/")]
+    [InlineData(false, "/foo/", "\\foo\\")]
+    [InlineData(true, "foo\\bar", "foo/bar")]
+    [InlineData(false, "foo\\bar", "foo\\bar")]
+    [InlineData(true, "foo\\bar\\", "foo/bar/")]
+    [InlineData(false, "foo\\bar\\", "foo\\bar\\")]
+    [InlineData(true, "/foo\\bar", "/foo/bar")]
+    [InlineData(false, "/foo\\bar", "\\foo\\bar")]
+    [InlineData(true, "foo/bar\\", "foo/bar/")]
+    [InlineData(false, "foo/bar\\", "foo\\bar\\")]
```
erri120, line 92: These inputs have to be updated, as they are not sanitized.

erri120 (changes requested):

erri120 (approved):

Issue #372, opened 2023-07-04
ID updates get skipped during high-throughput


erri120:
Our IPC system breaks down during high-throughput activities, like installing a mod, and the `SqliteDataStore.IdChanges` observable is unable to keep up. I found this while working on #362, as I hoped that ID-changes can be used to track when a new loadout-revision was created to re-run diagnostics.

The system works when debugging using breakpoints, which add latency and allow the system additional processing time between ID-updates. I'm unclear whether this is a bug in our code, or a limitation of Sqlite itself, or our entire IPC system.

Tasks:

- [x] #437
- [x] #501
- [ ] #502 
- [ ] fix ID updates being skipped

halgari:
I saw an issue like this once, where the IPC system (the memory mapped file) would say that there was a new ID, but SQLite may not have completely flushed the changes to disk and so the change isn't viewable from other processes yet. I thought I had resolved this via using transactions but it could still be an issue.

The best way to fix this is to probably write a few more tests around it and see if we can get it to break in the tests.

erri120:
This issue is also responsible for some UI bugs we have. When the user installs a mod, our code will quickly update the loadout a bunch of times, which can lead to the ID updates being skipped. Our UI is subscribed to those updates, so not getting any means the UI has to be refreshed manually for the new mods to appear.

erri120:
Batching might be a possible solution for this. Instead of pushing one update at a time, we collect a bunch of updates over a fixed period of time and then push multiple updates as one.

erri120:
Our code also often locks parts of Sqlite completely. This is especially the case with jobs, which get updated and removed frequently, leading to scenarios where a read and write happen at the same time. This slows down the entire thing, and Sqlite will throw an exception after a timeout.

erri120:
Adding a `CommandTimeout` to the `UpdateProgress` method will result in `SQLite Error 5: 'database is locked'`.

PR #437, opened 2023-07-18
Prevent database locking in reader loop

erri120:
Part 1 of #372.

Inside the `SqliteIPC.ReaderLoop` method, we're calling [`ProcessJobs`](https://github.com/Nexus-Mods/NexusMods.App/blob/e1455ec7aaef2faf3c8225cdbd3c158052496524/src/NexusMods.DataModel/Interprocess/SqliteIPC.cs#L239) which gets every job from the IPC DB and then updates or inserts the jobs into the `_jobs` `SourceCache`. The issue is this section:

https://github.com/Nexus-Mods/NexusMods.App/blob/e1455ec7aaef2faf3c8225cdbd3c158052496524/src/NexusMods.DataModel/Interprocess/SqliteIPC.cs#L263-L271

While reading from the database, we're updating the progress of the job, which will try to update the progress in the database:

https://github.com/Nexus-Mods/NexusMods.App/blob/e1455ec7aaef2faf3c8225cdbd3c158052496524/src/NexusMods.DataModel/Interprocess/Jobs/InterprocessJob.cs#L72-L81

However, this is unintended behavior. The fix is to add a check and only update the DB if we actually own the job.

I also a small performance issue:

https://github.com/Nexus-Mods/NexusMods.App/blob/e1455ec7aaef2faf3c8225cdbd3c158052496524/src/NexusMods.DataModel/Interprocess/SqliteIPC.cs#L254-L256

We're using `GetBytes` to get the amount of bytes the ID takes up, then we allocate a new array with the required length and read the bytes into the array. However, Job IDs are always 16-bytes long and we can simply use `GetBlob`, which returns a `ReadOnlySpan<byte>`, and pass that directly to the `Guid` constructor. This completely removes all allocations.

### DIFF
--- a/src/NexusMods.DataModel/Interprocess/Jobs/InterprocessJob.cs
+++ b/src/NexusMods.DataModel/Interprocess/Jobs/InterprocessJob.cs
@@ -12,6 +12,7 @@ public class InterprocessJob : IInterprocessJob
 {
     private readonly IInterprocessJobManager _manager;
     private Percent _progress;
+
     /// <summary>
     /// True if this instance is the "owning" instance of the job, where disposing
     /// the instance will auto-close the job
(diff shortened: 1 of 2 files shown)


Sewer56:
Had a quick read, looks good to me.

Sewer56 (approved):

halgari (approved):

PR #440, opened 2023-07-18
Added: A bunch of .nx related fixes. (fixes #410)

### LINKED ISSUE #410: Memory corruption when NX is used with the app
Enabling the NX backend for the app causes strange memory corruption issues in the app, swapping out to a Zip based archive manager removes these crashes. 

Replication:
* Swap the ArchiveManager out via DI
* Run the DataModel tests 6 or so times, you should see a hard crash about half the time, I noticed it more on Linux/OSX

Sewer56:
# Fixes

- Multiple bug fixes in .nx Archive library (updated to `0.3.3-preview`).  
- Removed unnecessary `.nx` dependencies in projects that don't depend on it.  
- Reduced `stackalloc` from 64-bytes to just needed amount in `UpdateIndexes`.  

# Breaking Changes

- Changed to using `V1` instead of `V0` file entry format.
  - This will allow for 4GB+ files to be used in stored mods.
  - Delete your DataStore after merging, due to incompatibility of changed format.
  - We could potentially save some space, here,

### DIFF
--- a/src/NexusMods.DataModel/NexusMods.DataModel.csproj
+++ b/src/NexusMods.DataModel/NexusMods.DataModel.csproj
@@ -3,6 +3,7 @@
     <Import Project="$([MSBuild]::GetPathOfFileAbove('NuGet.Build.props', '$(MSBuildThisFileDirectory)../'))" />
 
     <ItemGroup>
+        <ProjectReference Include="..\..\NexusMods.Archives.Nx\NexusMods.Archives.Nx\NexusMods.Archives.Nx.csproj" />
         <ProjectReference Include="..\ArchiveManagement\NexusMods.FileExtractor\NexusMods.FileExtractor.csproj" />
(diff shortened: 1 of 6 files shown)


halgari (approved):

src/NexusMods.DataModel/NexusMods.DataModel.csproj
```diff
@@ -3,6 +3,7 @@
     <Import Project="$([MSBuild]::GetPathOfFileAbove('NuGet.Build.props', '$(MSBuildThisFileDirectory)../'))" />
 
     <ItemGroup>
+        <ProjectReference Include="..\..\NexusMods.Archives.Nx\NexusMods.Archives.Nx\NexusMods.Archives.Nx.csproj" />
```
erri120, line 6: I don't know why CI doesn't fail, but this project doesn't exist.

Sewer56:
ooooops

PR #445, opened 2023-07-19
Fix broken version parsing for SMAPI Manifests

erri120:
Changes:
- fixes comments breaking JSON serialization for SMAPI manifests
- adds `SMAPIVersion` that correctly parses version numbers in SMAPI manifests


Al12rs (approved):
It looks fine to me

PR #458, opened 2023-07-20
Log the 7z.exe output to have more info on extract failures

Al12rs:


erri120 (changes requested):
Since we only want logging information when this fails, `PipeTarget.ToDelegate` should write to a `StringBuilder`:

```csharp
var sb = new StringBuilder();

//...
PipeTarget.ToDelegate(line =>
{
  if (string.IsNullOrWhiteSpace(line)) return;
  sb.Append(line);
  //...
});
```
We can log the contents of the `StringBuilder` if the result indicates an error:

```csharp
if (result.ExitCode != 0)
{
  _logger.LogInformation("Output from the extractor, trying to extract {File}:\n{Output}", file, sb.ToString());
  //...
}
```

We would sacrifice some performance for better logging, however, this is going to keep our logs more concise.

Al12rs:
Good idea, I'll get on that

erri120 (approved):

PR #466, opened 2023-07-24
Also log stdErr for SevenZip extractor in case of failure.

Al12rs:


src/ArchiveManagement/NexusMods.FileExtractor/Extractors/SevenZipExtractor.cs
```diff
@@ -141,7 +142,7 @@ private async Task ExtractAllAsync_Impl(IStreamFactory sFn, AbsolutePath destina
                 .WithStandardOutputPipe(PipeTarget.ToDelegate(line =>
                 {
                     if (string.IsNullOrWhiteSpace(line)) return;
-                    processOutput.Append($"[7z.exe] {line} \n");
+                    processStdOutput.Append($"[7z.exe stdout] {line} \n");
```
erri120, line 145: You can use [`StringBuilder.AppendLine`](https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.appendline?view=net-7.0#system-text-stringbuilder-appendline(system-string)). It's also probably best to just use `7z` since the App is cross-platform:

```csharp
processStdOutput.AppendLine($"[7z stdout] {line}");

processStdError.AppendLine($"[7z stderr] {line}");
```

erri120 (changes requested):

erri120 (approved):

PR #468, opened 2023-07-24
Improve Apply cli command help description

### LINKED ISSUE #415: Apply CLI command is confusing
To actually have files deployed to the game folder the Apply command requires the `--run` which is confusing.

Run also currently doesn't actually run any program.

Either the documentation needs clarification or the behaviour changed to be more intuitive, perhaps adding another option that does the deployment, separate from the run option.

```
nexusmods.app.exe apply -h
Options:
  -l, --loadout <My Loadout 1>  Loadout to apply
  -r, --run                     Run the application? (defaults to just printing the steps)
  -s, --summary                 Print the summary, not the detailed step list
  -?, -h, --help                Show help and usage information
```

Al12rs:
Fixes #415 
Let me know if this version is good


erri120 (approved):
Fine for now. We'll probably do a pass on the entire CLI before release anyways.

PR #469, opened 2023-07-24
Select home button on startup

### LINKED ISSUE #373: Home isn't selected on startup
On startup, the "Home" view isn't selected automatically. This means that the three buttons "Newsfeed", "My Games" and "Browse Games" are visible, but not clickable. The user first has to click on the home icon in the spine to be able to click on the other buttons.

Al12rs:
Fixes #373 

The LeftMenu is initially set to a DesignViewModel, which isn't interactable, that is what was happening before.

This just changes the view model to the correct one and activates the home button inside WhenActivated().

### DIFF
--- a/src/NexusMods.App.UI/Controls/Spine/SpineViewModel.cs
+++ b/src/NexusMods.App.UI/Controls/Spine/SpineViewModel.cs
@@ -93,25 +93,11 @@ public class SpineViewModel : AViewModel<ISpineViewModel>, ISpineViewModel
                 .SubscribeWithErrorLogging(logger)
                 .DisposeWith(disposables);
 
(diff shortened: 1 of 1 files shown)


erri120 (changes requested):
The commands that are being created inside the constructor should be updated. Instead of doing `ReactiveCommand.Create(() => //...)` they should point to a member method `Home.Click = ReactiveCommand.Create(NavigateToHome)`.

Al12rs:
> The commands that are being created inside the constructor should be updated. Instead of doing `ReactiveCommand.Create(() => //...)` they should point to a member method `Home.Click = ReactiveCommand.Create(NavigateToHome)`.

Done

erri120 (approved):

PR #484, opened 2023-07-27
Refactored content sending method and updated tests

halgari:
The content sending logic has been refactored into a separate method called "SendContent" for better maintainability. This simplifies the HTTP response handling section in the "LocalHttpServer.cs" file, as the responsibility to position the stream, create a buffer, and write to the response output stream has been moved to this new helper method. The range of data to be sent is now calculated within this method itself.  Also, included Debug.Assert statement in "AdvancedHttpDownloader.cs" for debugging purpose.

### DIFF
--- a/.github/workflows/clean_environment_tests.yaml
+++ b/.github/workflows/clean_environment_tests.yaml
@@ -29,7 +29,7 @@ jobs:
 
   build-and-test:
     if: github.event_name == 'push' || github.event.pull_request.draft == false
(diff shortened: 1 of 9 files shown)


halgari:
When this finally all passes, I'll write up a description and set it up for review, thus far it's just for getting things to pass CI

halgari:
Merging just to get tests running a bit smoother while we evaluate other options for test runners

PR #493, opened 2023-07-31
Fix how we handle generated file metadata during applying. 

### LINKED ISSUE #411: Fix the handling of plugins.txt with Skyrim/SkyrimSE
# Bug Report

Feel free to ping @halgari for help with this

## Summary
From Slack:

I noticed that I was failing the SSE game test and spent a few hours trying to figure out...on my branch but worked fine on main.
I finally isolated the problem to me fixing the path of plugins.txt so that it would point to appdata, but I couldn't find any inconsistencies with the paths in...both success and failure cases I finally realized that the problem was that the test fails if Plugins.txt already exists then the test fails due to trying to get metadata for the existing file and ending up in a not yet implemented code path.
Removing plugins.txt passes the test

halgari:
Fixes #411 - in a somewhat temporary way. Fixes how we get metadata for generated files. This fixes how generated files work during applying but not during ingesting. Ingesting generated files is part of a future task

### DIFF
--- a/src/NexusMods.DataModel/Loadouts/LoadoutSyncronizer.cs
+++ b/src/NexusMods.DataModel/Loadouts/LoadoutSyncronizer.cs
@@ -167,11 +167,23 @@ public class LoadoutSynchronizer
 
             if (flattenedLoadout.TryGetValue(gamePath, out var planned))
             {
-                var planMetadata = await GetMetaData(planned.File, existing.Path);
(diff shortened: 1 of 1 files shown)


Al12rs (approved):
The `continue` in the middle of the switch was confusing at first, had to look up to make sure it didn't interact with the switch statement in C# for some reason.

But yeah, should fix the crash on GetMetaData()

erri120 (approved):

PR #500, opened 2023-08-02
Added: Fully Featured support for Download Suspend & Resume (fixes #255 fixes #480)

### LINKED ISSUE #255: Allow users to pause downloads

### LINKED ISSUE #480: Improve DownloaderService
From our meeting notes:

* Make DownloadService fully stateful (have it reference IDataStore)
* DownloadJobs (in ITaskState) can be a single Job
* Download manager should analyze/archive downloads
* Convert IDownloadTask base into Entity and hand it to : coll.AddSingleton<JsonConverter,AbstractClassConverterFactory<IFileAnalysisData>>();
* Serialize as JSON
* Use ReadonlyObservableCollections on the DownloadService (connect IPC) index on IID
* Remove Loadout from the DownloadService
* Move the installer code into the UI

Sewer56:
# Changelog

## Download Persistence Mechanism

Download state is saved at the following moments:  

- Main Window Close [user closes app].  
- Download Task Started [download is idling in queue/background].  
- Download Started [adds download info including size, etc.].  
- On Suspend (pause).  

As a result, current state is fully persisted between reboots; for example, this screenshot shows the fact the App remembered downloaded amount, size, game name and other details without having to restart the download.

NexusMods App_hkZciTOnre

In the event of a crash, the downloads can successfully be resumed too; however will not show up to date completion status (number of downloaded bytes) as a crashed app had no way to persist them before closing.

This works for all current downloader implementations.

All state [including completed downloads] is persisted as JSON, in DataStore under `DownloadStates` category.

## UI Additions

- The pause/play button for individual items has been styled to match behaviour of other icon buttons in the UI.
  - Gray on hover, smooth shrink to 0.9 scale on click.
- All other remaining frontend buttons (Pause, Pause All) in toolbar were rigged to the backend/VM.  

## Backend Changes & Refactorings

- `DownloadService` should now be thread safe w.r.t. Task Initialisation.
- Observable Change Sets were introduced to `DownloadService`'s Public API.
  - With a private `ReadOnlyObservableCollection` for some internal logic that requires fetching all elements.
  - Saved some lines of code in `InProgressViewModel` as a result.
- Jobs are now hidden away from `IDownloadTask` interface; only the info needed from them is now exposed.
- `DownloadService` now handles archive analysis.
- App/UI now handles loadout selection and mod installation.
  - App can now 'suggest' recommended loadouts, for installing mods from download tasks which expose 'Game' (domain) property via `IHaveGameName`.
  - For now, installs to first suggested or available loadout (until we get designs for loadout picker). 

## Note

For now, resuming a download will restart it from 0 (in the UI). 

This is not a bug in the code for this PR; but is on the downloader side; as the same path is successfully passed to downloader between runs.

I don't consider it a 'bug' on downloader side either however we don't currently have a clearly defined strategy for download termination. For example, we have no way of specifying if a download task that's cancelled via CancellationToken should be truly cancelled (i.e. files deleted) or simply suspended for later use.

I will open an issue for this soon.

I have not confirmed whether progress resetting to 0 currently (as download suspends without deleting), is a bug where the download actually resets; or whether it's a reporting error related to the Job the download exposes.


erri120 (changes requested):
https://github.com/Nexus-Mods/NexusMods.App/assets/16577545/249a5654-91ad-471b-9068-8cde68a788db

There is an issue with the UI and the buttons. They aren't clickable some of the time.

src/Networking/NexusMods.Networking.Downloaders/Tasks/State/DownloaderState.cs
```diff
@@ -0,0 +1,117 @@
...
+    /// Size of the file being downloaded in bytes. A value of less than 0 means size is unknown.
+    /// </summary>
+    /// <remarks>Provided by <see cref="IHaveFileSize"/> trait.</remarks>
+    public long? SizeBytes { get; private set; }
+
+    /// <summary>
+    /// Version of the mod; can sometimes be arbitrary and not follow SemVer or any standard.
+    /// </summary>
+    /// <remarks>Provided by <see cref="IHaveDownloadVersion"/> trait.</remarks>
+    public string? Version { get; private set; }
```
halgari, line 76: if we don't set these members outside of the `Create` we should be using `required` and `init` vs `private set`. This helps enforce immutability.

src/NexusMods.App/Listeners/NxmRpcListener.cs
```diff
@@ -39,12 +39,12 @@ private async Task ListenAsync()
...
         {
             var item = await _urlsBlock.ReceiveAsync(_cancellationSource.Token);
-            
+
             // The following code is wrapped in an exception handler such that on possible unhandled exception,
             // the service keeps running rather than terminated.
             try
             {
-                _downloadService.AddNxmTask(item.Value);
+                _ = _downloadService.AddNxmTask(item.Value);
```
halgari, line 47: This seems odd to me, we should either name this return variable or not assign it.

halgari (changes requested):

src/Networking/NexusMods.Networking.Downloaders/Tasks/State/DownloaderState.cs
Sewer56: Resolved in 8180e754007c1696c438a84853f162f171bad718

src/NexusMods.App/Listeners/NxmRpcListener.cs
Sewer56: If you have any stylistic preference here, let me know.

PR #501, opened 2023-08-02
Publish Job updates via the IPC

erri120:
Related to #372.

Job updates via the IPC were never published. We only added new jobs or removed completed ones. This PR allows job updates to be observed. I've added some tests for a single-thread and multi-thread scenario. These tests have a high enough timeout for the IPC reader loop to pick up the changes and notify the subscribers.


halgari (approved):

halgari:
Good catch, I wonder why I hadn't seen this before. Must have been how we used to call `ProcessJobs` or something, at any rate this will make the code propagate job info much more quickly

PR #502, opened 2023-08-02
Sqlite IPC Rework

### LINKED ISSUE #372: ID updates get skipped during high-throughput

erri120:
Fixes #372. This is a follow-up to #501.

**Important Changes**:

- Use in-built connection pool, instead of our custom one for the IPC.
- Add `WriterLoop` and batching.
- Notify subscribers in-process about immediate changes of in-process jobs and messages.
- Optimize the hell out of the IPC.

Previously, sending a message, creating, update or deleting a job would always write to the database immediately. This caused a lot of lock contention in the database and slowed things down, as threads had to wait for the database to be available again.

This PR changes how we write to the IPC database. Instead of writing to it immediately, the job/message will be put into a queue. The `WriterLoop` will run periodically and batch insert/update/delete the items from the queue into the database. This solves the database locking issue.

The second issue is updates being skipped (#372). This was also caused by locking. The `ReaderLoop` could wait multiple seconds before it was able to read the database again. During this time, inserts/updates/deletes could occur, which the `ReaderLoop` would be unable to pick up. Thus, we skipped updates.

While the locking issue was resolved using batching, this introduced another issue. If a job gets created, updated and deleted within the sleep period of the `WaiterLoop`, all those changes will be batch inserted at once into the database. The `ReaderLoop` would only see the final flattened result, which is either going to be the last update or the last delete. With our current data model, this issue can't be resolved completely. However, it can be resolved for the _current_ process only.

Jobs and messages that are created in the current process will be broadcasted to all subscribers in the current process. This means that only jobs that come from another process are affected by the "update skipping" behavior.

**Misc Changes**:

- Updated the table definitions:
  - Using `TEXT` instead of `VARCHAR` (`VARCHAR` doesn't exist in Sqlite)
  - Force ascending aliased row ID values for message IDs.

### DIFF
--- a/src/NexusMods.DataModel/Interprocess/SqliteIPC.cs
+++ b/src/NexusMods.DataModel/Interprocess/SqliteIPC.cs
@@ -228,7 +228,7 @@ public class SqliteIPC : IDisposable, IInterprocessJobManager
         }
 
         using var cmd = conn.Value.CreateCommand();
-        cmd.CommandText = "CREATE TABLE IF NOT EXISTS Ipc (Id INTEGER PRIMARY KEY AUTOINCREMENT, Queue TEXT, Data BLOB, TimeStamp INTEGER)";
+        cmd.CommandText = "CREATE TABLE IF NOT EXISTS Ipc (Id INTEGER PRIMARY KEY ASC AUTOINCREMENT, Queue TEXT, Data BLOB, TimeStamp INTEGER)";
         cmd.ExecuteNonQuery();
 
         using var cmd2 = conn.Value.CreateCommand();
--- a/src/NexusMods.DataModel/Interprocess/SqliteIPC.cs
+++ b/src/NexusMods.DataModel/Interprocess/SqliteIPC.cs
@@ -311,20 +311,23 @@ public class SqliteIPC : IDisposable, IInterprocessJobManager
         try
         {
             _logger.SendingByteMessageToQueue(Size.FromLong(message.Length), queue);
-            using var conn = _pool.RentDisposable();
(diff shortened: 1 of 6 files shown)


src/NexusMods.DataModel/Interprocess/SqliteIPC.cs
```diff
@@ -207,14 +207,20 @@ private long ProcessMessages(long lastId)
...
             }
         }
         catch (Exception ex)
         {
             _logger.LogError(ex, "Failed to process messages after {LastId}", lastId);
         }
 
+        while (_updates.TryPop(out var update))
+        {
+            _subject.OnNext(update);
```
Al12rs, line 221: If I'm not mistaken this is effectively reversing the order of updates. Was the code already processing from newer to older and so this makes it right or will observers be notified of newer stuff first this way?

erri120: Good catch, I forgot about LIFO/FIFO for a moment.

halgari (approved):

PR #505, opened 2023-08-02
Support for SKSE launching when running Skyrim

halgari:
Modifies the RunGameTool so that it supports launching SKSE instead of Skyrim. Also waits for the game itself ot exit after running the launcher.

### DIFF
--- a/src/Games/NexusMods.Games.BethesdaGameStudios/Services.cs
+++ b/src/Games/NexusMods.Games.BethesdaGameStudios/Services.cs
@@ -12,8 +12,8 @@ public static class Services
     {
         services.AddAllSingleton<IGame, SkyrimSpecialEdition>();
         services.AddAllSingleton<IGame, SkyrimLegendaryEdition>();
-        services.AddSingleton<ITool, RunGameTool<SkyrimLegendaryEdition>>();
-        services.AddSingleton<ITool, RunGameTool<SkyrimSpecialEdition>>();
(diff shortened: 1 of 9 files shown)


Sewer56 (approved):
Looks fine to me.

src/Games/NexusMods.Games.BethesdaGameStudios/Services.cs
```diff
@@ -12,8 +12,8 @@ public static IServiceCollection AddBethesdaGameStudios(this IServiceCollection
     {
         services.AddAllSingleton<IGame, SkyrimSpecialEdition>();
         services.AddAllSingleton<IGame, SkyrimLegendaryEdition>();
-        services.AddSingleton<ITool, RunGameTool<SkyrimLegendaryEdition>>();
-        services.AddSingleton<ITool, RunGameTool<SkyrimSpecialEdition>>();
+        services.AddSingleton<ITool, SkyrimLegendaryEditionGameTool>();
```
Al12rs, line 15: @erri120 and I discussed about this the other day, and we don't really like the fact that game specific stuff is added to the DI system, like modInstallers and ITools here. 

All these singletons are floating around the code even if the game isn't even being actively managed by the user.
The code also needs to do some gymnastics to get the objects that are actually relevant for the current game.

It just makes more sense if these things were just owned by the game instead, and so the relevant things could be obtained by querying the game.

If the objects in question are optional (not all games will have/require them) or if we don't want to add an accessor for them in the main IGame interface, we use an approach Like GameCapabilities.

PR #511, opened 2023-08-03
Try and fix flakey tests

halgari:
Two fixes

* Flush .ZIP files before returning from a test function, this might help with cases were we get ".zip is not an archive"
* Move the number of socket retries from 4 -> 16, because the network for the test runners on GH seem super flakey at times

### DIFF
--- a/src/Networking/NexusMods.Networking.HttpDownloader/AdvancedHttpDownloader.cs
+++ b/src/Networking/NexusMods.Networking.HttpDownloader/AdvancedHttpDownloader.cs
@@ -46,7 +46,7 @@ namespace NexusMods.Networking.HttpDownloader
         /// </summary>
         private readonly Size _chunkSize = Size.MB * 128;
 
-        private const int MaxRetries = 4;
+        private const int MaxRetries = 16;
 
         /// <summary>
         /// The size of the buffer used to read from the network stream, no need to make this too large as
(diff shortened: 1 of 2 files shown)


Sewer56 (approved):
LGTM

src/Networking/NexusMods.Networking.HttpDownloader/AdvancedHttpDownloader.cs
```diff
@@ -46,7 +46,7 @@ public class AdvancedHttpDownloader : IHttpDownloader
         /// </summary>
         private readonly Size _chunkSize = Size.MB * 128;
 
-        private const int MaxRetries = 4;
+        private const int MaxRetries = 16;
```
erri120, line 49: We can use the `CI` environment variable to check whether we're running in CI or not. Instead of setting the max retries to `16` for everyone, we can set it to `4` for users and `16` for CI.

halgari: We can, but is there a problem with having a higher retry count on all platforms?

erri120: Probably not. It's just something to keep in mind when tuning other settings for CI specifically.

PR #519, opened 2023-08-07
Cleanup `AdvancedHttpDownloader`

### LINKED ISSUE #512: Reduce logging spam from `AdvancedHttpDownloader`
Reduce logging spam introduced in #481.

erri120:
Changes:

- Resolves #512 by removing unnecessarily verbose logging statements.
- Minor code cleanup.


Al12rs (approved):
Bit perplexed at the change to static, but I guess there is no reason for the methods not to be static. LGTM

PR #528, opened 2023-08-08
Abstraction namespace

halgari:
Working on the auto-updater, I realized I wanted CLI verbs for the updater but didn't want them in the CLI project. After our conversations today about moving interfaces into a generalized abstraction project, I took the liberty to rough this out for the CLI. 

This PR does the following:

* Adds `NexusMods.Abstractions`
* Moves the CLI interfaces into `NexusMods.Abstractions.CLI`
* Reworks all existing verbs to reference these new interfaces (a bit of untangling here was needed as the interfaces can't reference CLI logic)
* Moved one verb `nexus-api-verify` into the project it belongs in, the Nexus API
* Deletes the Configurator class. This class was used to late-bind the Renderer and allow for CLI options to define the renderer. This was never a very clean implementation, so instead I opted for a trait in this case: the verbs implement `IRenderingVerb` and the CLI configurator modifies this renderer before calling the verb. I wish there was a better way to do this, but Rendering can change after the verb is constructed based on the options passed into the CLI parser. What we have now with this patch is clean enough.

The only library required by the Abstractions namespace (thus far) is the Microsoft.Extensions.DependencyInjection.Abstractions library, which itself is lightweight. 

As time goes on we can move other verbs into their proper libraries, but this unblocks my work and introduces an abstraction library which I think is a positive change for the project.


src/NexusMods.CLI/OptionDefinitionExtensions.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace NexusMods.CLI;
+
+public static class OptionDefinitionExtensions
+{
+
+
+}
```
erri120, line 7: Empty class should be removed.

erri120 (changes requested):

src/NexusMods.Abstractions/CLI/OptionDefinition.cs
```diff
@@ -0,0 +1,31 @@
...
+/// <param name="ShortOption"></param>
+/// <param name="LongOption"></param>
+/// <param name="Description"></param>
+public abstract record OptionDefinition(string ShortOption, string LongOption, string Description)
+{
+
+    /// <summary>
+    /// Returns the type this option is parsed into
+    /// </summary>
+    /// <param name="provider"></param>
```
Al12rs, line 15: Other leftover parameter in comment

src/NexusMods.Abstractions/CLI/RegisteredVerb.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace NexusMods.Abstractions.CLI;
+
+/// <summary>
+/// Represents an individual action, e.g. 'Analyze Game'
+/// </summary>
+public class RegisteredVerb
```
Al12rs, line 6: This lives outside the IVerb, AVerb and implementations hierarchy but the comment is identical to IVerb.
Some clarification regarding their relationship could be useful in the comment.
If I read the code right, this something like a wrapper for verbs when registered in the DI system.

Al12rs (commented):
Few comments that need cleanup. The CI has highlighted them

Issue #316, opened 2023-05-11
As a modder, I want to use a FOMOD installer to make mod choices


Greg-Nexus:
# User story

**As a** modder

**I want to** use a FOMOD installer

**So that** I can make customisation choices when installing a mod

## Requirements
We need to support FOMOD installers and present users with a generic UI/UX.

## Design

[Prototype ](https://www.figma.com/proto/Ntr9PttR5VrqpTRwwZHzLu/%F0%9F%93%B1-Mod-list?page-id=1479%3A27452&type=design&node-id=1548-4037&viewport=-520%2C-329%2C0.74&t=R5gbLaVaAKIzHHfh-1&scaling=min-zoom&starting-point-node-id=1548%3A4037&mode=design)

[Figma](https://www.figma.com/file/Ntr9PttR5VrqpTRwwZHzLu/%F0%9F%93%B1-Mod-list?type=design&node-id=1548%3A4036&mode=design&t=6wszPNE5HLzhudKQ-1)

## DOD
_(Describe the expected behaviour from the users point of view. Create multiple DoD's if required to fully describe what should happen. Number each DoD)_

## Scenario 1
**Given** am downloading a FOMOD for Skyrim

**When** I install the FOMOD

**Then** I am prompted to run the FOMOD installer

## Scenario 2
**Given** I am using the FOMOD installer

**When** I complete an installation step

**Then** I am taken to the next FOMOD installation step

## Scenario 3
**Given** I am using the FOMOD installer

**When** I go back

**Then** I am taken to the previous FOMOD installation step

## Scenario 4
**Given** I am using the FOMOD installer

**When** I complete the installation

**Then** I am shown an installation complete message

## Extra

Example FOMODs:

- https://www.nexusmods.com/skyrimspecialedition/mods/659
- https://www.nexusmods.com/skyrimspecialedition/mods/19518
- https://www.nexusmods.com/skyrimspecialedition/mods/2017
- https://www.nexusmods.com/skyrim/mods/17802

halgari:
Mod Organizer 2:

Static Mesh Improvement Mod:
![mo2_1](https://github.com/Nexus-Mods/NexusMods.App/assets/654621/c8c76478-5686-441d-b7b6-4df0529d2763)

Falmer Armors and Weapons Retexture:
![mo2_2](https://github.com/Nexus-Mods/NexusMods.App/assets/654621/735c07a2-2d12-48d4-8287-2a1684cf5a6f)
[...]

PR #541, opened 2023-08-09
Update guided installer abstraction and FOMOD installer

### LINKED ISSUE #544: Can't select many options from many groups in CLI FOMOD
Example: 

Blocked by #541.

erri120:
Followup on #532 and part 2 of #316. Blocked by #540 and #542.

Fixes #544.

**Changes:**

- Complete rework of `IOptionSelector`. We now have the `IGuidedInstaller` interface and more internal types that aren't directly related to FOMODs.
- The CLI installer received massive changes and needs to be tested before merging.
- Contains a new project `NexusMods.Games.FOMOD.UI` that isn't complete yet, but will be worked on in the next PR.

Originally, I wanted to implement the required ViewModels for the FOMOD UI. However, I had to rework some of our existing FOMOD code because the API of the library we're using is pretty bad.


src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
```diff
@@ -69,74 +79,130 @@ public void EndDialog()
...
+        return options.Select(option =>
         {
-            Id = option.id,
-            Name = option.name,
-            Description = option.description,
-            ImageUrl = option.image != null ? AssetUrl.From(option.image) : null,
-            Type = MakeOptionState(option),
-        }).ToArray();
+            var optionId = OptionId.From(Guid.NewGuid());
+            optionIdMappings.Add(new KeyValuePair<int, OptionId>(option.id, optionId));
```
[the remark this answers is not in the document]
erri120: The FOMOD library sends us integers. The `IGuidedInstaller` implementation and everything around it shouldn't be specific for that library. As such, I decided to use GUIDs for IDs and use mappings to convert FOMOD library IDs to our internal ID format.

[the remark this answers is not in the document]
erri120: If we were to add more installers, then we can always map to and from a GUID. If we were to us IDs, we'd be more restricted.

src/NexusMods.Common/GuidedInstaller/ValueObjects/AssetUrl.cs
```diff
@@ -1,10 +1,14 @@
+using JetBrains.Annotations;
 using Vogen;
 
-namespace NexusMods.Common.UserInput;
+namespace NexusMods.Common.GuidedInstaller.ValueObjects;
+
+// TODO: convert this into a discriminated union using OneOf
```
Al12rs, line 6: What do you mean here? What would the alternatives of the union be?

erri120: > URL of an asset that's tied to an option. This can be a relative path, or a URL to a remote resource.

That is the description of `AssetUrl`. It can be:

1) a relative path
2) a URL

As such, you could model this type as a union: `AssetUrl : OneOfBase<RelativePath, URI>`.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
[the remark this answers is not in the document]
Al12rs: I see, the problem with GUIDs is that running the installer again will generate different IDs.
Right now it's not a problem, but it might be inconvenient if we later might want to keep track of user choices.

I see that the previous code was using generics for the ID type. Perhaps that could be an option.

Otherwise for the purpose of making the id persistent across reinstalls, we could try to generate a unique id from discriminatory data of the option (group+name+ID or something like that), using a hash function. But I guess that would be out of scope for this PR.

Thoughts on using generics?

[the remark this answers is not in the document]
erri120: We had generics previously with `Option<TOptionId>` and `OptionGroup<TGroupId, TOptionId>`. This PR removes that. They are a pain to work with, especially when dealing with the VMs. Tracking user choices and making it persistent is out of scope for this PR and #316. If we wanted to add that, we can easily change our code, since the actual value is hidden behind a custom type.

src/NexusMods.Common/GuidedInstaller/ValueObjects/AssetUrl.cs
Al12rs: Ah, I understand now. I missed the RelativePath option. Yeah a OneOf would make this much clearer

src/NexusMods.App/Services.cs
```diff
@@ -48,6 +49,7 @@ public static IServiceCollection AddListeners(this IServiceCollection services)
         services.AddCLI()
             .AddFileSystem()
             .AddUI()
+            // .AddFomodInstallerUi()
```
Al12rs, line 52: ?

erri120: Yup, that's not done yet. I didn't want the PR to explode in size, so I'll work on the VMs in the next PR.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
```diff
@@ -1,11 +1,11 @@
...
-using FomodInstaller.Interface.ui;
 using Microsoft.Extensions.Logging;
-using NexusMods.Common.UserInput;
+using NexusMods.Common.GuidedInstaller;
+using NexusMods.Common.GuidedInstaller.ValueObjects;
 
 namespace NexusMods.Games.FOMOD.CoreDelegates;
 
-public class UiDelegates : IUIDelegates
+public class UiDelegates : FomodInstaller.Interface.ui.IUIDelegates
```
Al12rs, line 8: Any way to avoid the full qualifier each time?

erri120: No, I made those explicit in this file to show what's from the external library and what's ours.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
```diff
@@ -29,17 +29,24 @@ public class UiDelegates : IUIDelegates
     private delegate void ContinueToNextStep(bool forward, int currentStepId);
     private static void DummyContinueToNextStep(bool forward, int currentStepId) { }
 
+    /// <summary>
+    /// Created by the executor, this delegate notifies the executor that the installer
+    /// should exist.
+    /// </summary>
+    private delegate void CancelInstaller();
```
Al12rs, line 36: exist or exit?

erri120: Exit, it's a typo.

Al12rs:
I tested installing a mod from CLI locally, and I found that the experience was rather confusing.

Would it be possible to print a description of the expected operation that the user needs to take depending on the type of option group? (ExactlyOne, AtMostOne, AtLeastOne, Any)

Like: "Select at least one of the following options"

erri120:
> I tested installing a mod from CLI locally, and I found that the experience was rather confusing.
> 
> Would it be possible to print a description of the expected operation that the user needs to take depending on the type of option group? (ExactlyOne, AtMostOne, AtLeastOne, Any)
> 
> Like: "Select at least one of the following options"

The CLI shouldn't be different from `main`. I did change the internal logic and adapted the code to the new system, but the presentation and usage should be the same.

Al12rs:
Also the installer shows "Back" instead of "Next" I think:
image

erri120:
> Also the installer shows "Back" instead of "Next" I think: image

Always has been https://github.com/Nexus-Mods/NexusMods.App/blob/1a21c5d3cb27bf448ed83165a7d8ca0c6744a07f/src/NexusMods.CLI/CliOptionSelector.cs#L12-L16

src/NexusMods.CLI/CliGuidedInstaller.cs
```diff
@@ -0,0 +1,211 @@
...
+
+/// <summary>
+/// Implementation of an option selector for the CLI.
+/// </summary>
+[UsedImplicitly]
+public class CliGuidedInstaller : IGuidedInstaller
+{
+    private const string ReturnInput = "x";
+    private static readonly string[] TableOfOptionsHeaders = { "Key", "State", "Name", "Description" };
+    private static readonly object[] TableOfOptionsFooter = { ReturnInput, "", "Back", "" };
```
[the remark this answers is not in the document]
erri120: Fixed.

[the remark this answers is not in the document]
Al12rs: Now it's like this:

image

It doesn't suggest a way to proceed. This is with LeanWolf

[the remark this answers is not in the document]
erri120: Yeah, working on it. I'll ping you when you can check again.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
```diff
@@ -69,74 +82,138 @@ public void EndDialog()
...
+                    // TODO: verify that this calls EndDialog
+                    _cancelInstaller();
+                } else if (result.IsGoToPreviousStep)
+                {
+                    _continueToNextStep(forward: false, currentStepId);
+                } else if (result.IsGoToNextStep)
                 {
-                    var (selectedGroupId, selectedOptionIds) = tuple;
-                    _selectOptions(currentStepId, selectedGroupId, selectedOptionIds.ToArray());
+                    var selectedOptions = result.AsT2.SelectedOptions;
```
Al12rs, line 115: Use switch case?

src/NexusMods.CLI/CliGuidedInstaller.cs
```diff
@@ -0,0 +1,215 @@
...
+        var currentGroup = installationStep.Groups.Length == 1
+            ? installationStep.Groups[0]
+            : null;
+
+        var selectedOptions = new List<SelectedOption>();
+
+        while (true)
+        {
+            RenderStep(installationStep, currentGroup, selectedOptions);
+            var input = SkipAll ? "0" : GetUserInput();
```
Al12rs, line 51: I'd honestly prefer something that wasn't a number for this, like 's'.
`0` could be typed in by accident when trying to select the first option.

src/NexusMods.CLI/CliGuidedInstaller.cs
```diff
@@ -0,0 +1,215 @@
...
+/// </summary>
+[UsedImplicitly]
+public class CliGuidedInstaller : IGuidedInstaller
+{
+    private const string ReturnInput = "x";
+    private static readonly string[] TableOfOptionsHeaders = { "Key", "State", "Name", "Description" };
+    private static readonly object[] TableOfOptionsFooterBack = { ReturnInput, "", "Back", "" };
+    private static readonly string[] TableOfGroupsHeaders = { "Key", "Group" };
+    private static readonly object[] TableOfGroupsFooterContinue = { ReturnInput, "Continue" };
+    private static readonly object[] TableOfGroupsFooterFinish = { ReturnInput, "Finish" };
```
Al12rs, line 21: I would have a 'n' or 'x' for "Next" and "Continue", and perhaps 'b' for "Back"

src/NexusMods.CLI/CliGuidedInstaller.cs
```diff
@@ -0,0 +1,241 @@
...
+namespace NexusMods.CLI;
+
+/// <summary>
+/// Implementation of an option selector for the CLI.
+/// </summary>
+[UsedImplicitly]
+public class CliGuidedInstaller : IGuidedInstaller
+{
+    private const string CancelInput = "x";
+    private const string BackInput = "b";
```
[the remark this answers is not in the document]
erri120: Having both "Back" and "Previous" would probably be ideal. "Back" is used for the option's table to go _back_ to the group selection, but also in the group's table to go _back_ to the previous step,

src/NexusMods.CLI/CliGuidedInstaller.cs
```diff
@@ -0,0 +1,241 @@
...
+            ).ToList();
+
+        while (true)
+        {
+            if (currentGroup is null)
+            {
+                // if the installation step has multiple groups
+                // the user has to select which group they want to use
+
+                RenderTableOfGroups(installationStep);
```
Al12rs, line 75: So this is always null the first time around. I'm guessing that even if there is just one group, you still show the table to let the user select it first?
The comment might be confusing, but perhaps only because I know how it worked before and that influenced me.

src/NexusMods.CLI/CliGuidedInstaller.cs
```diff
@@ -0,0 +1,241 @@
...
+
+                        selectedOptions.AddRange(requiredOptions);
+
+                        // proceed to the next step
+                        return Task.FromResult(new UserChoice(new UserChoice.GoToNextStep(selectedOptions.ToArray())));
+                    }
+                }
+
+                var groupIndex = ParseNumericalUserInput(input, installationStep.Groups.Length);
+                if (groupIndex < 0) continue;
```
Al12rs: For better code clarity, shouldn't the following code be part of the switch? Either as default case or I think C# has some cool tricks with switch to allow handling the parse output and the index out of ranges as separate cases, but just the default would already be clearer I think.

src/NexusMods.CLI/CliGuidedInstaller.cs
```diff
@@ -0,0 +1,244 @@
...
+    private static string GetUserInput()
+    {
+        return (Console.ReadLine() ?? "").Trim();
+    }
+
+    private static int ParseNumericalUserInput(string input, int upperLimit)
+    {
+        try
+        {
+            var idx = int.Parse(input) - 1;
```
Al12rs, line 233: Add comment to explain the -1
It might be missed at a glance and cause confusion

src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
```diff
@@ -69,74 +86,187 @@ public void EndDialog()
...
-    public void UpdateState(InstallerStep[] installSteps, int currentStepId)
+    public void UpdateState(FomodInstaller.Interface.ui.InstallerStep[] installSteps, int currentStepId)
     {
-        var currentStep = installSteps[currentStepId];
-        var choices = ToChoices(currentStep.optionalFileGroups).ToArray();
+        // NOTE(erri120): The FOMOD library we're using can send us stupid inputs.
+        if (currentStepId < 0 || currentStepId >= installSteps.Length) return;
+
+        // NOTE(erri120): This fuckery is explained further below when we call _selectOptions()
+        if (Interlocked.Read(ref _taskFuckeryState) == WaitingForCallback)
```
Al12rs, line 95: You can leave the comments, but I think the `_taskFuckeryState` variable should have a more expressive name, something like `_taskWaitingState`

Al12rs (changes requested):

Al12rs (approved):

PR #542, opened 2023-08-09
Allow games to set target path for Fomod installer

### LINKED ISSUE #431: Fomod installer targets game folder instead of "Data" for bethesda games
### The Problem:
The fomod installer is generating mods that target the Root folder of whatever the currently managed game is. 
This is usually not the correct target location for the installed mod files.

For bethesda games the target location would be the "Data" folder.

Fomod installers are mainly used for bethesda games, but their popularity for those games has made them see use for others as well (eg. Stalker Anomaly).

### Expected behaviour:
- The fomod installer should be able to install mod files to the correct directories for each game.
- The fomod installer should not be required to have hardcoded game specific information. 
- Uninterested games should not be subjected to a complex interface to implement or at least be offered valid default implementations.
- Would be better if the installer didn't need to have checks everywhere to make sure the current game has a valid implementation of the interface.

Al12rs:
Fixes #431 

Fomod installer now installs to GamePathType.Game by default unless games specify a different path using a capability.

To simplify making changes for all Bethesda games I also added `ABethesdaGame`, and made the two skyrim games inherit from it.

Also did a bit of cleanup of capabilities related code.

! Requires #540 to be merged first (this is already rebased on those changes).


src/Games/NexusMods.Games.BethesdaGameStudios/Capabilities/BethesdaFolderMatchInstallerCapability.cs
```diff
@@ -96,8 +96,9 @@ public class BethesdaFolderMatchInstallerCapability : AFolderMatchInstallerCapab
...
 
-    protected override IEnumerable<InstallFolderTarget> InstallFolderTargets { get; } = new[]
+    /// <inheritdoc />
+    public override IEnumerable<InstallFolderTarget> InstallFolderTargets()
     {
-        GameRootInstallFolderTarget
-    };
+        return new[] { GameRootInstallFolderTarget };
+    }
 }
```
erri120, line 104: Use a static read-only field instead of allocating a new array on every call.

src/NexusMods.App/Program.cs
```diff
@@ -49,6 +50,12 @@ public static async Task<int> Main(string[] args)
 
             return await builder.InvokeAsync(args);
         }
+        else
+        {
+            var selector = host.Services.GetRequiredService<CliOptionSelector>();
+            var renderers = host.Services.GetServices<IRenderer>();
+            selector.Renderer = renderers.FirstOrDefault(r => r.Name == "console") ?? renderers.First();
+        }
```
erri120, line 58: From #540.

src/Games/NexusMods.Games.FOMOD/FomodXmlInstaller.cs
```diff
@@ -40,6 +42,13 @@ public Priority GetPriority(GameInstallation installation, EntityDictionary<Rela
         EntityDictionary<RelativePath, AnalyzedFile> archiveFiles,
         CancellationToken cancellationToken = default)
     {
+
+        // TODO: For now FOMOD installer is enabled for all games and assumes the Game Root folder as the install path.
+        // Consider changing this in the future so games need to opt-in to FOMOD installers and provide a custom install path.
+        var gameTargetPath = gameInstallation.Game.SupportedCapabilities().Values
+            .OfType<AFomodCustomInstallPathCapability>()
+            .FirstOrDefault()?.ModInstallationPath() ?? new GamePath(GameFolderType.Game,"");
```
erri120, line 50: This should do a lookup by ID: `TryGetValue(AFomodCustomInstallPathCapability.CapabilityId, out var capability)`. Afterward, you should do a safe cast: `if (capability is AFomodCustomInstallPathCapability fomodCapability)`. Since this will be used a couple of times, you can make an extension methods out of this:

```csharp
public static bool TryGetCapability<TCapability>(this IDictionary<> capabilities, CapabilityId Id, out TCapability capability);
```

erri120 (changes requested):

Al12rs:
Thanks for reviewing, that's great feedback!

src/NexusMods.DataModel/Games/IGame.cs
```diff
@@ -53,10 +53,10 @@ public interface IGame
...
     public IStreamFactory GameImage { get; }
-    
-    
+
+
     /// <summary>
     /// Collection of <see cref="IGameCapability"/> that this game supports.
     /// </summary>
-    public Dictionary<GameCapabilityId, IGameCapability> SupportedCapabilities { get; }
+    public Dictionary<GameCapabilityId, IGameCapability> SupportedCapabilities();
```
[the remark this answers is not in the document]
Al12rs: Great point, we don't want to populate the dictionary each time, but I wanted to have a cascade structure, where each element in the inheritance could add capabilities. So I guess I'll have to compose the dictionaries instead of just adding the the existing dictionary.

Alternatively, what do you think of using a list instead of a dictionary. We still have the GameCapabilityId inside each capability. We would manly loose out on lookup speed, but would simplify the overall thing.

src/Games/NexusMods.Games.BethesdaGameStudios/ABethesdaGame.cs
```diff
@@ -0,0 +1,31 @@
...
+
+    /// <inheritdoc />
+    public override Dictionary<GameCapabilityId, IGameCapability> SupportedCapabilities()
+    {
+        var capabilities = base.SupportedCapabilities();
+
+        // Support for installing simple Data and GameRoot level mods.
+        capabilities.Add(AFolderMatchInstallerCapability.CapabilityId, new BethesdaFolderMatchInstallerCapability());
+        // Configure FOMOD install to install to Data folder instead of GameRoot.
+        capabilities.Add(AFomodCustomInstallPathCapability.CapabilityId, new FomodDataInstallPathCapability());
```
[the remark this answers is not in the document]
Al12rs: Ah, in case we are replacing a capability of a parent with a more specialized version. Great catch

src/Games/NexusMods.Games.BethesdaGameStudios/Capabilities/BethesdaFolderMatchInstallerCapability.cs
Al12rs: Pushed some changes, to only allocate once. What do you think of the new version?

src/NexusMods.DataModel/Games/IGame.cs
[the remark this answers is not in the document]
Al12rs: I didn't go with the immutable dictionary in the end since it wasn't really worth it considering I planned to update it in each inheritance child. 
Let me know what you think of the new version. 
If we confirm capabilities as the design of choice going forward I think I'll add a CapabilitiesCollection class with Register() and TryGetCapability() methods.

PR #546, opened 2023-08-10
App Updater

halgari:
First steps of the app updater, disabled in the code, but the main parts can be run from the CLI and the unit tests can be run. Description of the service can be found here: https://github.com/Nexus-Mods/NexusMods.App/pull/546/files?short_path=d82896c#diff-d82896c18eb5f8f050e974d17f1209f8e859c7b6a069ab38b0e9ee1513b88ba7

After this is in, I'll release a v0.2.1 version of the app with this included, and then we can test the end-to-end process of updating.


erri120:
Some things to note: the updater should fetch the correct release asset, depending on the [`InstallationMethod`](https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/NexusMods.Common/CompileConstants.cs). The update process will also be different depending on the method and the OS. There should also be an "opt-out" compile constant that completely disables the automatic update checking, since we might publish the App on package systems that do updating for us (winget, choco, AUR).

Sewer56:
I had a look over the code; I have nothing else to add other than what @erri120 already said above.
i.e. In some situations, the App might end up shipping in external package systems.

halgari:
> Some things to note: the updater should fetch the correct release asset, depending on the [`InstallationMethod`](https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/NexusMods.Common/CompileConstants.cs). 

Why would we want inno setup involved on updates just because the original version was installed that way? I don't know enough about Flatpack to comment on that, but with Windows I don't think we want to download and run inno setup for updates.

halgari:
After reviewing this with the team, we're going to go down a different route of showing a dialog and instructing users where to go to update the app themselves

PR #548, opened 2023-08-11
Use explicit commit

halgari:
Another flakey test fix, the idea here is that we might be running into a delay between when a transation ends and a commit is invoked.


erri120:
It might be worth applying the Sqlite IPC optimizations from #502 to the Date Store.

PR #559, opened 2023-08-16
Switch to xxhash64 nuget

### LINKED ISSUE #550: Use external xxHash64 library
The xxHash64 library was moved to https://github.com/Nexus-Mods/NexusMods.Hashing.xxHash64, but our solution still contains the entire code.

halgari:
resolves #550


Directory.Packages.props
erri120, line 1: Sadly, using the in-built package manager of Rider and VS will completely scramble this file. I'd suggest keeping the same formatting as before or merging all groups into one and sort it lexically.

erri120 (changes requested):

Directory.Packages.props
halgari: I'm confused at what you mean by this, looking at the history it doesn't look like this file has changed much much in this PR. What am I missing?

halgari: done

PR #566, opened 2023-08-22
Chunked seekable stream reader

### LINKED ISSUE #565: Add support for seekable read only streams on top of a chunked loader.
In order to allow for just-in-time file analysis (part of tect debt removal) we need a way to get a readonly seekable stream from files in the `ArchiveManager`. 

## Requirements

* Implement a chunked seekable stream reader, use a LRU cache chunked streams
* Add support for the chunked seekable stream reader to the IArchiveManager, have the NX implementation throw (for now)
* Add tests for all the methods

halgari:
resolves #565 

This implements a system for lazy loading of data via chunked streams, main classes implemented in this PR:

* ChunkedStream<T> - takes a IChunkedStreamSource for loading chunks with a size and count specified by the source. ChunkedStream is a seekable, read-only stream that supports both async and sync read methods. Chunks are stored in a LRU cache that defaults to a size of 16. This should allow for random access to be rather fast, as long as the current working set of data fits within the chunks that fit in the LRU cache
* IChunkedStreamSource - a interface for defining a chunked stream, has both sync and async `GetChunk` methods so that each type of threading model remains performant
* LightweightLRUCache - a very lightweight LRU cache that does not support threading in any manner (we don't need it here), but because of that can be a struct embedded in ChunkedStream.

### DIFF
--- a/src/NexusMods.DataModel/ChunkedStreams/IChunkedStreamSource.cs
+++ b/src/NexusMods.DataModel/ChunkedStreams/IChunkedStreamSource.cs
@@ -3,7 +3,8 @@
 namespace NexusMods.DataModel.ChunkedStreams;
 
 /// <summary>
-/// A source of data for a <see cref="ChunkedStream"/>
+/// A source of data for a <see cref="ChunkedStream{T}"/>. Sizes of chunks should be no larger
+/// than 1MB or the chunked stream should be updated to not use a memory pool in those cases.
 /// </summary>
 public interface IChunkedStreamSource
 {
--- a/src/NexusMods.DataModel/ChunkedStreams/IChunkedStreamSource.cs
+++ b/src/NexusMods.DataModel/ChunkedStreams/IChunkedStreamSource.cs
@@ -1,11 +1,11 @@
 ﻿using NexusMods.Paths;
 
-namespace NexusMods.DataModel.ChunkedReaders;
+namespace NexusMods.DataModel.ChunkedStreams;
 
 /// <summary>
-/// A source of data for a <see cref="ChunkedReader{T}"/>
+/// A source of data for a <see cref="ChunkedStream"/>
 /// </summary>
-public interface IChunkedReaderSource
+public interface IChunkedStreamSource
 {
     /// <summary>
     /// Gets the size of the source in bytes.
--- a/src/NexusMods.DataModel/ChunkedStreams/LightweightLRUCache.cs
+++ b/src/NexusMods.DataModel/ChunkedStreams/LightweightLRUCache.cs
@@ -1,4 +1,6 @@
-﻿namespace NexusMods.DataModel.ChunkedStreams;
+﻿using NexusMods.Paths.Extensions;
+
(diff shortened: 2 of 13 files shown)


src/NexusMods.DataModel/ChunkedStreams/LightweightLRUCache.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// <summary>
+    /// Moves the given index to the front of the cache.
+    /// </summary>
+    /// <param name="index"></param>
+    private void Touch(int index)
+    {
+        if (index == 0) return;
+        var key = _keys[index];
+        var val = _values[index];
+        Array.Copy(_keys, 0, _keys, 1, index);
```
Sewer56, line 59: Avoid `Array.Copy` in this code. Because that code is non-generic, it needs to do some extra pre-processing; and we're dealing with small cache sizes.

Use `Buffer.MemMove` private API.

You can either do it with `Buffer.MemoryCopy();` or `_keys.AsSpanFast().CopyTo()`.

Sewer56 (changes requested):
All's good aside from the small notes.

src/NexusMods.DataModel/ChunkedStreams/IChunkedStreamSource.cs
```diff
@@ -0,0 +1,42 @@
+﻿using NexusMods.Paths;
+
+namespace NexusMods.DataModel.ChunkedStreams;
+
+/// <summary>
+/// A source of data for a <see cref="ChunkedStream"/>
```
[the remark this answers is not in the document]
halgari: rofl, it had {T} then I removed the {T}, then I made ChunkedStream generic anyways 🤦

src/NexusMods.DataModel/ChunkedStreams/LightweightLRUCache.cs
halgari: Where are these private API methods at? I can seem to find them.

Sewer56: `Buffer.MemMove` is internal to the runtime. It's basically `memcpy`.

`Buffer.MemoryCopy();` and `_keys.AsSpanFast().CopyTo().` would be two efficient ways to access that method with minimal overhead.

Not sure which one's faster, the second one technically avoids pinning.

Note: `AsSpanFast` needs is an extension used in some of our other projects (e.g. Paths), just grab it from there.

src/NexusMods.DataModel/ChunkedStreams/IChunkedStreamSource.cs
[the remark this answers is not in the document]
Sewer56: :sweat:

Sewer56 (approved):
This is fine for merging in current state.

Do consider eliminating some additional bounds checks with `DangerousGetReferenceAt` where you feel applicable; as it would shave some instructions here and there. That said, all's good now.

PR #570, opened 2023-08-23
GamePath FileTree support

### LINKED ISSUE #553: Add a file tree abstraction for mod installers (and other parts of the app)
# Improvement

## Currently 
Files are passed into installers as a `Dictionary<RelativePath, Hash>` while this is a very simple interface each installer ends up writing their own tree traversal code.

## We want
To consolidate common operations on these entries into a simple data model. This file tree class should:

* Be somewhat path agnostic, should work for relative and absolute paths
* Support:
  * Finding the parent of an entry
  * Finding sibblings of an entry
  * Finding all decendents of an entry
  * Getting information of an entry (hash, size, etc)
* Recommended: keep the tree as a generic, so that different "value entry" types can be provided, so we can quickly convert a `Dictionary<RelativePath, Hash>` or a `Dictionary<GamePath, FileEntry>` to a `FileTree<TPath, TValue>`
* May require implementing a more detailed `IPath` interface

Al12rs:
closes #553 

Adds support for GamePath based FileTrees (`FileTreeNode<GamePath>`).

- Made game path extend `IPath<GamePath>` so the file tree can use it.
- Implemented IPath<> interface methods
- Added tests for new methods
- Added simple test for `FileTreeNode<GamePath>` (Paths library has more extensive testing already).


Al12rs:
Follow up to https://github.com/Nexus-Mods/NexusMods.Paths/pull/11

halgari (approved):

PR #571, opened 2023-08-23
Hide remaining time in UI when paused

### LINKED ISSUE #510: Erroneous Time Remaining Estimate on Paused or Completed Downloads inside InProgressView
# Bug Report

## Summary

When all downloads are paused or completed, the estimated time remaining for downloads is inaccurately reported, showing either extremely high or low values, such as -241132 minutes. 

The current behaviour appears to be related to how the input values for the time remaining calculation are reported to the function.

## Steps to reproduce
1. Start a download  `AdvancedHttpDownloader`.
2. Pause all active downloads or wait until all active downloads are finished.
3. Observe the time remaining estimate for downloads.

When no active downloads are running, the estimated time remaining should be reset to 0, rather than showing extreme values based on last reported values.

## Other information

This issue is not a bug in the actual time remaining calculation, but just a side effect of the way we fetch values via polling. 

Fix here should be relatively trivial, just reporting as 0 in the UI if no downloads are active.

Example(s)

NexusMods App_v3JmmlvrF8

NexusMods App_wHK1vxVA7o

NexusMods App_hkZciTOnre

JanuarySnow:
closes #510 

Hides the UI element text for time remaining when throughput is effectively zero (paused)

Thought it was better hidden rather than saying "0 seconds remaining" which might be a bit misleading.

Normal:

unpaused

Paused:

paused

### DIFF
--- a/src/NexusMods.App.UI/RightContent/Downloads/InProgressCommonViewModel.cs
+++ b/src/NexusMods.App.UI/RightContent/Downloads/InProgressCommonViewModel.cs
@@ -176,7 +176,7 @@ public class InProgressCommonViewModel : AViewModel<IInProgressViewModel>, IInPr
         // Calculate Remaining Time.
         var throughput = Tasks.Sum(x => x.Throughput);
         var remainingBytes = totalSizeBytes - totalDownloadedBytes;
-        SecondsRemaining = (int)(remainingBytes / Math.Max(throughput, 1));
+        SecondsRemaining = throughput < 1.0 ? 0 : (int)(remainingBytes / Math.Max(throughput, 1));
     }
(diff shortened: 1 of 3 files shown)


Sewer56:
CC @captainsandypants (<= UX/design).

-----

The code part is fine, because paused jobs return AdditiveIdentity as their throughput (i.e. 0).

Sewer56 (approved):
Code is OK, however please don't merge without input from design.

JanuarySnow:
Updated test code to check for the empty string instead of the numerical value when downloads are paused

Al12rs:
Thanks for the contribution!

Issue #489, opened 2023-07-27
Installer architecture revamp


erri120:
As pointed out by @Al12rs, our current architecture and handling for installers doesn't allow for customizations. This issue tracks possible solutions to this problem.

## Possible Solution: Interface Markers

Proposed by @Al12rs, this would allow `AGame` instances to implement a marker:

```c#
interface MyMarker { }

class MyGame : AGame, MyMarker
```

Generic installers that work on multiple games can check for this interface. We might have an installer that always extract archive files into the game folder, or one of its subdirectories. This is useful for things like SKSE and ENB. We would create an interface `RootDirectoryInstallerCompatible` which our `AGame` instance for Skyrim would implement. This interface could have additional methods like `GetInstallationDirectory` to allow for some configuration.

## Possible Solution: Capabilities using value objects

Proposed by me, @erri120, as an alternative to interface markers. The drawbacks of using interface markers is that you don't know they exist unless you closely follow documentation or have seen the source code. I wrote a plugin for a C# project once that had this exact issue, where you could choose from dozens of interfaces to implement, and you had almost no idea which one to choose.

Instead of using interfaces, we could make use of value objects:

```csharp
[ValueObject<string>]
public readonly struct Capability { }

class MyInstaller
{
    public static readonly Capability RequiredCapability => Capability.From("some-unique-identifier");
}

class MyGame
{
    public readonly Capability[] Capabilities => { MyInstaller.RequiredCapability };
}
```

This has the benefit of allow anything to create a new capability, since it's just a typed-string under the hood. The drawback is that this solution doesn't allow for customization, since you essentially can only have flags.

## Possible Solution: Make the Game call the Installer

Proposed by me, @erri120, as something completely different. Currently, the loadout manager gets all registered `IModInstaller` instances from DI. When you install a mod, the loadout manager calls the `GetPriority` method on every mod installer it knows of, and then proceeds to use the installer that returns the highest priority:

```mermaid
flowchart TD
    LoadoutManager[Loadout Manager]
    ModInstaller1[Mod Installer 1]
    ModInstaller2[Mod Installer 2]
    LoadoutManager --> ModInstaller1
    LoadoutManager --> ModInstaller2
```

The core issue we have is that our game implementation has to supply the mod installer with additional information. So why don't we just call the mod installer from the game instead of the loadout manager?

```mermaid
flowchart TD
    LoadoutManager[Loadout Manager]
    Game[Game]
[...]

PR #573, opened 2023-08-24
Installer Architecture Revamp

halgari:
resolves: #489 

Reworks mod installers, so that they are no longer global, but come from a sorted list on the game itself. Also removes the no longer needed capabilities system. Instead now, information for each installer is added to it via the game that creates the installer.


Al12rs (approved):
This looks good.
To note, this creates direct dependencies between each game project and whatever installer projects they are using. 
Once we will get to the point of having games or installers not hosted in the same solution this could become less desirable.

Issue #476, opened 2023-07-25
[UI Design] Show application version somewhere on the UI


Al12rs:
It's important for testing and debugging to have a visible version of the application listed somewhere so that users and tester may immediately know which version of the app they are running.

Proposed places are the top windows bar or a new bottom status bar. Top windows bar could apparently be absent on Linux.

Greg-Nexus:
Related: https://github.com/Nexus-Mods/NexusMods.App/issues/567

Pickysaurus:
The banner at the bottom of the UI shows the version number

Issue #567, opened 2023-08-22
Display alpha build banner


Greg-Nexus:
# Improvement

## Currently 
We do not display anywhere on the App that it is in a pre-alpha build state and general modding use is not yet supported.

## We want
To display a persistent banner at the bottom of the App, reminding users that we are in a pre-alpha build state.

## Supporting Information
Please use the following copy, dynamically pulling through the version number if possible:

> {insert version number} - DEVELOPMENT USE ONLY

Example:

> v0.1 - DEVELOPMENT USE ONLY

## Design
**Note:** please use the text above (not the text in the image), the design styling should be followed.

![image (7)](https://github.com/Nexus-Mods/NexusMods.App/assets/100707075/a246cd3d-5e4d-4b90-85a2-54bf9542a34e)

Greg-Nexus:
Related issue: https://github.com/Nexus-Mods/NexusMods.App/issues/476

This issue solves 476.

PR #574, opened 2023-08-24
Add banner for development builds (#567)

JonathanFeenstra:
This adds a banner at the bottom of the screen to show that the current build is a development/debug build as described in #567. In case of a debug build, it looks like this:

Screenshot 2023-08-24 231037

In case of an official (pre-)release, it shows the product version instead of the "Debug build" text.

For now, the banner is hard-coded into the MainWindow as discussed on Discord. This means the code needs to be removed before making a stable release, or updated to only display the banner if a stable release version is detected. The text is also hard-coded an not localised.

I noticed that Rider automatically made some whitespace changes in IconStyles.xaml. Let me know if I should remove those from the commit.


JonathanFeenstra:
Note that this does not fully resolve #476 since the banner is only shown for development builds. For stable releases, it would still make sense to display the version elsewhere in the UI.

Al12rs (approved):

halgari:
For now, all builds are development, so this is fantastic. Thanks for the contribution!

Issue #452, opened 2023-07-19
Implement the translation backend


halgari:
# Improvement

## Currently 
We have no defined way of implementing localization/internationalization

## We want
A standard framework for implementing these so we can start implementing it in the app and rework previously written code to use it. 

Based on previous discussions we will need:
* Support for `.axaml` file translations
* Support for formatted strings in C# code
* Translations should be pulled from translation resource files (to conform with industry standards)

We will not need:
* Exception name/contents translation
* Log translation (including our high performance logging code)

## Design
See (https://github.com/Nexus-Mods/NexusMods.App/issues/330) and use the approaches there to develop the framework, nothing more, for translation.

PR #587, opened 2023-08-30
Implemented localization support throughout the app

### LINKED ISSUE #453: Implement localization/internationalization throughout the app
# Improvement

## Currently
After (#452) we have the tools we need to implement localization/internationalization throughout the app. Next step is to convert old code. 

## We want
Every UI text element, and every UI facing message needs to be abstracted to go through the translation code. 

What does not need to support translation:
* Game names
* Log messages
* Exceptions
* Class names

We *do not* intend to translate any of this on our own, simply abstract these strings into the resource files. Someone else can translate as they find the need.

Al12rs:
closes #453

- Implemented translations for relevant strings in the app using backed implemented with #452
- Localized strings should update themselves live on language change for the most part, or close enough in some cases. 
- I did not test live updating as we discussed it wasn't a hard requirement.
- Added a way to create avalonia bindings in code for localized strings and added documentation for that use case.
- Added rough but serviceable Italian translation, which I used to test the if translations were working.
- Reduced some of the flakyness of Localization tests by making them sequential and resetting the locale to English after each test.
- The ui related localization test are still flaky due to accessing ui from wrong thread or something.
- Fixed a DownloadTaskViewModel test that didn't work after the localization changes due to there being an additional Poll() call in the viewModel constructor now.

I tried finding all of the string, but I might have missed some.


halgari:
This is a lot of mind-numbing work. Thanks for cranking through all this code.

PR #594, opened 2023-08-31
More Skyrim docs fixes

JonathanFeenstra:
* Change SSEEdit to TES5Edit in LE docs
* Fix local appdata path in SE docs
* Use USSEP as example instead of LE-version in SE docs


Sewer56 (approved):
LGTM, I'll let someone else look before merging.

halgari (approved):

PR #599, opened 2023-09-04
Update docs regarding translation live updates.

Al12rs:


Al12rs:
This is no longer true as for decision to move to static localization: #601

PR #608, opened 2023-09-05
FOMOD validation for CLI

Al12rs:
- Added static GuidedInstallerValidation class containing two static validation methods, one for a single group, one for an entire step.

- Adds checks before proceeding with steps to make sure each group selection respects the requirements of the group (At least one, at most one, exactly one).

- Fixed Fomod installer creating GamePaths using unsanitized paths.

### DIFF
--- a/src/NexusMods.Common/GuidedInstaller/GuidedInstallationStepValidator.cs
+++ b/src/NexusMods.Common/GuidedInstaller/GuidedInstallationStepValidator.cs
@@ -17,17 +17,14 @@ public static class GuidedInstallerValidation
     /// <param name="installationStep">The installation step containing all the groups</param>
     /// <param name="selectedOptions">A collection of <see cref="SelectedOption"/> for all the groups.</param>
     /// <returns>
-    /// A collection of <see cref="GroupId"/>s that have invalid selections.
-    /// An empty collection means all groups have valid selections.
+    /// An array of <see cref="GroupId"/>s that have invalid selections.
+    /// An empty array means all groups have valid selections.
     /// </returns>
-    public static IEnumerable<GroupId> ValidateStepSelections(GuidedInstallationStep installationStep,
-        IEnumerable<SelectedOption> selectedOptions)
+    public static GroupId[] ValidateStepSelections(GuidedInstallationStep installationStep,
+        IReadOnlyCollection<SelectedOption> selectedOptions)
     {
-        var selectedOptionsList = selectedOptions.ToArray();
-
-        return (from @group in installationStep.Groups
-            where !IsValidGroupSelection(@group, selectedOptionsList)
-            select @group.Id).ToList();
+        return installationStep.Groups.Where(group => !IsValidGroupSelection(group, selectedOptions))
+            .Select(group => group.Id).ToArray();
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


src/NexusMods.Common/GuidedInstaller/GuidedInstallationStepValidator.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </returns>
+    public static IEnumerable<GroupId> ValidateStepSelections(GuidedInstallationStep installationStep,
+        IEnumerable<SelectedOption> selectedOptions)
+    {
+        var selectedOptionsList = selectedOptions.ToArray();
+
+        return (from @group in installationStep.Groups
+            where !IsValidGroupSelection(@group, selectedOptionsList)
+            select @group.Id).ToList();
+    }
```
erri120, line 31: While `IEnumerable<T>` is a nice interface, it often leads to unwanted allocations, as seen by the unnecessary `ToArray` and `ToList` calls. This should return `GroupdId[]` and accept `IReadOnlyCollection<T>` or similar.

For consistency, it'd be better to use extension methods instead of LINQ expressions.

src/NexusMods.Common/GuidedInstaller/GuidedInstallationStepValidator.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// A selection is valid if it satisfies the group's <see cref="OptionGroup.Type"/> requirements.
+    /// </remarks>
+    /// <param name="group"></param>
+    /// <param name="selectedOptions">
+    /// A collection of <see cref="SelectedOption"/>, could contain selections for other groups as well
+    /// </param>
+    /// <returns>True if the group has a valid selection, false otherwise</returns>
+    public static bool IsValidGroupSelection(OptionGroup group, IEnumerable<SelectedOption> selectedOptions)
+    {
+        var selectedOptionsList = selectedOptions.ToArray();
```
erri120, line 46: Similarly, this method can just accept `IReadOnlyCollection<T>` and doesn't need the `ToArray` call.

src/NexusMods.Common/GuidedInstaller/GuidedInstallationStepValidator.cs
```diff
@@ -0,0 +1,61 @@
...
+                selectedOption => selectedOption.GroupId == group.Id) == 1,
+
+            OptionGroupType.AtMostOne =>
+                selectedOptionsList.Count(selectedOption => selectedOption.GroupId == group.Id) <= 1,
+
+            OptionGroupType.AtLeastOne =>
+                selectedOptionsList.Any(selectedOption => selectedOption.GroupId == group.Id),
+
+            _ => true
+        };
```
erri120, line 59: You should write a unit test for the method. This kind of logic can easily be tested.

erri120 (changes requested):

src/NexusMods.Common/GuidedInstaller/GuidedInstallationStepValidator.cs
Al12rs: Good idea

Issue #610, opened 2023-09-05
Advanced installer backend


Greg-Nexus:
## Requirements

In order to deliver https://github.com/Nexus-Mods/NexusMods.App/issues/449, we need to build a backend to handle the advanced installer logic.

This issue should not handle basic UI functionality or styling.

### Issues

1. Backend (this issue)
2. Functionality (https://github.com/Nexus-Mods/NexusMods.App/issues/611)
3. Styling (https://github.com/Nexus-Mods/NexusMods.App/issues/449)

PR #628, opened 2023-09-14
Advanced Installer Documentation & Tasks

Sewer56:
This PR provides the documentation and task breakdown for the behind the scene details for
-  https://github.com/Nexus-Mods/NexusMods.App/issues/610

These are provided as a series of ADRs; which detail the context, decision drivers, requirements and implementation details for each of the sub-tasks that require to be done to complete the aforementioned issue.

The provided tasks are (mostly) dependency free and can be done in parallel; thus are up for grabs.
Text for all of the individual sub-tasks (which will be posted as separate issues upon merge), is available below.

Feedback would be appreciated.

----------

# Issue 1: Implement Base Advanced Installer Design

Approx Time: 2 Days.

Implement the base [Advanced Installer design described here](https://github.com/Nexus-Mods/NexusMods.App/blob/advanced-installer/docs/decisions/backend/0009-advanced-installer-design.md#advanced-installer-design).

## Checklist

- [ ] Auto-Invoke `AdvancedInstaller` if every other installer is no-op (does nothing).

- [ ] `AdvancedInstaller` must work in the absence of suggestion data (in case it is incomplete).

- [ ] `AdvancedInstaller` returns appropriate error if the user attempts to map a path which is already mapped. (This can be an exception, just include extra context data, e.g. target path, which can later be displayed in UI).

- [ ] `AdvancedInstaller` does not give error when mapping an entire folder whereby an item inside this folder has already been mapped. (i.e. If you map file, then folder, do not throw double mapping error)

- [ ] `AdvancedInstaller` emits correct `IEnumerable<AModFile>` operations.

- [ ] Tests cover all aforementioned cases.

----------
# Issue 2: Implement Advanced Installer Suggestion Algorithm

Approx Time: 1-4 hours.

Implement the base [Advanced Installer suggestion algorithm described here](https://github.com/Nexus-Mods/NexusMods.App/blob/advanced-installer/docs/decisions/backend/0010-advanced-installer-suggestions.md).

## Prerequisites:

None. This code can be created and tested before Issue 1. 
It's essentially a method with signature `(string filePath, InstallFolderTarget rootTarget)`.

## Checklist

- [ ] `IGame` exposes `InstallFolderTarget`.
- [ ] Tests cover all aforementioned edge cases.

----------
# Issue 3: Expose `InstallFolderTarget` in `IGame`

Approx Time: 1-2 hours.

In order to use the [Advanced Installer Suggestion Algorithm](https://github.com/Nexus-Mods/NexusMods.App/blob/advanced-installer/docs/decisions/backend/0010-advanced-installer-suggestions.md), we need to expose `InstallFolderTarget` during the deployment step (`GetModsAsync`).

To do this, we extend the `IGame` interface to expose `InstallFolderTarget`(s) for the game's most common directories. Then we populate that interface export with `InstallFolderTarget`(s) used in the game's common mod installer(s).

This interface can be accessed during the deploy step under `GameInstallation` structure.

## Prerequisites

Issue 5.

## Checklist

- [ ] `IGame` exposes `InstallFolderTarget`.
- [ ] Current games are updated to include their `InstallFolderTarget`(s).

----------
# Issue 4: Implement Advanced Installer Persistence Mechanism

Approx Time: 1 day.

Documented here: [Advanced Installer Persistence](https://github.com/Nexus-Mods/NexusMods.App/blob/advanced-installer/docs/decisions/backend/0011-advanced-installer-persistence.md).

## Prerequisites

Issue 1.

This can be *partially* completed before Issue 1. 

For instance you can create the new IDs; and add serialization/deserialization tests. Maybe even use the code for FOMOD persistence until Issue 1 is done.

Completion of this task can be delayed until either Issue 1 is done; or the integration between Issue 1 and Issue 3 can be added as a separate PR.

----------
# Issue 5: Migrate Remaining Games to use `InstallFolderTarget`

Approx Time: 1 day.

Some installers still use more legacy mechanisms for installation, rather than the newer `InstallFolderTarget` of GameCapabilities. In order to leverage the suggestions system for `AdvancedInstaller`, we should migrate these games to use `InstallFolderTarget` where possible. 

## Prerequisites

None


docs/decisions/backend/0009-advanced-installer-design.md
```diff
@@ -0,0 +1,246 @@
...
+│  ├─ model.dae
+│  └─ hair.dds
+Green Hair
+├─ Data
+│  ├─ model.dae
+│  └─ hair.dds
+Blue Hair
+└─ Data
+   ├─ model.dae
+   └─ hair.dds
```
Al12rs, line 77: Mo2 actually has an installer for mods packaged like this, I think it's the BAIN installer. 
It recognizes that multiple folders on the same level contain "valid" files or folder structures, then shows a simple installer ui with a checkbox for each option with a view on the side that shows any readme.txt contents inside each "option" folder.

We could add something like that in the future.

The main purpose of the manual installer is to handle any case where the archive is not recognized, so as you mentioned above, new mod type, wrongly packaged stuff, niece game, new formats, one of a kind mods etc.

docs/decisions/backend/0009-advanced-installer-design.md
```diff
@@ -0,0 +1,246 @@
...
+    /// Value: The relative path where the file should be placed in the game directory.
+    /// </summary>
+    /// <example>
+    /// If a mod file 'texture.dds' in the archive should go to 'Game/Data/Textures/texture.dds',
+    /// then the KeyValuePair would be like ("texture.dds", "Game/Data/Textures/texture.dds").
+    /// </example>
+    /// <remarks>
+    /// Paths follow internal Nexus Mods App path standards: they use a "/" as a separator, trim whitespace, and do not alter "..".
+    /// </remarks>
+    private Dictionary<RelativePath, RelativePath> _archiveToOutputMap { get; init; } = new();
```
Al12rs, line 186: Use `Dictionary<RelativePath, GamePath>` instead. We don't want to just target the game directory. 
Some games might need files to be installed in Appdata or Documents/Mygames or somewhere else.
We can map this with GamePaths. 

For now GamePaths can only target 4 locations, defined by the GameFolderType enum, plan is to expand this in the future. For now check which of the GameFolderTypes were defined by the game (or just add them to all games) and show the ones used as suggested folders.

docs/decisions/backend/0010-advanced-installer-suggestions.md
```diff
@@ -0,0 +1,81 @@
...
+in the following mockup below:
+
+![](./images/0009-advanced-installer-location.png)
+
+## Context Statement
+
+When using the Advanced Installer, the user should be provided with 'hints' dictating where a file might
+require to be placed based on a number of heuristics.
+
+This is part of the UX effort to 'Make Modding Easy' and a general requirement of our [Advanced Installer Design](./0009-advanced-installer-design.md).
```
Al12rs, line 13: The purpose of the Suggested locations area is not to suggest possible locations but rather list *All * the available locations.
"Suggested locations" is just a user facing name for it.

So the objective here isn't to help the user choose, but rather to offer all the locations they might want/need to install the files to. 

Intelligent sorting of the options based on potential knowledge is outside the scope for now.

docs/decisions/backend/0009-advanced-installer-design.md
Sewer56: Good point, seems I missed this when doublechecking the code.

An earlier version of this actually had the GamePath base path as part of Relative path (in the same way we do for configs). But I decided to scrap that for GamePath, it seems I didn't update this code snippet. Let me do that.

docs/decisions/backend/0010-advanced-installer-suggestions.md
```diff
@@ -0,0 +1,81 @@
...
+
+Rather than splitting the metadata in two, we can leverage the existing
+`InstallFolderTarget` system in order to implement Advanced Installer suggestions;
+as that system already has required metadata.
+
+### Consequences
+
+- [Good] Strong code reuse as `InstallFolderTarget` already has required metadata to support this functionality.
+- [Neutral] Each `InstallFolderTarget` folder will now need a description.
+- [Neutral] All games will need to be converted to new `InstallFolderTarget` system.
```
Al12rs, line 30: `InstallFolderTarget` was meant to be used by the Generic Folder installer to recognize folders.
If we unify the two things, we could have games that define InstallFolderTargets that are not actually recognizable.

So either we separate the two concepts with one that just wraps the other adding the information necessary for the Generic Installer to recognize the files, or we keep it a single thing but with a clear way for the Generic installer to know which InstallFolderTarget it should actually use (so perhaps two different methods on IGame returning the two different lists?)

docs/decisions/backend/0010-advanced-installer-suggestions.md
```diff
@@ -0,0 +1,81 @@
...
+tl;dr Include any folders where any `substring` of the in-archive directory path, matches an `InstallFolderTarget`.
+
+## Implementation Algorithm (Files)
+
+The suggestion algorithm for files functions in the following manner:
+
+- Create initial list of 'suggestions' from all `InstallFolderTarget`(s) recursively.
+- Filter by `KnownValidFileExtensions`. (Remove if unknown extension).
+- Filter by `FileExtensionsToDiscard`. (Remove if extension to discard).
+- Add any directories from [Implementation Algorithm (Directories)](#implementation-algorithm-directories)
```
Al12rs, line 74: This is outside the scope for now. A simple static list of available options is enough.

Al12rs: Just making sure it doesn't get pulled in now.

docs/decisions/backend/0010-advanced-installer-suggestions.md
Sewer56: > Game folder should always be suggested, if Appdata and Mygames exist for that game install, they should be suggested as well, even if the game does not declare any InstallFolderTarget for them.

I agree, however the designs don't currently accomodate this, so feedback from Design ( @captainsandypants ) is requested.

The current designs assume the only output folder is game directory. 

e.g. From a style perspective, are we supposed to make separate bordered regions for other types of output paths?

Al12rs: You just add the entry here:
image

For example for starfield you would add MyGames/Starfield as another target location.
For Baldur's Gate 3 you would have Appdata/Larian and the Game folder.

So you might need some short name or description of the Target paths from the Games.

You would add another expandable section here for MyGames, under the SkyrimSpecialEdition section.
image

We discussed this during some of the design meetings we had for this.

docs/decisions/backend/0011-advanced-installer-persistence.md
```diff
@@ -0,0 +1,147 @@
...
+
+After manually installing a mod via the Advanced Installer, the application needs to save the details of the user's
+last set of choices made in the Installer.
+
+This is such that, when the application needs to perform a re-deployment occurs, the end user does not have to go through
+the manual installation steps again.
+
+Note: Re-deployment can happen due to a variety of factors, such as a mod deployed before the Advanced Installed mod
+being removed from a loadout. Not all re-deployments are due to the user's direct actions/expectations. Therefore a
+persistence mechanism here is crucial.
```
Al12rs, line 16: I don't understand, why is this needed?
Don't we already store the archive source and destination in IToFile as a GamePath, what do we need more?

docs/decisions/backend/0010-advanced-installer-suggestions.md
Sewer56: Alright, that works.

docs/decisions/backend/0010-advanced-installer-suggestions.md
Sewer56: I'm not sure that I quite understand the problem here.

The initial set of 'suggested folders' (w/o suggestion algorithm) should consist of:
- All `GamePath` entries.
- All deploy-able folders not in `GamePath` (specified via `InstallFolderTarget`).

Are you suggesting that we might run into a situation where we might want to recommend subfolder(s) to the users which we may not necessarily want the Generic Installer to deploy to?

e.g.

```
Mod
├─ texture.dds
```

Where `texture.dds` should go to a folder like `{Game}/Data/Textures`, but `{Game}/Data/Textures` is not considered a folder under which the Generic Installer might try deploying to?

That doesn't happen often, but I'm happy to add an interative task to export the needed properties as an interface (trait), and expose another getter from `IGame` which provides custom suggestions that are not used as part of the deployment step. [And update the documentation accordingly]

docs/decisions/backend/0011-advanced-installer-persistence.md
Sewer56: It's because of the inherent design of our Data Store.

Suppose we have a loadout of mods like the following (arrows indicate order in which they were added):

```mermaid
flowchart LR
    A[Mod A] --> B[Mod B]
    B --> C[FomodOrAdvanced]
    C --> D[Mod D]
```

Where `FomodOrAdvanced` is our mod in question.

If the user removes `Mod B`, the DataStore performs a rollback to before `Mod B` was added, so the flowchart becomes

```mermaid
flowchart LR
    A[Mod A]
```

and then, the remaining items are then freshly 're-deployed' from scratch on top of `Mod A`.

```mermaid
flowchart LR
    A[Mod A]  --> C[FomodOrAdvanced]
    C --> D[Mod D]
```

This means that for all mods after `Mod A`, we have to drop existing information and re-run all of the installers from scratch (i.e. this accommodates for things like FOMOD file based dependencies). 

So for `FomodOrAdvanced` and `Mod D`, we end up having to run all of the installers again (`GetModsAsync`); which means that the user will have to make their Advanced Installer (and/or FOMOD) choices from scratch again.

Al12rs: No, we don't ask users to re-run installers in this case. 
I think we need to have a discussion about this because I don't think we are on the same page on how the main app should handle mods and deployment.

Each installed mod is a collection of file sources and destinations. The source will be a .NX archive containing all the downloaded mod files, destination is where the file needs to be deployed.

If a user removes a mod this doesn't really influence all the other mods installed after it. Each mod is an independent entity. 
Deployment phase just takes the list of files from each enabled mod, if there are two mods wanting to install different files to the same file path, then we solve this using some user defined priority or conflict resolution rules.

This is all virtual, then when the user wants to actually play the game, we try to match the filesystem state with our virtual loadout state (Adding new files, removing files that don't appear in the current loadout (backing them up in NX archive if we don't already have their hash in the db).

Sewer56: Seems you're right, my knowledge of the internals is outdated in this scenario. I had a walk through the source to confirm.
I'll adjust the PR accordingly.

docs/decisions/backend/0010-advanced-installer-suggestions.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Acquiring `InstallFolderTarget` During Deploy Step (a.k.a. `GetModsAsync`)
+
+Extend the `IGame` interface to expose `InstallFolderTarget` for the game's most common directories.
+Then populate that with `InstallFolderTarget`(s) used in the game's common mod installers.
+
+This interface can be accessed during the deploy step under `GameInstallation` structure.
```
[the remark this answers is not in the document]
Sewer56: This should hopefully be addressed with latest commit through the use of abstraction (`ISuggestionFolder`)

docs/decisions/backend/0011-advanced-installer-persistence.md
Sewer56: Resolved in ADR.

Sewer56:
Updated the OP to reflect latest commit.

docs/decisions/backend/0010-advanced-installer-suggestions.md
Sewer56: Made this explicit in OP and ADR text.

docs/decisions/backend/0009-advanced-installer-design.md
```diff
@@ -0,0 +1,246 @@
...
+logic does not translate well to `Advanced Installer`,
+
+On the other hand, a good chunk of the UI styling/logic should be reusable.
+In other words, we're going for a custom solution.
+
+### Consequences
+
+- Good, because custom solution allows for easier serialization of the data.
+- Good for maintenance, as the logic for persisting deployment steps could be reused in the FOMOD code.
+    - Note: Both are a file->file mapping effectively, so we can use serialization format for Advanced Installer in FOMOD.
```
Al12rs, line 109: The deployment steps are already shared between all mods, mod installation has no baring on it. Installer just returns a list of source/target mappings and everything else is handled by common code.

docs/decisions/backend/0009-advanced-installer-design.md
```diff
@@ -0,0 +1,246 @@
...
+
+  As part of our slogan named 'Make Modding Easy', it's important that the finished product is easy to use.
+
+  Namely, we have to consider that non-technical users may be using the app, and therefore we must make sure
+  to provide appropriate hints (suggestions) for them to deploy an unknown mod; even in presence of install
+  instructions from mod authors.
+
+- Must support arbitrary path mapping.
+
+  This system must support mapping any path from the source archive into any location in the target game directorie(s).
```
Al12rs, line 45: To clarify, not just the game install directory, but also other game locations such as Appdata, MyGames etc

docs/decisions/backend/0009-advanced-installer-design.md
```diff
@@ -0,0 +1,246 @@
...
+There may be temptation to dynamically emit the extract (`FromArchive`) instructions as the user selects the install
+target in the UI:
+
+![](./images/0009-advanced-installer-location.png)
+
+This however, is not a good idea; as `undo` operations and unmaps are more difficult to process in a solution like this.
+(Might also be costly in huge mods).
+
+Instead, we should maintain a mapping of `Source` -> `Target`, as a dictionary; and only emit the needed operations 'Just in Time'
+at the deployment stage.
```
Al12rs, line 141: I need some clarification about what you mean with deployment stage

docs/decisions/backend/0010-advanced-installer-suggestions.md
```diff
@@ -0,0 +1,123 @@
...
+
+tl;dr Include any folders where any `substring` of the in-archive directory path, matches an `InstallFolderTarget`.
+
+## Filtering Suggestions (Files)
+
+The filtering algorithm for suggestions where to place a file functions in the following manner:
+
+- Filter by `KnownValidFileExtensions`. (Remove if unknown extension).
+- Filter by `FileExtensionsToDiscard`. (Remove if extension to discard).
+- Add any directories from [Implementation Algorithm (Directories)](#implementation-algorithm-directories)
```
Al12rs, line 123: Remember that the whole point of the manual installer is to allow users to install stuff wherever, we can't filter out options for them. 

Honestly the best thing we could possibly ever do in the future is to sort the options in order by most likely, potentially show that as a special suggested option at the top. 
But all of this should first go through Design, we should not include this ADR as if we already accepted this idea.
I think you should just keep the parts about how to get the target locations, remove everything else as this is not something we discussed or agreed on. 

If you want you can open an issue to request sorting of the target locations, so that design can take a look at it.

docs/decisions/backend/0009-advanced-installer-design.md
Sewer56: Flattening the dictionary into a list of files to deploy in `IModInstaller`

docs/decisions/backend/0010-advanced-installer-suggestions.md
Sewer56: Note: We've decided to put the non-core ADRs as separate issues; so they can be merged with their corresponding implementations later. The ADR text will be preserved inside each issue.

Al12rs (approved):

PR #643, opened 2023-09-18
Load remote and local images for FOMOD

### LINKED ISSUE #614: Rework `AssetUrl` into a discriminated union for different types of assets
Right now, the `AssetUrl` is just a string underneath, but it can be a relative path or a URL to some remote image. A discriminated union would be ideal, as it allows you to have one type that is a union of other types, which you can check at runtime.

This can be done using [`OneOf`](https://github.com/mcintyre321/OneOf) or similar libraries.

The new type should also make loading the data easier. Either integrate with the existing `IStreamFactory` interface or just add a method that returns `byte[]`, so the FOMOD UI can easily create a `Bitmap` out of the data.

erri120:
Fixes #614.


src/Games/NexusMods.Games.FOMOD/CoreDelegates/InstallerDelegates.cs
```diff
@@ -12,12 +13,19 @@ public sealed class InstallerDelegates : ICoreDelegates
     public IContextDelegates context { get; }
     public IIniDelegates ini => throw new NotImplementedException();
     public IPluginDelegates plugin { get; }
-    public IUIDelegates ui { get; }
 
-    public InstallerDelegates(ILoggerFactory loggerFactory, IGuidedInstaller guidedInstaller)
+    public IUIDelegates ui => UiDelegates;
+    public readonly UiDelegates UiDelegates;
```
Al12rs, line 18: Why was this needed?

Al12rs:
Apart from the question above, everything else looked good to me, but some parts will require rewriting when #615 is merged.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/InstallerDelegates.cs
erri120: The FOMOD installer needs to set a field inside the `UiDelegates` class.

Al12rs (approved):

PR #644, opened 2023-09-18
Updater functionality

### LINKED ISSUE #465: As a user, I want to update the Nexus Mods App
# User story

**As a** user

**I want to** update to the latest version of the App

**So that** I can benefit from new features, bug fixes and improvements

## Requirements
We'll need to add an update mechanism for users, with a UI interface.

Updating the app should download and install the latest version.

## Design

[Figma design for basic first iteration](https://www.figma.com/file/oWcjirFQDdwyeVPGiCUgWP/%F0%9F%93%B1-Update-application?type=design&node-id=0%3A1&mode=design&t=rWBWMF7SjxGeoB48-1)

image

## DOD

## Scenario 1
**Given** that there is an app update available

**When** I start the app

**Then** I see an 'Update available' modal
**And** I see a "[View changelog on GitHub](https://github.com/Nexus-Mods/NexusMods.App/releases)" link

## Scenario 2
**Given** I am viewing the 'Update available' modal

**When** I select "Later"

**Then** the modal closes

## Scenario 3
**Given** I am viewing the 'Update available' modal

**When** I select "Update"

**Then** the app updates to the latest version

halgari:
resolves #465 

Adds the functionality described below. Shows the updater if an update is available and if the the metrics opt-in wasn't shown. This is to not spam the user with multiple overlays. 

updater


src/NexusMods.App.UI/Overlays/Updater/UpdaterView.axaml.cs
```diff
@@ -0,0 +1,54 @@
...
+                .DisposeWith(d);
+
+            this.WhenAnyValue(view => view.ViewModel!.ShowSystemUpdateMessage)
+                .Select(show => !show)
+                .BindToUi(this, view => view.UpdateButton.IsEnabled)
+                .DisposeWith(d);
+
+            this.WhenAnyValue(view => view.ViewModel!.ShowSystemUpdateMessage)
+                .BindToUi(this, view => view.UseSystemUpdater.IsVisible)
+                .DisposeWith(d);
```
erri120, line 37: These should use `OneWayBind` instead of `BindToUi`. That method also has an optional lambda argument to transform the inputs before they are passed to the component.

src/NexusMods.App.UI/Overlays/Updater/UpdaterView.axaml.cs
```diff
@@ -0,0 +1,54 @@
...
+                .BindToUi(this, view => view.ChangelogButton.Command)
+                .DisposeWith(d);
+
+            this.WhenAnyValue(view => view.ViewModel!.UpdateCommand)
+                .BindToUi(this, view => view.UpdateButton.Command)
+                .DisposeWith(d);
+
+            this.WhenAnyValue(view => view.ViewModel!.LaterCommand)
+                .BindToUi(this, view => view.LaterButton.Command)
+                .DisposeWith(d);
```
erri120, line 49: These should use `BindCommand` which automatically binds to the command of the given control:

```csharp
this.BindCommand(ViewModel, vm => vm.LaterCommand, view => view.LaterButton).DisposeWith(d).
```

src/NexusMods.Common/ApplicationConstants.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <summary>
+    /// The current version of the app.
+    /// </summary>
+    public static Version CurrentVersion
+    {
+        get
+        {
+
+            if (CompileConstants.InstallationMethod == InstallationMethod.Manually || CompileConstants.IsDebug)
+                return Version.Parse("0.0.0.1");
```
erri120, line 19: What's this magic number?

src/NexusMods.App.UI/Overlays/Updater/UpdaterDesignViewModel.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    public UpdaterDesignViewModel()
+    {
+        this.WhenActivated(d =>
+        {
+            this.WhenAnyValue(vm => vm.Method)
+                .Select(m => m is InstallationMethod.Archive or InstallationMethod.InnoSetup)
+                .Select(c => !c)
+                .BindToUi(this, view => view.ShowSystemUpdateMessage)
+                .DisposeWith(d);
```
erri120, line 45: `BindToUi` shouldn't be used in a VM. `Subscribe` or `Do` + `Subscribe` should be used instead:

```csharp
SubscribeWithErrorLogging(value => ShowSystemUpdateMessage = view)
```

src/NexusMods.App.UI/Overlays/Updater/UpdaterViewModel.cs
erri120, line 1: This file contains the VM, DTOs and functionality for getting the releases from GitHub. This should probably be split up.

erri120 (changes requested):

src/NexusMods.App.UI/Overlays/Updater/UpdaterViewModel.cs
halgari: I thought about that, but where would we put it? It's a single call and the only thing that uses those DTOs, and isn't even used in the tests. Thoughts?

erri120: They could probably go into the common library, which already contains the compile constants. The DTOs don't have to public and can be marked as internal.

src/NexusMods.App.UI/Overlays/Updater/UpdaterView.axaml.cs
halgari: Unfortunately this binds on whatever thread fires the event in the VM, so it'll throw errors if *anything* modifies the VM but isn't on the UI thread.

src/NexusMods.App.UI/Overlays/Updater/UpdaterView.axaml.cs
halgari: Unfortunately this binds on whatever thread fires the event in the VM, so it'll throw errors if *anything* modifies the VM but isn't on the UI thread.

src/NexusMods.Common/ApplicationConstants.cs
halgari: Low enough to trigger a update no matter what version we are on.

halgari:
Handling the feedback as part of #661

PR #656, opened 2023-09-20
Allow Games to define arbitrary GameFolderTypes (game path Ids)

### LINKED ISSUE #492: Use an extensible solution for IDs instead of Enums for GameFolderType
GameFolderType currently is an Enum defined in the main App, that identifies a small list of common game paths that the main application and other components might need to reference in a game installation agnostic way.

The issue is that some games might have particular paths they need to expose that aren't present in other games.
If new game needs to add one such identifiable path it would need to edit the shared Enum. 
The problem is that the enum would quickly get polluted with game specific path ids that might be interpreted differently for different games.

A different solution is needed, one that doesn't rely on a central shared list of IDs but instead allows each game to define its own list in addition to the common ones.

Many of these paths could likely be useful for other extension components (mod installers, diagnostics, etc), while the main application could realistically not need to be aware of them.

So it would be convenient for games to be able to define new PathIds without having to change the main app code, while still allowing other components to reference them. 
The developer of the extension or component will need to know the static ID of the path it needs and check for its existence in the currently managed game.
This admittedly isn't very elegant, but would allow bypassing visibility of the ID definition.

It can make it harder for developers to find the correct ID that is needed to be implemented for a game. 

For example all bethesda games have the `Data` folder, to add a new Bethesda game, the developer would need to look for what ID was used in the other games to define Data and make sure to reuse the same. Any extension developer wanting to reference Data would need to look up the exact ID to use. This is somewhat simplified in the case of human readable IDs like strings. while more error prone in the case of GUIDs for example.

## Proposal:
Use statically defined GUIDs wrapped in ValueObjects (for Nominal Typing) as GamePathTypes (or GamePathIds)

To define a new GUID statically on Rider, you can press Shift twice, type in GUID and there is an option to generate a GUID in various formats.

App defines a default list of common GamePathTypes (game, saves, confic, appdata)
Game then exposes a collection of GamePaths with either the App defined IDs or custom IDs.

Upside is fixed size of IDs and very low chance of collision and ability for any component to check for a path if they know the ID, even if the game type that defined it isn't visible from the extension or component.

Downside is non human readable nature of the ID. 
You can name the value object instance in the code that references a particular path ID instead.

Al12rs:
fixes #492 

TODO:
- [x] Change GameFolderType to ValueObject<string>
  - [x] Fix Serialization/deserialization
- [x] Add a LocationsManager
  - [x] Detect nested locations and allow easy query
  - [x] Properly initialize LocationsManager on setup
  - [x] Add XML doc comments
  - [x] Add Tests
- [x] Update AGame.GetLocations() to return dictionary to ensure unique GameFolderTypes
- ~[ ] How to handle nested GamePaths~ moved to #658 
- [x] Fix tests (They work locally on both Linux and Windows)
- [x] Remove IGameLocator parameter from AGame.GetLocations()
- [x] Handle Location Ids that point to the same path ? Added to #658 as well.
- [x] Rename GameFolderType to either LocationId 
- [x] Investigate why CI is failing (CI was choking on the implicit conversion of paramerters to LocationId in the [Theory] tests)


src/NexusMods.DataModel/Games/AGame.cs
```diff
@@ -100,8 +101,7 @@ private List<GameInstallation> GetInstallations()
     /// <param name="locator">The locator used to find this game installation.</param>
     /// <param name="installation">An installation of the game found by the <paramref name="locator"/>.</param>
     /// <returns></returns>
-    protected abstract IEnumerable<KeyValuePair<GameFolderType, AbsolutePath>> GetLocations(
-        IFileSystem fileSystem,
+    protected abstract IReadOnlyDictionary<GameFolderType, AbsolutePath> GetLocations(IFileSystem fileSystem,
         IGameLocator locator,
```
Sewer56, line 105: The `locator` field is currently unused. Do we have a known use case where we need to search for Game B from Game A?

Al12rs: Yeah, not sure on that one, might be a leftover from previous versions of the code, but I can't be sure there aren't some valid use cases

Al12rs:
Removed draft to let tests run

src/NexusMods.DataModel/Games/AGame.cs
Sewer56: I think that's probably an accidental leftover, since we already have a `GameLocatorResult` available as a parameter.

@erri120 seems to have written that code; can you confirm if it's intended?

src/Games/NexusMods.Games.DarkestDungeon/DarkestDungeon.cs
```diff
@@ -45,22 +46,22 @@ public override GamePath GetPrimaryFile(GameStore store)
...
 
-        yield return new KeyValuePair<GameFolderType, AbsolutePath>(GameFolderType.Preferences, globalSettingsFile);
+        var result =  new Dictionary<GameFolderType, AbsolutePath>()
+        {
+            { GameFolderType.Game, installation.Path },
+            { GameFolderType.Preferences, globalSettingsFile }
+        };
+
+        if (installation.Metadata is SteamLocatorResultMetadata { CloudSavesDirectory: not null } steamLocatorResultMetadata)
+            result[GameFolderType.Saves] = steamLocatorResultMetadata.CloudSavesDirectory.Value;
```
Sewer56, line 64: There's a `// TODO: Steam Cloud Saves` comment below inside the branch, but this code is right above. Do you know if there a specific known reason why this specific statement is left outside of the branch? 

CC. @erri120 (<= wrote the original code)

I imagine `SteamLocatorResultMetadata` can only be returned when the store is Steam.

src/NexusMods.DataModel/Games/GameLocationDescriptor.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// If true, no other game location contains this game location.
+    /// </summary>
+    public bool IsTopLevel { get; private set; }
+
+    private GameFolderType? _topLevelParent;
+
+    /// <summary>
+    /// The top level location that contains this location, if there is any.
+    /// </summary>
+    public GameFolderType? TopLevelParent => _topLevelParent;
```
Sewer56, line 45: Fun fact: You can actually do `{ get; internal set; }` here, save some lines.
Feel free to change it if you feel like, will be a bit cleaner that way.

src/Games/NexusMods.Games.DarkestDungeon/DarkestDungeon.cs
Al12rs: I changed it to something that makes more sense to me

src/NexusMods.DataModel/Games/GameLocationsRegister.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// Computes the nested and top level parents relationships.
+    /// </remarks>
+    /// <param name="newLocations">A Dictionary of <see cref="GameFolderType"/>, <see cref="AbsolutePath"/> to register</param>
+    public GameLocationsRegister(IReadOnlyDictionary<GameFolderType, AbsolutePath> newLocations)
+    {
+        foreach (var (newId, newPath) in newLocations)
+        {
+            var newLocation = new GameLocationDescriptor(newId, newPath);
+
+            var isDuplicate = !_locations.TryAdd(newId, newLocation);
```
Sewer56, line 35: Inline this method call into the assert, because chances are this will still execute in Release mode; the dead code eliminator isn't super sophisticated and might not consider this pure enough to remove.

src/NexusMods.DataModel/Games/AGame.cs
erri120: The `IGameLocator` interface is pretty old, you can probably remove it if it's not used.

src/Games/NexusMods.Games.DarkestDungeon/DarkestDungeon.cs
Sewer56: Sure that works.

src/NexusMods.DataModel/Games/GameLocationsRegister.cs
Al12rs: I always want to add, I just check whether it failed due to duplicated ID in debug mode. Unless I'm getting your comment wrong.

src/NexusMods.DataModel/Games/GameLocationsRegister.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            Debug.Assert(isDuplicate == false,
+                $"Duplicate location found for {newId} at {_locations[newId].ResolvedPath}: {newPath}");
+
+            foreach (var existingId in _locations.Keys)
+            {
+                if (existingId == newId) continue;
+
+                if (this[existingId].GetFullPathLength() < newPath.GetFullPathLength())
```
Sewer56, line 44: This code appears to check if the existing path is a parent to the new path; however this is not clear at first.
Can you leave a comment in the code here? (for future maintainers)

Al12rs: Yeah, good idea.

src/NexusMods.DataModel/Games/GameLocationDescriptor.cs
```diff
@@ -0,0 +1,70 @@
...
+
+    /// <summary>
+    /// If true, no other game location contains this game location.
+    /// </summary>
+    public bool IsTopLevel { get; private set; }
+
+    /// <summary>
+    /// The top level location that contains this location, if there is any.
+    /// </summary>
+    public GameFolderType? TopLevelParent { get; private set; }
```
Sewer56, line 43: This one can probably also be `internal set;`

src/NexusMods.DataModel/Games/GameLocationsRegister.cs
Sewer56: Ah, okay. In that case that's no problem. I assumed it was just for debug since the variable went unused in Release; all good.

Al12rs:
Tests are failing strangely in the Advanced Installer project, but I can't reproduce locally. Any clues what is happening there?

Sewer56 (approved):
After latest commit (f445583d5c6251e9567d0c2fe1bfe3d7a03e9985) , seems good to merge

Sewer56:
> Tests are failing strangely in the Advanced Installer project, but I can't reproduce locally. Any clues what is happening there?

No idea, but no issues on Linux locally either. 

The code in this PR shouldn't affect current Advanced Installer code on Main branch (only Suggestions/Common Folders, which is a future PR). Advanced Installer code doesn't have anything special; no shared, state, concurrency, etc. so if it failed, that would have been weird.

Sewer56:
I can't repro it locally either. Not in release or debug mode.

We've been updating a lot of dependencies in `main` lately. Maybe something we use broke specifically in CI under some weird circumstance.

PR #659, opened 2023-09-21
Move back to nx archivemanager

halgari:
Currenty this still fails due to a ZStd decompression error in the SDV tests. The unit tests for the archive manager work fine. But something with how SDV is doing things is causing this to fail.


halgari:
@Sewer56 mind taking a look at `NxArchiveManager.ChunkedArchiveStream.*`? I'm sure I'm missing something. If you run the SDV tests you'll see "no installer found for X". But if you then run with debugging enabled you'll see a ZStd error being thrown by the NX `Compression.Decompress` method. I'm sure I'm doing something wrong here, but I can't figure out what it is.

Sewer56:
Sure, I got some time to spare. 

The advanced installer suggestion stuff I'm finishing up is dependent on AL's current work (their work changes API I have to interface with); so it'd be more efficient for me to fully wait until that's finalised anyway, as I've been working off of an incomplete PR.

Al12rs:
Is this ready for review?

Sewer56 (changes requested):
Address the changes above then it's good for merging.
I'll detail the performance gimmick I mentioned earlier in a new issue. We can adjust the text of that issue as needed.

src/NexusMods.DataModel/NxArchiveManager.cs
```diff
@@ -185,30 +184,206 @@ public async Task ExtractFiles(IEnumerable<(Hash Src, AbsolutePath Dest)> files,
...
+            for (var chunkIndex = 0; chunkIndex < chunkCount; chunkIndex++)
+            {
+                var blockIndex = _entry.FirstBlockIndex + chunkIndex;
+                var block = new ExtractableBlock
+                {
+                    BlockIndex = blockIndex,
+                    DecompressSize = _entry.DecompressedBlockOffset + (int)Math.Min(remainingDecompSize, (ulong)chunkSize)
+                };
+
+                _blocks.Add(block);
```
Sewer56, line 367: This can be preallocated as array rather than using list; since the number of blocks (chunkCount) is known ahead of time.

Issue #308, opened 2023-05-08
As a user, I want to run a modding tool


erri120:
### Requirements

**Must haves:**

- Users should be able to run modding tools from the mod manager, without having to manually apply the loadout and run the tool inside the game folder.
- Ideally, these tools should be detected automatically during mod installation by keeping a list of known tool executable file names, or hashes. Users should also be able to manually add/remove/change this "tool list".
- The UI should display all currently installed tools and allow the user to easily launch and modify them.

**Other:**

The output of these modding tools should be captured. How this is done is up for debate. _Most_ tools that produce files have an output directory option (eg: [xEdit](https://www.nexusmods.com/skyrimspecialedition/mods/164) has the `-o` command line option, [FNIS](https://www.nexusmods.com/skyrimspecialedition/mods/3038) has `RedirectFiles`) but others update files in-place or add new files to the game directory. Without MO2 integration, [BethINI](https://www.nexusmods.com/skyrimspecialedition/mods/4875) will just update the original INI files and editors like the Creation Kit will run inside the game directory and output to the game directory.

Another thing to consider is that mods and the _game_ itself also create and update files. If the user has a [crash logger](https://www.nexusmods.com/skyrimspecialedition/mods/59596) installed and the game crashes, it will create a crash log inside the game directory.

For Stardew Valley, every SMAPI mod is able to [read](https://github.com/Pathoschild/SMAPI/blob/8d600e226960a81636137d9bf286c69ab39066ed/src/SMAPI/IModHelper.cs#LL67C23-L67C23) from a config, if this JSON config file doesn't exist, it will be [created](https://github.com/Pathoschild/SMAPI/blob/8d600e226960a81636137d9bf286c69ab39066ed/src/SMAPI/Framework/ModHelpers/ModHelper.cs#L149-L155).

**Related:**

- #152

Aragas:
Mods are also used to distribute tools. In case of Bannerlord, we have BLSE or custom launchers.
With BLSE, we could hardcode the tool in the game's project, but we'll need a way to determine whether to show a tool to the user or not, based on OS or other conditions.

This could be expanded to automatic tool discovery, but it will definitely require some kind of a metadata file distributed with the mod, much like FOMOD?

halgari:
Waiting on #312

erri120:
Related to #463.

MattSturgeon:
[...]

PR #681, opened 2023-09-27
Run games through Steam

### LINKED ISSUE #152: Run Windows builds of games from Steam through Steam
If the current game is not native for Linux and was installed via Steam, we have to run the game through the Steam created Proton instance.

erri120:
Resolves #152.

### DIFF
--- a/src/NexusMods.DataModel/Games/RunGameTool.cs
+++ b/src/NexusMods.DataModel/Games/RunGameTool.cs
@@ -115,6 +115,7 @@ public class RunGameTool<T> : IRunGameTool
 
     private async Task RunThroughSteam(uint appId, CancellationToken cancellationToken)
     {
+        var existingReaperProcesses = Process.GetProcessesByName("reaper");
         await _osInterop.OpenUrl(new Uri($"steam://rungameid//{appId.ToString(CultureInfo.InvariantCulture)}"), cancellationToken);
 
         if (OSInformation.Shared.IsWindows)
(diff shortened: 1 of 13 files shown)


src/NexusMods.DataModel/Games/RunGameTool.cs
```diff
@@ -107,6 +113,61 @@ public async Task Execute(Loadout loadout, ApplyPlan applyPlan, CancellationToke
...
+        CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            var start = DateTime.UtcNow;
+            while (!cancellationToken.IsCancellationRequested && start + timeout < DateTime.UtcNow)
+            {
+                var processes = Process.GetProcessesByName(processName);
+                var target = processes.FirstOrDefault();
+                if (target is not null) return target;
```
Sewer56, line 153: Not sure why this requires LINQ when the returned value is an array, but it's not a big deal.

Sewer56 (approved):
Good for merging, no complaints.
Note that this only covers basic case of running main game binary. 

That said, I doublechecked, and it seems the more advanced required cases are already present as checkboxes in:
- https://github.com/Nexus-Mods/NexusMods.App/issues/308

So that's a 👍

src/NexusMods.DataModel/Games/RunGameTool.cs
```diff
@@ -107,6 +113,61 @@ public async Task Execute(Loadout loadout, ApplyPlan applyPlan, CancellationToke
         _logger.LogInformation("Finished running {Program}", program);
     }
 
+    private async Task RunThroughSteam(uint appId, CancellationToken cancellationToken)
+    {
+        await _osInterop.OpenUrl(new Uri($"steam://rungameid//{appId.ToString(CultureInfo.InvariantCulture)}"), cancellationToken);
```
Sewer56, line 118: ☝️ Likely Bug Here

If a user has already been running an application via Steam, a `reaper` process might already exist; thus there is risk that `WaitForProcessToStart` returns the wrong `reaper` process (one belonging to an existing application).

While it may be uncommon for a user to launch 2 games at the same time; Steam also ships regular software these days. Running Steam shipped software in background may be much more common.

Some examples that come to mind:  
- Blender (Exporting 3D Models for Testing in Custom Mods)
- VTubing Software (e.g. VTube Studio, VSeeFace, FaceRig)
- VR Software (SteamVR, etc.)

Sewer56 (changes requested):

Sewer56 (approved):
Looks good to me now.

Not sure how the OAuth stuff got here, maybe you committed to wrong branch. If it's a mistake, just fix it up before merging :p

Issue #674, opened 2023-09-27
Create panel container


Greg-Nexus:
# Chore

## Implementation TODOs

- [ ] Workspace
  - [x] has panels (#688)
  - [x] panels can be added (#688, #704)
    - [ ] offer a "default" option
      - [ ] prefer columns to rows when horizontal
      - [ ] prefer rows to columns when vertical
    - [x] images are auto-generated (#704)
    - [x] support for 2x2 (#704)
      - [x] 1 panel -> add column or row
      - [x] 2 columns/rows -> add new panel in one of the corners
      - [x] 2 columns -> add new panel as a top/bottom row (#814)
      - [x] 2 rows -> add new panel as a left/right column (#814)
  - [x] can be serialized and deserialized (#731)
- [ ] Panels
  - [x] can be closed (#704)
    - [x] prefer columns to rows when horizontal (#708)
    - [x] prefer rows to columns when vertical (#708)
  - [x] can be swapped (#704)
  - [x] ~~can be resized~~ (handled in #705)
  - [ ] can be popped out into its own window
  - [x] contains tabs (#708)
- [x] Tabs
  - [x] has an icon (#708)
  - [x] has a title (#708)
  - [x] has contents (#708)
  - [x] can be selected (#708)
  - [x] can be closed (#708)
  - [x] empty tabs contain a selection screen (#768)
- [ ] Styling
  - [ ] Workspace
    - [ ] Add panel icons
    - [ ] Add panel icon drop-down menu
  - [x] Panels
    - [x] tab header strip
      - [x] tabs (#708)
      - [x] plus button (#708)
      - [x] close button (#708)
      - [x] popout button (#708)
      - [x] close and popout button only visible when you have multiple panels (#768)
  - [x] Tabs
    - [x] tab header
      - [x] icon (#708)
      - [x] title (#708)
        - [x] long titles should be truncated with `...` (#708)
      - [x] close button (#708) 

## Requirements
This issue is to create a new panel container UI pattern, which can be extended across existing views within the app in follow-up issues.

Containers will need to contain UI grids, with up to four panels. Opening a new panel should be possible via the newly updated top navigation bar.

New panels should split the largest existing container in half, either vertically or horizontally (as chosen the by user).

Panels will need to contain tabs, opened or closed from within the panel. Eventually new tabs will display list menus of views that could be opened, but for this issue we're only concerned with opening and closing tabs.

[...]

PR #688, opened 2023-10-02
Initial support for Workspaces and Panels (split and join)

erri120:
Part of #674.

- Adds a design-only playground for testing. This is currently the only place where the system is used, so the feature is completely isolated. The code for the playground is in constant change, so don't read into it too much.
- Adds a `Workspace` and `Panel` component:
  - Panels can be added (split) and removed (join).
  - Math is outside the VM and fully tested.

### DIFF
--- a/src/NexusMods.App.UI/WorkspaceSystem/Panel/IPanelViewModel.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/Panel/IPanelViewModel.cs
@@ -8,9 +8,28 @@ public interface IPanelViewModel : IViewModelInterface
 
     public IViewModel? Content { get; set; }
 
+    /// <summary>
+    /// Gets or sets the logical bounds the panel.
+    /// </summary>
+    /// <remarks>
+    /// The logical bounds describe the ratio of space the panel takes up inside the workspace.
+    /// The values range from 0.0 up to and including 1.0, where 0.0 means 0% of the space and 1.0
+    /// means 100% of the space.
+    /// </remarks>
+    /// <seealso cref="ActualBounds"/>
     public Rect LogicalBounds { get; set; }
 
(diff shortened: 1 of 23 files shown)


src/NexusMods.App.UI/WorkspaceSystem/Panel/IPanelViewModel.cs
```diff
@@ -0,0 +1,35 @@
+using Avalonia;
+
+namespace NexusMods.App.UI.WorkspaceSystem;
+
+public interface IPanelViewModel : IViewModelInterface
```
Al12rs, line 5: I would add a comment describing what a panel is in this context and its purpose.

erri120: I'll add documentation once I add tabs to the system.

src/NexusMods.App.UI/WorkspaceSystem/Workspace/IWorkspaceViewModel.cs
```diff
@@ -0,0 +1,26 @@
+using System.Collections.ObjectModel;
+using System.Reactive;
+using Avalonia;
+using ReactiveUI;
+
+namespace NexusMods.App.UI.WorkspaceSystem;
+
+public interface IWorkspaceViewModel : IViewModelInterface
```
Al12rs, line 8: Here as well, a comment describing what a Workspace is and its purpose. We had some naming issues on these so having some clarification in the code is important I think.

erri120: I'll add documentation once I add tabs to the system.

Al12rs (approved):

Issue #626, opened 2023-09-12
(WIP) Rework the loadout syncronizer 


halgari:
# Improvement

## Currently 
Our loadout syncronizer has several shortcomings and issues. Namely it can lose data when ingesting changes from the data folder. The structure is overly opaque due to being rather abstract. Since we use traits it's never clear what combination of interfaces may show up in what locations. And it's often unclear what will happen when two files collide

## We want
A clean extensible design for managing ingestion and application of files. Something that's easy to extend with various games. 

## Design
In keeping with the more recent designs of the application we'll fully delegate ingestion and application to the IGame class, then extend AGame to provide common reusable logic. 

The quick TL;DR view of the new design is rather simple: we will rework the syncronizer to deal primarily with merging two modlists. Application of files and ingestion of files are quite similar: they merge two loadouts. The only thing that changes is how the loadouts are created. During ingestion one loadout is created from the disk state, in application the loadout is created by loadout manipulation routines. 

#### Disk State
Given a disk state we can generate a new tree by hashing (or using cached hashes) the files. But this does not tell us what sort of files these are, or how they are generated. By default the Disk State tree generator will use `FromArchive` for each file and backup each file. However each game can override this behavior on a file by file basis. For example skyrim may override `plugin.txt` to be a EnabledPlugins file. This override is important as it insures that the tree comparing functionality can treat both sides of the comparison as the same type. The EnabpledPluginsFile can compare lines from plugins.txt instead of one side being FromArchive. Since the game itself is responsible for handling this state creation, we can throw a "unreachable error" if one size of the comparison is the wrong type.

Al12rs:
Some older points on this topic:
https://github.com/Nexus-Mods/NexusMods.App/issues/422

Al12rs:
Couple of notes on the above design:
- During Ingest, against which Loadout revision is the Fork created? Does the design consider cases of users switching loadouts?
- For `ALoadoutSynchronizer`, since these functions can end up having to be reimplemented by Game developers, it will require attention to make the part visible to them easy to understand even to people not aware of the intricacies of how the Apply logic works. Potentially including more specialized interfaces in addition to the top level `ALoadoutSynchronizer`, for specific functionality. 
- This design is focused on ingesting and merging outside changes, so this might be out of scope, but in the future it would be desirable for games to be able to define alternative methods of deployment and detection of outside changes (Links, VFS, compacting into game archives), so for now I would just keep an eye out that the design doesn't impede this.

[...]

PR #697, opened 2023-10-05
Rework the Loadout Synchronizer

halgari:
Implementation of the structure and tests of #626. This does not convert exsting games over to the new syncronizer, that will be done in a future pull request. All of the code is in the ALoadoutSyncronizer, start there or in ALoadoutSyncronizerTests to see usage.


src/NexusMods.DataModel/Abstractions/EntityDictionary.cs
erri120, line 1: `EntityDictionary<TK, TV>` can be made `readonly`.

src/NexusMods.DataModel/Abstractions/EntityDictionary.cs
```diff
@@ -238,6 +238,35 @@ public TState Walk<TState>(Func<TState, Entity, TState> visitor, TState initial)
...
+        {
+            // We've already processed colliding keys, so skip them.
+            if (_coll.ContainsKey(key)) continue;
+            var newValue = mergeFn(null, _store.Get<TV>(value));
+            if (newValue is not null)
+                result = result.With(key, newValue);
+        }
+
+        return result;
+    }
```
erri120, line 268: `result = result.With(key, newValue)` uses `ImmutableDictionary.SetItem`. Doing this inside two for-each loops and creating new immutable dictionaries on possible every iteration is not very efficient. `ImmutableDictionary.SetItems` can be used instead.

src/NexusMods.DataModel/Loadouts/DiskStateRegistry.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <returns></returns>
+    public DiskState? GetState(LoadoutId loadoutId)
+    {
+        var iid = loadoutId.ToEntityId(EntityCategory.DiskState);
+        var data = _dataStore.GetRaw(iid);
+        if (data == null) return null;
+        using var ms = new MemoryStream(data);
+        using var compressed = new GZipStream(ms, CompressionMode.Decompress);
+        return JsonSerializer.Deserialize<DiskState>(compressed);
+    }
```
erri120, line 59: Might be worth creating a helper method for doing JSON+GZip if we're going to be using it more.

src/NexusMods.DataModel/LoadoutSynchronizer/DiskModificationPlan.cs
erri120, line 1: This entire class is unused, is this a left-over?

src/NexusMods.DataModel/LoadoutSynchronizer/ALoadoutSynchronizer.cs
```diff
@@ -0,0 +1,561 @@
...
+            await foreach (var entry in _hashCache.IndexFolderAsync(location))
+            {
+                var gamePath = installation.LocationsRegister.ToGamePath(entry.Path);
+                if (prevState.TryGetValue(gamePath, out var prevEntry))
+                {
+                    // If the file has been modified outside of the app since the last apply, we need to ingest it.
+                    if (prevEntry.Value!.Hash != entry.Hash)
+                    {
+                        HandleNeedIngest(entry);
+                        throw new UnreachableException("HandleNeedIngest should have thrown");
```
erri120, line 132: `HandleNeedIngest` is `virtual`, so implementations can not throw. Is that method supposed to only throw or do something else?

src/NexusMods.DataModel/LoadoutSynchronizer/ALoadoutSynchronizer.cs
```diff
@@ -0,0 +1,561 @@
...
+
+        // Write the generated files (could be done in parallel)
+        foreach (var entry in toWrite)
+        {
+            await using var outputStream = entry.Key.Create();
+            var hash = await entry.Value.Write(outputStream);
+            if (hash == null)
+            {
+                outputStream.Position = 0;
+                hash = await outputStream.HashingCopyAsync(Stream.Null, CancellationToken.None, async x => {});
```
erri120, line 208: `reportFn` can be `null`.

erri120 (changes requested):

src/NexusMods.DataModel/LoadoutSynchronizer/ALoadoutSynchronizer.cs
halgari: Anything that overrides that method is supposed to throw, as at this point the state of the apply process is not recoverable after a ingest. I'll leave it for now, but we can easily rework that code in the future if it's a problem.

src/NexusMods.DataModel/Loadouts/DiskStateRegistry.cs
halgari: I hope to remove that code sometime in the future when I do work on the datamodel, and at that point we'll probably look into some sort of better compression method as well.

erri120 (approved):

Al12rs:
Can I have a look before merging?

src/NexusMods.DataModel/Loadouts/Mods/Mod.cs
```diff
@@ -24,6 +24,21 @@ public record Mod : Entity, IHasEntityId<ModId>
...
+
+    /// <summary>
+    /// Category used for 'Save Games'.
+    /// </summary>
+    public const string SavesCategory = "Saved Games";
+
+    /// <summary>
+    /// Category used for 'Save Games'.
+    /// </summary>
+    public const string OverridesCategory = "Overrides";
```
Al12rs, line 40: Comment needs fixing

src/NexusMods.DataModel/LoadoutSynchronizer/FlattenedToLoadoutTransformer.cs
```diff
@@ -0,0 +1,168 @@
...
+                        continue;
+                    }
+                    else
+                    {
+                        AddToValues(_moveFrom, prevPair.Value.Mod.Id, prevPair.Value.File);
+                        AddToValues(_moveTo, newPair.Mod.Id, newPair.File);
+                        continue;
+                    }
+
+                    continue;
```
Al12rs, line 81: This is Unreachable, can be removed

Al12rs (approved):

PR #700, opened 2023-10-08
Game: Blade & Sorcery support

nepellion:
Implemented support for PCVR version of Blade & Sorcery.
As far as I can see ingesting existing mods from game files isn't currently possible, but it is one of the features important for this game as mentioned in the docs file.


halgari:
@nepellion 
Now that the code for game support is a bit more stable, we'd like to look at getting this merged. What is left to be done, and is there something I can do to help?

nepellion:
> @nepellion Now that the code for game support is a bit more stable, we'd like to look at getting this merged. What is left to be done, and is there something I can do to help?

There is the issue mentioned in #130 about ingesting mods from the in-game mod manager, but as far as I understand there is no support for that for now, right?

The icons/images I used are very placeholder-y, not sure if that's ok or needs to be changed, and the doc file doesn't have a number because I wasn't sure where it would end up.

Other than that it should be ready.

PR #708, opened 2023-10-11
Workspace Tabs

### LINKED ISSUE #676: Create new icons for panels
# Chore

## TODOs

- [x] auto-generate layout images (#704)
- [x] `icon/navigation/arrow_drop_down_24px` for the panel multi-select drop-down (#708)
- [x] `icon/content/add_24px` for adding new tabs and panels (#708)
- [x] `icon/navigation/close_24px` for closing tabs and panels (#708)
- [x] `icon/action/open_in_new_24px` for popping out panels into a separate window (#708)
- [x] `tab_24px` new tab icon (#708)

## Requirements

We need to review the new design (see below) and ensure we've created icons for anything missing, this will need to include any in the top, left and spine navigation, as well as any used by the new panels.

erri120:
Part of #674 (after #688, and #704). Closes #676.

TODO:
- [x] virtualize/cache views
- [x] Adds tabs to panels.
- [x] Adds missing icons.
- [x] Some styling.

https://github.com/Nexus-Mods/NexusMods.App/assets/16577545/d8d40cda-7f79-48f8-86fe-7711563312c9


docs/decisions/frontend/0002-Workspaces.md
```diff
@@ -0,0 +1,62 @@
...
+    - Example 1: `X: 0, Y: 0, Width: 1, Height: 1` is a panel that is positioned in the top left corner and takes up 100% of the width and height of the workspace.
+    - Example 2: `X: 0.5, Y: 0, Width: 0.5, Height: 1` describe a column on the right side, it takes up 100% of the height but only 50% of the width of the workspace and is positioned on the right.
+  - Actual bounds are the exact pixel sizes of the panel on the monitor. They are derived by multiplying the logical bounds with the pixel size of the workspace.
+
+### Adding a Panel
+
+- Additional panels can only be added to the workspace if the maximum number of columns and rows aren't reached yet.
+- An existing panel must be split in half to create space for a new panel.
+- An existing panel can only be split vertically if the number of columns in the current row isn't greater than the maximum amount of allowed columns.
+- An existing panel can only be split horizontally if the number of rows in the current column isn't greater than the maximum amount of allowed rows.
```
Al12rs, line 53: I'm confused, when splitting a panel vertically I understand creating two panels that are stacked vertically, so this sounds like it should the the other way around.

docs/decisions/frontend/0002-Workspaces.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Closing a Panel
+
+- Closing a panel is done by joining the current panel with **one or more** adjacent panel.
+- Columns are preferred to rows when the workspace is horizontal (width > height).
+- Rows are preferred to columns when the workspace is vertical (height > width).
+- An adjacent panel is a panel that is in the same row or column as the current panel.
```
Al12rs, line 60: Adjacent means that they share a border, while this definition seems to consider all panes on the same row or column, even if not directly connected.

Currently this can't happen since we are limited to 4 panels, but since the system should be generic, it's a good idea to generalize it. 

I can suggest "aligned" perhaps as an alternative to adjoining, to indicate all elements sharing the same row or column. I would leave adjacent to panels that actually share a border.

PR #720, opened 2023-10-19
Update Language.pl.resx

Patriot99:
[#708](https://github.com/Nexus-Mods/NexusMods.App/pull/708)

`PanelTabHeaderViewModel_Title_New_Tab`


Sewer56 (changes requested):
(see above comment, it was probably made in error)

Sewer56:
As a fun note.

I always found it funny how 'new tab' in Polish is 'nowa karta'; because 'nowa karta' means 'new card'. (Note that 'card' can also mean 'greeting card', 'card with information' etc.)
In English it's derived from 'folder tabs', but in Polish the commonly used word for that (IIRC) is 'zakładka' which is bookmark.

PR #722, opened 2023-10-19
Use .NET 8.0

halgari:
Mostly here to see what happens when we want to migrate, do the tests pass? Does the world fall apart?


halgari:
Redoing this PR

PR #726, opened 2023-10-20
Polish for FOMOD UI

Patriot99:
> #600
> 
> `Language.pl.resx.`
> 
> Polish UI for FOMOD.

### DIFF
--- a/src/Games/NexusMods.Games.FOMOD.UI/Resources/Language.pl.resx
+++ b/src/Games/NexusMods.Games.FOMOD.UI/Resources/Language.pl.resx
@@ -1,0 +1,55 @@
+
+<root>
+
+    <resheader name="resmimetype">
+        <value>text/microsoft-resx</value>
+    </resheader>
+    <resheader name="version">
+        <value>1.3</value>
+    </resheader>
+    <resheader name="reader">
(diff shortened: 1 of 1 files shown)


Sewer56 (approved):
Looks good to me.

----------

Note: 
 
The original English localization uses 'Next' [Następny] and 'Previous' [Poprzedni]; which is more formal/robotic.

The PR uses the words `Dalej` [onward/proceed] and `Wstecz` [back(wards)], which are more informal and commonly found on install wizards (thus within user expectation). Since the FOMOD installer is effectively an install wizard, I think it's more appropriate to use what is in this PR.

In that vein, I'm wondering if we should change the English locale to say 'Back' instead of 'Previous'. It 'feels' more user friendly, so I figured I'd ask @captainsandypants

Patriot99:
I will add from myself:
For me, `Back (Wstecz)` is more suitable than `Previous (Poprzedni)`. I also took into account that this is an installation wizard where you usually see `Back (Wstecz)` and `Next (Dalej)` rather than `Previous (Poprzedni)`. When translating, I change the words that fit in Polish, rather than literally translating a word from English to Polish. A normal user will not check the source code, but only what he sees at the moment. It's all about good word choice and making sure everything is understandable. I know from experience that people will complain if the word does not fit in a particular place.

captainsandypants:
This all makes sense, happy to change.

PR #734, opened 2023-10-25
Implement "Standard Button" using Control Themes

erri120:
This was mostly an experiment to see how well the new [Control Themes](https://docs.avaloniaui.net/docs/next/basics/user-interface/styling/control-themes) work in Avalonia.

Part of #735

### DIFF
--- a/src/NexusMods.App.UI/Controls/Buttons/StandardButtonTheme.axaml
+++ b/src/NexusMods.App.UI/Controls/Buttons/StandardButtonTheme.axaml
@@ -1,49 +1,103 @@
 <ResourceDictionary
     xmlns="https://github.com/avaloniaui"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:buttons="clr-namespace:NexusMods.App.UI.Controls.Buttons">
+    xmlns:icons="clr-namespace:Projektanker.Icons.Avalonia;assembly=Projektanker.Icons.Avalonia"
+    xmlns:buttons="clr-namespace:NexusMods.App.UI.Controls.Buttons"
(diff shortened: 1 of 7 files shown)


src/NexusMods.App.UI/Controls/Buttons/StandardButtonTheme.axaml
```diff
@@ -0,0 +1,370 @@
...
+                        CornerRadius="{TemplateBinding CornerRadius}">
+                    <StackPanel x:Name="PART_Panel">
+                        <icons:Icon x:Name="PART_Icon"
+                                    Foreground="{TemplateBinding Foreground}"
+                                    Value="{TemplateBinding IconType, Converter={StaticResource IconTypeConverter}}" />
+                        <Image x:Name="PART_Image"
+                               Source="{TemplateBinding Property=Image}" />
+                        <TextBlock x:Name="PART_TextBlock"
+                                   Foreground="{TemplateBinding Foreground}"
+                                   Text="{TemplateBinding Text, Converter={StaticResource TextCaseConverter}, ConverterParameter=upper}" />
```
Sewer56, line 192: In the interest of efficiency, it might be a good idea to reference the converters via `x:Static` instances, given that we might reuse them elsewhere.

It should also speed up instantiation of the control, given we avoid a dictionary lookup (per control), and reduce memory use.

-----

Might also be handy to make a converter for each case (makes usage easier from xaml as use doesn't need to know the 'magic' parameters). Should be doable fairly easy with abstract class; and abstract method (which parent implements) to do the conversion. Then make that sealed and you got both convenience and efficiency at around same LoC.

Sewer56 (approved):

Issue #735, opened 2023-10-25
UI style consolidation


erri120:
# Epic

## Issues

- [x] create a separate UI project for only styles, resources and components
- [x] Convert `Nexus.xaml` (Style) to resources dictionaries (#736)
  - [x] Remove all uses of `Nexus.xaml` and use the new resources
- [ ] ~~Use Control Themes~~ (We ended up using Style Classes)
  - [x] #734
- [x] Consolidate icons
- [x] Remove global styles
- [x] Look into Text centering issues (line height related problems)
- [x] Consolidate Text styles
- [x] Cleanup

PR #736, opened 2023-10-25
Add Colors and Fonts resources

erri120:
Part of #735.

### DIFF
--- a/src/NexusMods.App.UI/App.axaml
+++ b/src/NexusMods.App.UI/App.axaml
@@ -2,8 +2,7 @@
     x:Class="NexusMods.App.UI.App"
     xmlns="https://github.com/avaloniaui"
     xmlns:local="using:NexusMods.App.UI"
-    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:system="clr-namespace:System;assembly=System.Runtime">
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
     <Application.DataTemplates>
         <local:ViewLocator />
     </Application.DataTemplates>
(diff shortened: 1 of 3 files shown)


src/NexusMods.App.UI/Palatte/Fonts.axaml
```diff
@@ -0,0 +1,22 @@
...
+        put them in the correct folder, and link them in as shown below.
+    -->
+
+    <FontFamily x:Key="MontserratRegular">/Fonts/Regular#Montserrat</FontFamily>
+    <FontFamily x:Key="MontserratSemiBold">/Fonts/Bold#Montserrat</FontFamily>
+    <FontFamily x:Key="MontserratBold">/Fonts/Bold#Montserrat</FontFamily>
+
+    <FontFamily x:Key="RobotoRegular">/Fonts/Regular#Montserrat</FontFamily>
+    <FontFamily x:Key="RobotoSemiBold">/Fonts/Bold#Montserrat</FontFamily>
+    <FontFamily x:Key="RobotoBold">/Fonts/Bold#Montserrat</FontFamily>
```
Al12rs, line 20: These should be Roboto

src/NexusMods.App.UI/App.axaml
```diff
@@ -13,6 +12,9 @@
     </Application.Styles>
 
     <Application.Resources>
+        <ResourceInclude x:Key="Colors" Source="avares://NexusMods.App.UI/Palatte/Colors.axaml"/>
+        <ResourceInclude x:Key="Fonts" Source="avares://NexusMods.App.UI/Palatte/Fonts.axaml"/>
+
```
Al12rs, line 17: The idea is that we have everything in a single file that then references the rest. 
We want to be able to swap out one file and get a completely different styling for the app.

Is that possible?

erri120: You can just load a different resource file later and it will overwrite the previous values.

halgari (approved):

PR #753, opened 2023-11-06
More UI coding conventions

erri120:
- Adds more context to OAPH
- Adds a section about design view models

### DIFF
--- a/docs/UICodingConventions.md
+++ b/docs/UICodingConventions.md
@@ -1335,6 +1335,55 @@ public class GoodExampleViewModel
 }
 ```
 
+### Creating and using Design View Models
+
+**Never** create empty design view models that aren't different from the normal view model.
+
+The design view model should **only** contain design data. If necessary, it can inherit from the normal view model and set properties in the constructor:
+
+```csharp
+// if the view model has a construtor that accepts data:
+public class MyViewModel
+{
+    public MyViewModel(SomeData data)
+    {
+        // ...
+    }
+}
+
+public class MyDesignViewModel : MyViewModel
+{
+    public MyDesignViewModel() : base(SetupData()) { }
+
+    private static SomeData SetupData()
+    {
+        // ...
+    }
+}
+
+// if the view model has a default constructor but properties with data:
+public class MyViewModel
+{
+    [Reactive]
+    public SomeData Data { get; set; }
+}
+
+public class MyDesignViewModel : MyViewModel
+{
+    public MyDesignViewModel()
+    {
+        base.Data = SetupData();
+    }
+
+    private static SomeData SetupData()
+    {
+        // ...
+    }
+}
+```
+
+If a view model creates other view models, those view models **should** not be design view models. Don't try and get fancy with factory methods and other unnecessary abstractions.
+
 ### View to View Model Bindings
 
 **Always** populate the View directly with values from the View Model **if** the values don't change over time (eg. if they are read-only):
--- a/docs/UICodingConventions.md
+++ b/docs/UICodingConventions.md
@@ -1315,7 +1315,7 @@ public class GoodExampleViewModel
 }
 ```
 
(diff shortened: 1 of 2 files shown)


docs/UICodingConventions.md
```diff
@@ -1335,6 +1335,55 @@ public class GoodExampleViewModel
...
+    }
+
+    private static SomeData SetupData()
+    {
+        // ...
+    }
+}
+```
+
+If a view model creates other view models, those view models **should** not be design view models. Don't try and get fancy with factory methods and other unnecessary abstractions.
```
Al12rs, line 1385: What if a Design VM has properties containing other VMs, that require passing stuff in the constructor, can I just instantiate the Design version of the child VMs instead since we are already in a design VM? 

This would avoid having to create mock data for the non design version in the parent DesignVM, on top of having to create that mock data already in the child DesignVM

erri120: As stated above, the design view model should only set up data. The normal VM would then take that data and create other VMs.

erri120: You can check the FOMOD changes I recently did. I massively cleaned up the design VMs: https://github.com/Nexus-Mods/NexusMods.App/pull/738/files#diff-9867405e9fbd8549cf89ffca330cee4eef1c8a421dba0cb637e5e5b36f59d897

Al12rs: But then you end up in a situation where if a standard VM has a constructor with custom parameters, you need to fabricate mock version of that data twice, once in the DesignVM to allow for a constructor without arguments, and then again in any parent DesignVM that needs to instantiate the standard VM.

This would then call for getting "fancy" to generate that data to avoid ending up with replicated code.

erri120: If the parent has a list of children, you can make a data generation method on the child and re-use it in the parent.

Al12rs: I would be using the data generation method located in the DesignVM to populate the main VM. 
Why not just instantiate the DesignVM? Especially if it's exactly as the example where it just sets up the data for the main one.
What is the downside in that case?

erri120: I want to avoid a case where the design view model and the normal view model do the same thing by instantiating child view models, but the difference is in the type. This was an issue with the previous FOMOD code, where I'd had duplicate code that did the same thing but with a different type.

Al12rs: But if the alternative is 
```csharp
// inside a DesignVM constructor
ChildVM = new ChildVM(ChildDesignVM.GenerateDummyData());
```
Is it really any better than just:
```csharp
// inside a DesignVM constructor
ChildVM = new ChildDesignVM();
```

erri120: In that scenario there isn't much difference. My concern mainly lies with collection of child view models, since you usually construct those by connecting to a `SourceCache`:

```csharp
//MyViewModel:
_sourceCache
  .Connect()
  .Transform(data => new ChildViewModel(data))
  .BindTo(out _children);
```

If you'd want to create a `MyDesignViewModel` and you have a `ChildDesignViewModel`, then you **should not** copy this code and change the `Transform` call to create a `ChildDesignViewModel` instead of a `ChildViewModel`. The ideal solution would be to just create dummy data in the `MyDesignViewModel` and let the base class `MyViewModel` create instances of `ChildViewModel` using that dummy data, similar to how a `ChildDesignViewModel` uses dummy data to pass it to the base constructor of `ChildViewModel`.

Al12rs: I see, so the general idea is to reduce as much as possible the changes in logic between MyVM and MyDesignVM.
I guess the example you posted above then is a better one to illustrate the point the documentation is trying to make. 
If possible the DesignVM should just generate dummy data and feed it to the base VM.

erri120: I'll update the docs to be more precise.

PR #776, opened 2023-11-15
Upgrade deps

halgari:
Updates several dependencies that were still on .Net 7.0 versions


halgari:
I screwed this up somehow...reverting and trying again

PR #778, opened 2023-11-16
ArchiveInstaller propagate properties

Aragas:
Version was not propagated
I'm not sure whether other properties should be propagated also

### DIFF
--- a/src/NexusMods.DataModel/ArchiveInstaller.cs
+++ b/src/NexusMods.DataModel/ArchiveInstaller.cs
@@ -181,6 +181,7 @@ public class ArchiveInstaller : IArchiveInstaller
                             Status = ModStatus.Installed,
                             Enabled = true,
                             Name = mod.Name,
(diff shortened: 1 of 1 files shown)


Aragas:
Do I need to fix the coverage issue? And in your opinion, should other properties like `SortRules` be propagated?

Al12rs:
No the coverage thing is fine, we might remove it in the future. 
I think we will propagate sort rules if we have an installer that actually applies some.
Thanks for the PR!

Al12rs (approved):

Aragas:
@Al12rs the Bannerlord PR has such code
https://github.com/BUTR/NexusMods.App/blob/3d5d87f77cb68f403d2ceb5e02c8830b240e3fe9/src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Installers/MountAndBlade2BannerlordModInstaller.cs#L118-L121

PR #783, opened 2023-11-21
ADR for Default App Locations

Sewer56:
## Contest / Problem Statement

The App currently stores its files in the same folder as the application (`{EntryFolder}`).

This approach has issues with certain packaging formats and installers. For example, when the application is run from
an `.AppImage`, the `EntryFolder` becomes read-only, and our application fails to run.

We need to identify alternative storage locations that are both accessible and writable, regardless of how the
application is deployed or executed.

## About The Document

This document (WIP ADR) presents us with existing options as to where we *could* potentially store our content, including some of the benefits and drawbacks for each location.

It also introduces a potential new problem that may arise should we decide to switch to a new user data location.

## Feedback Requested

This is intentionally submitted as a draft PR ***in request for feedback***.
 
More specifically, I request that everyone reads this, and forms their own preference/opinion on the subject matter ahead of time, as it is probably for the best that we discuss this (on call) as a team down the road.

If you have any other suggestions, fixes, etc. feel free to comment or even amend this PR directly (no force pushes, let's be civilized here !!).

### DIFF
--- a/docs/decisions/backend/0013-default-datastore-locations.md
+++ b/docs/decisions/backend/0013-default-datastore-locations.md
@@ -1,0 +1,136 @@
+```
+# These are optional elements. Feel free to remove any of them.
+status: {proposed}
+date: {2023-11-21 when the decision was last updated}
+deciders: {App Team}
+```
+
+# Storage Locations for NexusMods.App Files
+
+## Context and Problem Statement
+
+The App currently stores its files in the same folder as the application (`{EntryFolder}`).
+
+This approach has issues with certain packaging formats. For example, when the application is run from
+an AppImage, the `EntryFolder` becomes read-only, and our application fails to run.
+
+We need to identify alternative storage locations that are both accessible and writable, regardless of how the
+application is deployed or executed.
+
+## The Current Situation
+
+| Packaging System        | Can Write Entry Directory? | Notes                                                                       |
+|-------------------------|----------------------------|-----------------------------------------------------------------------------|
+| Windows (User Folder)   | ✅                          | No issues.                                                                  |
+| Windows (Program Files) | ⚠️                         | Requires elevated permissions.                                              |
+| macOS (User Folder)     | ✅                          | No issues.                                                                  |
+| macOS (/Applications)   | ⚠️                         | Requires elevated permissions. Reportedly affects code signing workflow.    |
+| Linux (AppImage)        | ⚠️                         | Read only. Can write to `.AppImage` folder via `$OWD` environment variable. |
+| Linux (Flatpak)         | ❌                          | Read only.                                                                  |
+
+## Per Distribution Method Notes
+
+| Packaging System | Notes                                                                                                             |
+|------------------|-------------------------------------------------------------------------------------------------------------------|
+| Windows          | Requires elevated permissions for non-user folders.                                                               |
+| macOS            | Requires elevated permissions for non-user folders.                                                               |
+| Linux (AppImage) | Single file app. Mounted on launch.                                                                               |
+| Linux (Flatpak)  | Can only access certain directories by default. Full FileSystem permissions may be requested via manifest change. |
+
+## Decision Drivers
+
+* On uninstall, the application should be able to remove all of its files from the system.
+* Application data should be user accessible.
+    * In other words, easy to find/navigate to.
+    * Because application data, also contains logs and user accessible files.
+    * e.g. If the App crashes on boot, finding logs should be easy.
+* Cross-platform consistency
+    * The paths should be consistent across different operating systems.
+    * For example, if on one platform it is in entry directory, it should do the same on other platforms too.
+* The user must have write access to the directory.
+
+## Considered Options
+
+* **Option 0: Keep using Entry Directory**
+* **Option 1: User's Home Directory**
+* **Option 2: Operating System's AppData Directory**
+
+## Decision Outcome
+
+{TBD}
+
+### Consequences
+
+{TBD}
+
+## Pros and Cons of the Options
+
+### Option 0: Keep using Entry Directory
+
+* Good, because it makes the application portable (e.g. can carry on USB stick).
+* Good, because it is consistent across deployment methods.
+* Good, because it makes user data very easy to find.
+* Bad, because it is incompatible with many packaging systems.
+
+### Option 1: User's Home Directory
+
+In a subfolder of:
+
+- `~` on Linux/macOS
+- `C:\Users\{User}` on Windows
+
+* Good, because it is easy to find.
+* Good, because it remains consistent across different deployment methods.
+* Bad, because it can clutter the user's personal space with application data. ('yet another folder in my home
+  directory')
+
+### Option 2: Operating System's AppData Directory
+
+In a subfolder of:
+
+- `AppData/Roaming` on Windows
+- `~/Library/Application Support` on macOS
+- `~/.config` on Linux a.k.a. `XDG_CONFIG_HOME`
+
+* Good, because it's the idiomatic (intended) location for application data.
+* Good, because it separates application data from user files, reducing clutter.
+* Bad, because user might struggle to find these locations.
+
+## Issue: Separate Per User and Per Machine Data
+
+The current issue with idiomatically using [AppData](#option-2-operating-systems-appdata-directory) is
+the presence of multi user systems. For example, offices, LAN centres, etc.
+
+There are two specific issues (we will be using Windows as an example):
+
+***1. Network Syncronization***
+
+Files in `AppData/Roaming` are usually downloaded upon login in these configurations, and this download typically
+happens on every login.
+
+This is bad because it means having to potentially wait a very long time (even at Gigabit speed) to download a lot of
+data on login. In the case of something like Starfield, game backup + mods could exceed 100GB, meaning a 10+ minute
+download if no data is locally cached.
+
+***2. Disk Space (Local)***
+
+If multiple users are using the same local machine, all mod + game backup data is duplicated on storage, wasting
+potentially a huge amount of space
+
+***Additional Context***
+
+The idiomatic approach for this kind of problem (using Windows as example) is storing mods + backup game files
+in a machine wide location such as `C:\ProgramData`. Per user data (loadouts, mod configs etc.) in `AppData/Roaming`.
+
+When using in a multi-machine setup, like an office, the App would load the loadout from `AppData/Roaming` and start
+downloading any mod assets it may be missing. If they are modding a game which has been modded by another user on the
+same local system, the shared backup game files in `C:\ProgramData` would be used to first restore the game to its
+original state.
+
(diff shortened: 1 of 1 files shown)


Al12rs:
So one of the biggest complaints people had with Vortex when it first came out was the fact that it put its files in Appdata by default and there was no way to avoid that.
You could move out some things but people weren't happy with it. 

From MO2 side as well, being able to set custom locations and supporting Portable installations was something very important for a lot of users.

Knowing this I think it's important that whatever solution we come up with for the default can be completely customized and overruled by users. In particular choosing a default location on C: drive for everything might incur into drive capacity issues. 

Ideally during installation or first setup the user would be able to choose whether to create a portable instance or a normal instance. A portable instance would keep all files inside the installation folder, which would need to be write enabled. The user would still need to be able to change the locations for modfile backups etc, as those can take up hundreds of GBs and they might want to put the FileStore of different games on different drives (e.g. most game mods on NVME SSD for performance, but not the Skyrim 500GB setup).

Could we get away with having the user configure this on first startup, so we don't have to choose the default ourselves?
I imagine for a lot of people Appdata could be fine for the small stuff (our sqlite database and a few config files), but they probably want to put all the big stuff in some secondary drive.

Sewer56:
> Knowing this I think it's important that whatever solution we come up with for the default can be completely customized and overruled by users. 

Note that this is already possible today 😉, I did it with some config work a while back. This ADR specifically focuses on what we should suggest to the user as a default location, rather than anything else.

As another point of note, our DataModel also supports (or used to support last I checked) having multiple archive locations; i.e. multiple places where archives are stored. This worked at the reader side, but at the writer site, there was currently not yet a mechanism to configure where new archives go. I'm not sure if the situation on that changed.

> Ideally during installation or first setup the user would be able to choose whether to create a portable instance or a normal instance

> Could we get away with having the user configure this on first startup, so we don't have to choose the default ourselves?

That's actually a great piece of feedback, ultimately it would require some UX designs from @captainsandypants down the road.

Regarding the point of when this configuration is set, it would have to be at runtime. It's not really possible to have custom install logic for every single packaging format we should strive to support and even if we could; that would be a lot of repeated work to get set up for every single packaging format.

------

Anyway, lemme add a footnote to the ADR regarding these things.

Sewer56:
Updated with some additional potential discussion topics regarding UX and information regarding running the app 'portably'.

halgari:
I think it's important to note that the reason people want portable modes is often tied to the inflexability of the app in other areas. Having to have instances for MO2, Vortex treating mods as only installable once for all loadouts, etc. Solving these issues removes a lot of the need for a truely portable app. Clearly we'll always want some sort of protable system for testing and development, but I'm not sure it's a feature that we have to ship with the MVP

erri120:
> I think it's important to note that the reason people want portable modes is often tied to the inflexability of the app in other areas. Having to have instances for MO2, Vortex treating mods as only installable once for all loadouts, etc. Solving these issues removes a lot of the need for a truely portable app. Clearly we'll always want some sort of protable system for testing and development, but I'm not sure it's a feature that we have to ship with the MVP

Our loadout system is much more advanced. Once we have a proper settings system, there won't be a need to have multiple instances of the App, since you'd be able to change per-loadout locations.

Al12rs:
I agree with LocalAppdata as the default location in windows for things that shouldn't balloon in size.

If we put something that could potentially become hundreds of GBs in Appdata by default, without explicitly asking the users we will get a lot of complaints, even if it is possible to move the folder after the fact, simply because people will not know that it can be changed or how (even if it is just there staring at them in settings).

So in my opinion we should explicitly ask the user on first setup where to put the heavy stuff, to avoid the user complaints, while the smaller things like the database can be put in the default location.

src/NexusMods.App/AppConfig.cs
```diff
@@ -96,28 +97,52 @@ public class LoggingSettings : ILoggingSettings
...
+        FilePath.ToAbsolutePath().Parent.CreateDirectory();
+        ArchiveFilePath.ToAbsolutePath().Parent.CreateDirectory();
+    }
+
+    private static AbsolutePath GetDefaultBaseDirectory(IFileSystem fs)
+    {
+        return fs.OS.MatchPlatform(
+            () => fs.GetKnownPath(KnownPath.LocalApplicationDataDirectory).Combine("NexusMods.App/Logs"),
+            () => fs.GetKnownPath(KnownPath.XDG_DATA_HOME).Combine("NexusMods.App/Logs"),
+            () => throw new NotSupportedException(
```
erri120, line 138: Logs should go into the intended XDG base directory: `XDG_STATE_HOME` as defined by the [specs](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

Sewer56:
I undrafted this by accident, oops. Please ignore.

src/ArchiveManagement/NexusMods.FileExtractor/IFileExtractorSettings.cs
```diff
@@ -21,26 +21,41 @@ public interface IFileExtractorSettings
...
+    }
+
+    private static AbsolutePath GetDefaultBaseDirectory(IFileSystem fs)
+    {
+        return fs.OS.MatchPlatform(
+            () => fs.GetKnownPath(KnownPath.LocalApplicationDataDirectory).Combine("NexusMods.App/Temp"),
+            () => fs.GetKnownPath(KnownPath.XDG_DATA_HOME).Combine("NexusMods.App/Temp"),
+            () => throw new NotSupportedException(
+                "(Note: Sewer) Paths needs PR for macOS. I don't have a non-painful way to access a Mac."));
+    }
```
erri120, line 48: The temporary extraction folder should go somewhere else. `%Temp%` can be used on Windows without issues. On Linux things are a bit different due to `tmpfs` residing in memory and/or swap partition(s). However, I still think that `/tmp` should be used as a _default_ option, and we should probably have size limit checks in place anyways.

Sewer56: I feel robbed 😭, you made this comment literally at the same time I made a commit to address this.

Sewer56: In any case, are you sure `/tmp` is a good idea?

I had the same thing in mind, because `tmpfs` is in mem/swap, we risk a very easy case of memory starvation. 

Imagine for example the user queues downloads of multiple mods at once, with a large queue. `tmpfs` becomes 30+GB, and RAM/Swap are unable to handle all this space. Suddenly the system essentially grinds to a complete halt with the end user not knowing what really happened. 

But this happened directly after interacting with the App, thus we take the blame for it (not to mention more potential support requests). 

-------

Misc Note:

I've had a few cases of forced restarts earlier in App development, when tests didn't properly clean up after themselves. My `tmpfs` filled up without me noticing and only thing I had was the power button on my machine.

While the kernel does have an OOM killer and all, it is often pretty slow to react, and won't stop a user from halting the machine or by filling up that space by adding 'just another mod'.

Sewer56: We discussed this over and decided to use XDG_STATE_HOME

src/NexusMods.App/Program.cs
```diff
@@ -111,23 +111,18 @@ private static AppConfig ReadAppConfig(AppConfig existingConfig)
...
-        }
-        catch (Exception)
+
+        var fs = FileSystem.Shared;
+        if (fs.OS.IsLinux)
         {
-            // Config doesn't exist.
-            // Try the OWD environment variable in case this is an AppImage
+            // On AppImage (Linux), 'OWD' should take precedence over the entry directory if it exists.
             var owd = Environment.GetEnvironmentVariable("OWD");
```
erri120, line 119: Add a link to https://docs.appimage.org/packaging-guide/environment-variables.html.

src/ArchiveManagement/NexusMods.FileExtractor/IFileExtractorSettings.cs
```diff
@@ -21,26 +21,44 @@ public interface IFileExtractorSettings
...
+    private static AbsolutePath GetDefaultBaseDirectory(IFileSystem fs)
+    {
+        // Note: The idiomatic place for this is Temporary Directory (/tmp on Linux, %TEMP% on Windows)
+        //       however this can be dangerous to do on Linux, as /tmp is often a RAM disk, and can be
+        //       too small to handle large files.
+        return fs.OS.MatchPlatform(
+            () => fs.GetKnownPath(KnownPath.TempDirectory).Combine("NexusMods.App/Temp"),
+            () => fs.GetKnownPath(KnownPath.XDG_STATE_HOME).Combine("NexusMods.App/Temp"),
+            () => throw new NotSupportedException(
+                "(Note: Sewer) Paths needs PR for macOS. I don't have a non-painful way to access a Mac."));
```
erri120, line 50: Default to the current directory or throw `PlatformNotSupportedException`.

src/NexusMods.App/AppConfig.cs
```diff
@@ -96,28 +97,52 @@ public class LoggingSettings : ILoggingSettings
...
+        ArchiveFilePath.ToAbsolutePath().Parent.CreateDirectory();
+    }
+
+    private static AbsolutePath GetDefaultBaseDirectory(IFileSystem fs)
+    {
+        return fs.OS.MatchPlatform(
+            () => fs.GetKnownPath(KnownPath.LocalApplicationDataDirectory).Combine("NexusMods.App/Logs"),
+            () => fs.GetKnownPath(KnownPath.XDG_STATE_HOME).Combine("NexusMods.App/Logs"),
+            () => throw new NotSupportedException(
+                "(Note: Sewer) Paths needs PR for macOS. I don't have a non-painful way to access a Mac."));
```
erri120, line 139: Default to the current directory or throw `PlatformNotSupportedException`.

src/NexusMods.DataModel/IDataModelSettings.cs
```diff
@@ -90,34 +90,70 @@ public class DataModelSettings : IDataModelSettings
...
+        IpcDataStoreFilePath.ToAbsolutePath().Parent.CreateDirectory();
     }
+
+    private static AbsolutePath GetDefaultBaseDirectory(IFileSystem fs)
+    {
+        return fs.OS.MatchPlatform(
+            () => fs.GetKnownPath(KnownPath.LocalApplicationDataDirectory).Combine("NexusMods.App/DataModel"),
+            () => fs.GetKnownPath(KnownPath.XDG_DATA_HOME).Combine("NexusMods.App/DataModel"),
+            () => throw new NotSupportedException(
+                "(Note: Sewer) Paths needs PR for macOS. I don't have a non-painful way to access a Mac."));
```
erri120, line 145: Default to the current directory or throw `PlatformNotSupportedException`.

erri120 (changes requested):

erri120 (approved):

PR #787, opened 2023-11-23
Panel resizing

### LINKED ISSUE #705: Resizing panels in workspaces
# Spike

- add space between panels
- put resizers in the space between panels
- change mouse icon when hovering over the resizers

## Proposal
_(Document any existing proposals for solutions if already known here)_

## Impact
_(Describe the new behaviour you want to see for this feature)_

## Supporting Information
_(Include any supporting screenshots etc to support the above)_

erri120:
Resolves #705. Part of #674.

- [x] Panels can be resized.
- [x] Panels have space in between them.
- [ ] Future TODO: #791
- [x] Hovering and interacting with the resizer should change the mouse icon.
- [ ] Bugs:
  - [ ] Cursor icon doesn't change in previewer? Maybe doesn't change at all, I have no idea.
  - [x] Wrong control is removed from the `Children` list.
  - [x] Space isn't expanded for the other control correctly.
  - [x] Resizers are at the wrong position after a panel has been closed.
  - [x] New panel icons aren't updated when the user stops resizing
  - [x] No minimum size for panels.

### DIFF
--- a/src/NexusMods.App.UI/WorkspaceSystem/GridUtils.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/GridUtils.cs
@@ -267,7 +267,9 @@ internal static class GridUtils
         return res.SetItems(updates);
     }
 
-    internal static IReadOnlyList<ResizerInfo> GetResizers(ImmutableDictionary<PanelId, Rect> currentState)
+    internal static IReadOnlyList<ResizerInfo> GetResizers(
+        ImmutableDictionary<PanelId, Rect> currentState,
+        bool isWorkspaceHorizontal = true)
     {
         // TODO: make this less messy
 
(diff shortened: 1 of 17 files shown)


src/NexusMods.App.UI/WorkspaceSystem/GridUtils.cs
```diff
@@ -9,7 +9,7 @@ internal static class GridUtils
     /// <summary>
     /// Returns all possible new states.
     /// </summary>
-    internal static IEnumerable<IReadOnlyDictionary<PanelId, Rect>> GetPossibleStates(IReadOnlyList<IPanelViewModel> panels, int columns, int rows)
+    internal static IEnumerable<ImmutableDictionary<PanelId, Rect>> GetPossibleStates(IReadOnlyList<IPanelViewModel> panels, int columns, int rows)
```
Al12rs, line 12: What is the reason for the change to Immutable?

PR #797, opened 2023-11-29
Remove ipc code

### LINKED ISSUE #792: Remove all IPC logic
# Improvement

## Currently 
We have the IPC Job system, the IPC messaging system in the project, etc. 

## We want
Now that we have moved to a single process design we can yank all of this code. 

## Supporting Information
The IPC Job monitor requires the use of some serialization logic, we can delete this code. In addition much of the job queue deals with specifics of updating and removing jobs that were completed in other processes. These can also be deleted. 

The Sqlite database can be switched to single process mode further improving our datastore performance. 

In general we shoudl do a once-over of the code an yank any IPC logic

halgari:
This PR makes the following changes:

* Removes all IPC message and job systems.
* Replaces the IPC message queue with a simple Rx based pub/sub system. IMessageConsumer<T> and IMessageProducer<T> remain, but they are backed by Rx
* All serialization for the message queue is removed, nothing needs to be serialized so the serialization code is dead
* RateLimiting library has been removed, it was mostly unused, and rate limiting should be done by other parts of the app (like the downloader) or by the user (by not installing 1000 mods at one time). We can revisit this in the future.
* Ratelimiting jobs have been replaced with Activities
* Activities are simple markers for "work being done"
  * Completely contained behind abstractions
  * Projects only need `NexusMods.Abstractions` to participate in activites
  * The write-side `IActivityFactory` and the read side `IActivityMonitor` are separated via interfaces, only import what you need
  * All rate limiting code is gone so code can simply say `.SetProgress` to update the progress
  * Typed versions exist for activities that are based on `Size` or other such units
  * The `IActivityMontior` exposes a `IObservable<ActivityReport>` that can be used to get updates on the activity
  * The activity reports contain throughput information, estimated completion times, etc.  
* Reworks the app startup for a "Slim mode". In this mode, only the code needed to startup the SingleProcess code and kick off the starting of the main process is loaded in the DI system. Improves startup time a little. 
* Telemetry only loads in the "normal mode" not "slim". Same for the datamodel, game logic, etc.

Resolves #792


src/NexusMods.Abstractions/Activities/IActivitySource.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <param name="value"></param>
+    public void SetProgress(T value);
+
+    /// <summary>
+    /// Adds to the progress of the activity, by a specific value.
+    /// </summary>
+    /// <param name="value"></param>
+    public void AddProgress(T value);
+
+}
```
erri120, line 62: Why would you use this compared to the non-generic `IActivitySource`?

src/NexusMods.Abstractions/Activities/ActivityGroup.cs
```diff
@@ -0,0 +1,10 @@
+using TransparentValueObjects;
+
+namespace NexusMods.Abstractions.Activities;
+
+/// <summary>
+/// A name for a group of activities, groups are useful for filtering and sorting activities. For example
+/// a group could be "Download" or "Mod Install".
+/// </summary>
+[ValueObject<String>]
+public readonly partial struct ActivityGroup;
```
erri120, line 10: Should be augmented with the `JsonAugment`.

src/NexusMods.DataModel/Activities/Activity.cs
```diff
@@ -0,0 +1,206 @@
...
+
+    /// <summary>
+    /// The activity progress, stored as a double so we can handle clamping and overflow for rounding errors
+    /// </summary>
+    protected double Percentage;
+
+    /// <summary>
+    /// The unique identifier of the activity.
+    /// </summary>
+    public ActivityId Id { get; } = ActivityId.From(Guid.NewGuid());
```
erri120, line 38: This can use the new `ActivityId.NewId()` method that gets added by the source generator for `Guid` VOs.

src/NexusMods.Abstractions/Activities/IActivitySource.cs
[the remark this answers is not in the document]
halgari: Good point, and the concrete instance does constrain it, I'll put the cosntraints here as well

src/Networking/NexusMods.Networking.Downloaders/Tasks/NxmDownloadTask.cs
```diff
@@ -30,15 +31,15 @@ public class NxmDownloadTask : IDownloadTask, IHaveDownloadVersion, IHaveFileSiz
...
-    public long CalculateThroughput<TDateTimeProvider>(TDateTimeProvider provider) where TDateTimeProvider : IDateTimeProvider
+    public long CalculateThroughput()
     {
-        if (_state.Job == null)
+        if (_state.Activity == null)
             return 0;
 
-        return (long)_state.Job.GetThroughput(DateTimeProvider.Instance).Value;
+        return (long)(((ActivityReport<Size>?)_state.ActivityStatus?.GetReport())?.Throughput ?? Size.Zero).Value;
     }
```
erri120, line 43: `ActivityReport<T>.Throughput` appears to never be set by anything, so this will always return `0`.

src/NexusMods.Abstractions/Activities/ActivityGroup.cs
erri120: nvm, I thought we serialize this as part of the download status.

src/NexusMods.Abstractions/Activities/ActivityReport.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// The estimated total time for the activity, if any progress has been made.
+    /// </summary>
+    public TimeSpan? EstimatedTotalTime => PercentagePerSecond.HasValue ?
+        TimeSpan.FromSeconds(100 / PercentagePerSecond.Value.Value) : null;
+
+    /// <summary>
+    /// The end time of the activity, if it can be calculated.
+    /// </summary>
+    public DateTime? EndTime => IsFinished ?
+        StartTime + Elapsed : StartTime + EstimatedTotalTime;
```
erri120, line 60: These should use `Optional<T>`.

erri120 (changes requested):
All good besides the `T?` for value types in the activity types. We should use `Optional<T>` instead, as it's much easier to use.

PR #798, opened 2023-11-30
Game Support: Blade And Sorcery

halgari:
Clone of #700 so I can fix the builds


halgari:
Merged as part of #799

PR #801, opened 2023-12-01
Bannerlord DI integration

Aragas:
This is more of a discussion rather than an actual PR right now.
I've removed the whole callback mechanism and added a somewhat proper DI integration, but there is an issue with holding a context for the game installation.

The easiest would be to create a scope per game installation and expose a scoped service that will return the installation data

This is an example of how it possible could be done


Aragas:
The main idea is to create a scope for `GameInstallation` and configure `GameInstallationContextAccessor` with the proper data

halgari:
I'm not quite grasping what the issue is here. Do you want delayed startup of the BL code? Or do you need information in the BL code that isn't available at DI setup time? Do we need to be able to destroy the BL objects at certain points?

`but there is an issue with holding a context for the game installation`

Can you expound on that? What's the issue with holding the context?

Aragas:
> I'm not quite grasping what the issue is here. Do you want delayed startup of the BL code? Or do you need information in the BL code that isn't available at DI setup time? Do we need to be able to destroy the BL objects at certain points?
> 
> `but there is an issue with holding a context for the game installation`
> 
> Can you expound on that? What's the issue with holding the context?

When a game installation is created, BL needs to hold state that is specific to the game installation by creating an instance of the Launcher Manager. Its stateful and uses right now the install path. In the future it will manage the Load Order of the concrete installation.
Previously I used a dictionary that held an instance of the Launcher Manager based on the installation path string, but it's not a good practice.
https://github.com/Nexus-Mods/NexusMods.App/blob/8799284e9a4b1959a13aad6af509f83900f8af0b/src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Services/LauncherManagerFactory.cs#L12
I see as an alternative creating a scope per game installation where game extensions can hold their own scoped services that can hold state specific for the installation. It might be a full launcher manager like my case or might be an interface which chooses the implementation based on the store I guess?

Regarding the questions, no delay is needed, and neither do we need additional info on DI startup. I might see a point on disposing BL resources if a game installation is removed. This could be achieved by disposing the game installation scope.

halgari:
@Aragas are you happy with this code, or is there some work to be done on it yet?

Aragas:
Yep, I'm fine with the implementation. I just hope I didn't overcomplicate the game installation scope feature, but I believe the idea will be really useful for other game extensions.  
It's roughly follows the same what ASP.NET Core does with the HttpContext injection in controllers.

Aragas:
@halgari you still plan to merge the conflicts yourself? 😂

Aragas:
> Lol yeah, when I get done with all the abstraction changes, which I think happened last week.

I can re-do the PR if needed, have spare time to allocate

erri120:
I'm closing this PR since it's gone stale. I'd probably be easier to re-do the thing instead of trying to rebase. It'd probably be better to wait with a new PR until we have more features for you to work with.

PR #808, opened 2023-12-06
Fix Advanced installer incorrectly canceling when pressing Install

Al12rs:


Sewer56 (approved):
👌

Al12rs:
For some reason the first CI pass failed. Tried reproducing locally with no success, second pass worked. 
We should keep an eye out.

PR #812, opened 2023-12-11
UI Styling and Theming Approach ADR

Al12rs:
Creating this Draft to ask for opinions on best approach to take for styling.

Main issue is whether to use ControlThemes or Styles or a mix of the two.

### DIFF
--- a/docs/decisions/frontend/0003-UI-Styling-Approach.md
+++ b/docs/decisions/frontend/0003-UI-Styling-Approach.md
@@ -1,0 +1,150 @@
+﻿# UI Styling and Theming approach
+
+## Context and Problem Statement
+
+We need a standardized way to easily style and theme our UI components.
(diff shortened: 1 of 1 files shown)


Al12rs:
From actually trying to theme part of the app using the above methods, I have decided to use primarily Style classes. with an exception for TextBlocks, where the classes sit on top of ControlThemes, which allow easy setting of the Typography Styles inside other control styles.

For all other controls it just isn't worth the extra mess of having both control themes and style classes, since control themes can't affect child items and Styles can do everything control themes can do but are composable.

Main downside with this approach is that Style classes can't set Classes on child elements, meaning that code reuse could be impacted, but it shouldn't be a blocking problem.
There is an open issue on Avalonia to allow setting Style classes from a Style, so hopefully they will add this at some point.

Closing this for now. Will update the ADR and include it in the main PR with the style changes.

PR #813, opened 2023-12-11
Fix panel splitting

erri120:
Fixes the panel splitting algorithm to respect the joined panels, depending on the orientation.

Related to #674 and 791.

### DIFF
--- a/src/NexusMods.App.UI/WorkspaceSystem/WorkspaceGridState.AdjacentPanelEnumerator.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/WorkspaceGridState.AdjacentPanelEnumerator.cs
@@ -1,0 +1,86 @@
+using System.Collections;
+using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
+
+namespace NexusMods.App.UI.WorkspaceSystem;
+
+public readonly partial struct WorkspaceGridState
+{
+    [Flags]
+    public enum AdjacencyKind : byte
+    {
+        None = 0,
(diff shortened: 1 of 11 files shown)


src/NexusMods.App.UI/WorkspaceSystem/WorkspaceGridState.AdjacentPanelEnumerator.cs
```diff
@@ -0,0 +1,86 @@
...
+    public enum AdjacencyKind : byte
+    {
+        None = 0,
+        SameRow = 1 << 0,
+        SameColumn = 1 << 1
+    }
+
+    public record struct AdjacentPanel(PanelGridState Panel, AdjacencyKind Kind);
+
+    public struct AdjacentPanelEnumerator : IEnumerator<AdjacentPanel>
```
Al12rs, line 19: Is this supposed to enumerate all panels on the same column or on the same row or just the neighboring ones (ones that share a border)?
I think that it currently does the former.

erri120: It enumerates all panels that "touch" the anchor panel. I think adjacent is the right word for that, and the `AdjacenyKind` enum should make it clear that the panels can be in the same column and/or same row.

Al12rs: You are right, I read the code wrong, it checks to make sure they share a border.

Al12rs (approved):

PR #826, opened 2024-01-08
Migrate to Trees 2.0

Sewer56:
This migrates our existing tree code to the new, efficient 'create your own tree' system.

## General Changes

- Replaced some usages of LINQ over trees with source generation.
- Adds custom tree (`ModFileTree`) containing only the tree info needed by our DataModel.
- Fixes some incorrect asserts related to use of `.AllSatisfy` throughout our codebase.
  - After fixing these asserts, it turns out we had broken code, even on `main` branch. 
  - So I fixed any failing tests related to that.
- Cleaned up some dead tree code that was unused.

## Observations

- Total lines of code in game extensions has noticeably decreased with the new APIs.
  - Note total LoC in this project increased because we now have the tree construction here instead of paths; however excluding that file, we observe around -150 LoC in this PR.
- Each tree node now fits under a single x86 cache line (64 bytes), providing significant speedup on x86.

## Other Notes

There are still opportunities for perf gains, for instance the `GetFiles().Select` pattern in some places can be replaced with `GetChildrenRecursive` with custom `TFilter` and `TSelector`, at expense of a few extra LoC. For now I've refrained from rewriting those to keep the code simple, but if additional speedup is required, the possibilities are still there.

I just didn't consider them to be significant enough perf improvements to be worth implementing in place of readability; as the impact on UX would be negligible. (Gains are limited by `IEnumerable` return type on `ModInstallerResult` )

# Reviewer

@halgari - Self requested during meeting.


Sewer56:
Note that CI might fail due to `CanInstallAndApplyMostPopularMods` for SkyrimLE. This is broken on `main` too, the failure is unrelated to this PR.

src/Games/NexusMods.Games.AdvancedInstaller.UI/AdvancedManualInstallerUI.cs
```diff
@@ -70,7 +71,7 @@ public AdvancedManualInstallerUI(IServiceProvider provider)
 
     private async Task<(bool shouldInstall, DeploymentData data)> GetDeploymentDataAsync(
         GameInstallation gameInstallation, string modName,
-        FileTreeNode<RelativePath, ModSourceFileEntry> archiveFiles)
+        KeyedBox<RelativePath, ModFileTree> archiveFiles)
```
Al12rs, line 74: I just had a quick look through, 
since we use `KeyedBox<RelativePath, ModFileTree>` a lot, can we make a type alias for it that is shorter and either has Node or Tree in the name? 

I think something simple like `RelativeModFileTree` works. To improve readability and shorten it.

halgari: I agree with this, I had to dig through the path library to find out what KeyedBox was, it turns out it's a wrapper for `TVal` with an additional `TKey` *typed* tag, the Key isn't passed along with the box, just the type of the Key. We might want to find a better name for these in the mainline parts of the application.

halgari:
Aside from the KeyedBox comment this all looks solid

src/Games/NexusMods.Games.AdvancedInstaller.UI/AdvancedManualInstallerUI.cs
Sewer56: I'm open to suggestions personally.
Heck, even `ModFileTreeNode` works for me.

Al12rs: Go ahead with `ModFileTreeNode`. We can change it later if we come up with a better name.

halgari (approved):

PR #835, opened 2024-01-11
Use separate activity to report Chunk dl reading progress.

### LINKED ISSUE #834: Download ProgressBar goes over 100% up to 200%, reporting progress twice

Al12rs:
Fixes download getting double the progress value
Fixes #834


Sewer56 (approved):
This is probably good.
Just in case, CC @halgari

PR #859, opened 2024-01-23
Update the Avalonia group

Al12rs:
Played around with it a bit to check for any issues, didn't notice any.


halgari:
Handled manually

PR #867, opened 2024-01-24
FOMOD installer fixes

### LINKED ISSUE #860: FOMOD installer failing to show images present in archive
# Bug Report

## Summary
There is a bug in the code that searches for the images, it is comparing the root of the archive tree with the first section of the path, but the root of the tree is `""` so it always fails.

Likely a small regression from when the Trait based trees have been integrated.

### LINKED ISSUE #864: Fomod installer doesn't show images when switching from option without image to option with.

### LINKED ISSUE #865: Fomod installer show option details when pressing checkbox

### LINKED ISSUE #866: [Fomod installer] Fix styling to join borders of description section
Should look like this:
image

Currently the top and bottom part have separate border with rounding.

Al12rs:
### Changes:
- Fix #860 
- Fix #864 
- Fix #865 
- Fix #866 
- Improved Fomod progress bar accuracy when there are few steps or hidden steps.
- Fixed tests to detect failures in loading fomod images.
- Enabled Multithreading when extracting archives with 7zip.

#### EDIT: further changes after reviewing with Design:
- Switched highlighted option description and image, so that description would always be visible.
- Various padding/spacing fixes.
- Fix Caps on Finish button.
- Fix auto scrolling to last option when new step is loaded.

### DIFF
--- a/src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
+++ b/src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
@@ -139,7 +139,11 @@ public sealed class UiDelegates : FomodInstaller.Interface.ui.IUIDelegates, IDis
             optionIdMappings
         );
 
-        var progress = Percent.CreateClamped(currentStepId, installSteps.Length);
+
+        // Progress is computed as currentIndex + 1 /(all steps before the current step + future visible steps + 1).
+        // The +1 are to avoid having the bar at 0% when we're at the first step, and to account for the extra finish screen.
+        var progress = Percent.CreateClamped(currentStepId + 1,
+            currentStepId + 1 + installSteps.Skip(currentStepId + 1).Count(step => step.visible) + 1);
 
         _guidedInstaller
             .RequestUserChoice(guidedInstallationStep, progress, CancellationToken.None)
@@ -243,6 +247,7 @@ public sealed class UiDelegates : FomodInstaller.Interface.ui.IUIDelegates, IDis
             Id = StepId.From(Guid.NewGuid()),
             Name = installSteps[currentStepId].name ?? string.Empty,
             Groups = stepGroups.ToArray(),
+            IsVisible = installSteps[currentStepId].visible,
             HasPreviousStep = currentStepId != 0,
             HasNextStep = currentStepId != installSteps.Count - 1,
         };
--- a/src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
+++ b/src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
@@ -247,7 +247,6 @@ public sealed class UiDelegates : FomodInstaller.Interface.ui.IUIDelegates, IDis
             Id = StepId.From(Guid.NewGuid()),
             Name = installSteps[currentStepId].name ?? string.Empty,
             Groups = stepGroups.ToArray(),
-            IsVisible = installSteps[currentStepId].visible,
             HasPreviousStep = currentStepId != 0,
             HasNextStep = currentStepId != installSteps.Count - 1,
         };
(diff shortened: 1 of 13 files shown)


src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
```diff
@@ -133,7 +136,11 @@ public void UpdateState(FomodInstaller.Interface.ui.InstallerStep[] installSteps
             optionIdMappings
         );
 
-        var progress = Percent.CreateClamped(currentStepId, installSteps.Length);
+
+        // Progress is computed as currentIndex + 1 /(all steps before the current step + future visible steps + 1).
+        // The +1 are to avoid having the bar at 0% when we're at the first step, and to account for the extra finish screen.
```
erri120, line 141: Why should the progress bar not be at 0% when the user is at the first step? The progress bar indicates how far along the user is with the installation process. 0% means the user hasn't done anything yet, since they are at the first step.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
```diff
@@ -280,7 +291,9 @@ public void UpdateState(FomodInstaller.Interface.ui.InstallerStep[] installSteps
         if (CurrentArchiveFiles is null) return null;
 
         var asPath = RelativePath.FromUnsanitizedInput(image);
-        var node = CurrentArchiveFiles.FindByPathFromRoot(asPath);
+        var node = CurrentArchiveFiles.Path() != RelativePath.Empty
+            ? CurrentArchiveFiles.FindByPathFromRoot(asPath)
+            : CurrentArchiveFiles.FindByPathFromChild(asPath);
```
erri120, line 296: You should add a short comment explaining this.

src/Games/NexusMods.Games.FOMOD/FomodAnalyzer.cs
```diff
@@ -45,7 +45,9 @@ async Task AddImageIfValid(string? imagePathFragment)
                 byte[] bytes;
                 try
                 {
-                    var node = allFiles.FindByPathFromRoot(imagePath);
+                    var node = allFiles.Path() != RelativePath.Empty
+                        ? allFiles.FindByPathFromRoot(imagePath)
+                        : allFiles.FindByPathFromChild(imagePath);
```
erri120, line 50: You should add a short comment explaining this.

erri120 (changes requested):
See comments for changes.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
Al12rs: When the bar is at 0% there is no information for the user regarding how much stuff is left since it makes no difference on whether it is 0/2 or 0/100. Setting it to 1/3 1/101 fixes this, giving the user an idea of how much each step counts towards the finish.

erri120: The user will know how much each step counts towards the finish after the first step. Having the bar already filled to a certain point is more confusing than having it at 0%.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
Al12rs: I think I will change it to always  be `FindByPathFromChild`. The path of the image should always be relative to the root anyways, so root should never be considered, even if present.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/UiDelegate.cs
Al12rs: The issue I'm specifically trying to solve is when the Fomod installer closes after the first step, with the user expecting to have much more than just the first step since the progress bar was at 0%.

erri120: Ideally, the installer should jump to the "end" screen if that happens.

Al12rs: Haven't managed to fix that yet. I can look into it more, but no guarantees.

Al12rs: Had a chat with captainsandypants and he agreed with both starting with 1 for first page and counting the Finish page towards the total.

erri120 (approved):

PR #869, opened 2024-01-24
Additional Abstraction Work

### LINKED ISSUE #207: Unnecessary Project Dependencies: Introduce Abstractions or Common Package
# Improvement

## Currently

Our C# codebase has a problem where some projects have dependencies on other projects whose code is not used, but only their abstractions are required. For instance, the `NexusMods.Hashing.xxHash64` project has a dependency on `NexusMods.DataModel.RateLimiting` because it utilizes the `IJob` interface in an extension method. 

However, no actual logic from the `NexusMods.DataModel.RateLimiting` is used, resulting in an unnecessary dependency.  
We will use `IJob` interface as an example from this point onward.  

## We want

Eliminate these unnecessary dependencies by reorganizing the codebase and introducing a more streamlined approach to handle shared abstractions or extension methods that span across multiple projects.

### Abstractions

Create `.Abstractions` packages [where appropriate] that contain the shared interfaces and base classes [e.g. the same way `Microsoft.Extensions.DependencyInjection.Abstractions` exists]. This package can be referenced by other projects that only need the abstractions without adding any additional logic; making sizes of our dependencies small.

Example:

- Create a new package called `NexusMods.DataModel.RateLimiting.Abstractions`.  
- Move the `IJob` interface from `NexusMods.DataModel.RateLimiting` to `NexusMods.DataModel.RateLimiting.Abstractions`.  
- Update the `NexusMods.Hashing.xxHash64` project to reference `NexusMods.DataModel.RateLimiting.Abstractions` instead of `NexusMods.DataModel.RateLimiting`.  

### Common/Shared Extensions Package

Alternatively you can create an `Extensions` package that bridge the gap between different projects. This package can be used internally within the app and its components without introducing unnecessary dependencies into the standalone libraries.  

Example:

- `Create a new package called `NexusMods.Extensions`.
- Move the extension method(s) using the `IJob` interface from `NexusMods.Hashing.xxHash64` to `NexusMods.Extensions`.  
- Update the `NexusMods.Hashing.xxHash64` and any other projects to no longer use `NexusMods.DataModel.RateLimiting` and other libraries it only required abstractions from.  
- Reference `NexusMods.Extensions` in app's dependencies.

Drawback:  

- This limits the usage of these extension methods to the application; as we cannot introduce the `Extensions` project as a dependency of any of the libraries.

## Related Issues

- https://github.com/Nexus-Mods/NexusMods.App/issues/208
- https://github.com/Nexus-Mods/NexusMods.App/issues/120

[Please comment below if any issue is made that fits this category, so it can be added here]

## Other Information

This is labeled as a 'spike'; the exact extent/scale of this problem is not currently fully known.

Sewer56:
- Closes https://github.com/Nexus-Mods/NexusMods.App/issues/207 

Fell asleep before I could PR this yesterday 😅, so here we are.

In the meantime, while this is being reviewed, I'll:
- Write doc file of current project layout
- Solve merge conflicts with main.

Due to the nature of this PR, some bikeshedding is expected, this is ok.


halgari (approved):
This is massive, and edits 700 files, but it's 99% namespace renaming. Looks good

PR #877, opened 2024-01-28
Adds support for OSX in the App

halgari:
The app will run in debug mode, with the debugger running. What works:

 - [x] Can find games in Steam
 - [x] Can open the UI
 - [x] Tests Pass
 - [x] Running Games
 - ~~[ ]~~ Downloading files (need NXM handler) #962 
 - ~~[ ]~~ Updating the app (needs a release first) #964 
 - [x] Run it in release mode ~~(Single process code seems broken on OSX)~~ (fixed)


src/NexusMods.StandardGameLocators/Services.cs
```diff
@@ -112,6 +116,13 @@ public static class Services
                         ),
                     }
                 ));
+            },
+        onOSX: () =>
+            {
+    #pragma warning disable CA1416
+                services.AddSingleton<IHardwareInfoProvider, HardwareInfoProvider>();
+    #pragma warning restore CA1416
```
erri120, line 124: `IHardwareInfoProvider` is required for EA Desktop, and the `HardwareInfoProvider` implementation is Windows-only.

This should be removed.

src/Abstractions/NexusMods.Abstractions.FileExtractor/IFileExtractorSettings.cs
```diff
@@ -46,8 +46,7 @@ private static AbsolutePath GetDefaultBaseDirectory(IFileSystem fs)
         return fs.OS.MatchPlatform(
             () => fs.GetKnownPath(KnownPath.TempDirectory).Combine("NexusMods.App/Temp"),
             () => fs.GetKnownPath(KnownPath.XDG_STATE_HOME).Combine("NexusMods.App/Temp"),
-            () => throw new PlatformNotSupportedException(
-                "(Note: Sewer) Paths needs PR for macOS. I don't have a non-painful way to access a Mac."));
+            () => fs.GetKnownPath(KnownPath.TempDirectory).Combine("NexusMods.App/Temp"));
```
erri120, line 49: Should this be `NexusMods_App` instead?

erri120 (changes requested):
See comments.

halgari:
This should be good to merge. I had to implement some code to make sure the UI always runs on the main thread (the thread that calls `Main`. Turns out that the fact that this has worked this long is a side effect of our OS implementations not a design feature of Avalonia. https://github.com/AvaloniaUI/Avalonia/issues/12554 Avalonia has no plans to lift this restriction.

With that we're good to go for OSX support. I need to get in NXM link support, and figure out release/updating but I'll file other issues for that

src/Abstractions/NexusMods.Abstractions.FileExtractor/IFileExtractorSettings.cs
```diff
@@ -46,8 +46,8 @@ private static AbsolutePath GetDefaultBaseDirectory(IFileSystem fs)
         return fs.OS.MatchPlatform(
             () => fs.GetKnownPath(KnownPath.TempDirectory).Combine("NexusMods.App/Temp"),
             () => fs.GetKnownPath(KnownPath.XDG_STATE_HOME).Combine("NexusMods.App/Temp"),
-            () => throw new PlatformNotSupportedException(
-                "(Note: Sewer) Paths needs PR for macOS. I don't have a non-painful way to access a Mac."));
+            // Use _App vs .App so as not to confuse OSX into thinking this is
+            () => fs.GetKnownPath(KnownPath.TempDirectory).Combine("NexusMods_App/Temp"));
```
erri120, line 50: I think that comment got cut off.

src/NexusMods.App/GlobalFlags.cs
```diff
@@ -0,0 +1,24 @@
+using System.Collections.Concurrent;
+
+namespace NexusMods.App;
+
+/// <summary>
+/// Holds global information about the application.
+/// </summary>
+public static class GlobalFlags
+{
```
erri120, line 9: `GlobalFlags` doesn't really fit as a name. Maybe something like `ThreadData` or similar?

src/NexusMods.App/GlobalFlags.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// <summary>
+    /// Holds the thread that started the application.
+    /// </summary>
+    public static Thread StartingThread { get; internal set; } = null!;
+
```
erri120, line 14: This property should have some safeguards to prevent misuse. [`InvalidOperationException`](https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception) is thrown when a method call is invalid for the object's current state.

```csharp
private static Thread? _startingThread = null;

public static Thread StartingThread
{
  get
  {
    if (_startingThread is null) throw new InvalidOperationException();
    return _startingThread;
  }
  set
  {
    if (_startingThread is not null) throw new InvalidOperationException();
    _startingThread = value;
  }
}
```

erri120 (changes requested):
See comments.

src/NexusMods.App/GlobalFlags.cs
halgari: Decided to simplify this even more and put it behind an argumentless method that sets the private thread method to the current thread.

PR #891, opened 2024-01-30
Use java directly

erri120:


Sewer56:
I was going to say 'this is cursed'.
But it's just code signing

Sewer56 (approved):

PR #898, opened 2024-01-31
Workspace integration

erri120:
Part of #706.

- [x] add "add panel" button to toolbar
- [x] rebase and fix conflicts
- [x] resolve TODOs in code
  - [x] cleanup
  - [x] better API

PR #908, opened 2024-02-05
Convert "Right Content" to Pages

### LINKED ISSUE #409: Add the workspace "panel" system to the app and adapt existing panels
The following needs to be done:

* Finish up and PR the existing work for the structure defined in [UI ADR-0001](https://github.com/Nexus-Mods/NexusMods.App/blob/main/docs/decisions/frontend/0001-UI-Panel-Layout.md)
* Build a structure for defining and registering panels via the DI system
* Adapt the existing views to the panel system:
  * Downloads Panels
  * Mods loadout
  * Detected games

### LINKED ISSUE #675: Move 'My Games' to a workspace/panel
# Improvement

## Currently

The 'My Games' page is not currently set within a workspace or panel.

## We want

The 'My Games' page to be set within a workspace and panel.

For the purposes of this issue, when navigating directly to the page via the left-menu we should open it in a new workspace/panel.

## Acceptance Criteria

### LINKED ISSUE #678: Move 'Downloads' to a workspace/panel
# Improvement

## Currently

The 'Downloads' page is not currently set within a workspace or panel.

## We want

The 'Downloads' page to be set within a workspace and panel.

For the purposes of this issue, when navigating directly to the page via the spine-menu we should open it in a new workspace/panel.

## Acceptance Criteria

erri120:
**Requires #898**

- Resolves #409
- Resolves #678
- Resolves #675

TODO:

- [x] get the left menu back


src/NexusMods.App.UI/Pages/Downloads/InProgressPage.cs
```diff
@@ -0,0 +1,38 @@
...
+        return ServiceProvider.GetRequiredService<IInProgressViewModel>();
+    }
+
+    public override IEnumerable<PageDiscoveryDetails?> GetDiscoveryDetails(IWorkspaceContext workspaceContext)
+    {
+        yield return new PageDiscoveryDetails
+        {
+            // TODO: translations?
+            SectionName = "Downloads",
+            ItemName = "In-progress Downloads",
```
Al12rs, line 30: Should we go ahead and add these to translations?

src/NexusMods.App.UI/WorkspaceSystem/Context/DownloadsContext.cs
```diff
@@ -0,0 +1,3 @@
+namespace NexusMods.App.UI.WorkspaceSystem;
+
+public record DownloadsContext : IWorkspaceContext;
```
Al12rs, line 3: Downloads will need an optional Loadout context in the future, do you want to add it now?
Maybe we need to discuss it a bit with Design?

erri120: > Downloads will need an optional Loadout context in the future

What do you mean?

src/NexusMods.App.UI/WorkspaceSystem/Controllers/WorkspaceController.cs
```diff
@@ -137,32 +161,87 @@ public void ChangeActiveWorkspace(WorkspaceId workspaceId)
...
+            return newWorkspace;
+        }
+
+        ChangeActiveWorkspace(existingWorkspace.Value.Item1.Id);
+        return existingWorkspace.Value.Item1;
+    }
+
     public void AddPanel(WorkspaceId workspaceId, WorkspaceGridState newWorkspaceState, AddPanelBehavior behavior)
     {
+        Dispatcher.UIThread.VerifyAccess();
```
Al12rs, line 211: Is this temporary for debug or is that going to be in the final code?
I would rather not have it

erri120: The controller can only work on the UI thread. Using it in another thread is not allowed.

src/NexusMods.App.UI/WorkspaceSystem/Context/DownloadsContext.cs
erri120: > When opening the DL page in a Loadout Workspace it should automatically be filtered for that workspace.

That'd be for the page context `InProgressPageContext`. This is `DownloadsContext` which is a `WorkspaceContext`.

erri120: Either way, it's not needed right now.

src/NexusMods.App.UI/Controls/FoundGames/FoundGamesView.axaml
```diff
@@ -2,16 +2,17 @@
...
     mc:Ignorable="d"
-    x:Class="NexusMods.App.UI.RightContent.FoundGamesView"
-    x:TypeArguments="rightContent:IFoundGamesViewModel"
+    x:Class="NexusMods.App.UI.Controls.FoundGames.FoundGamesView"
+    x:TypeArguments="foundGames:IFoundGamesViewModel"
     xmlns="https://github.com/avaloniaui"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:reactiveUi="http://reactiveui.net"
     xmlns:rightContent="clr-namespace:NexusMods.App.UI.RightContent"
```
Al12rs, line 11: Leftover RightContent ref

Al12rs (changes requested):

src/NexusMods.App.UI/Controls/Spine/SpineViewModel.cs
```diff
@@ -71,11 +79,26 @@ public class SpineViewModel : AViewModel<ISpineViewModel>, ISpineViewModel
...
+                .Select(workspace => workspace?.Context)
+                .Select(context =>
+                {
+                    if (context is LoadoutContext loadoutContext)
+                    {
+                        return new LoadoutLeftMenuViewModel(
+                            loadoutContext,
+                            workspaceController,
+                            serviceProvider
+                        );
```
Al12rs, line 95: This will have the Apply button in the future, so I don't think that creating and discarding it each time the Loadout changes is a good idea. 

Would you consider having the left menu VM be part of the Workspace instead?
Something like a Property that contains the leftmenu that gets populated when the workspace is created and is kept alive in the workspace?

Keep in mind that these left menus will get busier as we add things: 
image

src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
```diff
@@ -49,11 +47,12 @@ public class MainWindowViewModel : AViewModel<IMainWindowViewModel>, IMainWindow
             serviceProvider: serviceProvider
         );
 
-        topBarViewModel.AddPanelDropDownViewModel = new AddPanelDropDownViewModel(WorkspaceController);
-        TopBar = topBarViewModel;
+        TopBar = serviceProvider.GetRequiredService<ITopBarViewModel>();
+        TopBar.AddPanelDropDownViewModel = new AddPanelDropDownViewModel(WorkspaceController);
+
+        Spine = serviceProvider.GetRequiredService<ISpineViewModel>();
+        DevelopmentBuildBanner = serviceProvider.GetRequiredService<IDevelopmentBuildBannerViewModel>();
```
Al12rs, line 54: Why not getting these through constructor injection instead of `GetRequiredService()`?

Al12rs (changes requested):

src/NexusMods.App.UI/Controls/Spine/SpineViewModel.cs
erri120: > This will have the Apply button in the future, so I don't think that creating and discarding it each time the Loadout changes is a good idea.

Right now, we only have a launch button.

> Would you consider having the left menu VM be part of the Workspace instead?

No, we have workspaces that don't have left menus. The left menu is a main-window-concept.

src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
erri120: This way they get instantiated at the `GetRequiredService` call instead before the constructor. If they get instantiated before the constructor, then the `MainWindowViewModel` won't be able to register itself as a new window.

Al12rs: Ah, I see. This might be worth adding a comment to then.

src/NexusMods.App.UI/Controls/Spine/SpineViewModel.cs
Al12rs: You are right, popped out workspaces don't have navigation. Doesn't mean they can't have the info anyways. Create the nav lazily if requested. 

If you still don't like that idea, I still think we want to keep the VMs alive. 

I would propose to keep the DL and MyGames directly in the Spine, while for the loadouts we could either have a map, one for each loadout, or a single one where the context gets replaced when switching. I think it would be better to keep a map for each loadout instead though.

erri120: We can just use the factory pattern again to let some factory create a left menu depending on the current workspace context. In the `MainWindowViewModel` we'd just react to context changes in the workspace and find a VM using a factory that fits that context.

But this isn't something we should do in this PR.

Al12rs: Ok, the factory idea works, as long as the spine or something then keeps the VMs alive after when the workspace is switched.

Al12rs (approved):

PR #912, opened 2024-02-07
Workspace integration improvements

Al12rs:
- Fixed various erratic behaviour of tab header area when resizing
- Disabled Add Panel button when it is empty
- Closed Add Panel Drop Down when item selection is changed
- Fixed missing bottom border of Scroll left button in tab are
- Basic MainWindow layout and border adjustments to match new designs

This PR:
image

Latest design:
image

Stuff still needs moving around


Al12rs:
CI tests fail with failed to open database file again

PR #913, opened 2024-02-07
Update Stardew Valley docs

erri120:


docs/games/0002-StardewValley.md
```diff
@@ -138,6 +142,59 @@ Regardless of platform, similarly to Bethesda Script Extenders, SMAPI comes with
...
+
+### File Conflicts
+
+Before SMAPI Content Packs and the Content Patcher were a thing, assets had to be changed by replacing the raw XNB files in the `Content` directory. Most XNB mods have been [migrated](https://forums.stardewvalley.net/threads/migrating-xnb-mods-to-content-patcher-packs.564/) to the Content Patcher format, but some older mods didn't receive an "official" update. The wiki contains a [list of unofficial updates](https://stardewvalleywiki.com/Modding:Using_XNB_mods#Alternatives_using_Content_Patcher).
+
+Two mods that replace the same XNB file are in conflict with each other and the order in which they are deployed determines the winner.
+
+### C# Mods
+
+C# Mods can target an incompatible version of SMAPI or of the game itself. Conflicts between SMAPI mods are logic/functionality based.
```
Al12rs, line 155: For design it might be good to have a list of possible incompatibilities:
- Can a mod require a newer version of SMAPI?
- Can a mod require an older version of SMAPI (or is forward compatibility guaranteed?)
- Same for Game versions
- Same for Mod dependencies versions
- Can mods define other mods they are incompatible with in the manifest?

docs/games/0002-StardewValley.md
```diff
@@ -138,6 +142,59 @@ Regardless of platform, similarly to Bethesda Script Extenders, SMAPI comes with
...
+    {
+      "UniqueID": "Foo"
+    }
+  ]
+}
+```
+
+SMAPI will first load the C# mod `Pathoschild.ContentPatcher`. The Content Patcher mod will get a list of owned Content Packs from SMAPI. This list is already sorted for dependencies. In the example above, `Foo` will be applied _before_ `Bar` because `Bar` has a dependency on `Foo`. If both `Foo` and `Bar` edit the same asset, the edits of `Bar` will be applied last.
+
+Detecting if two Content Packs are in conflict with one another depends on the Content Pack Framework. Each framework has its own way of modifying existing, or adding new assets.
```
Al12rs, line 196: Is this to say that this out of scope for the app or that we need to find out that info for each framework?

erri120: That's something I want to bring up today.

docs/games/0002-StardewValley.md
erri120: Mods can use the `manifest.json` to define a [minimum SMAPI version](https://stardewvalleywiki.com/Modding:Modder_Guide/APIs/Manifest#Minimum_SMAPI_version). Since SMAPI versions are tied to game versions, you indirectly specify what game versions the mod supports.

erri120: SMAPI also offers a mod compatibility list: https://smapi.io/mods

Al12rs: Ok, so it assumes that a mod working with version 1.0 of SMAPI will work on 1.1, and same for mod dependencies. This is good info to know. Could you add it to the doc? Same for the Game being indirect dependency.

SMAPI ver-> Game ver on the other hand is exact? So new game version is not compatible usually?

Al12rs (approved):

PR #914, opened 2024-02-07
Added: Basic File-level Deduplication in our FileStore

Sewer56:
Quick Summary:

- Tests will now clear tables between runs, making the tests less flaky (no leftover data from previous runs).
    - Note that this is slow. I didn't add a 'fast' routine for this because we will be switching underlying database code in Event Sourcing migration.
- `FileOriginRegistry` will only backup new files that are not already backed up anywhere.
  - e.g. Installing same mod twice results in no new `.nx` data written.
  - e.g. Installing an update will only write the files that are new in that specific archive.
  - e.g. If multiple mods have the same files, they will be deduplicated.
  
 Also added some basic tests around re-archiving our downloads.


src/NexusMods.DataModel/FileOriginRegistry.cs
```diff
@@ -60,26 +62,56 @@ public async ValueTask<DownloadId> RegisterDownload(AbsolutePath path, AArchiveM
...
+        {
+            var ent = _dataStore.Get<DownloadAnalysis>(id);
+
+            // shouldn't be null but technically possible in case of concurrent modifications
+            if (ent == null)
+                continue;
+
+            foreach (var content in ent.Contents)
+                knownHashes.Add(content.Hash.Value);
+        }
```
Al12rs, line 83: This feels like an suboptimal solution. We should have a map of all files somewhere that we can use for lookup.
Now the question is where. 

I would say in the `NxFileStore` is probably the correct place. 
This means that the map needs to be updated when new files are added and in the future when files will be deleted.

Then the implementation of `NxFileStore.HaveFile` can use the map directly, without having to iterate everything for each request, which sounds quite bad.

Does that make sense? Are there issues with the idea or is it something you could do?

src/NexusMods.DataModel/FileOriginRegistry.cs
```diff
@@ -60,26 +62,56 @@ public async ValueTask<DownloadId> RegisterDownload(AbsolutePath path, AArchiveM
...
+
         _logger.LogInformation("Analyzing archive: {Name}", path);
         foreach (var file in path.EnumerateFiles())
         {
             // TODO: report this as progress
             var hash = await file.XxHash64Async(token: token);
```
Al12rs, line 89: I wish this was multithreaded...

src/NexusMods.DataModel/FileOriginRegistry.cs
Sewer56: It is suboptimal, indeed. I didn't feel like introducing a table just for file hashes as that'd probably be overkill for a single use case.

Pulling this data from `DownloadMetadata` table isn't going to be a bottleneck, what is currently the bottleneck is sending `Get` requests for every single item from the `DataStore` as that involves a lot of back and forth (especially as we scale past 1000 items).

Having a `GetAll` API would solve this problem and be performant enough that it's make a negligible difference in the speed of the overall operation.

I just left it this way for the time being as I want to wait till event sourcing is done. EventSourcing will change database backend from SQLite to RocksDB. 

I'm not sure if @halgari plans to change the interface for this, but there's no point of me writing a `GetAll` for Sqlite if I'll just have to write a `GetAll` method for RocksDB in a month.

Just open an issue for this.

Al12rs: Wouldn't adding a `HashSet` of file hashes to `NxFileStore` bypass the need to access the data base? 

Then you get the hit on startup or on first use of `NxFileStore.HaveFile()` if you want to make it lazy, but after that you bypass the `DataStore` right?

This would make `HaveFile` fast, which is what we need here, and would enable us to use it more liberally in the future.

If later we can optimize the first hit on the database to get all the hashes after we switch DB, then that is great.

What do you think?

Al12rs: Cool, yeah making it lazy should help.

Al12rs (changes requested):
As discussed above: 
- move `HashSet` in `NxFileStore`, 
- make it lazy 
- if already populated, update it when new files are backed up

halgari:
@Sewer56 @Al12rs 

I'm pretty sure NxFileStore already has a exists check:

```csharp
    private unsafe bool TryGetLocation(Hash hash, out AbsolutePath archivePath, out FileEntry fileEntry)
    {
        var prefix = new Id64(EntityCategory.ArchivedFiles, (ulong)hash);
        foreach (var entry in _store.GetByPrefix<ArchivedFiles>(prefix))
        {
            foreach (var location in _archiveLocations)
            {
                var path = location.Combine(entry.File);
                if (!path.FileExists) continue;

                archivePath = path;

                fixed (byte* ptr = entry.FileEntryData.AsSpan())
                {
                    var reader = new LittleEndianReader(ptr);
                    FileEntry tmpEntry = default;

                    tmpEntry.FromReaderV1(ref reader);
                    fileEntry = tmpEntry;
                    return true;
                }
            }
        }

        archivePath = default;
        fileEntry = default;
        return false;
    }
```

So I don't think we need to keep anything in memory. That's backed by Sqlite and has a O(n) search time based on the number of duplicates in the system, which after this patch will make it O(1). We can just call `TryGetLocation` and discard the output.

Sewer56:
I'm half asleep but damn we really missed this. Well, that's very easy.

Sewer56:
Anyway, my doggo woke me up from sleep because she wanted outside >.<

While I wait half asleep, the following comes to mind:

- We decided earlier when talking that we'll deduplicate on archive level too (avoiding extraction if possible.
- Deduping on Archive level means we store Archive Hash as finalHash in `DownloadAnalysis`.
- Which means we have to fetch `DownloadAnalysis` either way, to extract the ID field.

Also add that `TryGetLocation` could potentially bottleneck hard if used with those rare mods that have over 30000 files (exact impact remains to be seen though), due to sheer overhead of separately hitting the DB each time.

So I think a `GetAll()` method for a given entity is still probably desirable. `GetByPrefix` with an empty prefix maybe does the job, if the underlying DB is smart enough to eliminate it, but I have no way of knowing till I wake up and get to my pc again. 

I think I'll try that as a temp workaround, and measure overhead (very roughly) with some quick logging. (I'll fill the data store with 100k files). If it's fast enough I'll just use that and switch over to a `GetAll()` method post DB migration.

Nap time.

Al12rs:
To test or profile you can try installing SMIM from Skyrim:
https://www.nexusmods.com/skyrim/mods/8655?tab=files
Very popular and reasonably large without being absurd.

Sewer56:
I don't like half assing a job, so I went the full way.

- Now dedupes on Archives AND Files.
- I added a `GetAll` method to `IDataStore` myself.

I did a very, very rough benchmark by modifying one of the tests. 

----------------
!!! tl;dr File Dedupe Overhead: ~150ms for 30000mods with 300,000 files. (Basically free)

Namely, I wrote a loop which:
- Generates a Mod with 7-14 random small files.
- Packs it.
- Installs it.

After around 45 minutes, I installed 30000 mods through this loop, with approx 300,000 files total.
This is pretty much *worst case scenario* for current implementation (it fares better with fewer mods with many files like SMIM).

All in all, once you hit 30000 mods, the overhead is around `150ms` (slow by my standards but still 'acceptable' for 30000 mods with 300,000 files). 

This was done with the SQL statement ``SELECT Id,Data FROM [{tableName}]`` which I added in `GetAll`.
Around 70% of CPU time was spent fetching items from the DataStore, around 29% was spent deserializing; remaining <0.4% was expanding HashSets to use for deduping.

---------------------
!!! Archive Dedupe Overhead: `sizeof(Archive) / DiskReadSpeed`

Approximate overheads:

| Storage Type  | 1GB | 5GB | 10GB | 20GB |
|----------------|------|------|-------|--------|
| HDD              |  7.1s |  35s  |  71.4s  | 142.9s | 
| SATA SSD      |  2.5s |  12.5s     | 25s | 50s  |
| NVMe           | 0.3s  |   1.92s      |  3.8s  | 7.6s |

Unlike File Dedupe which is basically free and in practice improves performance by compressing less on mod updates, archive dedupe unfortunately carries pure overhead as we don't really have much use for the archive hash outside of deduping future archives.

I'm not sure if this is the best way to go about this, part of me feels like getting the user to not download the same archive in the first place may instead perhaps be a better approach. Or hashing a subset of the file (e.g. 4096 bytes), and using that as the hash for deduplication instead; it'd probably bring better average performance long run, but that's an issue for another issue.

src/NexusMods.DataModel/FileOriginRegistry.cs
```diff
@@ -109,28 +157,39 @@ public async ValueTask<DownloadId> RegisterFolder(AbsolutePath path, AArchiveMet
...
+        // TODO: This may need some benchmarking, it's unknown how well this scales across huge loadouts.
+        var archiveHashes = new Dictionary<ulong, DownloadId>();
+        var fileHashes = new HashSet<ulong>();
+        foreach (var ent in _dataStore.GetAll<DownloadAnalysis>(EntityCategory.DownloadMetadata)!)
+        {
+            // Note: Mods added from raw folders have a hash of '0', this is fine, just means there's nothing to dedupe.
+            archiveHashes[ent.Hash.Value] = ent.DownloadId;
+            foreach (var content in ent.Contents)
+                fileHashes.Add(content.Hash.Value);
+        }
```
Al12rs, line 182: As we discussed in standup not all `ArchivedFiles` have a `DownloadAnalysis` so getting this doesn't cover all the game files or files that get Ingested.

You need to get all the `ArchiveFiles` directly.
You might want to add a method to `IFileStore` to do it instead of going directly for the database, just to keep things more organized.

tests/NexusMods.DataModel.Tests/FileOriginRegistryTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+public class FileOriginRegistryTests : ADataModelTest<FileOriginRegistryTests>
+{
+    public FileOriginRegistryTests(IServiceProvider provider) : base(provider)
+    {
+
+    }
+
+    [Fact]
+    public async Task RegisterFolder_ShouldRegisterCorrectly()
```
Al12rs, line 28: See if you want to add a test for checking against entries that are in the datastore without an associated `DownloadAnalysis`

src/NexusMods.DataModel/FileOriginRegistry.cs
Sewer56: Is this desirable?

This means we are doing two trips to the database, rather than one. 

If we want to return a previously downloaded archive to do Archive level dedupe we need to query `DownloadAnalysis` so we can get a DownloadId, so that one's unavoidable. 

While it is true, that `DownloadAnalysis` does not contain every single file that the App has ever touched, consider the context of the situation we have here.

Hear me out:

- Our goal (and the original issue this PR addresses) is to deduplicate mod files. 
- The FileOriginRegistry's `DownloadAnalysis` table should contain all of the mod files.
    - Including those installed manually via `Add Mod`.
    - In other words every single file that's tied to a mod. (Not ingested)

So the issue you raised is the fact that we are not deduping with the original game files. 

*The whole point of a mod*, is to modify the game. If you are shipping unmodified game files as part of mods then you're
- [Breaking our own (and other sites') ToS](https://help.nexusmods.com/article/28-file-submission-guidelines) 
- Probably doing something wrong.

I currently pull the file hashes from there because I have to pull `DownloadAnalysis` anyway and it just happens to contain the set of all mod files.

Yes, there are also ingested files, but those are a separate issue. There is no dedupe mechanism for ingest (yet), and how we will use our ingest in the future is not yet clearly defined. (For example, the user may choose to delete ingested files, put them inside an existing mod. etc.) For ingest we can handle the deduping at the time the user decides what to do with the files, we can revisit this once we get there.

src/NexusMods.DataModel/FileOriginRegistry.cs
```diff
@@ -109,28 +158,29 @@ public async ValueTask<DownloadId> RegisterFolder(AbsolutePath path, AArchiveMet
...
+    /// <summary>
+    ///     Gets hashes of all known archives.
+    /// </summary>
+    private bool TryGetDownloadIdForHash(ulong expectedHash, [NotNullWhen(true)] out DownloadId? analysis)
     {
-        return _dataStore.Get<DownloadAnalysis>(IId.From(EntityCategory.DownloadMetadata, id.Value))!;
-    }
+        // Build a Hash Table of all currently known files. We do this to deduplicate files between downloads.
+        // Table not indexed by hash, so we need to query all.
+        foreach (var ent in _dataStore.GetAll<DownloadAnalysis>(EntityCategory.DownloadMetadata)!)
```
Al12rs, line 174: Comments need update

Sewer56: True!

Al12rs (approved):

Al12rs:
@Sewer56 Are the test failures due to the hashing bug? Should we wait for a fix on that to get merged first?

Sewer56:
> @Sewer56 Are the test failures due to the hashing bug? Should we wait for a fix on that to get merged first?

No, they're completely unrelated to this PR. Maybe something on Main.

PR #923, opened 2024-02-08
Fix SMAPI installer

### LINKED ISSUE #645: Implement SMAPI support for real
SMAPI support is currently quite minimal due to lack of support in the datamodel for what we were trying to accomplish. We should fix this.

erri120:
Resolves #645.

- Fixes the SMAPI Installer
- Removes the outdated `GameFile` type
- Cleanup in various files

The tests could be expanded but probably in another PR. We should look into using the Verify package instead of manually writing these tests.


src/Games/NexusMods.Games.StardewValley/Installers/SMAPIInstaller.cs
```diff
@@ -79,71 +101,78 @@ private SMAPIInstaller(IOSInformation osInformation, IFileHashCache fileHashCach
...
+            {
+                to = new GamePath(LocationId.Game, "StardewValley");
+            }
+
+            // For Xbox Game Pass: replace "Stardew Valley.exe" with "StardewModdingAPI.exe"
+            // https://stardewvalleywiki.com/Modding:Installing_SMAPI_on_Windows#Xbox_app
+            if (isXboxGamePass && kv.Item.FileName.Equals("StardewModdingAPI.exe"))
+            {
+                to = new GamePath(LocationId.Game, "Stardew Valley.exe");
+            }
```
Al12rs, line 147: So "Stardew Valley.exe" is some kind of launcher that isn't needed for executing the game?

erri120: I'm not sure on the specifics, especially with the Xbox version.

erri120: I think the executable doesn't do much, the main code is in the `Stardew Valley.dll` (it's a C# game).

Al12rs: No chance that SMAPI relies on the backup of that launcher file being where its install script puts it?

PR #926, opened 2024-02-08
Spine UI fixes (workspaces improvements)

Al12rs:
- Highlight active workspace button on the spine
- Add tooltips for spine buttons


erri120 (approved):
This is fine for now. Similar to the current left menu solution, this doesn't scale.

I believe a better approach would be to delegate both the left menu and the active state to another component.

For the left menu, we could have a simple factory that creates the appropriate left menu via DI.

The active state could be on the spine item itself.

PR #936, opened 2024-02-13
Add GenericIcon

### LINKED ISSUE #934: Add generic icon control
Currently, we have multiple ways of displaying an icon:

- `Icon` control from `Projektanker.Icons.Avalonia`: [Styling](https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Icons/IconsStyles.axaml).
- `Svg` control from `Avalonia.Svg`: [Styling](https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Icons/IconsStyles.axaml).
- `Image` control from `Avalonia` that uses an `IImage` property to draw.
- `PathIcon` control from `Avalonia` that uses a `Geometry` property to draw.

Choosing the right control has been dependent on the icon source, which makes using icons very difficult. A solution to this problem is adding a generic icon control that provides a unified way of displaying icons.

erri120:
Resolves #934.


Al12rs (changes requested):
As discussed in call:
- Add `Size` attribute to manage both Height/Width and FontSize in a unified way.
- Make sure Size works for all of them, add Stretch if necessary etc.
- Add `GenericIconStyles` and add the examples there, use Classes in `GenericIcon.axaml`

erri120:
Fixed the sizing issue.

Al12rs:
For the record: 
I would have liked to have convenience property setters in `GenericIcon` to allow 
```xml
<local:GenericIcon SvgPath="avares://NexusMods.App.UI/Assets/Icons/add_circle_24px.svg"/>
```

as opposed to how you currently have to do it in the PR:
```xml
            <local:GenericIcon>
                <local:GenericIcon.Value>
                    <local:IconValue>
                        <local:IconValue.AvaloniaSvgSetter>
                            <local:AvaloniaSvg Path="avares://NexusMods.App.UI/Assets/Icons/add_circle_24px.svg"/>
                        </local:IconValue.AvaloniaSvgSetter>
                    </local:IconValue>
                </local:GenericIcon.Value>
            </local:GenericIcon>
```
, which would be hidden inside a Style Class.

The argument against from @erri120 was that this would defeat the purpose of having a unified Api, as it would be possible to OneWayBind to the other setter properties as well, ending up with different ways the control can be interacted with.

I still would have liked the convenience, regardless. We agreed to disagree.

Al12rs (approved):

PR #938, opened 2024-02-13
Fix SMAPI apply order

### LINKED ISSUE #930: SMAPI doesn't overwrite Game Files
The SMAPI installer is supposed to replace the existing game file `StardewValley` with a file from the SMAPI archive in `internal/linux/internal.dat/unix-launcher.sh`.

Debugging shows that the flattened loadout attributes the file at `{GamePath}/StardewValley` to the `Game Files` mod, not SMAPI. This happens because of sorting, the `Game Files`  "mod" overwrite the `SMAPI` "mod".

erri120:
Fix #930.

### DIFF
--- a/src/Games/NexusMods.Games.StardewValley/Sorters/SMAPISorter.cs
+++ b/src/Games/NexusMods.Games.StardewValley/Sorters/SMAPISorter.cs
@@ -1,0 +1,32 @@
+using System.Diagnostics.CodeAnalysis;
+using NexusMods.Abstractions.DataModel.Entities.Sorting;
+using NexusMods.Abstractions.Loadouts;
+using NexusMods.Abstractions.Loadouts.Mods;
+using NexusMods.Abstractions.Loadouts.Sorting;
+using NexusMods.Abstractions.Triggers;
+
+namespace NexusMods.Games.StardewValley.Sorters;
+
+[SuppressMessage("ReSharper", "InconsistentNaming")]
+public class SMAPISorter : IGeneratedSortRule, ISortRule<Mod, ModId>
+{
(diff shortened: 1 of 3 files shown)


src/Games/NexusMods.Games.StardewValley/Sorters/SMAPISorter.cs
```diff
@@ -0,0 +1,34 @@
...
+using NexusMods.Abstractions.Serialization.Attributes;
+using NexusMods.Abstractions.Triggers;
+
+namespace NexusMods.Games.StardewValley.Sorters;
+
+[JsonName("NexusMods.Games.StardewValley.Sorters")]
+[SuppressMessage("ReSharper", "InconsistentNaming")]
+public class SMAPISorter : IGeneratedSortRule, ISortRule<Mod, ModId>
+{
+    public ITriggerFilter<ModId, Loadout> TriggerFilter => throw new NotImplementedException();
```
Al12rs, line 15: Isn't this going to be a problem? If someone actually calls the TriggerFilter in the future?

erri120: Nothing is calling this atm and I'm not sure how to properly implement this since there is no usage.

Al12rs: I think the idea is that you provide a FingerPrint, if it is different from the previous one, you need to recompute the rule. I think you can return an empty fingerprint and it should never recompute it?

halgari: That trigger isn't being used currently as the cost of re-generating the rules is low enough, not sure if we'll ever use it in the future.

PR #939, opened 2024-02-13
Add back Left Menus and Left Menu Navigation

### LINKED ISSUE #932: Implement LeftMenu navigation
Workspace integration removed the old Left Menu navigation.
No replacement was put in place.

This needs to be implemented to allow adding new pages to any of the workspaces and allow users to navigate between them.

This is also required for the Apply components, that will live in the Loadout Left menus.

Each mainwindow workspace has a different LeftMenu, but a workspace can exist without the leftmenu (e.g. in a popped out panel), so the LeftMenuVM should not live inside the workspace itself.

A factory can be used to generate a LeftMenu given a WorkspaceContext perhaps. 
Ideally left menus VMs would not be created and destroyed each time they are switched, so some component needs to maintain a list of the active ones.

EDIT:
## TODO:
- [x] Implement LeftMenuFactory for Loaodout
    - [x] Implement OpenLoadout page 
    - [x] Add My Mods button (doesn't show up currently)
- [x] Implement leftMenuFactory for Home
    - [x] Implement open Home Page
- [x] Implement leftMenuFactory for Downloads
    - [x] Implement open InProgress Page
- [x] Implement WorkspaceAttachmentsFactoryController
- [x] Create LeftMenus for each Workspace
- [x] Switch LeftMenu depending on active workspace
- [x] Implement primary and secondary default open page behaviors
- [x] Switch the active tab to the first one when non replacing existing tab.
- [x] Add default navigation to Home when app opens
- [ ] Add Apply button to LoadoutLeft manu?
- [ ] Implement way to get the names for each workspace (probably a property on the Workspace itself, can be updated externally, like the TabName
- [ ] Add name at the top of the Mainwindow topbar

Al12rs:
closes #932 

- [x] Added `WorkspaceId` to `ILeftMenuViewModel` and pass `WorkspaceContructor` etc to `LeftMenuViewModel` implementations. 
    - [x] This is to allow Left Menus to have an associated workspace and have the ability to Open pages on it.
- [x] Added `ILeftMenuFactory`s for Loadout, Home and Downloads Left Menus.
- [x] Added `IWorkspaceAttachmentsFactoryManager` that uses `ILeftMenuFactories` to return attached `ILeftMenuViewModels`.
- [x] Added Pages entries for `My Games` `My Mods` and `In Progress` pages on LeftMenus
    - [x] Implemented opening the pages in the Workspace.
- [x] Added `OpenPageBehavior.PrimaryDefault` and `OpenPageBehavior.SecondaryDefault` behaviours, for right click and middle click.
- [x] Added methods to select a specific tab from inside `WorkspaceViewModel`, necessary to make sure to bring opened page into view if it replacing an existing tab.
- [x] Added check to avoid creating a new page when replacing a tab, if the tab has exactly the same data.
- [x] Automatically open the Home Workspace when opening the app (so we don't have empty workspace for now)

Skipped the workspace name stuff for this PR. 
I'm actually thinking of having a property on the Workspace itself, not unlike the TabName.


src/NexusMods.App.UI/LeftMenu/ILeftMenuFactory.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// <summary>
+    /// Generic version of <see cref="ILeftMenuFactory.CreateLeftMenu"/>, that makes implementing factories easier.
+    /// </summary>
+    /// <param name="context">The WorkspaceContext used to determine the type of LeftMenu to return</param>
+    /// <param name="workspaceId">Id of the workspace that the LeftMenu will open new tabs on</param>
+    /// <param name="workspaceController">WorkspaceController to pass on to the LeftMenu</param>
+    /// <returns></returns>
+    protected ILeftMenuViewModel CreateLeftMenuViewModel(TContext context, WorkspaceId workspaceId,
+        IWorkspaceController workspaceController);
```
erri120, line 45: Accessibility modifiers like `public`, and `protected` are useless on interface members since they are always public.

src/NexusMods.App.UI/WorkspaceSystem/Workspace/WorkspaceViewModel.cs
```diff
@@ -314,24 +316,36 @@ private PageData GetDefaultPageData()
         return pageData;
     }
 
-    internal void OpenPage(Optional<PageData> optionalPageData, OpenPageBehavior behavior)
+    internal void OpenPage(Optional<PageData> optionalPageData, OpenPageBehavior behavior, bool selectTab = true)
```
erri120, line 319: `selectTab` should have a default value here.

src/NexusMods.App.UI/WorkspaceSystem/Controllers/WorkspaceController.cs
```diff
@@ -214,12 +226,13 @@ public void AddPanel(WorkspaceId workspaceId, WorkspaceGridState newWorkspaceSta
         workspaceViewModel.AddPanel(newWorkspaceState, behavior);
     }
 
-    public void OpenPage(WorkspaceId workspaceId, Optional<PageData> pageData, OpenPageBehavior behavior)
+    public void OpenPage(WorkspaceId workspaceId, Optional<PageData> pageData, OpenPageBehavior behavior,
+        bool selectTab)
```
erri120, line 230: This should have the `selectTab = true` from the interface.

erri120 (approved):

PR #949, opened 2024-02-14
Add Support for Setting Execute Bit on Linux based on Extension (fixes #924)

### LINKED ISSUE #924: Set the execute permission bit on Linux
To run programs on Linux, the execute permission bit has to be set on the specific file. To prevent issues from arising, we should set the permission for executables.

As an example, SMAPI ships with an executable and a script on Linux.

[`File.SetUnixFileMode`](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.setunixfilemode) added in .NET 7 can be used for this.

[`FileSystemInfo.UnixFileMode`](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.unixfilemode) to get the current file mode.

Sewer56:
Fixes #924

Basic test included.
Draft until PR in Paths is merged, 
- https://github.com/Nexus-Mods/NexusMods.Paths/pull/38
so I can include new dependency version.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -152,7 +152,6 @@ public class ALoadoutSynchronizer : IStandardizedLoadoutSynchronizer
                     resultingItems.Add(newEntry.GamePath(), prevEntry.Item.Value);
                     switch (newEntry.Item.Value!)
                     {
-
                         case StoredFile fa:
                             // StoredFile files are special cased so we can batch them up and extract them all at once.
                             // Don't add toExtract to the results yet as we'll need to get the modified file times
@@ -189,7 +188,6 @@ public class ALoadoutSynchronizer : IStandardizedLoadoutSynchronizer
                 continue;
 
             var absolutePath = installation.LocationsRegister.GetResolvedPath(path);
-
             switch (item.Item.Value!)
             {
                 case StoredFile fa:
@@ -248,6 +246,22 @@ public class ALoadoutSynchronizer : IStandardizedLoadoutSynchronizer
                 Size = entry.Size,
                 LastModified = path.FileInfo.LastWriteTimeUtc
             };
+
+            // And mark them as executable if necessary.
+            // These below are constants, so on platforms like Windows, the execute bit setting
+            // code won't even get JIT-ted at all!
+            if (!OperatingSystem.IsLinux() && !OperatingSystem.IsMacOS())
+                continue;
+
+            var fullPath = path.GetFullPath();
+            var ext = path.Extension.ToString();
+            if (ext is not ("" or ".sh" or ".bin" or ".run" or ".py" or ".pl" or ".php" or ".rb" or ".out"
+                or ".elf")) continue;
+
+            // Note (Sewer): I don't think we'd ever need anything other than just 'user' execute, but you can never
+            // be sure. Just in case, I'll throw in group and other to match 'chmod +x' behaviour.
+            var currentMode = File.GetUnixFileMode(fullPath);
+            File.SetUnixFileMode(fullPath, currentMode | UnixFileMode.UserExecute | UnixFileMode.GroupExecute | UnixFileMode.OtherExecute);
         }
 
         // Return the new tree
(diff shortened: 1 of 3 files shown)


Al12rs (approved):
Seems reasonable to me, but didn't test being on windows.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -248,6 +246,22 @@ await foreach (var entry in _hashCache.IndexFolderAsync(location))
                 Size = entry.Size,
                 LastModified = path.FileInfo.LastWriteTimeUtc
             };
+
+            // And mark them as executable if necessary.
+            // These below are constants, so on platforms like Windows, the execute bit setting
+            // code won't even get JIT-ted at all!
+            // Don't use `path.FileSystem.OS` because that's not constant, and this is a hot path.
+            if (!OperatingSystem.IsLinux() && !OperatingSystem.IsMacOS())
+                continue;
```
erri120, line 255: The entire reason we have `IOperatingSystem` is so we don't use the `OperatingSystem` from .NET. If you are concerned about performance, just add a `isUnix` boolean before the loop:

```csharp
var isUnix = _os.IsUnix();

foreach (var (path, entry) in toExtract)
{
  // ...
  if (!isUnix) continue;
}
```

The JIT should be smart enough to deal with this for loop and realize that `isUnix` never changes.

erri120 (changes requested):
See comment.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
Sewer56: Kind of.

The difference between the code I wrote, and using `var isUnix = _os.IsUnix();` is that `OperatingSystem` is a true (JIT) constant.

When you make the proposed change, the JIT will (if it thinks it may be profitable), clone the entire loop, and simply make a branch based on `isUnix`.

i.e. the resulting code produced by the JIT would look something more like (after translating back to C#):

```csharp
var isUnix = _os.IsUnix();
if (isUnix)
{
    // Update the resulting items with the new file times
    foreach (var (path, entry) in toExtract)
    {
        resultingItems[entry.To] = new DiskStateEntry
        {
            Hash = entry.Hash,
            Size = entry.Size,
            LastModified = path.FileInfo.LastWriteTimeUtc
        };
        
        var ext = path.Extension.ToString();
        if (ext is not ("" or ".sh" or ".bin" or ".run" or ".py" or ".pl" or ".php" or ".rb" or ".out"
            or ".elf")) continue;

        // Note (Sewer): I don't think we'd ever need anything other than just 'user' execute, but you can never
        // be sure. Just in case, I'll throw in group and other to match 'chmod +x' behaviour.
        var currentMode = path.GetUnixFileMode();
        path.SetUnixFileMode(currentMode | UnixFileMode.UserExecute | UnixFileMode.GroupExecute | UnixFileMode.OtherExecute);
    }
}
else
{
    // Update the resulting items with the new file times
    foreach (var (path, entry) in toExtract)
    {
        resultingItems[entry.To] = new DiskStateEntry
        {
            Hash = entry.Hash,
            Size = entry.Size,
            LastModified = path.FileInfo.LastWriteTimeUtc
        };
    }
}
```

Even though `IsUnix` should be a true constant.
This increases JIT time and produces code that will never be executed, on a hot path. I'm not sure if that's quite desirable.

erri120: >  This increases JIT time and produces code that will never be executed, on a hot path. I'm not sure if that's quite desirable.

A single if/else in a method doesn't matter. This isn't inside a loop.

Sewer56: You're also forgetting about instruction cache and pre-fetching. If this was AOT'd (which we're not planning for anymore), this would also be all be dead code, inflating binary size.

That said, I'm not too obsessed about this. If it bothers me too much, I'll fix this issue at the root (i.e. replace `IOSInformation`).

erri120 (approved):

PR #953, opened 2024-02-15
Simplify crating `IconValue`s for `UnifiedIcon` from XAML 

Al12rs:
Simplified the various existing Setters on `IconValue` to be easier to use from XAML.

They now take the original types of `string`, `IImage`, `Geometry` instead of the custom types used internally by `IconValue`.
This allows them to be set them inline instead of having to wrap them in the custom types.

This maintains the single `UnifiedIcon.Value` Api, but reduces the number of lines needed to define the contents of an icon from XAML.

Before:
```xml
                <unifiedIcon:IconValue x:Key="ProjektankerIconValue">
                    <unifiedIcon:IconValue.ProjektankerIconValueSetter>
                        <unifiedIcon:ProjektankerIcon Value="mdi-cog" />
                    </unifiedIcon:IconValue.ProjektankerIconValueSetter>
                </unifiedIcon:IconValue>
```
After:
```xml
<unifiedIcon:IconValue x:Key="ProjektankerIconValue" MdiValueSetter="mdi-cog"/>
```

### DIFF
--- a/src/NexusMods.App.UI/Controls/UnifiedIcon/IconValue.cs
+++ b/src/NexusMods.App.UI/Controls/UnifiedIcon/IconValue.cs
@@ -23,24 +23,24 @@ public sealed class IconValue
 {
     public Union Value { get; set; }
 
-    public ProjektankerIcon ProjektankerIconValueSetter
+    public string? MdiValueSetter
     {
-        set => Value = value;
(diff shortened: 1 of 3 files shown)


erri120 (changes requested):
The types of the union should be changed back to `init-only` records.

erri120 (approved):

PR #954, opened 2024-02-15
Save and restore last window session

### LINKED ISSUE #907: Save and load workspaces
The state of an entire workspace can be aggregated and applied using an existing method. What's currently missing is a "save" and "load" mechanism:

- Should we save all workspaces in all windows when the application closes, or when anything inside the workspaces changes?
- [ ] Add API to load workspaces given a context. This allows spine items to load saved workspaces, since they can deterministically create a context.

erri120:
Resolves #907.

### DIFF
--- a/src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
+++ b/src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
@@ -116,6 +116,12 @@ public class MainWindowViewModel : AViewModel<IMainWindowViewModel>, IMainWindow
             {
                 vm._windowManager.UnregisterWindow(vm);
             }).DisposeWith(d);
+
+            if (!_windowManager.RestoreWindowState(this))
+            {
+                // NOTE(erri120): select home on startup if we didn't restore the previous state
+                Spine.NavigateToHome();
+            }
         });
     }
 
--- a/src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
+++ b/src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
@@ -112,8 +112,18 @@ public class MainWindowViewModel : AViewModel<IMainWindowViewModel>, IMainWindow
                 .BindTo(_windowManager, manager => manager.ActiveWindowId)
                 .DisposeWith(d);
 
-            Disposable.Create(this, vm => vm._windowManager.UnregisterWindow(vm)).DisposeWith(d);
+            Disposable.Create(this, vm =>
+            {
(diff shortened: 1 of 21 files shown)


src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
```diff
@@ -112,10 +112,26 @@ public class MainWindowViewModel : AViewModel<IMainWindowViewModel>, IMainWindow
                 .BindTo(_windowManager, manager => manager.ActiveWindowId)
                 .DisposeWith(d);
 
-            Disposable.Create(this, vm => vm._windowManager.UnregisterWindow(vm)).DisposeWith(d);
+            Disposable.Create(this, vm =>
+            {
+                vm._windowManager.UnregisterWindow(vm);
+            }).DisposeWith(d);
+
+            if (!_windowManager.RestoreWindowState(this))
```
Al12rs, line 120: Should this be executed off the UI thread? since it hits the db?

src/NexusMods.App.UI/Windows/WindowManager.cs
```diff
@@ -73,4 +80,49 @@ public void UnregisterWindow(IWorkspaceWindow window)
...
+        catch (Exception e)
+        {
+            _logger.LogError(e, "Exception while saving window state");
+        }
+    }
+
+    public bool RestoreWindowState(IWorkspaceWindow window)
+    {
+        var isFirstWindow = _windows.Count == 1;
+        if (!isFirstWindow) return false;
```
Al12rs, line 100: Does this mean that if you open a new MainWindow while you have an existing one open it will start from scratch but still save when closed?

Would there be a problem with always restoring here?

erri120: When a main window closes, we save the data to the data store. Since the data is written to the same location every time, the windows will overwrite whatever existing data is there. This means that only the data of the last closed main window will be restored.

As far as restoring goes, we only want to restore data for the first main window. Any subsequently opened main windows are going to have a clean slate.

Al12rs: That means that if you close one of the clean slates last you will loose all your setup.
Why shouldn't a new window also load the last saved state?

Al12rs (approved):

erri120:
TODO: exclude "new tab"-tab from serialization.

Al12rs (changes requested):
Save NewTabs as well, currently their context isn't serializable

Al12rs (approved):

PR #955, opened 2024-02-15
Fix build on main

Al12rs:
Main was broken after #949
`ALoadoutSynchronizer()` was expected to have a single parameter by many subclasses or users, which broke when the Os param was added


Al12rs:
Tests are broken because they don't have access to `IOSInformation`. Does anyone know what `.Add` function adds that in `Startup.cs`?

Sewer56 (approved):

PR #970, opened 2024-02-19
Update formatting and diagnostics

### LINKED ISSUE #944: Disable "convert to primary constructor" Rider hint
Rider will suggest the following code to be converted into a [primary constructor](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/instance-constructors#primary-constructors):

```csharp
public class MyService
{
  private readonly IServiceProvider _serviceProvider;

  public MyService(IServiceProvider serviceProvider)
  {
    _serviceProvider = serviceProvider;
  }

  public string? GetThing() => _serviceProvider.GetService(typeof(string)) as string;
}

// using primary constructor
public class MyService2(IServiceProvider serviceProvider)
{
    public string? GetThing() => serviceProvider.GetService(typeof(string)) as string;
}
```

While primary constructors can be useful in some limited scenarios, the above code example is not one of those scenarios. The two code classes above are not equal, the compiler generates a private field that is not read-only:

https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQAUEEsC2UECeAwgPYB2yMCcAxjCQsgD4ACATAIwCwAULywGYABOyEBZAgGUICAG5YaEXgG9eQoQAdss2BCFIoAE3IAbAkICS0uQoiYS8wzKEB9ZDPmL7jmQG5eaiLCEtaeEAAUVh623lhOCELuNl4IDnEyAJSBqjzq6m7RKWnxQgC8iYV2qT4I/rlCAL4B9YIiHAAMAPxCAOIQMAAqABZYZADm4RllAHyuSWGx8QB0fTChtuEwBBoQJGDhLB0ZU1DIbe11TXw8raIhlWyR60U1FclVxZkqgeq3Hd2rYajCZTUqzeYxarpBArfrPCJbHZ7A5HE5nQ4XXgNIA=

### LINKED ISSUE #945: Update formatter settings
Rider likes to format long lines by cutting off the first possible place it can:

```csharp
// before
var isResizerXAligned = lastLogicalValue.IsCloseTo(currentSize.X, tolerance: defaultTolerance);

// after
var isResizerXAligned =
    lastLogicalValue.IsCloseTo(currentSize.X, tolerance: defaultTolerance);
```

This algorithmic approach to formatting is done because we likely have some hard limit on the amount of columns allowed. Formatting is supposed to make a codebase easier to read by applying a unified coding style to all files. However, the example above is not easier to read, the output produced by Rider is quite frankly _ugly_ and harder to parse.

Another example is a long line with a method declaration or a method invocation that has a lot of arguments:

```csharp
// before
public void Something(IReadOnlyList<string> someStrings, IReadOnlyList<int> someIntegers, bool isSomething, Vector math);

// after
public void Something(IReadOnlyList<string> someStrings, IReadOnlyList<int> someIntegers,
    bool isSomething, Vector math);
```

The Rider formatter splits the argument list at some random position that might make sense to the formatter but makes it, again, harder to read. In this case, putting each argument on its own line looks cleaner:

```csharp
public void Something(
    IReadOnlyList<string> someStrings,
    IReadOnlyList<int> someIntegers,
    bool isSomething,
    Vector math);
```

erri120:
Closes #945.
Closes #944.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,11 +2,13 @@ root = true
 
 [*]
 charset = utf-8
+end_of_line = lf
 insert_final_newline = true
-trim_trailing_whitespace = true
-max_line_length = 120
+indent_style = space
+indent_size = 4
 
 # Microsoft .NET properties
+csharp_indent_switch_labels = false
 csharp_new_line_before_members_in_object_initializers = false
 csharp_preferred_modifier_order = public, private, protected, internal, file, new, static, abstract, virtual, sealed, readonly, override, extern, unsafe, volatile, async, required:suggestion
 csharp_style_prefer_utf8_string_literals = true:suggestion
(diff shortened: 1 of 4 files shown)


Al12rs:
I tried this out a bit, found this auto formatting:
```
        this.WhenActivated(
            d =>
            {
                this.WhenAnyValue(x => x.ViewModel!.Items)
                    .BindTo(this, x => x.MenuItemsControl.Items)
                    .DisposeWith(d);
            }
        );

erri120:
> Do you think it can be avoided? Seems a bit pointless for labdas like that . This happens for all multiline lambdas in subscriptions for example. You can test this in SpineViewModel.cs

The formatter is somewhat limited when it comes to lambdas in invocations. You get this result because the setting for invocations was set to "chop if long or multiline", which chopped the lambda up like this.

Al12rs:
> > Do you think it can be avoided? Seems a bit pointless for labdas like that . This happens for all multiline lambdas in subscriptions for example. You can test this in SpineViewModel.cs
> 
> The formatter is somewhat limited when it comes to lambdas in invocations. You get this result because the setting for invocations was set to "chop if long or multiline", which chopped the lambda up like this.

Still happens, should that commit have fixed it?
It is not the end of the world tbh

Al12rs (approved):
I'm fine with this

Issue #965, opened 2024-02-16
Opening a window after the last window closes is not supported (5 second bug)


halgari:
# Opening a window after the last window closes is not supported

## Summary
There's a bug that we haven't encountered yet in the app, but yet still exists. When the UI starts, we do some one-shot startup code, that starts Avalonia and sets some flags. Once the last window exits, that code also exits. If we open two windows at once, that's fine, as the Avalonia code will stay alive as long as at least one window stays alive. But if we open a window, close it, and before the app exits (due to an inactivity timer) we try to open a new window, nothing will happen since Avalonia's one-shot routines have stopped and there is no window dispatcher. 

## How to fix
I think we should probably find a way to keep the Avalonia lifecycle alive as long as the single process server thread is alive. However we only want to start Avalonia if we need to open a window. So 

- [ ] Keep Avalonia lazy-loading
- [ ] Don't exit the Avalonia lifecycle until the single process server thread exits
- [ ] Make sure this all still works in the single-thread debug, debugger attached mode

halgari:
Relevant code is in : https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/NexusMods.App/StartupHandler.cs#L37-L54

Al12rs:
This is still an issue even after #972
Sometimes it works, other times it opens a console and fails after 5 seconds, depending on the timing of when the new instance is opened.

halgari:
Fixed in #1345, we no longer interact with the main process this way

PR #972, opened 2024-02-19
Fix multi-window interaction with the single process work

### LINKED ISSUE #968: Regression: Opening a second instance of the App doesn't create a new Window
Regression introduced with #877

When opening a second instance of the App the current mainwindow is brought into focus instead of creating a second main window.

Reproduced this on Windows 11.
Edit: erri confirmed same issue on Linux as well.

Use manually started instances of the App to reproduce this, starting the App with debugger makes it not reachable by normal instances of the app, meaning the bug would not reproduce.

halgari:
Fixes #968 and #965. 

This fixes the above bugs by reworking how we open multiple windows. Namely:

* Whenever a window opens, create a main process director token to keep the single process director alive
* Startup Avalonia in a mode that only shuts down when a cancellation token exits
* Link that cancellation token to the single process director cancellation token
* Only run the Avalonia startup and the first window on the creation thread, run the rest via the UI dispatcher (this fixes the #968 regression)


src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
```diff
@@ -64,11 +64,6 @@ public class MainWindowViewModel : AViewModel<IMainWindowViewModel>, IMainWindow
 
         this.WhenActivated(d =>
         {
-            // When the user closes the window, we should persist all download state such that it shows
-            // accurate values after a reboot.
-            // If we ever plan to remove and re-add main window (unlikely), this might need changing to not dispose but only save.
-            downloadService.DisposeWith(d);
```
halgari, line 70: This shouldn't be needed, the service is disposed by DI on shutdown

src/NexusMods.App/MainThreadData.cs
```diff
@@ -9,7 +9,22 @@ public static class MainThreadData
...
+    /// </summary>
+    public static CancellationToken GlobalShutdownToken => _cancellationTokenSource.Token;
+    
+    /// <summary>
+    /// Shuts down the application.
+    /// </summary>
+    public static void Shutdown()
+    {
+        _cancellationTokenSource.Cancel();
+    }
```
erri120, line 26: `Cancel` will throw all aggregated exceptions, those should be handled somewhere.

`OperationCanceledException` instances can be ignored.

src/NexusMods.App/Startup.cs
```diff
@@ -17,9 +21,27 @@ namespace NexusMods.App;
...
+
+#pragma warning disable CS0028 // Disables warning about not being a valid entry point
+    
+    /// <summary>
+    /// Main entry point for the application, the application loop will only exit when the token is cancelled.
+    /// </summary>
+    /// <param name="provider"></param>
+    /// <param name="args"></param>
+    /// <param name="token"></param>
     public static void Main(IServiceProvider provider, string[] args)
```
erri120, line 36: Empty `param` XML docs should be removed, they just clutter up the code.

src/NexusMods.App/Startup.cs
```diff
@@ -41,14 +66,45 @@ public static void Main(IServiceProvider provider, string[] args)
...
 
+    private static void AppMain(Application app, string[] args)
+    {
+        _logger.LogInformation("Starting UI application");
+        ShowMainWindow();
+
+        // Start the app, linking to the global shutdown token
+        _logger.LogInformation("Starting application loop.");
+        app.Run(MainThreadData.GlobalShutdownToken);
+        _logger.LogInformation("Application loop ended.");
```
erri120, line 77: Log event messages should be fragments, not sentences. Avoid a trailing period/full stop.

src/NexusMods.App/Program.cs
```diff
@@ -54,13 +54,24 @@ public static int Main(string[] args)
...
 
 
         // The UI *must* be started on the main thread, according to the Avalonia docs, although it
         // seems to work fine on some platforms (this behavior is not guaranteed). So when we need to open a new
         // window, the handler will enqueue an action to be run on the main thread.
-        while (!managerTask.IsCompleted)
+        while (!MainThreadData.GlobalShutdownToken.IsCancellationRequested)
         {
             if (MainThreadData.MainThreadActions.TryDequeue(out var action))
             {
```
erri120, line 77: The `action()` call inside the loop should probably be surrounded by a `try/catch`.

erri120 (changes requested):
See comments.

The app now doesn't open anymore, it like the following exists immediately:

```csharp
_logger.LogInformation("Starting application loop.");
app.Run(MainThreadData.GlobalShutdownToken);
_logger.LogInformation("Application loop ended.");
```

```log
00:00:01.793 [INFO] Starting UI application
00:00:02.104 [INFO] Starting application loop.
00:00:02.105 [INFO] Application loop ended.
00:00:02.105 [DEBUG] Startup handler returned 0
```

Edit: `MainThreadData.GlobalShutdownToken` appears to already be canceled.

halgari:
Is this in debug mode? Works fine in debug without a debugger attached on my machine (`dotnet run`). Looks like with a debugger attached it no longer works, I'll look into that

erri120:
> Is this in debug mode? Works fine in debug without a debugger attached on my machine (`dotnet run`). Looks like with a debugger attached it no longer works, I'll look into that

https://github.com/Nexus-Mods/NexusMods.App/blob/9232312b074836fa3aeedb4199b55f79c7072d50/src/NexusMods.App/StartupHandler.cs#L50-L63

`tcs.SetResult(0)` will "finish" `StartUiWindowAsync` which finishes `await startup.Start(args, _isDebug);` in `Program.cs`.

src/NexusMods.App/Program.cs
halgari: handled

halgari:
This gets us up back and running. I don't like how there's two paths for the code now, where some things only happen under Debug mode, but I need to think how to better manage this in a way where the code is a bit cleaner. Perhaps a lot of this logic can be moved into the `SingleProcess` project

erri120 (changes requested):
Multiple windows now work again, however, you're unable to close any windows. If you close a window, an exception is thrown due to https://github.com/AvaloniaUI/Avalonia/issues/14437 and the window stays frozen on screen. The only way to solve this is by manually killing the main process of the app.

halgari:
What process are you going through to get this to happen? I just opened and closed a bunch of windows with this branch and didn't see this exception?

erri120:
> What process are you going through to get this to happen? I just opened and closed a bunch of windows with this branch and didn't see this exception?

The exception should always appear in the log, since the underlying issue https://github.com/AvaloniaUI/Avalonia/issues/14437 hasn't been fixed yet.

However, the last two commits of this PR introduced another issue: the window doesn't close properly, and the main process is still running. I had to manually kill the main process before the windows disappeared. This happens when running with a debugger and without.

halgari:
I can't get this error at all, it all cleanly shuts down for me, I'll switch over to my ubuntu machine and see if I can get it to happen there.

halgari:
Ah I haven't managed a game yet, let me try that

Al12rs:
@halgari The app crash on close issue was fixed in:
- #1068 

That should unblock work on this PR again.
The changes from main need to be merged in.

erri120 (changes requested):
Running the app without an attached debugger via `dotnet run`, the processes remain running after closing the window.

When running the app with an attached debugger and running two instances, you get two processes, not a single process.

Sewer56:
When Running on Linux, this PR makes it so you can't run the App without the debugger attached.

First process spawns 2nd process, then it dies, as far as I can tell.

Issue #871, opened 2024-01-24
Deploying one loadout triggers an ingestion in another loadout


halgari:
# Bug Report

## Summary
Trying to apply loadout A while loadout B is applied results in A throwing an error because it needs an ingest. Instead we would expect some sort of ingest of B and then A to recognize B was up-to-date and not trigger a ingest. 

Not exactly sure how this should work, I'll take a look.

Al12rs:
Notes from meeting:
- Ingest needs to know the LoadoutId that was last applied, we can pass in a GameInstallation to help track that down.
- We should save the Last Applied Loadout ID as metadata.
- Are there any cases where we should ingest into a loadout that is different from the one that was last applied?
    - Maybe if the user clones their loadout?
    - User forgets they didn't apply new loadout before making changes, but they want to apply to the new loadout. Should still work.
    - So we might still need target LoadoutId to handle these future use cases. This will require attention to which loadout gets passed to which function, as it would mean potentially dealing with two different ones.

PR #973, opened 2024-02-20
Add initial Apply UI implementation

Al12rs:
This PR adds an initial quite rough Apply button implementation.

- [x] Added `ApplyService` to be used later to queue Apply operations and ensure only one loadout is being applied to a game at a time.
- [x] Added `ApplyControlView` that contains the Apply and Launch buttons.
- [x] Toggled visibility of buttons based on whether there are things to apply or things are being applied.

So for now the Apply button shows up after the user makes at least one new Loadout revision since starting the app.
Pressing the button Applies the loadout while showing a spinner and hiding the Launch button.
After applying is done the button is hidden waiting for new loadout changes.

Limitations of this PR:

- [ ] We currently don't save the last applied Loadout or the last applied Loadout revision, which is necessary to determine whether  there are things that need to be applied or not (see #871 for details).
- [ ] Currently the newest loadout revision on startup is used as the last applied revision arbitrarily.
- [ ] Likewise the number of changes (loadout revisions) to be applied is currently not available.
- [ ] In this implementation Apply always ingests outside changes on top of the latest version of the loadout.
- [ ] Launch button currently still applies (main reason is that we want to do an ingest after the program returns).
- [ ] Ui design for the button only vaguely matches the ones on figma for now.

All of these limitations should be handled after event sourcing has been integrated, together with #871


src/NexusMods.App.UI/Services.cs
```diff
@@ -99,6 +99,8 @@ public static IServiceCollection AddUI(this IServiceCollection c, ILauncherSetti
             .AddViewModel<ImageButtonViewModel, IImageButtonViewModel>()
             .AddViewModel<InProgressViewModel, IInProgressViewModel>()
             .AddViewModel<LaunchButtonViewModel, ILaunchButtonViewModel>()
+            .AddViewModel<ApplyControlViewModel, IApplyControlViewModel>()
+            .AddViewModel<ApplyControlDesignViewModel, IApplyControlViewModel>()
```
erri120, line 103: Don't add the design view model to the services.

src/NexusMods.App.UI/LeftMenu/Items/ApplyControlView.axaml.cs
```diff
@@ -0,0 +1,40 @@
...
+            this.WhenAnyValue(view => view.ViewModel!.IsApplying)
+                .OnUI()
+                .BindToView(this, view => view.InProgressBorder.IsVisible)
+                .DisposeWith(disposables);
+
+            this.WhenAnyValue(view => view.ViewModel!.IsApplying)
+                .Select(isApplying => !isApplying)
+                .OnUI()
+                .BindToView(this, view => view.LaunchButtonView.IsVisible)
+                .DisposeWith(disposables);
```
erri120, line 36: Move these `OnUI` calls to the VM and use simple `OneWayBind` calls instead.

Al12rs: I'm setting the values of the reactive properties from a task, should I use OnUi inside that task?

src/NexusMods.App.UI/LeftMenu/Items/ApplyControlViewModel.cs
```diff
@@ -0,0 +1,88 @@
...
+        LaunchButtonViewModel.LoadoutId = loadoutId;
+
+
+        _gameInstallation = _loadoutRegistry.Get(_loadoutId)?.Installation ??
+                            throw new InvalidOperationException("Loadout not found: " + _loadoutId);
+
+        (_lastAppliedLoadoutId, _lastAppliedRevisionId) = GetLastAppliedLoadout();
+
+        _newestLoadout = _loadoutRegistry.RevisionsAsLoadouts(loadoutId)
+            .ToProperty(this, vm => vm.NewestLoadout);
```
erri120, line 49: Set the scheduler to be `scheduler: RxApp.MainThreadScheduler`.

src/NexusMods.App.UI/LeftMenu/Items/ApplyControlViewModel.cs
```diff
@@ -0,0 +1,88 @@
...
+        this.WhenActivated(disposables =>
+        {
+            this.WhenAnyValue(vm => vm.NewestLoadout,
+                    vm => vm._lastAppliedLoadoutId,
+                    vm => vm.IsApplying)
+                .Subscribe(_ =>
+                {
+                    CanApply = !IsApplying && (
+                        !Equals(_lastAppliedLoadoutId, _loadoutId) ||
+                        !Equals(NewestLoadout.DataStoreId, _lastAppliedRevisionId));
```
erri120, line 63: Use either `a == b`, `a != b` or `a.Equals(b)` instead of `Equals(a, b)`.

src/NexusMods.App.UI/LeftMenu/Items/ApplyControlViewModel.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        _loadoutId = loadoutId;
+        _loadoutRegistry = serviceProvider.GetRequiredService<ILoadoutRegistry>();
+        _applyService = serviceProvider.GetRequiredService<IApplyService>();
+        LaunchButtonViewModel = serviceProvider.GetRequiredService<ILaunchButtonViewModel>();
+        LaunchButtonViewModel.LoadoutId = loadoutId;
+
+
+        _gameInstallation = _loadoutRegistry.Get(_loadoutId)?.Installation ??
+                            throw new InvalidOperationException("Loadout not found: " + _loadoutId);
```
erri120, line 44: This should probably be an `ArgumentException`.

erri120 (changes requested):
See comments.

src/NexusMods.App.UI/LeftMenu/Items/ApplyControlView.axaml.cs
erri120: `IsApplying` is the same as `ApplyCommand.IsExecuting`, is it not?

Al12rs: It is in this implementation, but in the future the ApplyService will determine that, since there might be another loadout applying to the same GameInstallation

erri120: Then just have the `IsApplying` property be set in the VM by subscribing to the `ApplyCommand.IsExecuting` property for now.

erri120 (approved):

PR #974, opened 2024-02-20
Ingest required error

### LINKED ISSUE #871: Deploying one loadout triggers an ingestion in another loadout

halgari:
Resolves #871 

It seems like the problem was mostly due to us storing the previous disk state keyed by LoadoutId instead of on the game version. One we switch to using the game version it all seems to work correctly. I'll do more testing on the CLI and we can create a new PR if this doesn't go far enough for some reason

### DIFF
--- a/NexusMods.App.sln.DotSettings
+++ b/NexusMods.App.sln.DotSettings
@@ -46,6 +46,10 @@
 	<s:String x:Key="/Default/CodeStyle/Naming/XamlNaming/Abbreviations/=XL/@EntryIndexedValue">XL</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/XamlNaming/Abbreviations/=XS/@EntryIndexedValue">XS</s:String>
 	<s:String x:Key="/Default/CustomTools/CustomToolsData/@EntryValue"></s:String>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpKeepExistingMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpPlaceEmbeddedOnSameLineMigration/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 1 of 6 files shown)


halgari:
Confirmed by the CLI: 

* Managed Skyrim into a list called "BaseList1"
* Managed Skyrim into a list called "MainList1"
* Applied both lists (works fine)
* Installed SKSE into "MainList1"
* Applied both lists and verified files were added/removed with each apply

Al12rs:
I want to test this locally, specifically adding outside changes to the test case above.

Al12rs:
Tested the PR merge with main using the new Apply button (has the same logic for apply as LaunchButton had).
I tested a number of scenarios with two loadouts with and without outside changes.

### Testing results:
- [ ] Removing mods and Applying leaves empty folder structures in game folder
- [ ] Creating a second loadout ingests the previously deployed loadout as the game files for that loadout.
- [ ] Changes are ingested into the loadout currently being applied instead of being ingested into the last applied loadout
- [ ] Outside changes are ingested and a new loadout is create but it is never used
    - [ ] When applying, outside changes are ingested, but then the original loadout is Applied
        - Outside changes are removed from file system and are not integrated in the loadout shown in the LoadoutGrid
    - Options:
        - [ ] Update user loadout with outside changes and apply the new loadout
        - [ ] Ingest outside changes into a branch of last applied loadout revision and keep that for users to decide what to do with.
- [ ] The loadout revision ID saved with a `DiskState` is not guaranteed to have a flattened loadout that matches the DiskState (unapplied changes).
- [ ] No way to get the Last Applied LoadoutId and the last applied loadout revision without also getting the entire last DiskState.
    - This is needed for Apply button UI logic (show which loadout is being applied and change diff) and knowing in which loadout to do the ingest.

Sewer56:
I also reviewed the code before falling asleep, just didn't comment on case tired me would say something silly.

In any case, there is one potential caveat I see here. I've heard mentions in the past of 'supporting multiple installations of the same game' at once within the team. 

Because we serialize with key of `Name|Version|Store`, there's.no way to uniquely identify a second copy of the game with thebsame version.

Personally I think the point is moot since a user can only currently have 1 copy of the game per store anyway in the App (since we pull from e.g. Steam info and Steam can only have 1 copy). 

I don't think it's worth investing effort in this area, but I figured it was worth pointing out since I've heard of the desire of multi copies per game a few times; and that would require changing ID format.

PR #991, opened 2024-02-22
Update icons to use `UnifiedIcon`

### LINKED ISSUE #979: Update styles to use the new unified icon
Follow-up to #934.

erri120:
Resolves #979.

### DIFF
--- a/src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/StandardButtonStyles.axaml
+++ b/src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/StandardButtonStyles.axaml
@@ -1,6 +1,6 @@
 ﻿<Styles xmlns="https://github.com/avaloniaui"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        xmlns:avalonia="https://github.com/projektanker/icons.avalonia">
+        xmlns:unifiedIcon="clr-namespace:NexusMods.App.UI.Controls.UnifiedIcon;assembly=NexusMods.App.UI">
 
     <!-- Preview -->
     <Design.PreviewWith>
@@ -36,13 +36,13 @@
                     <TextBlock>Standard Primary + Icon</TextBlock>
                     <Button Classes="Standard Primary">
                         <StackPanel>
-                            <avalonia:Icon Classes="Play" />
+                            <unifiedIcon:UnifiedIcon Classes="Play" />
                             <TextBlock Text="BUTTON" />
                         </StackPanel>
                     </Button>
                     <Button Classes="Standard Primary" IsEnabled="False">
                         <StackPanel>
-                            <avalonia:Icon Classes="Play" />
+                            <unifiedIcon:UnifiedIcon Classes="Play" />
                             <TextBlock Text="BUTTON" />
                         </StackPanel>
                     </Button>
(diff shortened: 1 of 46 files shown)


src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Icons/IconsStyles.axaml
```diff
@@ -81,231 +73,429 @@
...
 
     <!-- NOTE: Material icon names are not standardized well, use this site to look them up:  -->
     <!-- https://pictogrammers.com/library/mdi/ -->
-    <Style Selector="icons|Icon.Alert">
-        <Setter Property="Value" Value="mdi-alert" />
+    <Style Selector="unifiedIcon|UnifiedIcon.Alert">
+        <Setter Property="Value">
+            <Setter.Value>
+                <unifiedIcon:IconValue MdiValueSetter="mdi-alert"/>
+            </Setter.Value>
```
Al12rs, line 104: I shouldn't need the `<Setter.Value>`
so just this should work:
```xml
    <Style Selector="unifiedIcon|UnifiedIcon.AlertOutline">
        <Setter Property="Value">
                <unifiedIcon:IconValue MdiValueSetter="mdi-alert-outline"/>
        </Setter>
    </Style>
```

Al12rs (changes requested):

Al12rs (changes requested):
I see that there are still a lot of `icons:Icon` around, specifically inside buttons.
I think we need those ported over and the button styles updated to use `UnifiedIcon` instead of `icons:Icon`

Apart from that I think there might be broken due to the FontSize vs Size thing.
Game Widget had an Icon on the Add Game button that I think was broken even before this PR and needs updating.

erri120:
> I see that there are still a lot of `icons:Icon` around, specifically inside buttons. I think we need those ported over and the button styles updated to use `UnifiedIcon` instead of `icons:Icon`

I see them now. I was updating the files by searching for the Projektanker namespace, but apparently some files are using ` xmlns:icons="https://github.com/projektanker/icons.avalonia"` instead. I'll fix those.

> Apart from that I think there might be broken due to the FontSize vs Size thing. Game Widget had an Icon on the Add Game button that I think was broken even before this PR and needs updating.

A lot of icons didn't have sizes set before, and I didn't want to touch the styling. I changed `FontSize` to `Size` and I can fix some sizes.

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/StandardButtonStyles.axaml
```diff
@@ -205,21 +211,17 @@
...
             <!-- If the distance from the text is too low, try setting this to 8 as design -->
             <Setter Property="Margin" Value="0,0,3,0" />
         </Style>
 
-        <Style Selector="^ > StackPanel > Image">
+        <Style Selector="^ > StackPanel > unifiedIcon|UnifiedIcon">
             <Setter Property="Margin" Value="0,0,8,0" />
-            <Setter Property="Width" Value="24" />
-            <Setter Property="Height" Value="24" />
+            <Setter Property="Size" Value="24" />
```
Al12rs, line 224: We have two styles here for UnifiedIcon.
Basically I noticed that there was a difference in the margin when an image was used rather than an icon and tried to account for that. 

The supposed margin should have been 8 (should actually have been `StackPanel.Spacing`) but the final result was a bigger gap than what was on figma.

For now set the spacing to 8 and remove the double style

Al12rs (changes requested):

Issue #977, opened 2024-02-21
Diagnostic improvements


erri120:
#327 and #362 added basic diagnostics, and #975 added a basic source generator to improve the creation of diagnostics. This issue contains various improvements for our diagnostics system.

## Rendering and Embeddings

A difficult part of diagnostics is the rendering. Using #933 to render markdown, we can create custom a superset of Markdown with custom "embeddings":

```markdown
Mod {{% ModReference ModA %}} is not compatible with {{% ModReference ModB %}}!
```

This will be stored as a string and the UI can use #933 to render this using custom Avalonia controls. Instead of having `{{% ModReference ModA %}}` in the text, this would be replaced with a custom Avalonia control.

A `ModReference` embedding could be a custom image button containing the icon for mods on the left and the name of the mod on the right. Clicking the button will navigate the user to the mod.

Using source generators, we can use these text embeddings to generate a data type and to verify the string.

## Description and Details

Diagnostics currently only have a `Message` property. To provide the user with more information, like how to fix the issue, it'd be better to have a short `Description` and a potentially long `Details` property. The `Description` can be shown in the UI at various places, while the `Details` property would only be shown when the user clicks on the diagnostic to get more information.

The `Details` property can potentially be very elaborate, so it should be rendered in some scrollable text block.

## Diagnostic definitions outside of code

With the source generator added in #975, we can generate diagnostics using definitions outside of code. As an example, we could have a `001_MyDiagnostic.md` file with the following contents:

```markdown
---
Source: "NexusMods.Games.Example"
ID: 1
Severity: "Critical"
Description: "Mod {{% ModReference ModA %}} is not compatible with {{% ModReference ModB %}}!"
---

Mod {{% ModReference ModA %}} is not compatible with {{% ModReference ModB %}} because {{% ModReference ModB %}} contains file {{% FileReference FileA %}} that disables {{% ModReference ModA %}}.

This can be fixed by {{% EditAction FileA "editing FileA" %}} and removing the following line:

`disabled: "ModA"`
```

This builds heavily on the previous two improvements of having embeddings and a `Details` property.

### Publish to docs

If the diagnostic definitions are inside some markdown file, we can easily generate docs for them and put them on a website. This would allow anyone to quickly link to a specific diagnostic and help other users. Ideally, users would be able to right-click a diagnostic and get a permalink to the online documentation.

### Translations

Translations could be provided by having multiple files following this schema:

- Original English version: `001_MyDiagnostic.md`
[...]

Issue #915, opened 2024-02-07
Stardew Valley Diagnostics


Al12rs:
- Show missing dependencies, missing mod name and missing mod Version
- Show SMAPI ver incompatibility
- Show Game version incompatibility
- Show this in diagnostics

Requires:
- [ ] #327 
- [ ] #1002 
- [ ] #977

PR #1004, opened 2024-02-27
More diagnostic improvements

erri120:
Part of #977 and required for #915.

Changes:

- Replaced `Diagnostic.Message` with `Diagnostic.Summary` and `Diagnostic.Details`.
- Added `AddValue<T>` to the message builder. This allows you to include any kind of data in the diagnostic message data struct.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Games.Diagnostics/DiagnosticTemplateBuilder.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Games.Diagnostics/DiagnosticTemplateBuilder.cs
@@ -48,18 +48,45 @@ public static class DiagnosticTemplateBuilder
         /// <summary>
         /// Sets the <see cref="DiagnosticSeverity"/>.
         /// </summary>
-        IWithMessageStep WithSeverity(DiagnosticSeverity severity);
+        IWithSummaryStep WithSeverity(DiagnosticSeverity severity);
     }
 
     /// <summary>
-    /// Message Step.
+    /// Summary step.
     /// </summary>
-    public interface IWithMessageStep
+    public interface IWithSummaryStep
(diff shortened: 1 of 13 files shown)


src/Abstractions/NexusMods.Abstractions.Games.Diagnostics/DiagnosticTemplateBuilder.cs
```diff
@@ -71,6 +96,11 @@ public interface IMessageBuilder
         /// Adds data references to the message.
         /// </summary>
         IMessageBuilder AddDataReference<T>(string name) where T : IDataReference;
+
+        /// <summary>
+        /// Adds a simple value to the message.
+        /// </summary>
+        IMessageBuilder AddValue<T>(string name) where T : notnull;
```
Al12rs, line 103: I'm not really clear on what the use cases for `AddValues` are supposed to be. 
Is this to later support custom handlers? I would expect T to implement some kind of interface TBD in that case, or is this for something else?

erri120: This is to add any kind of data you want to the diagnostic. This can be a `string`, `int`, or whatever makes sense in the context of the diagnostic.

Al12rs: Is the only purpose of that data is for displaying in the description?

erri120: For now, yes.

Al12rs (approved):

PR #1019, opened 2024-03-04
Add project code style file

erri120:


.idea/.idea.NexusMods.App/.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,7 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <option name="LINE_SEPARATOR" value="&#10;" />
+    <option name="RIGHT_MARGIN" value="1000" />
```
Al12rs, line 4: I object to 1000 here, if left to developer discretion lines are unlikely to be kept under 120 characters (which I would like for GitHub readability).
I do not wish to tell people to shorten their lines in PR review, this should be handled by the formatter.

Al12rs (approved):
As per Halgari ruling

Sewer56 (approved):

PR #1026, opened 2024-03-05
Add diagnostic for missing SMAPI

erri120:


src/Games/NexusMods.Games.StardewValley/Emitters/MissingSMAPIEmitter.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace NexusMods.Games.StardewValley.Emitters;
+
+public class MissingSMAPIEmitter : ILoadoutDiagnosticEmitter
+{
+    private static readonly Uri NexusModsSMAPIUri = new("https://nexusmods.com/stardewvalley/mods/2400");
+
+    public async IAsyncEnumerable<Diagnostic> Diagnose(Loadout loadout)
+    {
+        await Task.Yield();
```
Al12rs, line 14: I looked up `Task.Yield()` a bit, for what purpose are you using it here?

src/Games/NexusMods.Games.StardewValley/Emitters/MissingSMAPIEmitter.cs
```diff
@@ -0,0 +1,27 @@
...
+        var smapiModCount = loadout.Mods.Count(kv => kv.Value.Metadata.OfType<SMAPIModMarker>().Any());
+        if (smapiModCount == 0) yield break;
+
+        var hasSMAPI = loadout.Mods.Any(kv => kv.Value.Metadata.OfType<SMAPIMarker>().Any());
+        if (hasSMAPI) yield break;
+
+        yield return Diagnostics.CreateMissingSMAPI(
+            ModCount: smapiModCount,
+            NexusModsSMAPIUri: NexusModsSMAPIUri
+        );
```
Al12rs, line 25: The marker on the mod would only work if SMAPI was installed through the App.
If it was already installed in the game folder at the moment of creation of the loadout then this check would fail to detect it.

Would it be possible to check for the presence of one of the SMAPI files instead? 

Ideally this check would also make sure that the mod containing SMAPI is enabled, or that could be a separate Diagnostic

src/Games/NexusMods.Games.StardewValley/Emitters/MissingSMAPIEmitter.cs
erri120: The return value of the method is `IAsyncEnumerable<T>` which is a bit clunky to work with when the entire method is synchronous.

src/Games/NexusMods.Games.StardewValley/Emitters/MissingSMAPIEmitter.cs
erri120: >  If it was already installed in the game folder at the moment of creation of the loadout then this check would fail to detect it.

We don't support that.

> Would it be possible to check for the presence of one of the SMAPI files instead?

Yes, but it's not worth supporting an unsupported use-case.

Al12rs: > We don't support that.

I saw you mention this before, I don't remember us deciding on straight up not supporting this. 
I personally don't agree with not supporting cases where the game folder isn't strictly vanilla. 
We should go over this on standup

Al12rs (approved):

PR #1033, opened 2024-03-06
Extract SMAPI version

### LINKED ISSUE #1027: Extract SMAPI version during installation
Mods can require a minimum SMAPI version, and SMAPI itself also requires a minimum game version.

erri120:
Resolves #1027.


Al12rs (approved):
I don't particularly like the fact that we need to extract the file after we just packaged it, but I don't know if we can do any better right now.

erri120:
> I don't particularly like the fact that we need to extract the file after we just packaged it, but I don't know if we can do any better right now.

If we have a custom version parser for managed assemblies, we wouldn't need to extract the file from the archive. However, we currently use `FileVersionInfo` from .NET which requires an actual file on disk.

PR #1034, opened 2024-03-06
Link mod pages

### LINKED ISSUE #1028: Add mod link to missing dependency diagnostic for SMAPI mods
The website `https://smapi.io` can be used to query a link to a mod page on Nexus Mods depending on the unique SMAPI mod ID found in the manifest. The SMAPI toolkit, which we use, already has a web client to access this information.

This requires #1027 and #1029.

erri120:
Resolves #1028.

- Adds `ISMAPIWebApi` for interacting with the SMAPI Web API
- Updates existing diagnostics with links to mod pages

### DIFF
--- a/src/Games/NexusMods.Games.StardewValley/WebAPI/SMAPIWebApi.cs
+++ b/src/Games/NexusMods.Games.StardewValley/WebAPI/SMAPIWebApi.cs
@@ -1,3 +1,4 @@
+using System.Collections.Immutable;
 using Microsoft.Extensions.Logging;
 using NexusMods.Paths;
 using StardewModdingAPI.Toolkit;
@@ -19,7 +20,7 @@ internal sealed class SMAPIWebApi : ISMAPIWebApi
     private bool _isDisposed;
     private WebApiClient? _client;
 
-    private readonly Dictionary<string, Uri> _knownModPageUrls = new(StringComparer.OrdinalIgnoreCase);
+    private ImmutableDictionary<string, Uri> _knownModPageUrls = ImmutableDictionary<string, Uri>.Empty.WithComparers(StringComparer.OrdinalIgnoreCase);
 
     public SMAPIWebApi(ILogger<SMAPIWebApi> logger)
     {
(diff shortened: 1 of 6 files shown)


src/Games/NexusMods.Games.StardewValley/WebAPI/SMAPIWebApi.cs
```diff
@@ -0,0 +1,118 @@
...
+        Version smapiVersion,
+        string[] smapiIDs)
+    {
+        ObjectDisposedException.ThrowIf(_isDisposed, typeof(SMAPIWebApi));
+
+        var semanticGameVersion = new SemanticVersion(gameVersion);
+        var semanticSMAPIVersion = new SemanticVersion(smapiVersion);
+        var platform = ToPlatform(os);
+
+        _client ??= new WebApiClient(
```
Al12rs, line 41: I don't know if it can really be a concern, but is `WebAbiClient` thread safe?

erri120: `WebApiClient` is just a wrapper around `FluentClient` which is a wrapper around `HttpClient` which is thread safe, meaning using it from multiple threads is okay.

src/Games/NexusMods.Games.StardewValley/WebAPI/SMAPIWebApi.cs
```diff
@@ -0,0 +1,118 @@
...
+{
+    private const string ApiBaseUrl = "https://smapi.io/api";
+    private const string NexusModsBaseUrl = "https://nexusmods.com/stardewvalley/mods";
+
+    private readonly ILogger<SMAPIWebApi> _logger;
+
+    private bool _isDisposed;
+    private WebApiClient? _client;
+
+    private readonly Dictionary<string, Uri> _knownModPageUrls = new(StringComparer.OrdinalIgnoreCase);
```
Al12rs, line 22: Could this cause issues if changed while being iterated in another thread?

erri120: Yes and no. Yes, because it's _documented_ to not be threadsafe, no because we probably won't be referencing this from multiple threads at once.

Al12rs: I`d prefer to be safe than sorry. I also think that the chances of a data race aren't high, but then again I'm sure there are lots of developers that thought so as well just to run into hard to debug issues later on.

erri120: Resolved using an atomic `ImmutableDictionary`. This makes the most sense for the very read heavy behavior.

Al12rs (approved):

Al12rs:
Restarting the CI, was giving an error, maybe check locally in the meantime

Al12rs (changes requested):
Build failing:
```
/home/runner/work/NexusMods.App/NexusMods.App/src/Games/NexusMods.Games.StardewValley/WebAPI/SMAPIWebApi.cs(85,53): error CS8619: Nullability of reference types in value of type '(string?, Uri?)' doesn't match target type '(string id, Uri uri)'. 
```
Edit: Ah, you beat me to it

Al12rs (approved):

Issue #906, opened 2024-02-05
Workspace quirks and improvements


erri120:
A list of minor bugs, glitches, quirks, and general areas for improvements.

- [ ] Items in the tab header sometimes glitch out and become invisible. This is especially noticeable when resizing the panels.
- [ ] Items in the tab header stay invisible after resizing.
- [ ] The list of "Add Panel" buttons gets cleared whenever the state changes. The Views can be recycled instead.
- [ ] The `ComboBox` containing the "Add Panel" buttons will load the Views for the buttons when you open the drop-down menu. This is jarring for users and make the App feel laggy. If we recycle the Views, then only the initial opening of the drop-down menu will feel slower than usual. There are probably some ways to optimize this for a smoother experience.

LukeNexusMods:
@captainsandypants

PR #1037, opened 2024-03-06
Reorder grid states

erri120:
Part of #906.

Fixes the order in which the possible grid states are generated. Now, we prioritize and put the primary action first, rather than last, depending on the orientation.

### DIFF
--- a/src/NexusMods.App.UI/WorkspaceSystem/GridUtils.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/GridUtils.cs
@@ -66,8 +66,8 @@ public static class GridUtils
         int maxColumns,
         int maxRows)
     {
-        var res1 = new List<WorkspaceGridState>();
-        var res2 = new List<WorkspaceGridState>();
+        var step1Res = new List<WorkspaceGridState>();
+        var step2Res = new List<WorkspaceGridState>();
 
         var (columnCount, maxRowCount) = currentState.CountColumns();
 
@@ -87,8 +87,8 @@ public static class GridUtils
             {
                 foreach (var panelToSplit in rows)
                 {
-                    res1.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: false));
-                    res1.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: true));
+                    step1Res.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: false));
+                    step1Res.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: true));
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/NexusMods.App.UI/WorkspaceSystem/GridUtils.cs
```diff
@@ -85,8 +87,8 @@ internal static bool IsPerfectGrid(WorkspaceGridState gridState)
             {
                 foreach (var panelToSplit in rows)
                 {
-                    res.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: false));
-                    res.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: true));
+                    res1.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: false));
+                    res1.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: true));
```
Al12rs, line 91: So I tested this and works great when you have a single panel.
If you have two panels side by side, then it offers to split the first panel on the left horizontally first, instead of offering the right most panel first, which I think would be more desirable.

I think it can be fixed by populating a list of lists in this While loop, and then inverting the order of the parent list. So the left panel split results are presented first.

Al12rs: This is my test code for reference:
```csharp
        var newRowsList = new List<List<WorkspaceGridState>>();
        while (columnEnumerator.MoveNext(rowBuffer))
        {
            var column = columnEnumerator.Current;
            var rows = column.Rows;
            
            var newRows = new List<WorkspaceGridState>();
            
            // NOTE(erri120): In a horizontal layout, the rows can move independent of rows in other columns.
            if (rows.Length != maxRows)
            {
                foreach (var panelToSplit in rows)
                {
                    newRows.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: false));
                    newRows.Add(SplitPanelInHalf(currentState, panelToSplit, splitVertically: false, inverse: true));
                }
            }

            // NOTE(erri120): If these conditions are met, we can split all rows of the current column in half
            // to create a new column. Note that this requires knowing the current amount of columns, so two
            // iterations are required (that's what CountColumns does).
            if (rows.Length > 1 && columnCount != maxColumns)
            {
                newRows.Add(AddColumn(currentState, column.Info, rows, inverse: false));
                newRows.Add(AddColumn(currentState, column.Info, rows, inverse: true));
            }
            newRowsList.Add(newRows);
        }

        res1.AddRange(Enumerable.Reverse(newRowsList).SelectMany(x => x));
```

erri120: The same can be achieved using a simpler LINQ chain:

```csharp
step1Res = step1Res.Chunk(size: 2).Reverse().SelectMany(x => x).ToList();
```

erri120:
# Horizontal

erri120:
# Vertical

Al12rs:
That good to me, Great work!

Al12rs (approved):

PR #1051, opened 2024-03-12
Dep updates

Al12rs:
Built, ran tests, opened the App played around a bit to check for obvious issues, didn't notice any problems.


erri120:
The `Microsoft.CodeAnalysis.CSharp.*` packages are somewhat special in that they are directly tied to the Roslyn version used during compilation. I don't think there are any issues with upgrading to latest since we're using .NET 8.

erri120 (approved):

Issue #1048, opened 2024-03-11
Mod Contents file tree view Refinements


Al12rs:
- [x] Move `IsExpanded` from `TreeNodeVM` to the inner `.Item` view model
 - [x] Add file counts status bar [figma design](https://www.figma.com/file/kbQxAayUbZPw27RtGS4MBc/%F0%9F%93%B1Tree-data-grid?type=design&node-id=97-650&mode=design&t=X3KbZcoIkxYGMFwW-4)
 - [x] #1050
     - https://github.com/Nexus-Mods/NexusMods.App/pull/1056
 - [x] Use single Unified Icon in Advanced installer instead of two
     - https://github.com/Nexus-Mods/NexusMods.App/pull/1057
 - [x] Make section headers of Advanced Installer scroll up again as per design
 - [x] Use `GetDefaultOpenPageBehavior` instead of using a custom behavior in `LoadoutGridViewModel`
 - [x] Remove custom path code from `GamePathNode`
   - [x] https://github.com/Nexus-Mods/NexusMods.App/pull/1067
 - [x] Add file type specific icons to ModFilesView
   - [x] https://github.com/Nexus-Mods/NexusMods.App/pull/1072
 - [x] Add file type specific icons to AdvancedInstaller
   - [x] https://github.com/Nexus-Mods/NexusMods.App/pull/1072
- [x] Consider unifying `TreeNodeVM` with the underlying flat item view model. So the view model itself would have a Children collection that gets populated inside `TransformToTree`. I (Al12rs) am still on the fence about this. 
    - https://github.com/Nexus-Mods/NexusMods.App/pull/1055
- [x] #1049
- [ ] #1354 
- [ ] Mod info file tree: Folders Size still sums reified delete file sizes
- [ ] #1014 
- [x] #2004
- [ ] #2005
- [ ] #2006
- [ ] #2007

Sewer56:
Reminder to self: Also add that API in paths

Al12rs:
> AdvancedInstaller is missing OpenFolderIcon on opened nodes. And diff is planned to have the extra icons, but not ViewModFiles currently.

I'm pretty sure that the ModFilesView is also intended to have the extra icons. The extra icons are honestly something we can add in phase 2 or later. They are just nice to have. I'll add items. 

Btw, feel free to edit the top post and add items directly.

PR #1023, opened 2024-03-05
View Mod Contents First Draft & TreeDataGrid Improvements

Sewer56:
fixes #989

Implementation includes:

- Dynamically changing icons on open/close.
- Custom theming for row headers and contents.
- Sorting, by name and size.
    - Including putting folders first, before files, to match common file explorer behaviour.
- Designer previews for everything.
- Backend code to support all of this.
- View correctly restores after app reboot.
- Asynchronous view loading.
- Cached ViewModels.
- A new 'Loading' and 'Error' view/panel.
- LoadoutGrid buttons are now disabled if unselected.
- Slightly improved LoadoutSynchronizer perf (side effect of making a change to `AGamePathTree`).
- App no longer displays trace logging in release builds (improves perf on small DB reads by ~70%).
- Subscribing to Tree Node VM in DynamicData is now lazy, improving perf to show a view.

Additional changes:
- Existing `AdvancedInstaller` dialogs now use a new `TreeDataGrid` helper to avoid repeated code.
- Fixed broken Virtualization in some cases of TreeDataGrid.
- TreeDataGrid columns can now be DI injected, instead of hardcoding them for each grid.
- Removed 1px header hack from 'AdvancedInstaller' by avoiding unnecessary `Grid` use.
- Removed unused `x:Name` declarations.

PR #1053, opened 2024-03-12
Move IsExpanded to Child Items

Sewer56:
Checks off second box in 
https://github.com/Nexus-Mods/NexusMods.App/issues/1048

Successor to:
- #1023 

Am on my phone, so no legendary elaborate descriptions. Besides, I've been meming hard enough today.


src/NexusMods.App.UI/TreeNodeVM.cs
```diff
@@ -57,6 +67,14 @@ public TreeNodeVM(Node<TItem, TKey> node, TreeNodeVM<TItem, TKey>? parent = null
...
         Id = node.Key;
         Parent = parent;
+        
+        this.WhenActivated(d =>
+        {
+            // re-broadcast 'PropertyChanged' event from the inner item.
+            Item.WhenAnyValue(x => x.IsExpanded)
+                .Subscribe(_ => this.RaisePropertyChanged(nameof(IsExpanded)))
+                .DisposeWith(d);
+        });
```
Al12rs, line 77: @erri120 Wanted to check in with you to see if there was a better suited way to handle this.
Basically we need to have `IsExpanded` duplicated both on the flat item and the tree wrapper (because treeDataGrid source can't bind to the internal item...)

Is there a clean way with reactive ui to do a two way bind between two view models? Or is the approach here ok?

Al12rs (approved):
erri said this is fine if not ideal

PR #1075, opened 2024-03-18
Convert diagnostic messages into plain text

erri120:
This is the first step towards rendering diagnostics in the UI. Diagnostic messages can now be written as plain text with all their fields populated.


Sewer56 (approved):
Seems good to me.

I only have one suggestion, the method `GetOutput` implies (by name) that it is a getter. 

Even if specified in docs, most programmers don't expect something that sounds like a getter to mutate the state of the current item. It may be better to rename it to something like `GetOutputAndReset` or `Finish`/`Finalize`. Neither of these suggestions are particularly great, so name it whatever you feel like. It's just a hint.

As for the generated formatting logic, it could be sped up by using `IndexOf` which defaults to SIMD under the hood. That said, I expect the bracket matching to be in the range of 900MB/s, so I don't expect that to ever be a bottleneck. The SIMD version for reference I imagine is above 13GB/s. Either way, no issue there, it's probably already fast enough, but if you ever want to optimize it, that's where you wanna strike.

Sewer56:
I'm down for `ToOutput` honestly.

PR #1078, opened 2024-03-18
Rework diagnostic system

erri120:
- Emitters are now per-game, similar to mod installers.
- Removed `IModDiagnosticEmitter` and `IModFileDiagnosticEmitter` because they haven't been used, and I'm not sure how they'll be used at all. If we ever need them, we can re-add them.
- Reworked `DiagnosticManager` to be much simpler.
- Added cancellation support to the emitters.


src/Abstractions/NexusMods.Abstractions.Games/AGame.cs
```diff
@@ -121,37 +126,29 @@ public void ResetInstallations()
...
             .ToList();
     }
 
     /// <summary>
     /// Returns the locations of known game elements, such as save folder, etc.
     /// </summary>
-    /// <remarks>
-    /// TODO: (Al12rs) Games can return Locations that point to the same AbsolutePath, a way is needed to decide which to use.
-    /// Current code will use the first declared one but relies on undefined ordering of Dictionary.
-    /// </remarks>
```
Al12rs, line 149: Keep the TODO remark, this is still a relevant issue

src/NexusMods.DataModel/Diagnostics/DiagnosticManager.cs
```diff
@@ -1,190 +1,63 @@
...
-            .SelectAsync(async emitter => await emitter.Diagnose(loadout).ToArrayAsync())
-            .ToArrayAsync();
+        // TODO: cancellation token
+        var cancellationToken = CancellationToken.None;
 
-        AddDiagnostics(newDiagnostics.SelectMany(vals => vals));
+        return _loadoutRegistry
+            .RevisionsAsLoadouts(loadoutId)
+            .DistinctUntilChanged(loadout => loadout.DataStoreId)
+            .Throttle(dueTime: TimeSpan.FromMilliseconds(250))
```
Al12rs, line 36: Use `Debounce` rather than `Throttle` https://reactivex.io/documentation/operators/debounce.html

I think it's more suited to our needs, as it always emits the last value after a while, rather than potentially ignoring it as happens with `Throttle`

Al12rs (changes requested):

Al12rs (approved):

PR #1082, opened 2024-03-19
View diagnostics

### LINKED ISSUE #1083: `DiagnosticWriter` output is mangled

### LINKED ISSUE #1084: Fix `MissingSMAPIEmitter`
- `MissingSMAPIEmitter`:
  - Shows both "SMAPI is not installed" and "SMAPI is not enabled" when SMAPI isn't installed.
  - Shows "SMAPI isn't installed" when SMAPI is installed.

erri120:
Allows users to view and interact with diagnostics. Part of #327.

- Fixes #1083
- Fixes #1084

Stuff that needs to be done in other PRs:

- Text wrapping in the list isn't working
- The horizontal line isn't styled with the severity color
- Filter by severity and show chips in a "top bar"
- Left menu icon for diagnostics


src/NexusMods.App.UI/Pages/Diagnostics/List/IDiagnosticListViewModel.cs
```diff
@@ -0,0 +1,13 @@
...
+using NexusMods.App.UI.Controls.Diagnostics;
+using NexusMods.App.UI.WorkspaceSystem;
+
+namespace NexusMods.App.UI.Pages.Diagnostics;
+
+public interface IDiagnosticListViewModel : IPageViewModelInterface
+{
+    public LoadoutId LoadoutId { get; set; }
+
+    public ReadOnlyObservableCollection<IDiagnosticEntryViewModel> DiagnosticEntries { get; }
```
Al12rs, line 12: Do we need this? Shouldn't this work like the LoadoutGrid, where we just have the LoadoutId, and get the list of mods or in this case Diagnostics form the LoadoutRegister or the DiagnosticManager?

src/NexusMods.App.UI/Services.cs
```diff
@@ -175,7 +176,13 @@ public static IServiceCollection AddUI(this IServiceCollection c, ILauncherSetti
             .AddView<FileTreeNodeView, IFileTreeNodeViewModel>()
             .AddView<LoadingView, ILoadingViewModel>()
             .AddView<ErrorView, IErrorViewModel>()
+
             .AddView<DiagnosticEntryView, IDiagnosticEntryViewModel>()
+            .AddViewModel<DiagnosticEntryViewModel, IDiagnosticEntryViewModel>()
```
Al12rs, line 181: Do we need to register `DiagnosticEntryViewModel`? It needs to be constructed from code as it takes a `Diagnostic` in the constructor, is there an other reason to register it?

src/NexusMods.App.UI/Pages/Diagnostics/DiagnosticIcons.cs
```diff
@@ -0,0 +1,19 @@
...
+            "M2 9.87439V6.87439C2 6.32439 2.19583 5.85356 2.5875 5.46189C2.97917 5.07022 3.45 4.87439 4 4.87439H20C20.55 4.87439 21.0208 5.07022 21.4125 5.46189C21.8042 5.85356 22 6.32439 22 6.87439V9.87439H20V6.87439H4V9.87439H2ZM4 20.8744C3.45 20.8744 2.97917 20.6786 2.5875 20.2869C2.19583 19.8952 2 19.4244 2 18.8744V15.8744H4V18.8744H20V15.8744H22V18.8744C22 19.4244 21.8042 19.8952 21.4125 20.2869C21.0208 20.6786 20.55 20.8744 20 20.8744H4ZM10 17.8744C10.1833 17.8744 10.3583 17.8286 10.525 17.7369C10.6917 17.6452 10.8167 17.5077 10.9 17.3244L14 11.1244L15.1 13.3244C15.1833 13.5077 15.3083 13.6452 15.475 13.7369C15.6417 13.8286 15.8167 13.8744 16 13.8744H22V11.8744H16.625L14.9 8.42439C14.8167 8.24106 14.6917 8.11189 14.525 8.03689C14.3583 7.96189 14.1833 7.92439 14 7.92439C13.8167 7.92439 13.6417 7.96189 13.475 8.03689C13.3083 8.11189 13.1833 8.24106 13.1 8.42439L10 14.6244L8.9 12.4244C8.81667 12.2411 8.69167 12.1036 8.525 12.0119C8.35833 11.9202 8.18333 11.8744 8 11.8744H2V13.8744H7.375L9.1 17.3244C9.18333 17.5077 9.30833 17.6452 9.475 17.7369C9.64167 17.8286 9.81667 17.8744 10 17.8744Z"
```
Al12rs, line 16: I'm not happy with this approach, we are duplicating the definition of the icons, which isn't great.
I think this is ok for this PR but we should think about a better approach for this.

I have thought about it over the weekend and I couldn't find many solutions that don't end with an untyped reference in the code to something that is in the Theme project. 
Maybe we can have a chat about this

src/NexusMods.App.UI/Pages/Diagnostics/List/IDiagnosticListViewModel.cs
Al12rs: So I just read this wrong I think

Al12rs (approved):
I just have a question above, the rest is fine considering a lot of stuff is to be done later

src/NexusMods.App.UI/Services.cs
erri120: This is just there to keep the services consistent. It'd look weird if we only register the View and not the View Model.

src/NexusMods.App.UI/Pages/Diagnostics/DiagnosticIcons.cs
erri120: Ideally, we'd have static values in the code project and the themes project would just import those as resources, but I haven't gotten that to work yet.

Al12rs: The Icons should be defined in the Theme project, as those are likely to change with the theme.

Al12rs: You can use theme resources from code without importing the theme project, as long as the resources are added to the App.xaml or whatever. 
It isn't ideal because you have to reference them by key which is just a string. You also need to have access to a View, either from a codebehind or by getting the mainwindow

PR #1085, opened 2024-03-19
Better `IDiagnosticWriter` API

erri120:
Requires #1082.

- makes `IDiagnosticWriter` a singleton
- requires `StringBuilder` to be passed as state


Sewer56 (approved):
Does exactly as says on the tin

PR #1096, opened 2024-03-20
Cancel login after set timeout

erri120:


Al12rs (approved):

Sewer56:
Consider discussing with Design before merging. 

A possible user journey may include:
- Typing their Password into Password Manager to Get Nexus Password
- Getting authentication code from authenticator. 
  - Either on phone, so phone unlock + check.
  - Or on Desktop.

That can take a while, so the situation can be awkward if the user goes through the whole login process, and finds out things failed because the process timed out.

PR #1100, opened 2024-03-20
Run git builds on schedule

### LINKED ISSUE #1079: 'Git Builds' Workflow not always Producing a New Build when Merging to Main
# Bug Report

## Summary

https://github.com/Nexus-Mods/NexusMods.App/actions/workflows/git-builds.yaml

20240318_16h49m33s

The 'Git Builds' Action is not always being triggered after PRs are merged to main.

## Steps to reproduce

<Unknown, Not Investigated>

A build should be made after every change makes it to the 'Main' branch.

## Other

[git-builds.yaml](https://github.com/Nexus-Mods/NexusMods.App/blob/main/.github/workflows/git-builds.yaml) has the following
```
on:
  push:
    branches: [ main ]
    paths:
      - ".github/workflows/clean_environment_tests.yaml"
      - "./src/**"
      - "./tests/**"
      - "**.props"
      - "**.targets"
```

[clean-environment-tests.yaml](https://github.com/Nexus-Mods/NexusMods.App/blob/main/.github/workflows/clean_environment_tests.yaml) has the following
```
on:
  workflow_dispatch:
  push:
    branches: [ main ]
    paths:
      - ".github/workflows/clean_environment_tests.yaml"
      - "src/**"
      - "tests/**"
      - "**.props"
      - "**.targets"
  pull_request:
    branches: [ main ]
    paths:
      - ".github/workflows/clean_environment_tests.yaml"
      - "src/**"
      - "tests/**"
      - "**.props"
      - "**.targets"
    types: [ opened, synchronize, reopened, ready_for_review ]
```

The latter appears to work fine, the former not.
It may be due to `./` prefix in the former, but I haven't yet confirmed.

erri120:
Fixes #1079. Using [`17 08,15 * * *`](https://crontab.guru/#17_08,15_*_*_*) as a schedule:

> At minute 17 past hour 8 and 15

### DIFF
--- a/.github/workflows/git-builds.yaml
+++ b/.github/workflows/git-builds.yaml
@@ -2,7 +2,9 @@ name: Git Builds
 
 on:
   schedule:
-    - cron: '17 08,15 * * *'
+    # At minute 17 past hour 8 and 14 (UTC) on every day-of-week from Monday through Thursday
+    # https://crontab.guru
+    - cron: '17 08,14 * * 1-4'
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/git-builds.yaml
+++ b/.github/workflows/git-builds.yaml
@@ -1,14 +1,9 @@
 name: Git Builds
 
 on:
-  push:
-    branches: [ main ]
-    paths:
-      - ".github/workflows/clean_environment_tests.yaml"
-      - "./src/**"
-      - "./tests/**"
-      - "**.props"
-      - "**.targets"
+  schedule:
+    - cron: '17 08,15 * * *'
+  workflow_dispatch:
 
 jobs:
   build:


Sewer56 (approved):
Assuming our staff stick closer to the standard 9-5 most of the time, would running a 2nd CRON job just after 17:00 not be more suitable?

Just curious.

Al12rs:
Times seem pretty arbitrary, some comments for why would be good

erri120:
We could technically further limit the schedule to only run on work days.

Al12rs:
> We could technically further limit the schedule to only run on work days.

Yeah, we should do that.
Running before active hours makes sense as it is guaranteed to get the latest stuff from the previous day. 
I wouldn't mind moving the other build to before standup, so `14:17` UTC or a bit sooner, as we often have meetings after that and having a build to test the changes of the morning could be useful.

erri120:
> > We could technically further limit the schedule to only run on work days.
> 
> Yeah, we should do that. Running before active hours makes sense as it is guaranteed to get the latest stuff from the previous day. I wouldn't mind moving the other build to before standup, so `14:17` UTC or a bit sooner, as we often have meetings after that and having a build to test the changes of the morning could be useful.

Updated to [`17 08,14 * * 1-4`](https://crontab.guru/#17_08,14_*_*_1-4):

> At minute 17 past hour 8 and 14 on every day-of-week from Monday through Thursday

Al12rs (approved):

Sewer56:
> I picked the hours 08 and 15 because it's _before_ most of our start and end times. The idea is that we'd be able to fix any issues without going into overtime.

My train of thought was the exact same when I made the suggestion.

- From what I heard, most of office starts at 9.00am
- We work 8 hours per day, not 7 (4x8 in our contracts)
- Thus most of office should be clocking out at 17:00 

I commented, because I thought the end time was actually set too early, rather than too late.

PR #1101, opened 2024-03-20
Temp fix for advanced installer not starting

### LINKED ISSUE #1094: Installation Cancelled when using "Add Mod (Advanced)", Advanced installer is not opened
```
00:01:03.917 [INFO] Loadout A8A052F01C3E7042B2B87D2873D0839A altered: Cancelled installation of SMIM 2-08-8655-2-08.7z
```

That is the only log line, seems to be a recent regression. Needs investigation.

Al12rs:
fixes #1094 
This is a temporary solution that uses a normal window rather than a Modal dialog for the Advanced installer, so that it can be used again for testing and debugging


Sewer56 (approved):
Fine.

However (quick reminder) please do make an issue for replacing this code for when we get a proper modal system.

PR #1102, opened 2024-03-20
Markdown writer

erri120:
Part of #327.

Allows value formatters to write either plain text or markdown depending on the state.

This PR doesn't include a markdown renderer.


src/Abstractions/NexusMods.Abstractions.Games.Diagnostics/Writer/DiagnosticWriterMode.cs
```diff
@@ -0,0 +1,20 @@
+using JetBrains.Annotations;
+
+namespace NexusMods.Abstractions.Diagnostics;
+
+/// <summary>
+/// Represents the valid modes of a <see cref="IDiagnosticWriter"/>.
+/// </summary>
+[PublicAPI]
+public enum DiagnosticWriterMode
```
Sewer56, line 9: Should be marked as `byte`.

src/Abstractions/NexusMods.Abstractions.Games.Diagnostics/Writer/DiagnosticWriterState.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// <summary>
+    /// Gets the mode of the writer.
+    /// </summary>
+    public readonly DiagnosticWriterMode Mode;
+
+    /// <summary>
+    /// Constructor.
+    /// </summary>
+    public DiagnosticWriterState(DiagnosticWriterMode mode)
```
Sewer56, line 25: An optional builder capacity might be handy here.

Sewer56 (changes requested):
This is fine. Just extremely minor nitpicks.

Quick Question:
- Is the `Mode` intended to be dynamically mutated?

Usually for code like this, I see a design where you have a writer for a specific 'mode' and then you invoke methods for that. This isn't a request to change the code, just curious more than anything.

src/Abstractions/NexusMods.Abstractions.Games.Diagnostics/Writer/DiagnosticWriterMode.cs
erri120: I used to put `byte` on every enum out of habit writing binary serialization stuff, but in most cases it doesn't matter due to alignment.

erri120:
> Quick Question:
> 
>     * Is the `Mode` intended to be dynamically mutated?
> 

Nope.

Sewer56 (approved):

Issue #1091, opened 2024-03-20
Diagnostics UI refinements


Al12rs:
Might need to separate between things to do in first version and what to leave for future improvements.

### Simplified version:
- [x] DiagnosticList page subscription does not provide the last recorded list, so it is always empty on first load (fixed in #1105)
- [x] #1114
- [x] The horizontal line isn't styled with the severity color (fixed in #1115)
- [x] Filter by severity and show chips in a "top bar" (implemented in #1115)
- [x] #1116
- [x] Show some message in DiagnosticListPage if there are no diagnostics to show
- [x] Count badge of number of identified diagnostics on the Left Menu Diagnostics item 
- [x] Left menu icon for diagnostics
- [x] Text wrapping in the list isn't working. (fixed in #1949)
- [ ] Add icons to the severity filter chips
- [ ] Replace severity filter buttons with Chip buttons instead of rounded buttons
- [x] Diagnostic items should not be bold, should be BodyMDNormal, NeutralModerate (fixed in #1949)
- [x] Link color on Diagnostic details page is wrong (should be neutral strong, underline, on hover go to subdued). (fixed in #1960 )
- [x] Diagnostic entry in diagnostic list page should have more padding (fixed in #1949)

erri120:
Closing this as being stale.

PR #1105, opened 2024-03-20
Use `Replay` instead of `Publish`

erri120:
Part of #1091, supersedes #1103.

Uses `Replay` to replay the last value to all current and future subscribers.


erri120:
> Also like Replay rather than the broadcast solution from before.

Not sure what you mean with "broadcast solution" but `Replay` is literally just `Publish` except instead of using a normal `Subject<T>` we use a `ReplaySubject<T>`.

Al12rs:
I meant #1103 with `Multicast`

erri120:
> I meant #1103 with `Multicast`

`Replay` and `Publish` are specialized versions of `Multicast`.

PR #1111, opened 2024-03-21
Better "cancel download" window in Polish UI

Patriot99:
I added a couple of changes to the strings so that the "cancel download" window will always match any situation in the Polish version. I tested two scenarios: Canceling one download and canceling at least two downloads or more at the same time. The UI now looks more professional than the previous one, which looked amateurish and did not match every situation.

One download:
Screenshot_20240321-103632~3

Two or more downloads:
Screenshot_20240321-104536~2


Sewer56:
Strangely the indent didn't update after the last commit.
Want me to fix it up myself?

Patriot99:
> Strangely the indent didn't update after the last commit. Want me to fix it up myself?

Yes, I see that there is another edition. The previous code I did on a PC and the current one is from a smartphone. You can compare the code more easily if you use a PC.

Patriot99:
> image
> 
> Accidental edit here. Aside from that, looks good to me.

I did it specifically to make the UI look better.

Check Screenshot 1:
CancelDownloadOverlayView_Description_download_will_be_cancelled will only be used if you cancel one download (the app will then generate the mod name).

Check Screenshot 2:
CancelDownloadOverlayView_Description_downloads will only be used if you cancel two or more downloads (the app will then generate a number depending on the canceled downloads).

CancelDownloadOverlayView_Description__Download is empty because it is unnecessary, and it acts as a dummy (his job is done by CancelDownloadOverlayView_Description_download_will_be_cancelled).

Sewer56 (approved):

PR #1114, opened 2024-03-21
Show diagnostic count

erri120:
Part of #1091.


Sewer56 (approved):
Looks good to me.

Aesthetically speaking, try using C#12 primary constructors for returning the observable (if you can)

```csharp
public struct DiagnosticCount(int NumSuggestions, int NumWarnings, int NumCritical)
```

This may make it easier to work with long term.
Make it `readonly`/`record struct` depending on your preference. It's a value type, so it gets copied around, so whatever floats your boat.

Al12rs (approved):

PR #1117, opened 2024-03-25
Move FileTreeView out of ModFiles

Al12rs:
This moves out the tree view from the ModInfo Files view, to make it reusable.
This is for #1113 

This removes the Locations section at the top, and always creates root nodes with the full path, this is for consistency (easier to reuse this) and regaining vertical space at the top.


Sewer56:
This removes the Location bar from the top of the view.

The view looks like this by default:

20240325_09h58m47s_Nexus_Mods_App

With the root node manually expanded:

20240325_09h50m13s_Nexus_Mods_App

-----

Before this PR: 

20240325_09h53m13s_Nexus_Mods_App

I'm happy to merge this, but it needs OK from Design ( @captainsandypants ).
The only other suggestion I have, is auto expanding the root node, if it's the only root.

Sewer56 (approved):

PR #1124, opened 2024-03-25
FileTreeView Status Bar

Al12rs:
- Adds the total file count and file sizes in a Status bar at the bottom.
- Rewritten the code that creates the tree to remove some superfluous code
- Cleaned up code
- Added a File Count column (I can comment it out, mostly needed to test if the file counts were accurate) 

image

Part of #1113


src/NexusMods.App.UI/Controls/Trees/ModFileTreeViewModel.cs
```diff
@@ -205,12 +202,47 @@ public ModFileTreeViewModel(LoadoutId loadoutId, ModId modId, ILoadoutRegistry r
                     node => node.IsExpanded
                 ),
 
+                // File Count column
+                new TextColumn<IFileTreeNodeViewModel, string?>(
```
Sewer56, line 206: There's also some other ways to achieve this which may be better.

## Returning the raw type

```csharp
new TextColumn<IFileTreeNodeViewModel, uint?>(
    Language.ModFileTreeViewModel_CreateTreeSource_File_Count,
    x => x.FileCount,
````

The columns will call `ToString` on the returned data, so unless there's a specific need to format the file count, returning the uint should be sufficient.

## Formatting via method on interface

```csharp
string ToFormattedFileCount()
{
    return FileCount.ToString();
}
```

Could put this in the interface itself, and thus you avoid having to add any more lines to the `ViewModel` or `DesignViewModel`.

```
new TextColumn<IFileTreeNodeViewModel, string?>(
            Language.ModFileTreeViewModel_CreateTreeSource_File_Count,
            x => x.ToFormattedFileCount(),
```

## Formatting via static method

```csharp
static string ToStringCount(IFileTreeNodeViewModel input) => input.FileCount.ToString()!;
```

```csharp
// File Count column
new TextColumn<IFileTreeNodeViewModel, string?>(
    Language.ModFileTreeViewModel_CreateTreeSource_File_Count,
    x => ToStringCount(x),
```

-------------

I think the last one is the best approach out of the bunch, since it's really the view's responsibility to choose how to display an item. That code shouldn't really be delegated to the ViewModel if its only purpose is to show it in the view.

Sewer56:
I've read through the code, but I'll make the remaining comments tomorrow, as I'm sleepy.

Good job with adjusting `ModFileTreeViewModel` constructor, you've managed to simplify it down a tiny bit more.

src/NexusMods.App.UI/Controls/Trees/ModFileTreeViewModel.cs
Al12rs: Good suggestions, I didn't like it myself. I have gone with option B, with a method on `IFileTreeNodeViewModel`. It makes the most sense for reusability later as the view only knows about the name, not the other columns, which are all defined on the `TreeNodeVM`,

src/NexusMods.App.UI/Controls/Trees/ModFileTreeViewModel.cs
```diff
@@ -205,12 +202,47 @@ public ModFileTreeViewModel(LoadoutId loadoutId, ModId modId, ILoadoutRegistry r
...
+                    Language.ModFileTreeViewModel_CreateTreeSource_File_Count,
+                    x => x.ToFormattedFileCount(),
+                    options: new TextColumnOptions<IFileTreeNodeViewModel>
+                    {
+                        // Compares IsFile, to show folders first, then by file count for folders, and file name for files.
```
Sewer56, line 211: This comment is a leftover from above code which was pasted as reference

Al12rs: What is the issue with the comment? I'm noting that we use FileCount for Folders and alphabetical for Files

Sewer56 (approved):

PR #1129, opened 2024-03-26
Diagnostics for incompatible SMAPI mods

### LINKED ISSUE #1119: Add SMAPI diagnostic for incompatible mods

erri120:
Resolves #1119.

This PR adds two new diagnostics for Stardew Valley:

The data for these diagnostics are sourced from the internal "mod database" file located at `smapi-internal/metadata.json`. SMAPI itself reads this file at startup to issue warnings and errors, we mirror this behavior in the App with diagnostics.

Other changes:

- Upgrades the SMAPI submodule from 3.18.6` to `4.0.2`.
- Fixes the diagnostic source generator when using raw string literals, multi-line strings, or escaped strings.
- Made the `DiagnosticManager` more robust when one emitter throws an exception.

### DIFF
--- a/src/Games/NexusMods.Games.StardewValley/Installers/SMAPIModInstaller.cs
+++ b/src/Games/NexusMods.Games.StardewValley/Installers/SMAPIModInstaller.cs
@@ -26,17 +26,10 @@ public class SMAPIModInstaller : AModInstaller
     /// DI Constructor
     /// </summary>
     /// <param name="serviceProvider"></param>
-    private SMAPIModInstaller(IServiceProvider serviceProvider) : base(serviceProvider)
+    public SMAPIModInstaller(IServiceProvider serviceProvider) : base(serviceProvider)
     {
         _logger = serviceProvider.GetRequiredService<ILogger<SMAPIModInstaller>>();
     }
-
-    /// <summary>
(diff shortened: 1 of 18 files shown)


src/Games/NexusMods.Games.StardewValley/Installers/SMAPIModInstaller.cs
```diff
@@ -26,18 +26,11 @@ public class SMAPIModInstaller : AModInstaller
...
         _logger = serviceProvider.GetRequiredService<ILogger<SMAPIModInstaller>>();
     }
 
-    /// <summary>
-    /// Creates a new instance of <see cref="SMAPIModInstaller"/>.
-    /// </summary>
-    /// <param name="serviceProvider"></param>
-    /// <returns></returns>
-    public static SMAPIModInstaller Create(IServiceProvider serviceProvider) => new(serviceProvider);
-
```
Al12rs, line 40: Was this simply unused?

erri120: It was used, but it's a weird pattern.

Al12rs (approved):

PR #1131, opened 2024-03-26
Add Ingest button

### LINKED ISSUE #1127: Add simple ingest button
As per discussion, as a stopgap for #695 for 0.4 we will add a simple button that is always shown that ingests any outside changes if present.
Some requirements:
- Ingest will add any found differences to the current loadout, changing existing mods and adding new files to single External Changes mod.
- Disable Apply and Run Game buttons while ingesting
- Disable Ingest button while Applying

Al12rs:
closes #1127
This adds a button to ingest outside changes as a temporary stopgap solution to expose the functionality to users, while we figure out how Apply/Ingest should work.

- Button is shown regardless of whether there are changes or not (no detection or check)
- Ingest happens in whatever loadout was last applied, not necessarily into the current loadout
- Button is hidden while Applying
- Button is disabled while ingesting
- Apply is hidden while ingesting
- Renamed Overrides mod to "Outside Changes"
- No checks if other loadouts are currently ingesting/applying

This pretty bad as features go, but anything more would still be temporary anyways until the Apply/Ingest workflow is fully thought out.


Al12rs:
Ugh, translation conflicts

src/NexusMods.App.UI/LeftMenu/Items/IApplyControlViewModel.cs
```diff
@@ -5,10 +5,16 @@ namespace NexusMods.App.UI.LeftMenu.Items;
...
+    
+    ICommand IngestCommand { get; }
 
     bool CanApply { get; }
+    
+    bool CanIngest { get; }
 
     bool IsApplying { get; }
+    
+    bool IsIngesting { get; }
```
erri120, line 17: You don't really need `CanApply`, `CanIngest`, `IsApplying` and `IsIngesting`. All of those properties are included in `ReactiveCommand`. Changing `ICommand` to `ReactiveCommand<Unit, Unit>` should allow you to remove the other properties.

erri120 (changes requested):

src/NexusMods.App.UI/LeftMenu/Items/IApplyControlViewModel.cs
Al12rs: Can you hide the button with reactive command? Because I need to hide the apply button to show the Applying spinner

erri120: `ReactiveCommand` has `CanExecute` and `IsExecuting` properties that you can subscribe to.

Al12rs: I'll remove IsExecuting, which is a duplicate. I guess you want to use CanExecute and bind it to IsVisible?

erri120: Yeah, you essentially duplicated the behavior with those properties.

Al12rs: I'll try to clean that up

Al12rs: Had to fight with it for a while, as it wasn't behaving like expected. Should be working right now

erri120 (approved):

Issue #1138, opened 2024-03-27
Implement new Manage Game UI design


Al12rs:
# Summary:
Manage game experience has been pretty lackluster. It was redesigned to be simplified in
- #1069

## Design:
Copied over from the other issue:
[Figma link](https://www.figma.com/file/droQp7IUqRCq5AUGmXldvP/%F0%9F%93%B1-Add%2Fmanage-games?type=design&node-id=1072%3A2728&mode=design&t=1vYPsjdLresFPQtR-1)

[Prototype](https://www.figma.com/proto/droQp7IUqRCq5AUGmXldvP/%F0%9F%93%B1-Add%2Fmanage-games?page-id=201%3A43163&type=design&node-id=1086-4338&viewport=2490%2C-2213%2C0.47&t=BQtyXD2INPsxCGxD-1&scaling=min-zoom&starting-point-node-id=1086%3A4338&show-proto-sidebar=1&mode=design)

![image](https://github.com/Nexus-Mods/NexusMods.App/assets/118466382/07aad74a-605e-4fc2-869e-c7a7e86c4a5e)

## Breakdown:
- [x] UI layout update #1140 
- [x] Show currently managed games separate from detected, not yet managed games #1141
- [x] Update GameWidged ui component to add following states: #1144
    - [x]  DetectedGame
    - [x] AddingGame
    - [x] ManagedGame
    - [x] RemovingGame
- [x] Show ManagedGame ui state for managed games, show DetectedGame ui state for detected games  #1147
- [x] Implement navigating to the last Applied loadout when pressing "View" button #1147
- [x] Show loading ui state while a new game is being setup #1147
    - [x] Only add the new loadout to the `LoadoutRegistry` after all files have been backed up, instead of before ingest step. #1147
- [x] Do nothing on Remove button for now, see #1106 for the that part

PR #1141, opened 2024-03-28
Managed games

Al12rs:
Part of 

- #1138
- [x] Show currently managed games separate from detected, not yet managed games
- Added the currently managed games to the "Games Added" section (live updates when a new game is managed)
- Excluded the currently managed games from the Detected Games section

image


erri120 (approved):
Not the biggest fan of converting the games list into a change set that never changes, but this appears to be the easiest solution.

PR #1143, opened 2024-03-28
Removed Database Transaction Bottlenecks

Sewer56:
This is a small set of patches that improve performance of writes by a few orders of magnitude.
It's around 10x faster on an NVMe SSD, maybe a bit more on slower mediums, not a big deal.

-------

The jist of these patches is adding bulk INSERT operations into our data store.

Previously, when we wrote items to the data store, each write would be its own separate transaction, i.e. on every write, for safety, the DB would synchronize with disk, so in the event of a data loss, things won't commit total complete annihilation. 

Unfortunately that's a tiiiiiny bit slow.
Anyway, this PR adds some bulk add operations to the DataStore, and uses them whenever it's known we need to insert a bunch of items at once in some places previously bottlenecked.

------

Note: Don't think too hard about this PR. This is more of a temporary measure to make performance under the existing system bearable. Most of the code touched here will be rewritten or non-existent once EventSourcing kicks in. 

I'll do more proper optimization around that point, such as having non-enumerable APIs.
This is the bare minimum to make things bearable for now.

For now I extracted these from my benchmarks branch, to make review easier down the road.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Serialization/DataModel/EntityDictionary.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Serialization/DataModel/EntityDictionary.cs
@@ -113,8 +113,10 @@ public readonly struct EntityDictionary<TK, TV> :
     public EntityDictionary<TK, TV> With(IEnumerable<TV> val, Func<TV, TK> keyFn)
     {
         var builder = _coll.ToBuilder();
-        foreach (var v in val)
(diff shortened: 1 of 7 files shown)


Sewer56:
(Note: I had to force push just now because rebase made an error, and since I had no conflict show up, I didn't pick up on that)

erri120 (approved):

PR #1144, opened 2024-03-28
Add GameWidget UI states

Al12rs:
Part of 
- #1138

Completes:
- [x] Update GameWidged ui component to add following states:
    - [x] Detected game
    - [x] Adding game
    - [x] Managed game
    - [x] Removing game
    
### Details:
- `AddingGame` and `RemovingGame` states also make the entire widget disabled (opacity)
- Added a State enum to switch the UI with working bindings
- All the added buttons don't have any functionality for now, will do in other PRs
- Removed an old GameWidgetStyles files that seems to have survived the move to Theme


src/NexusMods.App.UI/Controls/GameWidget/GameWidget.axaml.cs
```diff
@@ -10,28 +10,57 @@ namespace NexusMods.App.UI.Controls.GameWidget;
...
+                
+                this.WhenAnyValue(view => view.ViewModel!.State)
+                    .Select(state => state == GameWidgetState.RemovingGame)
+                    .BindToView(this, view => view.RemovingGameStackPanel.IsVisible)
+                    .DisposeWith(d);
+                
+                this.WhenAnyValue(view => view.ViewModel!.State)
+                    .Select(state => state is GameWidgetState.AddingGame or GameWidgetState.RemovingGame)
+                    .BindToClasses(GameWidgetBorder, "Disabled")
+                    .DisposeWith(d);
```
erri120, line 43: In a perfect world, where subscriptions are cheap, this would be okay. However, subscriptions are sadly more expensive, so this entire block should be one subscription on `State`.

erri120 (changes requested):
See comment.

erri120 (approved):

PR #1167, opened 2024-04-03
Update logging

erri120:
- Improved logging messages containing more information
- We're now using two logging files:
  - `nexusmods.app.main` for the main process only
  - `nexusmods.app.slim` for the "slim" process only
- Updates the "bug" issue template to prompt the user to upload log files.

Having two logging files fixes the issue where both the main and slim process try to write to the same file. While we didn't have any corruptions of log files, we did have an issue where a slim process will move the current main log to an archive log while the main process would continue writing to the archive log. It made debugging and viewing logs difficult, this PR fixes that issue by clearly separating the two logs.

### DIFF
--- a/src/NexusMods.App/AppConfig.cs
+++ b/src/NexusMods.App/AppConfig.cs
@@ -64,12 +64,30 @@ public class AppConfig
 
 public interface ILoggingSettings
 {
+    /// <summary>
+    /// Gets the path to the current log file for the main process.
+    /// </summary>
     public ConfigurationPath MainProcessLogFilePath { get; }
 
+    /// <summary>
+    /// Gets the template path to the archive log file for the main process.
+    /// </summary>
+    /// <remarks>
+    /// For details, see https://nlog-project.org/documentation/v5.0.0/html/P_NLog_Targets_FileTarget_ArchiveFileName.htm
+    /// </remarks>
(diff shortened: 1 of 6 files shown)


src/NexusMods.App/AppConfig.cs
```diff
@@ -64,15 +64,13 @@ public void Sanitize(IFileSystem fs)
...
 
-    /// <summary>
-    /// Path to historical log files, with templated element e.g. 'PATH_TO_FILE/nexusmods.app.{##}.log'
-    /// </summary>
-    public ConfigurationPath ArchiveFilePath { get; }
+    public ConfigurationPath MainProcessArchiveFilePath { get; }
+
+    public ConfigurationPath SlimProcessLogFilePath { get; }
+
+    public ConfigurationPath SlimProcessArchiveFilePath { get; }
```
Al12rs, line 73: Missing docs

erri120: Do you really need/want docs for this?

Al12rs: `ArchiveFilePath` might not be immediate, some explanation for what is SlimProcess might be useful as well

Al12rs (approved):
Looks ok if there are no objections to splitting the log file from others

halgari (commented):
Looks good to me, good solution to the problem

PR #1172, opened 2024-04-04
Apply diff backend

Al12rs:
This adds an extension method to the LoadoutSyncrhonizer to compute a Diff tree between a loadout and a DiskState.
Part of 
- #1113 

Added a test to check if it worked and results seem correct

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/Diff/DiskDiffEntry.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/Diff/DiskDiffEntry.cs
@@ -1,0 +1,31 @@
+using NexusMods.Abstractions.GameLocators;
+using NexusMods.Hashing.xxHash64;
+using NexusMods.Paths;
+
+namespace NexusMods.Abstractions.Loadouts.Synchronizers;
+
+/// <summary>
+/// A file diff entry
+/// </summary>
+public record struct DiskDiffEntry
+{
+    /// <summary>
+    /// The game path of the file
+    /// </summary>
+    public required GamePath GamePath { get; init; }
+    
+    /// <summary>
+    /// The size of the file (newest if modified)
+    /// </summary>
+    public required Size Size { get; init; }
+    
+    /// <summary>
+    /// The hash of the file (newest if modified)
+    /// </summary>
+    public required Hash Hash { get; init; }
+    
+    /// <summary>
+    /// The type of change that occurred to the file path
+    /// </summary>
+    public required FileChangeType ChangeType { get; init; }
+}
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/Diff/FileChangeType.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/Diff/FileChangeType.cs
@@ -1,0 +1,30 @@
+namespace NexusMods.Abstractions.Loadouts.Synchronizers;
+
+/// <summary>
+/// The type of change of a file at a specific path
+/// </summary>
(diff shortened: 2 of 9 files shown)


Sewer56:
I self assigned this because this code could be handy for improving ingest performance.

Al12rs:
I'm using this for the UI implementation and already have moved a couple of things around, so I might handle some of the feedback directly in that PR if there is any

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/Diff/LoadoutDiskSateDiffUtils.cs
```diff
@@ -0,0 +1,124 @@
...
+        var flattenedLoadout = await loadoutSynchronizer.LoadoutToFlattenedLoadout(loadout);
+        return await FlattenedLoadoutToDiskDiff(flattenedLoadout, diskState);
+    }
+
+    private static ValueTask<FileDiffTree> FlattenedLoadoutToDiskDiff(FlattenedLoadout flattenedLoadout, DiskStateTree diskState)
+    {
+        var loadoutFiles = flattenedLoadout.GetAllDescendentFiles();
+        var diskStateEntries = diskState.GetAllDescendentFiles();
+
+        Dictionary<GamePath, DiskDiffEntry> resultingItems = new();
```
Sewer56, line 35: Can you set the initial capacity here?
`Max(loadoutFiles, diskStateEntries)` should be ok, that accounts for deletion.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/Diff/FileChangeType.cs
```diff
@@ -0,0 +1,30 @@
+namespace NexusMods.Abstractions.Loadouts.Synchronizers;
+
+/// <summary>
+/// The type of change of a file at a specific path
+/// </summary>
+public enum FileChangeType
```
Sewer56, line 6: Consider restricting this to a byte.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/Diff/DiskDiffEntry.cs
```diff
@@ -0,0 +1,31 @@
+using NexusMods.Abstractions.GameLocators;
+using NexusMods.Hashing.xxHash64;
+using NexusMods.Paths;
+
+namespace NexusMods.Abstractions.Loadouts.Synchronizers;
+
+/// <summary>
+/// A file diff entry
+/// </summary>
+public record struct DiskDiffEntry
```
Sewer56, line 10: Consider setting this to `Record Class` for the time being.

I generally consider the threshold of struct profitability to be 4 registers. This has 6 variables (2 in Size), (2 in GamePath), thus is considerably larger.

The one in `Size` should probably be patched out by switching away from Vogen in paths, but GamePath will always be 2 vars.

Sewer56 (approved):
Feel free to do the changes either on this branch directly, or on the UI side branch you're working with already.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/Diff/LoadoutDiskSateDiffUtils.cs
Al12rs: Deletions are included in the Dictionary, so Max wouldn't be accurate if the loadout has both deletions and additions.
I guess it would still be better than nothing.

Al12rs:
I addressed the feedback on my branch, will be in for next PR

PR #1185, opened 2024-04-09
Performance PR #2

Sewer56:
## Summary

Since the last PR was nicknamed 'Performance PR', this one shall be officially named 'Performance PR #2'.

This PR introduces a small series of performance optimizations and improvements in places where some of our core workloads (mainly Ingest related ones) were previously bottlenecking hard. 

The changes here focus on enhancing the efficiency of file extraction, packing and hashing.

## Key Changes

1. **Faster Memory Mapped Hashing**
   - Implemented a faster, memory-mapped hash helper improved performance.
   - Achieved at least 15x improvement in file hashing speed for datasets of small (<100 byte) inputs.
   - This is now used in places where we're purely hashing, without copying.

2. **Parallelism in Ingest Deduplication**
   - Introduced parallelism in `ALoadoutSynchronizer.BackupNewFiles` when querying DB to determine which new files should be hashed.
   - Improves performance by 4-5x. Very noticeable on huge loadouts.

3. **Preallocating Dictionaries**
   - Dictionaries in NxFileStore now have an initial capacity, to avoid unnecessary copying.

4. **Parallelized DB Queries in NxFileStore**
   - In methods like `ExtractFiles`, we query DB to determine to check which archive a file originates from.
   - As we can have multiple locations for archives.

5. **Renamed ArchivedFiles to ArchivedFile**
   - Renamed the `ArchivedFiles` class to `ArchivedFile` as it represents a singular file.

6. **Bug Fix: Duplicate Extraction Paths**
   - Tests would sometimes request extracting the same hash multiple times.
   - Or to same location twice, as byproduct
   - Added an assertion (in Debug Mode Only) to ensure correct runtime behaviour just in case.

7. **FileStore Returns Concrete `Dictionary` as opposed to `IDictionary`**
   - Because this is internal code, we can change this anytime.
   - Helps avoiding unnecessary virtual method calls.

8. **Removal of Excessive LINQ in NxFileStore**
   - Removed cases where large LINQ chains were used, in `NxFileStore` and parameters passed to `NxFileStore` as inputs.

9. **Parallel Opening of `MemoryMappedFile`s for Extraction**
   - Speeds up extracting many small files.

10. **Improved Download Registration Performance**
    - Parallelized `RegisterFolderInternal`, to greatly speed up hashing of newly added mods.
    - Speeds up hashing of small files greatly.
    - This includes the base game state, too.

11. **Pre-Preparing Directories for Nx Extract Setup**
    - NxFileStore's Extract Methods now prepare the directories ahead of time before sending them to Nx.
    - This speeds up the extract process for small files a bit, as Nx doesn't have to handle the error silently. (Error is avoided)

12. **Reusing Hash from Nx Headers**
    - Improved performance by reusing the hash from Nx headers instead of converting string to hash.

13. **Removal of Unused GUID from `NxFileStore` `ArchivedFile` Keys**
    - Removed an unused GUID from `NxFileStore`'s `ArchivedFile` keys.
    - This speeds up DB queries when determining if a file has already been backed up.

15. **Updated Nx Archive Library**
    - [Changelog here](https://github.com/Nexus-Mods/NexusMods.Archives.Nx/releases/tag/0.4.0), and on Discord.


Sewer56:
This is OK for review.
I set this as draft because I need to add an API to Paths to determine if the FileSystem is the real one, because InMemoryFileSystem does not expose mmap like functionality.

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
```diff
@@ -30,4 +31,45 @@ public static class AbsolutePathExtensions
...
+            // but we have to live with CreateFromFile being the fastest for now.
+
+            // Empty file.
+            if (new FileInfo(fullFilePath).Length != 0) 
+                throw;
+
+            // JIT turns this to a constant.
+            // xxh64sum --little-endian "empty"
+            var emptyFileHash = BitConverter.IsLittleEndian ? 0x99e9d85137db46ef : 0xef46db3751d8e999;
+            return Hash.From(emptyFileHash);
```
erri120, line 72: Why do you return these magic numbers instead of a zeroed hash?

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
```diff
@@ -30,4 +31,45 @@ public static class AbsolutePathExtensions
...
+    /// <param name="input">The path to the file.</param>
+    /// <param name="job">The job to report progress to.</param>
+    /// <returns>The xxHash64 hash of the file.</returns>
+    public static unsafe Hash XxHash64MemoryMapped(this AbsolutePath input, IActivitySource<Size>? job = null)
+    {
+        #if DEBUG
+        // TODO: Assert that AbsolutePath uses RealFileSystem.
+        // This needs API update in paths, and an corresponding issue.
+        // If it's not RealFileSystem, we can fallback to the stream based version.
+        // Since we can't memory map, unless we add mmap ti Paths library.
```
erri120, line 46: Just add something like `IFileSystem.SupportsMemoryMappedFiles`.

src/NexusMods.DataModel/NxFileStore.cs
```diff
@@ -127,44 +126,73 @@ await using (var outputStream = outputPath.Create())
...
+                // down the road. Usually the difference is negligible, but in
+                // extra special with 100s of directories scenarios, it can
+                // save a second or two.
+                var containingDir = file.Dest.Parent;
+                if (createdDirectories.TryAdd(containingDir, 0))
+                    containingDir.CreateDirectory();
+#if DEBUG
+                if (!destPaths.TryAdd(file.Dest, 0))
+                    throw new Exception($"Duplicate destination path: {file.Dest}. Should not happen.");
+#endif
```
erri120, line 170: This can be a `Debug.Assert` or an `UnreachableException`.

src/NexusMods.DataModel/NxFileStore.cs
```diff
@@ -127,44 +126,73 @@ await using (var outputStream = outputPath.Create())
...
+            }
+            else
+            {
+                throw new Exception($"Missing archive for {file.Hash.ToHex()}");
+            }
```
erri120, line 175: Throw a more sensible exception than just `Exception`.

erri120:
Can you post some numbers on the before and after performance?

src/NexusMods.DataModel/FileOriginRegistry.cs
```diff
@@ -113,23 +113,33 @@ private async ValueTask<DownloadId> RegisterFolderInternal(AbsolutePath path, AA
...
-            if (!knownHashes.Contains(hash.Value))
-                filesToBackup.Add(archivedEntry);
+            var relativePath = file.RelativeTo(path);
+            lock (paths)
+            {
+                paths.Add(relativePath);
+                files.Add(archivedEntry);
+                if (!knownHashes.Contains(hash.Value))
+                    filesToBackup.Add(archivedEntry);
+            }
```
Al12rs, line 140: Do you know if the lock is faster than using `ConcurrentBag`s for `paths`, `files` and `filesToBackup`?

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
Sewer56: The hash function is defined as 

```csharp
public static ulong HashBytes(ReadOnlySpan<byte> data, ulong seed = 0)
{
    var hash = new XxHash64Algorithm(seed);
    var initialSize = (data.Length >> 5) << 5;
    if (initialSize > 0)
        hash.TransformByteGroupsInternal(data[..initialSize]);

    return hash.FinalizeHashValueInternal(data[initialSize..]);
}
```

The numbers are equivalent of what we'd do if we ran 
```csharp
byte dummy = 0;
`HashBytes(new Span<byte>(&dummy, 0x0))`
```

I could replace the code with that, but it won't fold to a constant.
Could move the constant to the hash library though.

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
Sewer56: Could the `IFileSystem` not technically be used to e.g. redirect the file paths?
I think in this case, checking explicitly for `RealFileSystem` (raw operations) may be preferred.

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
erri120: Remove the magic numbers from the code. You can make them const fields and add some comments, or move them into the hash library.

src/NexusMods.DataModel/FileOriginRegistry.cs
Sewer56: It depends on the level of contention in terms of adding items to the collection.
The main factor is `Time in Lock` vs `Out of Lock`. And that's also affected by size & size distribution of input data. 

In this specific scenario, this is more efficient as for regular sized files the time `Out of Lock` (Hashing) is much greater than the time taken to add to the collections. So the contention on the lock is very, very low in practice.

Thus taking only 1 lock (instead of Locking in every `ConcurrentBag` method) is cheaper here, because `List.Add` is super fast and `Time in Lock` is minimized.

(if that makes sense)

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
Sewer56: The issue is the code here calls `mmap` directly, avoiding the `IFileSystem `abstraction.

Thus if `IFileSystem` that uses real filesystem would perform remapping, we would get the wrong file.

Which is why I suggested we either make an mmap abstraction in Paths, or simply a way of knowing if I'm working on a path on real FileSystem.

src/NexusMods.DataModel/NxFileStore.cs
Sewer56: Note: This was preserved from the original code. Behaviour wasn't changed here.

I changed it though to throw `FileNotFoundException`.

Sewer56:
> Can you post some numbers on the before and after performance?

Do you want rough numbers? Or precise numbers?

These changes were done when I was trying to get the benchmarks in the `loadoutsynchronizer-benchmarks` to start fast 'enough', so for some of the functionality here, I don't have isolated benchmarks. Measuring was done via profiling.

However for a test case process of:

- RegisterDownload of 60k file archive, ~60 bytes average file size.
    - Extract Archive (zip/7z etc.)
    - Hash all 60k files with xxHash64
    - Compress all files to Nx
    
- Apply Loadout with 1 Mod (Aforementioned 60k Files)
    - i.e. Flatten Loadout, Get File Tree etc.
    - Extract to Disk with Nx

- And some other minor overhead.

This process previously took over 4 minutes on my machine. Currently, this is reduced to 20 seconds.

To replicate my setup, run the `GetDiskState` benchmark on the `loadoutsynchronizer-benchmarks` branch, and set a breakpoint on `Init("Game Files", filePath);` and `_flattenedLoadout = Task.Run(() => _defaultSynchronizer.LoadoutToFlattenedLoadout(_datamodel.BaseList.Value)).Result.Result`.

Improvements were done incrementally, so I'd need to bench without patch at all and with patch to get good numbers for some processes. (There is a commit for each perf improvement patch). However, at least, specifically the process of hashing the 60k files dropped from 17 seconds to 2 seconds.

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
erri120: > Thus if IFileSystem that uses real filesystem would perform remapping, we would get the wrong file.

How woudl we get the wrong file?

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -239,8 +242,7 @@ await foreach (var entry in _hashCache.IndexFolderAsync(location))
 
 
         // Extract all the files that need extracting in one batch.
-        await _fileStore.ExtractFiles(toExtract
-            .Select(f => (f.Value.Hash, f.Key)));
```
Al12rs, line 243: Was this really bottlenecking?
Do you have data on the improvements for using the new function for curiosity?

Sewer56: It's complicated to state the improvement here in isolation.

Normally something like this alone won't be a bottleneck, what was a bottleneck was taking this `IEnumerable` and then doing multiple LINQ operations on that enumerable inside the `ExtractFiles` method, creating a much bigger enumerable.

And then with that enumerable, making singlethreaded DB calls per file to obtain the archive in which each file is stored.

That logic was parallelized with around ~5x improvement if I recall. Mostly due to parallel DB calls.

For the parallel logic, knowing the size (length) of the collection was also handy for the purposes of preallocating. So it's hard to narrow how much this specific line in isolation changes. It was part of a bigger set of patches.

Sewer56: If I were to have a completely random guess though, it's probably ~1.6x as fast, and with less allocations. But it won't amount into much, probably like 20-40 millis improvement on a 100K file input on this line.

I just did it as a I de-LINQ'ed the whole codepath, as the logic inside the `ExtractFiles` which used the enumerable was originally bottlenecking.

Sewer56:
This is blocked by:

- https://github.com/Nexus-Mods/NexusMods.App/pull/1235

Since it touched some of the same code.

I have merged said branch into here, and resolved conflicts to keep stuff up to date, but the former needs merging first.

halgari (approved):

Issue #1182, opened 2024-04-08
Epic: Settings


LukeNexusMods:
For work related to the settings page.
Discussion to be had around what the settings page will do and what options are needed.

- [x] #396
- [x] #1189
- [x] #1190
- [x] #920
- [x] #1243
- [x] #1370
- [x] #1571
- [ ] #1373
- [x] #1371
- [ ] Add Dialog on save press in there are changes that need user to restart (#1372)
- [ ] Highlight modified settings in the UI (settings that no longer have the default value)
- [ ] Add button to reset to default value in case of modified
- [ ] Add Dialog for Cancel button (You have unsaved changes)
- [ ] Filter search
- [ ] Add developer docs on how to add new settings, before DI settings, and new setting UIs
- [ ] #2303
- [x] #1422

Issue #1043, opened 2024-03-07
Create a system to save and restore View specific UI states (header sizes, current sorting, selected ChipTab etc)


Al12rs:
In Qt there is a nice `QWidged::saveGeometry()` and `QWidged::restoreGeometry()` that can be extended to save UI states.
This creates data that is keyed to a specific Widget (View) type.

This way, for each view, when opening and closing them the user can keep custom UI state that was changed.

Examples are:
- Resizing/reordering Headers in DataGrid and TreeDataGrid
- Sorting header selection for DataGrid and TreeDataGrid
- Selected sub tab (chips).

erri120:
> This way, for each view, when opening and closing them the user can keep custom UI state that was changed.
> Examples are:
> * Resizing/reordering Headers in DataGrid and TreeDataGrid
> * Sorting header selection for DataGrid and TreeDataGrid
> * Selected sub tab (chips).

Specifically for pages, the state is stored inside the page context. That is the only thing that gets persisted. You can call this "data state".

I think what you want is to persist the "visual state", which doesn't contain or reference any data. Saving and restoring this state would massively increase the amount of bindings required between the View Model and the View, since the values are purely visual.

While this sounds like a "nice to have", it'd require a lot of effort for something purely cosmetic.

Al12rs:
> I think what you want is to persist the "visual state", which doesn't contain or reference any data.

That is correct.

> While this sounds like a "nice to have", it'd require a lot of effort for something purely cosmetic.

You say that, but always having to click the Size header every time you open a ModInfo page, or every time you open the Downloads View gets old very fast.
It seems like a small thing but it can make a lot of difference for users.

Maybe we can talk about options during standup.

PR #1187, opened 2024-04-09
Settings Backend

### LINKED ISSUE #396: Settings - Implement Settings Backend
Using the modern [Configuration](https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration) approach, components in our DI system can require `IOption<MyOptions>` or `IOptionMonitor<MyOptions>` to access component-specific configuration without having to deal with loading, saving and updating them. The monitor is especially useful as it's fully reactive and allows for changes to be applied immediately (see #362 for an example).

These settings/options should be saved to disk as JSON to be persistent, but should also be presented to the user in some capacity. Such a system would provide implementations of `IOptionMonitor<T>` and serve as a configuration provider.

- #1182

erri120:
Part of #1182, resolves #396.

Initial implementation of the settings backend:

- Adds `ISettings` for all setting types.
- Adds `ISettingsManager` to get or change settings.
- Adds `ISettingsBuilder` to configure how settings appear in the UI.

Missing from this PR that are done in a separate PR as part of #1182:

- Moving existing setting types to the backend
- Persisting and loading the settings
- Allowing the user to change the settings via a UI

### DIFF
--- a/NexusMods.App.sln
+++ b/NexusMods.App.sln
@@ -219,6 +219,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NexusMods.Icons", "src\Nexu
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NexusMods.Abstractions.Settings", "src\Abstractions\NexusMods.Abstractions.Settings\NexusMods.Abstractions.Settings.csproj", "{F8E762B4-185E-4DC5-9651-B8EAF6139488}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NexusMods.Settings", "src\NexusMods.Settings\NexusMods.Settings.csproj", "{2F642097-B510-4E0E-8467-8BD288C47DC3}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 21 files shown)


halgari:
This all looks solid to me. Any thoughts on localization with this? I assume that's on the UI side?

I thought we might want the `Set` methods to be async, but now that I think about it, we can cache those values, and have a background task that syncs the settings to disk, no reason to make the `Set` call async just to wait for a disk write.

halgari (approved):

Al12rs:
Do you have examples of when we might want to use settings without UI? 
It's effectively a key value store at that point, for simple values at least, but we may still need to configure the backing store for them.
Should we consider using this for things like saving visual UI state ( #1043 e.g. what column the user last sorted the mod file tree by)

erri120:
> This all looks solid to me. Any thoughts on localization with this? I assume that's on the UI side?

`WithDisplayName` and `WithDescription` take strings, so you can just use resources, and it'll work just fine.

erri120:
> Do you have examples of when we might want to use settings without UI? It's effectively a key value store at that point, for simple values at least, but we may still need to configure the backing store for them. Should we consider using this for things like saving visual UI state ( #1043 e.g. what column the user last sorted the mod file tree by)?

#1043 can definitely benefit from this. Ideally, every kind of setting we have should move to this backend.

Issue #985, opened 2024-02-22
Apply/Ingest - Store last applied `LoadoutId` and revision ID separate from `DiskState`


Al12rs:
This is to be able to access that information without also having to load the entire `DiskState` from DB.

This can be done by having two tables:
- `<GameLocationId, DiskState>`
- `<GameLocationId, (LoadoutRevisionID, LoadoutId, more small data in the future if needed)>`

This is because with large loadouts or games with large number of files DiskState can be pretty big.

halgari:
This will be fixed by the additional features offered by #1173 and so I should start by supporting DiskState and the FileStore

PR #1194, opened 2024-04-09
Move DiskStateRegistry to MnemonicDB

halgari:
Reworks the DiskStateRegistry to use MnemonicDB. 

This is a first step to integrating MnemonicDB into the app, and shook out a lot of the small bugs I had yet to encounter. 

What this does:
* Creates 4 attributes for the four paramters of the DiskStateAchive
  * 4 of these attributes are marked as 'noHistory:true' as we only ever want the most recent value, so no reason to track changes over time
  * All the attributes except for `State` are marked as `IsIndexed=true` meaning we can look up the state by any of these, or any combination of these attributes 
* Groups the attributes together into a model
* Updates DiskStateRegistry to use this model
* Adds an `IConnection` class into the app, and sets up everything to use MnemonicDB with the RocksDB backend
* Disk state is stored in a [FlatBuffer](https://flatbuffers.dev/index.html) there's some nice properties this format offers us, read up on it if you're bored
* MnemonicDB supports HashedBlobs now where the values can be super large (technically GB in size) and only the hash will be used for sorting, so we use that for storing the state

Not in this PR but to revisit later:

* MnemonicDB needs to be updated to allow for entity paritions. We can get better performance by separating out entities into rough groups: loadouts, downloads, diskstate, settings, should likely be all in separate partitions. I need to add support to this to MnemonicDB. Once I do we can move DiskState to its own partition

* We are hand crafting queries in this code currently (see the use of `.FindIndexed`) not a horrible issue, but we'll want a query engine eventually. I'm working on this

This PR resolves: #985 we no longer need to load the entire state to get the current applied loadout Id, later on we can query any of these values independantly of any others, simply by using `.FindIndexed<>`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ This also includes our first "suggestion" diagnostic. This diagnostic will only 
 ### Other Changes
 
 - A new settings backend was added in preparation for a settings UI. See [#1182](https://github.com/Nexus-Mods/NexusMods.App/issues/1182) for more details.
+- Parts of the backend has been reworked to use our new database. See [#1194](https://github.com/Nexus-Mods/NexusMods.App/pull/1194) for details.
 - The App will now use two logging files, `nexusmods.app.main.log` and `nexusmods.app.slim.log`, instead of one `nexusmods.app.log` to prevent log collisions between processes ([#1167](https://github.com/Nexus-Mods/NexusMods.App/pull/1167)).
 - The default logging level has been changed from `Information` to `Debug` for Release builds to make it easier to debug issues ([#1209](https://github.com/Nexus-Mods/NexusMods.App/pull/1209)).
 


src/NexusMods.DataModel/Serializers/AbsolutePathSerializer.cs
erri120, line 1: The comparison as well as reading from and writing to byte spans should probably be in the path library. This serializer class can remain here.

src/NexusMods.DataModel/Serializers/DiskStateTreeSerializer.cs
```diff
@@ -0,0 +1,114 @@
...
+        Files files;
+
+        if (decompressedSize == buffer.Length - sizeof(uint))
+        {
+            files = Files.Serializer.Parse(compressedData.ToArray());
+        }
+        else
+        {
+            var decompressed = new byte[decompressedSize];
+            unsafe
```
erri120, line 52: Use `GC.AllocateUninitializedArray`. The entire array gets written to, so we don't care if it contains non-zero data.

src/NexusMods.DataModel/Serializers/DiskStateTreeSerializer.cs
```diff
@@ -0,0 +1,114 @@
...
+        
+        var kvs = files.All
+            .Select(f => new KeyValuePair<GamePath, DiskStateEntry>(new GamePath(LocationId.From(f.LocationId), f.Path),
+                new DiskStateEntry
+                {
+                    Size = Size.From(f.Size),
+                    Hash = Hash.From(f.Hash),
+                    LastModified = DateTime.FromFileTimeUtc(f.LastWriteTime)
+                }
+            ));
```
erri120, line 73: We can replace this LINQ query with a simple for-loop. This is faster and more efficient since it doesn't have unnecessary allocations here. The subsequent iteration of the `IEnumerable<>` in `AGamePathNodeTree` will also be faster because no enumerator has to be allocated for arrays.

```csharp
        var res = GC.AllocateUninitializedArray<KeyValuePair<GamePath, DiskStateEntry>>(files.All.Count);

        for (var i = 0; i < files.All.Count; i++)
        {
            var f = files.All[i];
            res[i] = new KeyValuePair<GamePath, DiskStateEntry>(
                new GamePath(LocationId.From(f.LocationId), f.Path),
                new DiskStateEntry
                {
                    Size = Size.From(f.Size),
                    Hash = Hash.From(f.Hash),
                    LastModified = DateTime.FromFileTimeUtc(f.LastWriteTime),
                }
            );
        }
```

src/NexusMods.DataModel/Serializers/GameDomainSerializer.cs
```diff
@@ -0,0 +1,32 @@
...
+        return a.SequenceCompareTo(b);
+    }
+
+    public Type NativeType => typeof(GameDomain);
+    public Symbol UniqueId => Symbol.Intern<GameDomainSerializer>();
+    
+    public GameDomain Read(ReadOnlySpan<byte> buffer)
+    {
+        return GameDomain.From(Encoding.UTF8.GetString(buffer));
+    }
```
erri120, line 23: This should use `_encoding` instead of `Encoding.UTF8`.

src/NexusMods.DataModel/IDataModelSettings.cs
```diff
@@ -59,13 +65,17 @@ public class DataModelSettings : IDataModelSettings
 
     private const string DataModelFileName = "DataModel.sqlite";
     private const string DataModelIpcFileName = "DataModel_IPC.sqlite";
+    private const string MnemonicDBFileName = "MnemonicDB";
     private const string ArchivesFileName = "Archives";
```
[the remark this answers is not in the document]
erri120: Also this is gonna conflict with #1201.

src/NexusMods.DataModel/Serializers/AbsolutePathSerializer.cs
halgari: Since MnemonicDB stores all entries as a large sorted set, every value int the database has to be comparable to itself, but that comparison can be arbitrary. In addition, the comparisons here work on spans instead of on values. We can move this code later if needed, but I think it's pretty specific to this use case.

src/NexusMods.DataModel/IDataModelSettings.cs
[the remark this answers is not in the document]
halgari: It's actually a folder, so I renamed it, and gave it an extension just so we know what the data is inside it.

src/NexusMods.DataModel/Attributes/DiskState.cs
```diff
@@ -0,0 +1,75 @@
...
+    
+    /// <summary>
+    /// The associated loadout id.
+    /// </summary>
+    public static readonly Attribute<IId> LoadoutRevision = new("NexusMods.DataModel.DiskStateRegistry/LoadoutRevision", isIndexed: true, noHistory: true);
+
+    /// <summary>
+    /// The state of the disk.
+    /// </summary>
+    public static readonly Attribute<NexusMods.Abstractions.DiskState.DiskStateTree> State = new("NexusMods.DataModel.DiskStateRegistry/State", noHistory: true);
```
erri120, line 34: I'm not the biggest fan of uses raw strings like this. Maybe use something like this:

```csharp

    private const string NamespaceString = "NexusMods.DataModel.DiskStateRegistry";

    /// <summary>
    /// The associated game id.
    /// </summary>
    public static readonly Attribute<GameDomain> Game = new($"{NamespaceString}/{nameof(Game)}", isIndexed: true, noHistory: true);

```

src/NexusMods.DataModel/Attributes/DiskState.cs
```diff
@@ -0,0 +1,75 @@
...
+    
+        /// <summary>
+        /// The state of the disk.
+        /// </summary>
+        public NexusMods.Abstractions.DiskState.DiskStateTree DiskState
+        {
+            get => State.Get(this);
+            set => State.Add(this, value);
+        }
+    }
```
erri120, line 74: There is definitely some amount of boilerplate code involved. I think model classes like these should have these attributes to stop Rider/ReSharper from complaining:

```csharp
    [PublicAPI]
    [SuppressMessage("ReSharper", "MemberHidesStaticFromOuterClass")]
    internal class Model(ITransaction tx) : AEntity(tx)
```

erri120 (changes requested):

src/NexusMods.App/Program.cs
```diff
@@ -24,6 +25,7 @@ public static int Main(string[] args)
     {
         MainThreadData.SetMainThread();
 
+        Debugger.Launch();
         var isMain = IsMainProcess(args);
```
erri120, line 29: Is this a leftover?

erri120:
Conflicts should be easily resolved by adding the new config value to `DataModelSettings`.

src/NexusMods.App/Program.cs
halgari: bleh, yes, I was testing some things 🤦

src/NexusMods.DataModel/Services.cs
```diff
@@ -32,10 +38,26 @@ public static class Services
...
+        }
         else
-            coll.AddSingleton(settings);
+        {
+            coll.AddSingleton(new DatomStoreSettings
+                {
+                    Path = settings!.MnemonicDBPath.ToAbsolutePath(),
+                }
+            );
+            coll.AddAllSingleton<IStoreBackend, MnemonicDB.Storage.RocksDbBackend.Backend>();
```
erri120, line 59: With the new settings backend, you'd probably want something like this:

```csharp
coll.AddSingleton<IStoreBackend>(serviceProvider => {
  var settingsManager = serviceProvider.GetRequiredService<ISettingsManager>();
  var settings = settingsManager.Get<DataStoreSettings>();
  return settings.UseInMemoryDataModel ? new InMemoryBackend.Backend() : new RocksDbBackend.Backend();
})
```

src/NexusMods.DataModel/Attributes/DiskState.cs
```diff
@@ -16,26 +17,25 @@ public static class DiskState
...
     /// The associated game id.
     /// </summary>
-    public static readonly Attribute<GameDomain> Game = new($"{Namespace}/{nameof(Game)}", isIndexed: true, noHistory: true);
+    public static readonly GameDomainAttribute Game = new(Namespace, nameof(Game)) { IsIndexed = true, NoHistory = true };
     
     /// <summary>
     /// The game's root folder
     /// </summary>
-    public static readonly Attribute<AbsolutePath> Root = new($"{Namespace}/{nameof(Root)}", isIndexed: true, noHistory: true);
+    public static readonly StringAttribute Root = new(Namespace, nameof(Root)) { IsIndexed = true, NoHistory = true };
```
erri120, line 28: This can't be an `AbsolutePath`?

src/NexusMods.DataModel/Attributes/DiskStateAttribute.cs
```diff
@@ -0,0 +1,60 @@
...
+using NexusMods.Hashing.xxHash64;
+using NexusMods.MnemonicDB.Abstractions.Attributes;
+using NexusMods.MnemonicDB.Abstractions.ElementComparers;
+using NexusMods.Paths;
+using NexusMods.Paths.Trees.Traits;
+using File = NexusMods.DataModel.Serializers.DiskStateTreeSchema.File;
+
+namespace NexusMods.DataModel.Attributes;
+
+public class DiskStateAttribute(string ns, string name) : HashedBlobAttribute<DiskStateTree>(ns, name)
```
erri120, line 17: Doesn't this need to implement `ToLowLevel`?

src/NexusMods.DataModel/Services.cs
```diff
@@ -56,16 +100,19 @@ public static IServiceCollection AddDataModel(this IServiceCollection coll)
...
+        coll.AddAttributeCollection(typeof(DiskState));
+        
         coll.AddAllSingleton<IApplyService, ApplyService>();
 
         coll.AddSingleton<ITypeFinder>(_ => new AssemblyTypeFinder(typeof(Services).Assembly));
         coll.AddAllSingleton<ISorter, Sorter>();
 
         // Diagnostics
         coll.AddAllSingleton<IDiagnosticManager, DiagnosticManager>();
-        coll.AddSettings<DiagnosticSettings>();
```
Al12rs, line 67: Was this supposed to be removed?

src/NexusMods.DataModel/Attributes/DiskState.cs
Al12rs: I think that is a LocationId

halgari: I really should have documented this better, but the issue with using AbolutePath here is that AbsolutePath requires a IFileSystem, and these attributes don't have access to a DI Container at this location. Since they are static and part of the system as a whole, whenever we read the attribute as a string, we'll need some sort of IFileSystem to convert it to an AbsolutePath, and it's not clear here what the exact filesystem should be. 

Well LocationId requires a game and a version to link the data to, and in this context we don't have the version and the game is stored in a separate attribute. What the DiskStateRegistry really stores is: here's the last state (DiskStateTree) of this folder (AbsoltuePath), that was last associated with this Game and Loadout Revision. We could change this interface, but the previous version was coded to use `GameDomain+AbsolutePath` and I've simply broken these out into separate attributes

src/NexusMods.DataModel/Attributes/DiskStateAttribute.cs
halgari: Blobs (and HashedBlobs) are different because what we need at the low level is a `ReadOnlySpan<byte>` to write to the disk. So these expose a lower level interface that hands the attribute a `IBufferedWriter<byte>` that they can use to serialize as they please. I would have prefered it to be a uniform interface, but there's not an easy way to do this without making the normal use case (value types) more complex just to make a rare use case (blobs) simpler.

src/NexusMods.DataModel/Services.cs
halgari: Don't think so.

src/NexusMods.DataModel/Attributes/DiskState.cs
Al12rs: Ah, yeah I read this wrong. Saving the absolute path is what we want since we need to identify the installation.
If it needs to be saved as a string oks

erri120 (approved):

PR #1203, opened 2024-04-10
Polish UI for Apply Diff View

Patriot99:

### DIFF
--- a/src/NexusMods.App.UI/Resources/Language.pl.resx
+++ b/src/NexusMods.App.UI/Resources/Language.pl.resx
@@ -139,7 +139,7 @@
 		<value>Dodaj grę</value>
 	</data>
 	<data name="DownloadsView__DOWNLOADS" xml:space="preserve">
-		<value>POBIERANIE</value>
+		<value>POBRANE</value>
 	</data>
 	<data name="DownloadsView__In_Progress" xml:space="preserve">
 		<value>Pobierane</value>
--- a/src/NexusMods.App.UI/Resources/Language.pl.resx
+++ b/src/NexusMods.App.UI/Resources/Language.pl.resx
@@ -181,7 +181,7 @@
 		<value>Wstrzymaj wszystko</value>
 	</data>
 	<data name="InProgressTitleTextBlock" xml:space="preserve">
-		<value>Pobrane</value>
+		<value>Pobieralnia</value>
 	</data>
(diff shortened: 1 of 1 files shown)


src/NexusMods.App.UI/Resources/Language.pl.resx
```diff
@@ -181,7 +181,7 @@
 		<value>Wstrzymaj wszystko</value>
 	</data>
 	<data name="InProgressTitleTextBlock" xml:space="preserve">
-		<value>Pobierane</value>
+		<value>Pobrane</value>
```
Sewer56, line 184: Doesn't `Pobrane` mean `Downloaded`?

The 'accurate' translation here is probably `W Trakcie Pobierania` or rather `W Trakcie`, but I imagine that may be long or unfriendly.

src/NexusMods.App.UI/Resources/Language.pl.resx
```diff
@@ -139,10 +139,10 @@
 		<value>Dodaj grę</value>
 	</data>
 	<data name="DownloadsView__DOWNLOADS" xml:space="preserve">
-		<value>POBRANE</value>
+		<value>POBIERANIE</value>
```
Sewer56, line 142: This should probably still say `POBRANE` as `POBERIANIE` is DOWNLOAD(ING), but this string is unused currently it seems.

src/NexusMods.App.UI/Resources/Language.pl.resx
Patriot99: In the UI, it looks like a "Downloads" folder (a folder that contains downloaded files), just like in Windows Explorer, so the translation in this case is correct.

Screenshot_20240411-004817~2

Patriot99: Optionally, I can restore the previous name  `Pobierane`. This is a shorter alternative to `W trakcie pobierania`.

Patriot99: Or `Pobieralnia`. Tab where you can download files.

Sewer56: `Pobieralnia` sounds great, actually

Sewer56 (approved):

PR #1205, opened 2024-04-11
Allow launched process to spawn Consoles

### LINKED ISSUE #1135: v0.4 Issue - Launch SMAPI with console enabled on Windows
# Bug Report

## Summary
Apparently when you run smapi you are supposed to get a Console, but starting it from the App does not produce it.
We might need to use https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.useshellexecute?view=net-8.0 to allow the console to open.

## Steps to reproduce
Install SMAPI, Apply, hit Launch button.

A console and the game both open.

## Other information
The console is apparently useful and desirable to have for modding, so no need to have a setting to enable or disable it.

Al12rs:
- fixes #1135 

## Details:
- Used `Process` directly instead of `CliWrap` since CliWrap didn't expose any options to support this behavior

Open to suggestions if there are better alternatives for awaiting the Process instead of `process.Exited` event

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Games/RunGameTool.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Games/RunGameTool.cs
@@ -1,7 +1,5 @@
 using System.Diagnostics;
 using System.Globalization;
-using System.Text;
-using CliWrap;
 using Microsoft.Extensions.Logging;
 using NexusMods.Abstractions.GameLocators;
 using NexusMods.Abstractions.Games.DTO;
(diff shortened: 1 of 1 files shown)


erri120 (changes requested):
I don't think running a raw process instead of using `CliWrap` for _every_ game is the right choice. This solution will run a shell when you want to launch _any_ game. Instead, `RunGameTool` should have a property `UseShell` or similar. This way, games can opt into this behavior.

I also believe we can still use `CliWrap` and spawn a shell. Instead of launching the program directly, we launch `cmd.exe` that then launches the program. This would be similar to how we open URLs on Windows.

If we have to use process, then we should add a new method to `IProcessFactory` that takes in an unstarted `Process`. The implementation of this method would be similar to the private `RunProcessAsync` method from this PR.

Al12rs:
@erri120 Something like this? Do you want more stuff to be moved to to ProcessFactory?
Do you want me to do anything for FakeProcessFactory?

src/NexusMods.CrossPlatform/Process/ProcessFactory.cs
```diff
@@ -25,4 +25,28 @@ public ProcessFactory(ILogger<ProcessFactory> logger)
...
+            if (!process.HasExited)
+            {
+                process.Kill();
+            }
+        });
+
+        process.Start();
+
+        return tcs.Task; 
+    }
```
erri120, line 51: Missing `process.EnableRaisingEvents = true`. Without this, the `Exited` event won't ever been raised. This method also needs some `try/catch` and some logging.

erri120 (changes requested):

erri120 (approved):

PR #1216, opened 2024-04-16
Add CHANGELOG

erri120:
Preview here: https://github.com/erri120/NexusMods.App/blob/docs/changelog/CHANGELOG.md


Al12rs:
You want the changelog committed? I thought we would keep it on the releases

Al12rs:
My main issue with a committed changelog is that you need to update the code to make a revision to the changelog, that is describing what changed in the code.
So you end up with tag v0.5 and a changelog that is edited after that because we forgot something. 

I usually kept this in the Wiki section. I'm ok either way, just I find this less clean and less easy to update.

erri120:
> My main issue with a committed changelog is that you need to update the code to make a revision to the changelog, that is describing what changed in the code. So you end up with tag v0.5 and a changelog that is edited after that because we forgot something.

What's the issue with that? It's okay to edit changelogs. With #1217 we can just pull the latest changelog. It's like we're embedding the changelog and need to cut a new release if we messed up the changelog.

Al12rs:
Well you need a PR and for it to be approved to be able to edit the changelog. If it was in the Wiki section or if we use the release as ground truth then that is easier and faster to update. Community can come in and add quick notes in case there is a bug or some important thing to add.

erri120:
> Well you need a PR and for it to be approved to be able to edit the changelog.

We're doing the same thing for code and all our other documentation as well. Why should changelogs be the only thing that don't have to adhere to this standard? Changelogs are often de-valued by programmers, even less than documentation. However, both are important to users. A properly maintained changelog goes a long way.

Al12rs:
It's because I value Changelogs *more* that I say that they should be easier to edit. Because they are so important, they should be easier to access.
I would do the same thing for documentation, but we can't have the pretty rendering without committing that as far as I know.

But either way it is fine

Sewer56:
If we're manually maintaining changelogs, can't we simply update the changelog as part of each PR?

Basically in each PR, you'd also append to the changelog.

I've done that in a project before, and I don't mind doing it here.

Al12rs:
You want merge conflicts, that is how you get merge conflicts (put Archer ants reference pic here)

erri120:
> If we're manually maintaining changelogs, can't we simply update the changelog as part of each PR?
> 
> Basically in each PR, you'd also append to the changelog.
> 
> I've done that in a project before, and I don't mind doing it here.

It does tend to create merge conflicts, and it's not really necessary because we'll clean up the changelog before a release anyways (like what this PR is doing).

There are two options:

1) Every PR adds to the changelog. This will be a very short description that summarizes the change and links to the PR and any issues. This can be technical and rough, somewhat similar style to the "Other Changes" or "Fixes" section from this PR. Before we cut a release, one of us creates a PR like this to clean up the changelog, to add the "New Features" section as well as any media like images or videos showcasing the new features.

2) PRs don't update the changelog, we only update the changelog before we cut a release.

While I like the first option, it doesn't actually save us time. If we add to the changelog on every PR, we'll have to deal with merge conflicts. The "cleanup" step will kinda re-do the changelog anyway. The only benefit that the first option has over the second one is having an up-to-date changelog will all changes, even if it's not in the final form.

Al12rs:
I'd go with option 2. 
Contributors are unlikely to follow the format and merge conflicts are annoying.
We can use the GitHub auto generate changelog feature to have a starting point by creating a draft PR

CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
+# Changelog
+
+## Unreleased
+
+### New Features
+
+The highlight of this PR is the new Apply Diff View ([#1202](https://github.com/Nexus-Mods/NexusMods.App/pull/1202)). You can now preview the changes made to disk before applying the Loadout:
+
+![Screenshot of the new Apply Diff View that displays a list of files without different states like "Contents modified" or "Added" as well as file counts for directories and file/directory sizes.](./docs/changelog-assets/3ee7ede1aafade7797185cb7f9f49b2a.webp)
```
Al12rs, line 9: I would suggest using a structure like what VS Code does with their Release Notes:
image

Where there is a bullet list of the highlights and then a more in depth overview.
Then we can add further changes and fixes as a bullet list after that.

docs/changelog-assets/README.md
```diff
@@ -0,0 +1,46 @@
...
+```bash
+ffmpeg -i input.png -pix_fmt yuv420p -c:v libwebp -quality 80 -compression_level 6 output.webp
+```
+
+### Videos
+
+- Videos should be cropped to contents, don't record your entire desktop.
+
+TODO: decide on a format, probably WebM for longer videos and animated WebP for shorter videos. Definitely won't be using GIFs.
+
```
Al12rs, line 46: This is a PITA. Adding an image to the changelog could be as simple as ctrl-v on a wiki page or on the release description page. 
I'm fine with the guidelines on the contents of the image, but asking to run ffmpeg on them and check summing them makes this simply ugh.

CHANGELOG.md
erri120: VS Code pushes multiple medium to big features per release. We don't do that. In this release, we added the apply diff view and three diagnostics. A list of two items followed by those two items being explained doesn't look and feel great.

When we have a release that has multiple medium to big features, we can style the changelog for _that_ release similar to how VS Code does it.

docs/changelog-assets/README.md
erri120: I was planing on adding a script as a one-click solution but didn't want to put that in this PR.

Al12rs: I think I'm less fine now about committing the changelog to the repo. This means also committing all the images and videos. Those could instead be put in the wiki on on the release page description without issues and without growing the repo itself.

erri120: Those four images are 67 KB in size. That's 16.75 KB per image. Even if you have one thousand of those images, they'd only take up 16 MB.

halgari (approved):

PR #1222, opened 2024-04-17
fix: Add networking required traits

### LINKED ISSUE #1223: More tests should have `RequiresNetworking`
# Improvement

## Currently 

Some tests require networking, but are not annotated as such.

## We want

`[Trait("RequiresNetworking", "True")]` annotations for tests which require networking.

l0b0:
Closes  #1223.


erri120:
The `RequiresNetworking` trait isn't very descriptive, but we're using this for tests that talk to the Nexus Mods API. A better name for the trait would be "RequiresNexusModsAPIKey".

PR #1227, opened 2024-04-17
Use b2sum for coreutils 8

erri120:
Ubuntu 22.04 ships with coreutils 8...


Al12rs (approved):
I still don't agree with some of the requirements for images. It simply makes not want to add images to the changelog.

PR #1235, opened 2024-04-17
Convert ArchivedFiles over to MnemonicDB

halgari:
Converts the ArchivedFiles over to MnemonicDB.

* Added `IConnection` to NxFileStore
* Deleted ZipFileStore, we only had this for when NX was buggy, which it isn't now
* Added `AarchivedFile` and `ArchivedFileContainer` as MnemonicDB models
  * Since we have reverse and secondary indexes, we can link the ArchivedFiles to their containers
  * Containers can have the path info, the entries just have the `Hash`, `NxFileEntry` and a pointer to the container
* Rewrote the NxFileStore internals to reference the new formats

Yes, I know this code is crap:

```csharp
        var entries = from id in db.FindIndexed(hash, ArchivedFile.Hash)
            let entry = db.Get<ArchivedFile.Model>(id)
            from location in _archiveLocations
            let combined = location.Combine(entry.Container.Path)
            where combined.FileExists        
            select (combined, entry.NxFileEntry, true);
```

The abstraction here is all wrong, we're doing point lookups and then resolving them to the archives they are in, but I need to build the query support for MnemonicDB yet to really optimize that code. The above offending block is doing joining several datasets: file hashes, containers, files that exist in the `Archives` folder. All of that really can and should be done in a single query instead of via a point by point lookup. Doing set based queries on that could drastically reduce the number of lookups we need to perform, and that's next on my list for the query support in MnemonicDB. For now it works just fine.


Sewer56 (approved):
No objections here.
If we have to be a bit slower for the time being, so be it. Just make sure there's an issue for the missing functionality.

I made sure to review this early, as Perf PR #2 also touches this code, so I'll need to adjust there (resolve conflicts) to adapt for the new changes. I wanted to get that merged last week so the conflict doesn't exist, but due to unforeseen circumstances, this week has unfortunately been full of delays 😅

Issue #920, opened 2024-02-07
Implement Settings UI


Al12rs:
Needs #396 

Details are in #396 as well.

### Design

https://www.figma.com/file/rR0I4C46je8VwwqCdrpc5U/%F0%9F%93%B1%E2%9A%99%EF%B8%8F-Settings-page?type=design&node-id=155%3A4907&mode=design&t=y96yEjSE4IjI3YSh-1

**Demo**

https://github.com/Nexus-Mods/NexusMods.App/assets/118466382/f4960247-c580-4084-b8c3-2975b59979da

Does not include search in the sidebar for the first iteration. 

# Implementation Breakdown:
### Settings to bring into UI:
- Language setting: Dropdown of strings
- Telemetry enable (bool toggle)
- Diagnostic setting minimum severity (enum)

## Backend:
- [x] Make settings builder create a UI digestible interface of individual settings (Name, description, section, needs restart, default value etc.) #1239

## Frontend:
- [x] Create Settings Page and general UI layout // skip sections in the central part as well (#1246)
- [ ] Implement specific setting value UIs:
  - [x] Show name, description (markdown), needs restart and value setter (#1246) : 
    - [x] Bool toggle (#1248)
    - [x] String dropdown (#1248)
- [ ] Settings UI - Add Dialog on save press in there are changes that need user to restart

### Simple version:
- [ ] Add Dialog for when you press Save and we require a restart (we can't restart for now, ask user to do it)

### For full version:
- [ ] Design how to describe each option for multi choice settings (combobox)
- [ ] Highlight modified settings in the ui (settings that no longer have the default value)
- [ ] Add button to reset to default value in case of modified

### For later:
- [ ] Settings UI - Add Dialog for Cancel button (You have unsaved changes)
- [ ] Settings UI - Add Dialog for Cancel button in case of required restart (You have unsaved changes with restart)
- [ ] Add sections
- [ ] Filter search
- [ ] Implement automatic restart of the App
    - [ ] Restart with our multi window approach might be a pain to implement
    - [ ] Add confirm dialog before restarting

Al12rs:
Copied from the issue body:

### Missing from design:
We need Design for displaying settings with a Warning, such as Experimental or Not Recommended.

This came up in the context of offering an option to not Backup the game files when managing the game for the first time, which exposes the users to potential issues if the game updates, since we don't have the backups of the old files if the user wants to rollback.

erri120:
[...]

PR #1239, opened 2024-04-18
Get all properties from `ISettingsManager` that should be exposed to the UI

erri120:
Required for #920.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Settings/APropertyValueContainer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Settings/APropertyValueContainer.cs
@@ -9,27 +9,36 @@ namespace NexusMods.Abstractions.Settings;
 [PublicAPI]
 public abstract class APropertyValueContainer<T> : AbstractNotifyPropertyChanged
 {
-    private T _previousValue;
     private T _currentValue;
     private bool _hasChanged;
+    private bool _isDefault;
 
     private IEqualityComparer<T> _equalityComparer;
 
     /// <summary>
     /// Constructor.
     /// </summary>
-    protected APropertyValueContainer(T value, IEqualityComparer<T>? equalityComparer = null)
+    protected APropertyValueContainer(T value, T defaultValue, IEqualityComparer<T>? equalityComparer = null)
     {
-        _previousValue = value;
+        PreviousValue = value;
+        DefaultValue = defaultValue;
+
         _currentValue = value;
-        _hasChanged = false;
         _equalityComparer = equalityComparer ?? EqualityComparer<T>.Default;
+
+        _hasChanged = false;
+        _isDefault = _equalityComparer.Equals(value, defaultValue);
     }
+
+    /// <summary>
+    /// Gets the default value.
+    /// </summary>
+    public T DefaultValue { get; }
 
     /// <summary>
     /// Gets the previous value.
     /// </summary>
-    public T PreviousValue => _previousValue;
+    public T PreviousValue { get; }
 
     /// <summary>
     /// Gets or sets the current value.
@@ -49,12 +58,22 @@ public abstract class APropertyValueContainer<T> : AbstractNotifyPropertyChanged
         private set => SetAndRaise(ref _hasChanged, value);
     }
 
+    /// <summary>
+    /// Gets whether <see cref="CurrentValue"/> equals <see cref="DefaultValue"/>.
+    /// </summary>
+    public bool IsDefault
+    {
+        get => _isDefault;
+        private set => SetAndRaise(ref _isDefault, value);
+    }
+
     /// <inheritdoc/>
     protected override void OnPropertyChanged(string? propertyName = null)
     {
         if (propertyName == nameof(CurrentValue))
         {
             HasChanged = _equalityComparer.Equals(PreviousValue, CurrentValue);
+            IsDefault = _equalityComparer.Equals(CurrentValue, DefaultValue);
         }
 
         base.OnPropertyChanged(propertyName);
--- a/src/Abstractions/NexusMods.Abstractions.Settings/APropertyValueContainer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Settings/APropertyValueContainer.cs
@@ -1,0 +1,62 @@
+using DynamicData.Binding;
+using JetBrains.Annotations;
+
+namespace NexusMods.Abstractions.Settings;
+
+/// <summary>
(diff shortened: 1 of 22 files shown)


src/Abstractions/NexusMods.Abstractions.Settings/SettingsBuilder/IPropertyUIBuilder.cs
```diff
@@ -10,12 +10,24 @@ public interface IPropertyUIBuilder<TSettings, TProperty>
     where TSettings : class, ISettings, new()
 {
     /// <summary>
-    /// Sets the display name of the property.
+    /// Ads the property to a section.
```
Aragas, line 13: Adds* 😁

src/Abstractions/NexusMods.Abstractions.Settings/ISettingsPropertyUIDescriptor.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    public bool RequiresRestart { get; }
+
+    /// <summary>
+    /// Gets the optional custom restart message.
+    /// </summary>
+    /// <remarks>
+    /// If this is <c>null</c>, use a default message.
+    /// </remarks>
+    public string? RestartMessage { get; }
```
Al12rs, line 37: I had not thought of us needing this, do you have a use case in mind?

src/Abstractions/NexusMods.Abstractions.Settings/APropertyValueContainer.cs
```diff
@@ -0,0 +1,62 @@
...
+/// Helper abstract class for <see cref="SettingsPropertyValueContainer"/>.
+/// </summary>
+[PublicAPI]
+public abstract class APropertyValueContainer<T> : AbstractNotifyPropertyChanged
+{
+    private T _previousValue;
+    private T _currentValue;
+    private bool _hasChanged;
+
+    private IEqualityComparer<T> _equalityComparer;
```
Al12rs, line 16: What about the default value?
If it is not contained in the ValueContainer then the UI will not have access to it. 
I think we do want to know whether the current value is different from the default and have the ability to revert to default.

src/Abstractions/NexusMods.Abstractions.Settings/SettingsBuilder/IPropertyUIBuilder.cs
```diff
@@ -29,28 +41,34 @@ public interface IWithDescriptionStep
         /// <remarks>
         /// This property allows for Markdown.
         /// </remarks>
-        IOptionalStep WithDescription(string description);
+        IConfigureValueContainerStep WithDescription(string description);
+    }
+
+    public interface IConfigureValueContainerStep
+    {
+        IRequiresRestartStep UseBooleanContainer();
```
Al12rs, line 49: Does this mean that this can be called on any value type? Even if it isn't bool?

src/Abstractions/NexusMods.Abstractions.Settings/ISettingsPropertyUIDescriptor.cs
Al12rs: Do we need one? Any case were the generic one would not suffice?

src/Abstractions/NexusMods.Abstractions.Settings/SettingsBuilder/IPropertyUIBuilder.cs
erri120: Sadly yes. You can't do `where TProperty : bool`, and the only way to have this check at compile time is by using an extension method:

```
public static class Extensions
{
    public static IPropertyUIBuilder<TSettings, bool>.IRequiresRestartStep UseBooleanContainer<TSettings>(
        this IPropertyUIBuilder<TSettings, bool>.IConfigureValueContainerStep step
    ) where TSettings : class, ISettings, new()
    {
        // ...
    }
}
```

However, this doesn't really work because the extension method has to have a body and has to return an actual value while the interface puts the method implementation requirement on the actual implementation, which isn't public.

src/Abstractions/NexusMods.Abstractions.Settings/APropertyValueContainer.cs
erri120: Right, forgot to add that.

src/NexusMods.Settings/SettingsUIBuilder.cs
```diff
@@ -0,0 +1,224 @@
...
+        _requiresRestart = true;
+        _restartMessage = null;
+        return this;
+    }
+
+    public IPropertyUIBuilder<TSettings, TProperty>.IRequiresRestartStep UseBooleanContainer()
+    {
+        _factory = new BooleanContainerFactory();
+        return this;
+    }
```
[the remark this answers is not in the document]
erri120: It's possible using extension methods, but it requires moving code around and making it even more complicated.

src/Abstractions/NexusMods.Abstractions.Settings/ISettingsPropertyUIDescriptor.cs
erri120: I don't know? I'm sure we can think of something. Doesn't really matter if it goes unused or not.

src/Abstractions/NexusMods.Abstractions.Settings/SingleValueMultipleChoiceContainer.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace NexusMods.Abstractions.Settings;
+
+/// <summary>
+/// Container for a singe value with multiple choices.
+/// </summary>
+[PublicAPI]
+public sealed class SingleValueMultipleChoiceContainer : APropertyValueContainer<object>
+{
+    private readonly object[] _allowedValues;
+    private readonly Func<object, string> _valueToTranslation;
```
Al12rs, line 12: Maybe rename `_valueToTranslation` to `_valueToString` or something more generic like that

src/Abstractions/NexusMods.Abstractions.Settings/ISettingsPropertyUIDescriptor.cs
Al12rs: 👍 I didn't plan to add such functionality in the UI for now

Al12rs (approved):

PR #1246, opened 2024-04-22
Main UI layouts for Settings page and setting entry

Al12rs:
- Added `SettingsView` page
- Added `SettingEntryView`: this is a reusable view that shows the name, description and restart requirements of any setting. It contains a ViewModelViewHost for the actual interaction control (e.g. toggleSwitch, ComboBox)
- Different implementations of `ISettingEntryViewModel` can provide different InteractionControls and handle value updates/validations etc.
- Downside is that we need a wrapper for simple controls like ToggleSwitch, just so we can swap them out using different VMs.


erri120 (approved):

erri120:
Do you also wanna do a combo-box view?

Al12rs:
> Do you also wanna do a combo-box view, since that's also a UI element we need immediately?

Yes, but as a separate PR. I'll pull in your stuff and use that to make the Bool and combobox VMs

PR #1252, opened 2024-04-23
Move FileHashCache to MnemonicDB

halgari:
* Assigns partitionIds to the various entities
* Moves FileHashCache over to MnemonicDB
* Currently we're not storing the name of the file, just the xxHash64 I don't think this is a major risk, but I'm open to arguments otherwise. 

~~Merge: #1251 first as this is based on that work~~


src/NexusMods.DataModel/FileHashCache.cs
```diff
@@ -34,44 +30,46 @@ public class FileHashCache : IFileHashCache
...
     /// <inheritdoc/>
-    public bool TryGetCached(AbsolutePath path, out FileHashCacheEntry entry)
+    public bool TryGetCached(AbsolutePath path, out HashCacheEntry.Model entry)
     {
-        var normalized = path.ToString();
-        Span<byte> span = stackalloc byte[Encoding.UTF8.GetByteCount(normalized)];
-        Encoding.UTF8.GetBytes(normalized, span);
-        var found = _store.GetRaw(IId.FromSpan(EntityCategory.FileHashes, span));
-        if (found != null && found is not { Length: 0 })
+        var nameHash = path.ToString().XxHash64AsUtf8();
```
Sewer56, line 50: Copy this class out of my branch and use it:

https://github.com/Nexus-Mods/NexusMods.App/blob/93f53060927f1a693b7c0ee7da59cd6972791409/src/Extensions/NexusMods.Extensions.Hashing/StringExtensions.cs#L6-L28

It should be a bit faster, since there's no string conversion done. Might speed up things a bit since this can be hit on `IndexFoldersAsync`, a hot path.

[Don't forget about the corresponding write!]

Sewer56 (approved):

Sewer56:
Good to merge, but there's a small improvement. (See comment above)

PR #1259, opened 2024-04-23
Added: Vanilla/Original Game State Reuse

### LINKED ISSUE #1197: Loadouts - Store "Vanilla" game state separately to be able to roll back to it or create new profiles
For details see:
- #1166 

## Summary:
We need to be able to revert the game to a vanilla state, to do this we need to know what the vanilla state is.
We need this for:
- Removing a loadout (uninstalling all the mods) #399 
- Creating multiple loadout without ingesting previous loadouts #982 
- Managing game updates and allowing users to switch between them

## Implementation:
- Use the state of the game when the game is managed the first time as the base game state.
- Store this state in the `DiskStateRegistry`
- Provide an API to perform a hard reset of the game folders to a pre-cached `DiskStateRegistry` entry. 

## Future work
- Don't worry about how this data will be used, we will leverage this feature in the future, for now we just need the `Hard Reset` feature

Sewer56:
fixes #1197

This PR adds very basic support for storing and reusing DiskStates across game manage operations.

After some familiarizing myself with MnemonicDB, and a prototype or two (in earlier commits); I settled on a very simple solution that will make it easy to clean up afterwards.

https://github.com/Nexus-Mods/NexusMods.App/assets/6697380/e075ad90-55a4-4ece-85c7-f2551749fa7f

---

This is OK to review, I drafted it because I'm still adding a test.


halgari (approved):
Solid work here, I wasn't sure about having a separate set of attributes for this, but they really are separate entities and we never want to use them together, so by all means, separate them. :)

PR #1266, opened 2024-04-24
Add NavigationControl

### LINKED ISSUE #942: Create custom navigation control
To make the user experience more consistent, we should develop a custom navigation control that handles navigation. Consumers of that control would only need to specify _what_ they want to do, while the navigation control handles _how_ that's going to happen.

The control should also have a context menu that may contain the following items:

- Open in new tab
- Open in new panel
- Open in new window

Avalonia already provides a [ContextMenu](https://docs.avaloniaui.net/docs/reference/controls/contextmenu) control.

erri120:
Resolves #942.

https://github.com/Nexus-Mods/NexusMods.App/assets/16577545/6f80bac5-a0e5-4ea0-942e-572ac34d1a66

### DIFF
--- a/src/NexusMods.App.UI/Controls/Navigation/NavigationControl.cs
+++ b/src/NexusMods.App.UI/Controls/Navigation/NavigationControl.cs
@@ -14,26 +14,32 @@ public class NavigationControl : Button
     public static readonly StyledProperty<ReactiveCommand<NavigationInput, Unit>?> NavigateCommandProperty =
         AvaloniaProperty.Register<NavigationControl, ReactiveCommand<NavigationInput, Unit>?>(nameof(NavigationCommand));
 
-    public ReactiveCommand<NavigationInput, Unit>? NavigationCommand
+    public ReactiveCommand<NavigationInformation, Unit>? NavigationCommand
     {
         get
         {
             var value = GetValue(CommandProperty);
-            return value as ReactiveCommand<NavigationInput, Unit>;
+            return value as ReactiveCommand<NavigationInformation, Unit>;
         }
         set => SetValue(CommandProperty, value);
     }
 
-    private NavigationInput NavigationInput
+    private NavigationInformation NavigationInformation
     {
+        get => (NavigationInformation)CommandParameter!;
         set => CommandParameter = value;
     }
 
+    private readonly ReactiveCommand<OpenPageBehaviorType, Unit> _contextMenuCommand;
+
     public NavigationControl()
     {
-        // TODO: design this context menu and decide what goes in it
-        // NOTE: we'll also want consumers to disable/enable certain items
-        // TODO: implement click handlers for these items
+        _contextMenuCommand = ReactiveCommand.Create<OpenPageBehaviorType>(openPageBehaviorType =>
+        {
+            if (NavigationCommand is null) return;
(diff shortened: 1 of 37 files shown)


Al12rs:
So this doesn't handle buttons that are inside a page right? e.g. DiagnisticList or LoadoutGrid

src/NexusMods.App.UI/WorkspaceSystem/WorkspaceController/IWorkspaceController.cs
```diff
@@ -112,6 +113,11 @@ public interface IWorkspaceController
     /// </summary>
     public void SetIcon(IconValue? icon, WorkspaceId workspaceId, PanelId panelId, PanelTabId tabId);
 
+    public OpenPageBehavior GetOpenPageBehavior(
+        PageData requestedPage,
+        NavigationInformation navigationInformation,
+        Optional<PageIdBundle> currentPage);
```
Al12rs, line 119: I would set a default here for currentPage:
```csharp
public OpenPageBehavior GetOpenPageBehavior(
        PageData requestedPage,
        NavigationInformation navigationInformation,
        Optional<PageIdBundle> currentPage = Optional<PageIdBundle>.None);
```
So that the users outside a page don't have to set it every time.

Al12rs:
I haven't finished review of the new button code. I might sneak that in tomorrow morning

src/NexusMods.App.UI/WorkspaceSystem/WorkspaceController/IWorkspaceController.cs
erri120: No, I'd agree if we mostly call this method from outside the workspace system, but that's more an exception than the rule.

Al12rs (changes requested):
Using "Open in new panel" context menu while you already have 4 panels throws an exception
image

Al12rs:
Exception should be fixed, the rest looks good. We are of course missing all the navigation from inside Pages. I'm guessing that will be in another pr

Al12rs:
Ah Settings button navigation in the top bar is missing

erri120:
Everything that opens pages should now be using the new navigation control.

src/NexusMods.App.UI/Controls/Navigation/NavigationControl.cs
```diff
@@ -0,0 +1,118 @@
...
+                    Header = "Open in new tab",
+                    Command = _contextMenuCommand,
+                    CommandParameter = OpenPageBehaviorType.NewTab,
+                },
+                new MenuItem
+                {
+                    Header = "Open in new panel",
+                    Command = _contextMenuCommand,
+                    CommandParameter = OpenPageBehaviorType.NewPanel,
+                },
```
Al12rs, line 59: I think we can should make this option disabled if you have MaxPanels already

erri120: That would be the most user-friendly, but I'd take too long to add. We can do this later, this PR is already big enough.

Al12rs: Ok, please open an issue or make a note in an open one for this

Al12rs (approved):

PR #1271, opened 2024-04-25
Disable multi-window support

### LINKED ISSUE #1267: Disable multi-window
Please can multi-window be disabled. 

Only one window of the app can be open at a time. 

Multi-window will be added back in later.

erri120:
Resolves #1267.

### DIFF
--- a/src/NexusMods.App/Startup.cs
+++ b/src/NexusMods.App/Startup.cs
@@ -24,7 +24,7 @@ public class Startup
     private static bool _hasBeenSetup = false;
     private static IServiceProvider _provider = null!;
     private static ILogger<Startup> _logger = null!;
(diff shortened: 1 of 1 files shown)


halgari (approved):

Sewer56 (commented):
This is technically not thread safe, as the increment is done elsewhere but I doubt we'd be likely to run into a situation where this would matter.

(I was a bit confused on 'how does it work' for a second, till I noticed it was incremented elsewhere).
(Normally you'd make a threadsafe lock like this with `Interlocked.CompareExchange` (`cmpxchg` in x86 ASM).)

erri120:
True, this could've been just one `CompareExchange`.

Issue #399, opened 2023-07-05
Loadouts - As a user, I want to remove a loadout


Greg-Nexus:
# User story

**As a** user

**I want to** remove a loadout

**So that** it no longer appears on the App

## Requirements
Users needs to be able to remove loadouts, for when they no longer want to continue managing/using them.

The Trashcan icon on the "Manage Game" panel should set the loadout to deleted (Tombstone it for now). 

We need to put the files back to the way they were before we managed the game (delete all the mods from disk). This will likely require #401 to be completed first.

Issue #1204, opened 2024-04-11
Loadouts - Tech debt: Cleanup how loadout revisions are represented in code


Al12rs:
# Tech debt

## Currently 
We have:
- `Loadout`
- `LoadoutId`
- `IId` or `DataStoreId`
- `LoadoutMarker`

Apart from `LoadoutId` these don't have an intuitive mapping to actual loadouts as the users interact with them.
- `LoadooutId` identifies one loadout as users are familiar with them
- `Loadout` is an object representing a Loadout, but only at a specific moment in time. What we have been referring to as `LoadoutRevision` in speech and other places of the code. 
- `DataStoreId` is the Id identifying a particular LoadoutRevision
- `LoadoutMarker`, had to look at the code to check what this actually did, and it is an object that simplifies getting the latest `Loadout` of a certain `LoaodutId`. Basically it is supposed to offer convenience due to the revisions always changing.

These names are quite confusing and make contributing to the project harder for third parties, as well as confuse current devs.

## We want
Better names and potentially even reconsider how these are structured. 
We already need to touch this code for integrating our new DataStore MnemonicDB (#1173 ) so it could be an opportunity to make our code easier to understand.

## Suggested solutions
- `Loadout` -> `LoadoutRevision`
- `DataStoreId` -> `LoadoutRevisionId`
- `LoadoutMarker` -> `Loadout`: I would consider fleshing this out a bit more and using this more, but open to other suggestions.
- `LoadoutId` -> no change, this is clear enough.

There are other concepts like LoadoutRoots that might also need looking at.

Epic:
- https://github.com/Nexus-Mods/NexusMods.App/issues/397

LukeNexusMods:
Is this something that would make contributing to the project more accessible? Are there any performance factors that might be affected by this?

Al12rs:
No performance factors. 
Yes this would make our code easier to understand and thus make it more accessible to contributors.

Sewer56:
As a random reference, even I made an error last Thursday.

I assumed that under the current system a certain value we stored was like a `git commit hash` (basically random unique number), when in actuality it was a unique identifier of an item in our database. Which led me to write code suboptimally.

Thankfully it was caught in code review.

Sewer56:
Might be worth visiting as part of the move to MneumonicDB (Tim's Database Library with Event Sourcing).
Since at that point the code would be touched.

Al12rs:
Handled with move to MnemonicDb

PR #1279, opened 2024-04-29
Added: Remove Loadouts & Unmanage Games

Sewer56:
fixes https://github.com/Nexus-Mods/NexusMods.App/issues/399

This PR adds the ability to remove loadouts and cleanup after a game.

Important Notes:

- Removing an Active Loadout will Reset the Game to Its Original Disk State.
- Unmanaging the game will Reset the Game to its Original Disk State
    - And remove the original state, so a new original state is created next time game is managed.
    - This logic will also trigger on removal of last loadout, not just explicit 'unmanage'.
- Existing Code was slightly tweaked to handle loadout removal more gracefully.
    - (Previously, UI did not like it)
- Added a CLI verb for deleting loadout(s).
- New functionality has accompanying simple tests.
- Hitting the Trash Can in Current UI Unmanages a Game
    -  Intended as a dev tool, until this part of UI is redesigned
- Updates some previous code to use newer MnemonicDB APIs


src/NexusMods.DataModel/Loadouts/LoadoutRegistry.cs
```diff
@@ -456,4 +458,65 @@ public bool Contains(LoadoutId loadoutId)
...
+            }
+
+            var value = currentVersion.Value;
+            if (value == null)
+                break;
+
+            currentVersion = value.PreviousVersion;
+        }
+
+        return isActiveLoadout;
```
erri120, line 520: Shouldn't this be just two lines?

```csharp
var loadout = GetLoadout(lastAppliedId);
return loadout.LoadoutId = id;
```

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -123,14 +130,14 @@ public ValueTask<FileTree> FlattenedLoadoutToFileTree(FlattenedLoadout flattened
...
-    public async Task<DiskStateTree> FileTreeToDisk(FileTree fileTree, Loadout loadout, FlattenedLoadout flattenedLoadout, DiskStateTree prevState, GameInstallation installation, bool skipIngest = false)
+    public async Task<DiskStateTree> FileTreeToDisk(FileTree fileTree, Loadout? loadout, FlattenedLoadout flattenedLoadout, DiskStateTree prevState, GameInstallation installation, bool skipIngest = false)
     {
         // Return the new tree
         return await FileTreeToDiskImpl(fileTree, loadout, flattenedLoadout, prevState, installation, true);
     }
     
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    internal async Task<DiskStateTree> FileTreeToDiskImpl(FileTree fileTree, Loadout loadout, FlattenedLoadout flattenedLoadout, DiskStateTree prevState, GameInstallation installation, bool fixFileMode, bool skipIngest = false)
+    internal async Task<DiskStateTree> FileTreeToDiskImpl(FileTree fileTree, Loadout? loadout, FlattenedLoadout flattenedLoadout, DiskStateTree prevState, GameInstallation installation, bool fixFileMode, bool skipIngest = false)
```
erri120, line 140: Why did you make loadout nullable?

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -817,8 +802,128 @@ public virtual async Task<Loadout> Manage(GameInstallation installation, string?
...
+                In the future, we may make a setting to change the behaviour,
+                if for example the user wants it to revert to last applied loadout
+                that isn't the one being deleted.
+            */
+
+            await ResetToInitialState(installation);
+        }
+
+        // Delete the loadout from the LoadoutRegistry
+        _loadoutRegistry.Delete(id);
```
erri120, line 882: Useless comment.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -817,8 +802,128 @@ public virtual async Task<Loadout> Manage(GameInstallation installation, string?
...
+        
+        // If this is the last loadout, remove persisted disk state.
+        if (clearInitialState)
+            await _diskStateRegistry.ClearInitialState(installation);
```
erri120, line 886: Code doesn't match the comment. Remove or change either one.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -817,8 +802,128 @@ public virtual async Task<Loadout> Manage(GameInstallation installation, string?
...
+    }
+
+    private async Task ResetToInitialState(GameInstallation installation)
+    {
+        // Apply initial disk state.
+        var (isCached, initialState) = await GetOrCreateInitialDiskState(installation);
```
erri120, line 892: Useless comment.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -817,8 +802,128 @@ public virtual async Task<Loadout> Manage(GameInstallation installation, string?
...
+            with the original file tree here.
+        */
+
+        // Create a temporary loadout instance.
+        // Note: We recreate 'Game Files' mod because 
+        var gameFilesMod = CreateGameFilesMod(initialState);
+        var fileTree = CreateFileTreeFromMod(gameFilesMod);
+
+        // Apply temporary loadout.
+        // The loadout is only used as a parameter for generated files.
```
erri120, line 920: Useless and incomplete comments.

erri120 (changes requested):

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
Sewer56: `ResetToInitialState` calls the API without a Loadout when reverting the game folder to its original state to help with code reuse.

The Loadout here is used as a parameter to the APIs used for creating `GeneratedFile`(s) in `FileTreeToDiskImpl.

In the case of reverting the folder to stock, there (naturally) can't be generated files, as the original disk state is all static files. So I made the parameter nullable and added a defensive throw if there are generated files and loadout is null down below.

src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
```diff
@@ -79,6 +79,12 @@ public class MyGamesViewModel : APageViewModel<IMyGamesViewModel>, IMyGamesViewM
                             await Task.Run(async () => await ManageGame(install));
                             vm.State = GameWidgetState.ManagedGame;
                         });
+                        vm.RemoveAllLoadoutsCommand = ReactiveCommand.CreateFromTask(async () => 
+                        {
+                            vm.State = GameWidgetState.RemovingGame;
+                            await RemoveAllLoadouts(install);
+                            vm.State = GameWidgetState.ManagedGame;
+                        });
```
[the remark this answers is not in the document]
Sewer56: Fixed by offloading to threadpool.

The latter was already present. Change to `vm.State` actually hides the control (`RemovingGame` hides all buttons), so the user can't start the operation twice.

src/NexusMods.DataModel/Loadouts/LoadoutRegistry.cs
Sewer56: I renamed this to `IsApplied` in retrospect, because the original `IsActive` may be misleading depending on context.

In any case, the situation here is that the last applied loadout is not always the latest revision of a loadout.
For instance, suppose you make a change to the currently applied loadout.

The latest revision is updated, but until you actually apply, the last applied ID is of a previous revision. This is why we need to check all revisions just in case.

erri120: > last applied loadout is not always the latest revision of a loadout

But we don't care about that, at least I don't think we do. By definition, the last **applied** loadout was applied at some point. It's not a method that checks if the last applied loadout was applied, that's kinda weird. It's a method that checks if the last applied revision was part of the given loadout. So it should be just checking whether the loadout IDs match. No need to walk all revisions.

Al12rs: erri is right, from a loadout revision Id you can get the LoadoutId with GetLoadout(revisionId).LoadoutId. 
Then you just compare the LoadoutId with the passed one. No need to check rev by rev

Sewer56: Yeah pretty much.

The problem here is a bit interesting.
Basically, we don't store the actual `LoadoutId` (unique ID for loadout) of the last applied loadout anywhere. Not in Disk State, not as Last Applied Loadout. etc. Instead we store the `LoadoutRevision`, which is the unique ID for revision of a loadout. 

And there's no mapping anywhere of `LoadoutRevision` to `LoadoutId` that I'm aware of.

Should I maybe rename it to `IsAnyRevisionApplied`?
Might be more descriptive.

Al12rs: The mapping is there and is this: you have a `RevisionId`
`LoadoutRegistry.GetLoadout(RevisionId)` -> `Loadout`
`Loadout.LoadoutId` - > `LoadoutId`

Al12rs: Yes our names are a mess: https://github.com/Nexus-Mods/NexusMods.App/issues/1204

Sewer56: I posted at the same time as you :sweat_smile: .
That's why I edited my post with `wait a moment`.

Seems you're right. I was always under the impression that a revision was a separate item, and not just an ID to a previous loadout; hence the confusion.

Let me patch this up.

Al12rs: No worries, I'm not surprised you were confused, it really isn't intuitive.

erri120 (approved):

PR #1286, opened 2024-04-30
Hide game files

### LINKED ISSUE #1195: Loadouts - Make "GameFilesCategory" mods non user interactable
When turning off your game files and ingesting it prevents you from being able to launch the game again and then creates another set of game files. 

To stop this from happening please can we remove the switch from Game Files so that they cannot be disabled from the loadouts page?

- #397

erri120:
Resolves #1195.


Al12rs (approved):

Al12rs:
We should probably sort the settings at some point

PR #1294, opened 2024-05-01
Add Marker Loadouts & Minor Improvements

Sewer56:
This PR is an iterative improvement over #1279

----------

https://github.com/Nexus-Mods/NexusMods.App/assets/6697380/bd329799-0a9d-479f-aa73-886883c29213

This PR introduces the concept of "Marker Loadouts" and includes some small edge case handling around removing loadouts.

## Marker Loadouts

A 'marker loadout' is an invisible loadout that contains only the original game state.
It is invisible to the UI, and loadout listing in CLI.

The purpose of it is to enforce the 'integrity'/'internal consistency' of the data store, such that
elements such as 'last applied loadout' are always valid, even if the original last
applied loadout was just deleted.

- Added support for creating and managing marker loadouts.
- A marker loadout is created from the original game state when the last applied loadout for a game is deleted.
- Marker loadouts are hidden and should not be shown in any user-facing elements.
- Marker loadouts are deleted when a non-marker loadout is applied.
- Updated the `Loadout` record to include an `IsMarkerLoadout` property.

Accordingly, CLI and UI were updated to filter out marker loadouts from visible elements.

## UI Improvements

Now if you delete a loadout via the CLI, and the UI is in the workspace of that loadout, the UI will automatically navigate away from that loadout.

- Modified `LoadoutLeftMenuViewModel` to navigate to the "Home" workspace when the currently active loadout is deleted.

If on start, we find any of the existing workspaces are no longer valid due to loadout removal; we will remove them to avoid having any errors down the road.

- Implemented the `SanitizeWindowData` method in `MainWindowViewModel` to clean up invalid workspaces and handle scenarios where the saved window state becomes invalid.

This is handy for example if user removes a loadout purely from CLI, or if the app were to crash during loadout removal.

## Other Changes
- Updated the `IWindowManager` to support a callback for sanitizing loaded window data.
- Fixed a typo in `remove-loadout` command description.
- Fixed a bug where hitting 'View' in MyGames may take you to wrong/invalid loadout after a loadout delete.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -584,6 +584,10 @@ public class ALoadoutSynchronizer : IStandardizedLoadoutSynchronizer
     /// <returns></returns>
     public virtual async Task<DiskStateTree> Apply(Loadout loadout, bool forceSkipIngest = false)
     {
+        // Note(Sewer) If the last loadout was a marker loadout, we need to guard
+        // in the case the user has modified the game folder since.
+        forceSkipIngest = forceSkipIngest || IsLastLoadoutAMarkerLoadout(loadout.Installation);
+
         var flattened = await LoadoutToFlattenedLoadout(loadout);
         var fileTree = await FlattenedLoadoutToFileTree(flattened, loadout);
         var prevState = _diskStateRegistry.GetState(loadout.Installation)!;
@@ -970,8 +974,21 @@ public class ALoadoutSynchronizer : IStandardizedLoadoutSynchronizer
     }
     #endregion
 
-
     #region Misc Helper Functions
+    /// <summary>
+    /// Checks if the last applied loadout is a 'marker' loadout.
+    /// </summary>
+    /// <param name="installation">The game installation to check.</param>
+    /// <returns>True if the last applied loadout is a marker loadout.</returns>
+    private bool IsLastLoadoutAMarkerLoadout(GameInstallation installation)
+    {
+        var lastId = _diskStateRegistry.GetLastAppliedLoadout(installation);
+        if (lastId == null)
+            return false;
+
+        return _loadoutRegistry.GetLoadout(lastId) is { IsMarkerLoadout: true };
+    }
+    
     /// <summary>
     /// Removes the first found 'marker' loadout from the data store.
     /// </summary>
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -926,7 +926,7 @@ public class ALoadoutSynchronizer : IStandardizedLoadoutSynchronizer
 
         var loadout = _loadoutRegistry.Alter(loadoutId, "Marker Loadout", loadout => loadout with
         {
-            Name = "Marker Loadout",
+            Name = "Temporary Marker Loadout",
             Installation = installation,
             Mods = loadout.Mods.With(gameFiles.Id, gameFiles),
             IsMarkerLoadout = true,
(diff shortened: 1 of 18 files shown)


src/NexusMods.App.UI/LeftMenu/Loadout/LoadoutLeftMenuViewModel.cs
```diff
@@ -113,6 +118,23 @@ public class LoadoutLeftMenuViewModel : AViewModel<ILoadoutLeftMenuViewModel>, I
...
+                    {
+                        workspaceController.ChangeOrCreateWorkspaceByContext<HomeContext>(() => new PageData
+                        {
+                            FactoryId = MyGamesPageFactory.StaticId,
+                            Context = new MyGamesPageContext(),
+                        });
+                    }
+                }, false)
+                .SubscribeWithErrorLogging()
+                .DisposeWith(disposable);
```
Al12rs, line 137: I think this should be on the spine rather than on the left menu

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -585,12 +584,20 @@ public virtual Loadout MergeLoadouts(Loadout loadoutA, Loadout loadoutB)
     /// <returns></returns>
     public virtual async Task<DiskStateTree> Apply(Loadout loadout, bool forceSkipIngest = false)
     {
+        // Note(Sewer) If the last loadout was a marker loadout, we need to guard
+        // in the case the user has modified the game folder since.
```
Al12rs, line 588: Reword this a bit, 
If the last loadout was a marker loadout, we we need to skip ingest and ignore changes, since marker loadouts should not be changed.

Al12rs (changes requested):
Two items above

src/NexusMods.App.UI/LeftMenu/Loadout/LoadoutLeftMenuViewModel.cs
Sewer56: Good point.
I had it in left menu due to how an earlier iteration works, but current code is more fitting for spine.

erri120:
I'm gonna review this code tomorrow.

erri120 (changes requested):
I've pushed changes for a better approach to fix workspace contexts with invalid data. Here's what you should change:

I don't agree with this approach. If we ever have to add another "type" of Loadout that is not a "normal" or a "marker" Loadout, everything you've done for this special property has to be repeated.

I recently had a similar situation and found that enums and flags are more appropriate here.

Create an enum called `LoadoutType`, or similar, that has two members: `UserLoadout` for Loadouts created by the user, and `MarkerLoadout`.

Turn the property `IsMarkerLoadout` into a member function that checks if the Type equals `MarkerLoadout`.

Also add a member function called `IsVisible`, or similar, that returns true if the Type equals `UserLoadout`.

With this approach, we don't have to change every single piece of code that displays Loadouts if we ever added a new type of Loadout. This can also be further expanded using flags if we ever need them (eg, a `IsVisible` or `IsImmutable` flag).

erri120:
@Al12rs you should review my commit as well

Al12rs:
Yeah commit looks great!

src/NexusMods.App.UI/WorkspaceSystem/Context/LoadoutContext.cs
```diff
@@ -11,6 +11,18 @@ public record LoadoutContext : IWorkspaceContext
...
+            
+            In App, we create a loadout workspace for each loadout.
+
+            It may be possible that this loadout no longer exists, as a result
+            of removal from the CLI (UI not running) or a crash during removal.
+            [i.e. in cases we may have not run explicit workspace delete]
+            
+            In such cases, we may need to discard invalid loadouts, which we
+            do here.
+        */
```
erri120, line 25: Remove useless comment.

Al12rs (approved):
I didn't test again, but looks fine to me

erri120 (approved):

PR #1295, opened 2024-05-02
Add slim settings manager host

### LINKED ISSUE #1243: Add JsonStore and slim startup settings manager
Some settings need to be available when the application starts, but you can't get settings without DI. As such, we want a JSON based storage backend and a slim settings manager that gets created on startup and injects the current values into whatever service needs it before DI exists.

erri120:
Resolves #1243.

- Adds `JsonStorageBackend` for saving configs to JSON files
- Allows telemetry and logging settings to be loaded before DI is available
- Adds logging settings to the UI


Al12rs (approved):
I tried it out, code seems ok to me but I'm not a DI expert

PR #1302, opened 2024-05-03
Convert Loadouts to MnemonicDB

halgari:
What this PR does:

* Converts *all* of the Loadout/Mod/File trees into MnemonicDB attributes
* Files now point to the Loadout they are in, as well as the mod. We have a graph database, so lets use it
* Introduces `IRepository<Model>` which is a common entry point for a given model in the system. For example if you want to get all updates to all loadouts, or get all loadouts, or get all updates to a specific loadout, use DI to pull in `IRepository<Loadout.Model`.
* Deletes are reified in this release, it was much easier to add this functionality as part of this work rather than trying to port to the old broken way of handling deletes. 
* Metadata for games installed on the machine are auto-inserted into MnemonicDB. This means where we want to save a `GameInstallation` we no longer have to save a tuple of game name/store/version. Instead we can store a DB Id that links to this metadata in MnemonicDB. This is both a performance and a simplicity optimization. 
  * To do this I introduced the IGameRegistry which performs these inserts on startup and allows for conversion from `GameInstallation` to `EntityId` and vice-versa
* ModMarker/LoadoutMarker are gone, as they don't make sense in this data model
  * ModId and LoadoutId exist, but these are just wrappers around EntityId
  * Entities change over time (but have history tracked in the database), so ModId/LoadoutId/EntityId represent a single loadout/mod/entity *over time*. You can rewind the database and get the state of the entity via it id.
  * When we do need a specific EntityId and Time pair we combine it with a database TxId, those two then represent an Entity *as of a specific point in time*
* Most of the UI now subscribes to entities over time. Places in the code that say `_conn.Revisions(modId) => IObservable<Mod.Model>` should be read as "The mod's data model as it changes over time"
* We can now store data as a graph and index any attribute we wish, so where it makes sense I've linked attributes from separate entities. For example, if a file has a hash that represents a file in the IFileStore, we can link via the Hash into the File Store's data structures.
* Mount & Blade is a really interesting game for this datastore as the module information is extremely rich and links via unique module Ids. This means we can crosslink installed modules. This means we can use normal C# attributes to walk from one mod's modules into the modules in a completely different mod in the same loadout. Lots of interesting usecases in this game. 

This PR seems to be mostly working. The tests pass, and I can go through the UI and install mods into SDV, and play the game. Going to do more testing on this, but for the most part I think it's ready to merge after a bit of code reivew of the 9000 changed lines.


src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs
```diff
@@ -1,150 +1,167 @@
...
     }
 
     /// <summary>
-    /// Remove an individual mod from this loadout, returning a new loadout.
+    /// Returns all revisions of a loadouts over time.
     /// </summary>
-    /// <param name="mod"></param>
-    /// <returns></returns>
-    public Loadout Remove(Mod mod)
+    public static IObservable<Model> LoadoutRevisions(this IConnection conn)
```
erri120, line 76: This returns an observable stream of all revisions of all loadouts, right? If that's the case, the `summary` should be updated, and the method should be named something like `AllLoadoutRevisions` or similar.

To prevent confusion between `LoadoutRevisions` and `Revisions` methods, we can also consider using names like `OneLoadoutAllRevisions` and `AllLoadoutsAllRevision` to make it clear that we're getting multiple revisions of one or all loadout(s).

src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs
```diff
@@ -1,150 +1,167 @@
...
-        return this with
+        /// <summary>
+        /// The unique identifier for this loadout, casted to a <see cref="LoadoutId"/>.
+        /// </summary>
+        public LoadoutId LoadoutId => LoadoutId.From(Id);
+        
+        /// <summary>
+        /// Gets the loadout id/txid pair for this revision of the loadout.
+        /// </summary>
+        public LoadoutRevisionId LoadoutRevisionId => new(LoadoutId, Db.BasisTxId);
```
erri120, line 97: I'm not sure if I fully understand this. What's the difference between `LoadoutRevisionId` and `ulong Revision`? From my understanding, `LoadoutRevisionId` is a pair of `LoadoutId` and `BasisTxId` (terrible name btw) which points to a specific Loadout with a specific database revision? Meanwhile, `Revision` is just a user-facing monotonic increasing number that essentially represents the number of changes done?

These two things need way better names to prevent confusion. @Al12rs thoughts?

src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs
```diff
@@ -1,150 +1,167 @@
...
+        /// <summary>
+        /// Gets all the files in this loadout.
+        /// </summary>
+        public Entities<EntityIds, File.Model> Files => GetReverse<File.Model>(File.Loadout);
+
+                
+        /// <summary>
+        /// Gets the mod with the given id from this loadout.
+        /// </summary>
+        public Mod.Model this[ModId idx] => Mods.First(m => m.ModId == idx);
```
erri120, line 151: Couldn't this be optimized like so:

```csharp
var mod = Db.Get<Mod.Model>(idx.Value);
if (mod.LoadoutId != LoadoutId) throw new KeyNotFoundException();
return mod;
```

Since the new relationship is in reverse, this should work, right?

src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs
Al12rs: Do we need the user facing revision number?

src/Abstractions/NexusMods.Abstractions.Loadouts/LoaderExtensions.cs
```diff
@@ -0,0 +1,28 @@
+using NexusMods.Abstractions.Loadouts.Ids;
+using NexusMods.Abstractions.Loadouts.Mods;
+using NexusMods.MnemonicDB.Abstractions;
+
+namespace NexusMods.Abstractions.Loadouts;
+
+/// <summary>
+/// Extensions to IDb/IConnection for loading loadouts and other entities
+/// </summary>
+public static class LoaderExtensions
```
erri120, line 10: This should go into the `Extensions` folder and be named `DBExtensions` or similar. The `Extensions` namespace can also be removed from that folder, so all extensions should be available from just `NexusMods.Abstractions.Loadouts`.

src/Abstractions/NexusMods.Abstractions.Loadouts/Files/File.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// <summary>
+        /// Gets the file id
+        /// </summary>
+        public FileId FileId => FileId.From(Id);
+
+        public EntityId LoadoutId
+        {
+            get => File.Loadout.Get(this);
+            set => File.Loadout.Add(this, value);
+        }
```
erri120, line 51: `Loadout.LoadoutId` is of type `LoadoutId` while both `File.LoadoutId` and `Mod.LoadoutId` return an `EntityId`. This inconsistency is also present with `ModId` in both `File` and `Mod` models. This seems like something that'll easily result in more inconsistencies and weird behavior. We have `FileId`, `ModId`, and `LoadoutId` for a reason, so we should probably use them.

src/Abstractions/NexusMods.Abstractions.Loadouts/Files/File.cs
```diff
@@ -0,0 +1,89 @@
...
+        
+        /// <summary>
+        /// If this file is a stored file, this will return true and cast the stored file to the out parameter.
+        /// </summary>
+        public bool IsStoredFile(out StoredFile.Model storedFile)
+        {
+            storedFile = this.Remap<StoredFile.Model>();
+            return Contains(StoredFile.Hash);
+        }
```
erri120, line 87: This should follow the try-get pattern:

```csharp
public bool IsStoredFile([NotNullWhen(true)] out StoredFile.Model? storedFile)
{
  storedFile = null;
  if (!Contains(StoredFile.Hash)) return false;
  storedFile = this.Remap<StoredFile.Model>();
  return true;
}
```

src/Abstractions/NexusMods.Abstractions.Loadouts/Files/StoredFile.cs
erri120, line 1: There is a leftover `IStoredFile` interface in a nearby file that should be deleted.

src/Abstractions/NexusMods.Abstractions.Loadouts/Files/DeletedFile.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class Model(ITransaction tx) : File.Model(tx)
+    {
+        /// <summary>
+        /// True if the file is deleted.
+        /// </summary>
+        public bool Deleted
+        {
+            get => DeletedFile.Deleted.TryGet(this, out var deleted) && deleted;
+            set => DeletedFile.Deleted.Add(this, value);
+        }
```
erri120, line 34: This very much feels like HTML `option` with `selected` attribute situation. In HTML, the `option` element has a `selected` tag. The presence of this tag indicates that the option is initially selected. However, the value is ignored. So these two elements are both selected:

```html
<option selected="true">

<option selected="false">
```

Is a deleted file, a file that has the `Deleted` attribute, or is a deleted file a file where `Deleted = true`? Other parts of the code use the presence of an attribute as an indicator for what the entity represents.

src/Abstractions/NexusMods.Abstractions.Loadouts/Files/FileTreeNode.cs
```diff
@@ -0,0 +1,20 @@
...
+using NexusMods.Hashing.xxHash64;
+using NexusMods.Paths;
+
+namespace NexusMods.Abstractions.Loadouts.Files;
+
+/// <summary>
+/// Clas version of IFileTreeNode
+/// </summary>
+public class FileTreeNode : IFileTreeNode
+{
```
erri120, line 11: Appears to be unused.

src/Abstractions/NexusMods.Abstractions.Loadouts/Extensions/ModExtensions.cs
```diff
@@ -13,16 +14,17 @@ public static class ModExtensions
...
     /// <summary>
     /// Returns an optional metadata object on the mod, if it exists.
     /// </summary>
-    public static Optional<T> GetMetadata<T>(this Mod mod) where T : AModMetadata
+    public static Optional<T> GetMetadata<T>(this Mod.Model mod, IAttribute attr) where T : notnull
     {
-        return mod.Metadata.OfType<T>().FirstOrDefault();
+        throw new NotImplementedException();
+        //return mod.Metadata.OfType<T>().FirstOrDefault();
     }
```
erri120, line 29: These methods can be removed, as they are unused.

src/Abstractions/NexusMods.Abstractions.Loadouts/Extensions/AModFileExtensions.cs
```diff
@@ -14,8 +16,8 @@ public static class AModFileExtensions
     /// Checks whether any items in <see cref="AModFile.Metadata"/> are of type
     /// <typeparamref name="T"/>.
     /// </summary>
-    public static bool HasMetadata<T>(this AModFile file) where T : IMetadata
+    public static bool HasMetadata(this File.Model file, IAttribute attribute)
     {
-        return file.Metadata.Any(m => m is T);
+        return file.Contains(attribute);
```
erri120, line 21: This can be removed. I'll probably end up reworking the Stardew Valley diagnostics anyway, to take advantage of the new system.

src/Abstractions/NexusMods.Abstractions.Loadouts/IApplyService.cs
```diff
@@ -15,12 +17,12 @@ public interface IApplyService
     /// </summary>
     /// <param name="loadoutId"></param>
     /// <returns></returns>
-    public Task<Loadout> Apply(LoadoutId loadoutId);
+    public Task<Loadout.Model> Apply(Loadout.Model loadout);
```
erri120, line 20: The doc-string of this method needs to be updated. Also, does this even need a return value? Looking at the implementation, we just return the input...

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/IGeneratedFile.cs
erri120, line 1: This can probably be removed, as it's unused.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/IFileGenerator.cs
```diff
@@ -0,0 +1,24 @@
...
+/// the application of a loadout.
+/// </summary>
+public interface IFileGenerator : IGuidClass
+{
+    /// <summary>
+    /// Not supported, as this is an interface
+    /// </summary>
+    /// <exception cref="NotSupportedException"></exception>
+    static UInt128 IGuidClass.Guid => throw new NotSupportedException();
+
```
erri120, line 19: If this doesn't support `IGuidClass.Guid`, why is it implementing that interface in the first place? The interface doesn't have anything else in it.

erri120: Maybe throw `UnreachableException` instead? Since it assumes that whatever implements `IFileGenerator` "overrides" this value, so this should never be called on `IFileGenerator`.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/GeneratedFile.cs
```diff
@@ -0,0 +1,65 @@
...
+
+public static class GeneratedFile
+{
+    private const string Namespace = "NexusMods.Abstractions.Loadouts.Files.GeneratedFile";
+    
+    /// <summary>
+    /// The generator unique ID for the file, this specifies the class that is used to generate
+    /// the contents of this file. 
+    /// </summary>
+    public static readonly ClassAttribute<IFileGenerator> Generator = new("NexusMods.Abstractions.Loadouts.Files.GeneratedFile", nameof(Generator)) { IsIndexed = true };
```
erri120, line 16: This should use the namespace constant.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/GeneratedFile.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+        
+        /// <summary>
+        /// Set the generator for this file.
+        /// </summary>
+        public void SetGenerator<TType>()
+        where TType : IFileGenerator
+        {
+            GeneratedFile.Generator.Add(this, TType.Guid);
+        }
```
erri120, line 35: This looks super weird, why isn't this one property with a getter and setter like all the other models?

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/GeneratedFile.cs
```diff
@@ -0,0 +1,65 @@
...
+    public static bool IsGeneratedFile<TType>(this File.Model model)
+    where TType : IFileGenerator =>
+        model.TryGet(GeneratedFile.Generator, out var val) && val == TType.Guid;
+    
+    /// <summary>
+    /// Returns true if the file is a generated file.
+    /// </summary>
+    public static bool IsGeneratedFile(this File.Model model) =>
+        model.Contains(GeneratedFile.Generator);
+}
```
erri120, line 65: This should go into a separate file. Similar to the `File.IsStoredFile` method, `IsGeneratedFile` should use the try-get pattern, see the comment on `File.IsStoredFile` for details. Also, please be consistent with these remappings. These methods are in an extension class while the `File.IsStoredFile` method is a class method on `File` itself. Do one or the other, not both.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -750,48 +838,110 @@ public virtual async Task BackupNewFiles(GameInstallation installation, FileTree
...
+        return file.TryGet(DeletedFile.Deleted, out var deleted) && deleted;
+    }
+
+    /// <summary>
+    /// True if the file is a stored file.
+    /// </summary>
+    public bool IsStoredFile(File.Model file)
+    {
+        return file.Contains(StoredFile.Hash);
+    }
```
erri120, line 874: As written multiple times already, this is inconsistent with everything else and somewhat duplicate with existing methods.

src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/BooleanAttribute.cs
erri120, line 1: Remove the duplicate `BoolAttribute.cs` file.

src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/ClassAttribute.cs
```diff
@@ -0,0 +1,44 @@
...
+    
+    /// <summary>
+    /// Gets the instance of the class from the DI container.
+    /// </summary>
+    public TType GetInstance(Entity entity) 
+    {
+        var key = Get(entity);
+        var item = entity.Db.Connection.ServiceProvider.GetKeyedService<TType>(key);
+        if (item == null)
+            throw new NullReferenceException($"The item {typeof(TType)} with key {key} was not found in the DI container.");
```
erri120, line 30: Should be a `KeyNotFoundException`.

src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/EnumAttribute.cs
```diff
@@ -0,0 +1,23 @@
...
+    protected override string ToLowLevel(T value)
+    {
+        return value.ToString();
+    }
+
+    /// <inheritdoc />
+    protected override T FromLowLevel(string value, ValueTags tag)
+    {
+        return (T) Enum.Parse(typeof(T), value);
+    }
```
erri120, line 22: Converting Enums to and from strings is notoriously slow, with tons of allocations. Using strings also means that we'll never be able to change the name in the enum or we'll break this. Instead of using names, it'd be better to use the underlying integer values. The integer values are more resilient to change because developers are already accustomed to not changing existing values.

@Sewer56 probably has some more details on this.

src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/Extensions/EntityExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public static bool TryParseFromHex(string hex, out EntityId id)
+    {
+        if (hex.StartsWith("EId:"))
+            hex = hex[4..];
+        
+        if (ulong.TryParse(hex, NumberStyles.HexNumber, null, out var parsed))
+        {
+            id = EntityId.From(parsed);
+            return true;
```
erri120, line 32: This method should use Spans to prevent unnecessary allocations.

src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/GamePathAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        // TODO: make this a reference or something
+        return $"{value.LocationId.Value}|{value.Path}";
+    }
+
+    /// <inheritdoc />
+    protected override GamePath FromLowLevel(string value, ValueTags tags)
+    {
+        var parts = value.Split('|');
+        return new GamePath(LocationId.From(parts[0]), RelativePath.FromUnsanitizedInput(parts[1]));
```
erri120, line 24: Add a `Debug.Assert(parts.Length == 2)`.

src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/UrlAttribute.cs
erri120, line 1: Duplicate file: `UriAttribute.cs`. Remove one or the other. Also don't throw a custom exception when Uri parsing fails, just use the `new Uri()` as it has better exceptions.

src/NexusMods.DataModel/Repository/Repository.cs
```diff
@@ -0,0 +1,109 @@
...
+    private readonly ReadOnlyObservableCollection<TModel> _observable;
+
+    /// <inheritdoc />
+    public ReadOnlyObservableCollection<TModel> Observable => _observable;
+
+    /// <inheritdoc />
+    public IObservable<TModel> Revisions(EntityId id)
+    {
+        if (!Exists(id))
+            return Empty<TModel>();
```
erri120, line 71: This feels like a weird implementation detail that might result in hours of wasted debugging. It might very well be possible to want all changes to a yet not-existing model. The observable stream should return no elements either way. Either remove this branch or add a parameter `ignoreMissingModel` with a default value of `false` but definitely document this somewhere.

src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/Operations.cs
```diff
@@ -0,0 +1,52 @@
...
+            );
+            old.Loadout.Revise(tx);
+            
+            old.Revise(tx);
+            await tx.Commit();
+        }
+        
+        /// <summary>
+        /// Toggle the enabled status of the mod.
+        /// </summary>
```
erri120, line 35: That's just wrong.

src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/ModCategory.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    GameFiles,
+    Mod,
+    Saves,
+    Overrides,
+    /// <summary>
+    /// Files maintained by the game's framework, not intended for direct user modification.
+    /// </summary>
+    Metadata,
+}
```
erri120, line 13: Our previous iteration of categories had an open set, this is a closed set. We might want to discuss if this is the best approach. If we want to keep this as an enum, then all of these items should have explicit values.

@Al12rs thoughts?

src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/Mod.cs
```diff
@@ -17,85 +23,172 @@ namespace NexusMods.Abstractions.Loadouts.Mods;
...
+        }
+        
+        /// <summary>
+        /// The revision number for this mod, increments by one for each change.
+        /// </summary>
+        public ulong Revision
+        {
+            get => Mod.Revision.Get(this, 0);
+            set => Mod.Revision.Add(this, value);
+        }
```
erri120, line 112: Same issue as Loadout revision.

src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs
halgari: This was unused in the code, as I swapped over to IRepository<> later on in the PR. Removed.

src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs
halgari: yep, good catch.

src/Abstractions/NexusMods.Abstractions.Loadouts/Ids/IdAsOf.cs
```diff
@@ -0,0 +1,35 @@
+using NexusMods.MnemonicDB.Abstractions;
+using NexusMods.MnemonicDB.Abstractions.Models;
+
+namespace NexusMods.Abstractions.Loadouts.Ids;
+
+public struct IdAsOf<TId, TModel> 
+    where TId : ITypedId<TModel> 
+    where TModel : IEntity
```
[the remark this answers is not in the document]
halgari: removed

src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/ModCategory.cs
Al12rs: If this is a closed set we are only going to use this for GameFiles and Overrides and promoting those mods as even more special.  Games will not able to define their own special mods. So yeah I think this should be an open set.
I thought before we could have multiple categories on a mod?

src/Abstractions/NexusMods.Abstractions.Loadouts/Files/File.cs
halgari: Same with the Mod's LoadoutId

src/Abstractions/NexusMods.Abstractions.Loadouts/Files/StoredFile.cs
halgari: Removed

src/Abstractions/NexusMods.Abstractions.Loadouts/Files/DeletedFile.cs
halgari: This is the reified deletes thing, File already has a `To` and so all we really need is `Deleted`. So yes a deleted file in this case, is a object in the winning "layer" that has "Deleted=True"

src/Abstractions/NexusMods.Abstractions.Games/RunGameTool.cs
```diff
@@ -60,10 +61,12 @@ public RunGameTool(IServiceProvider serviceProvider, T game)
     public string Name => $"Run {_game.Name}";
 
     /// <inheritdoc />
-    public async Task Execute(Loadout loadout, CancellationToken cancellationToken)
+    public async Task Execute(Loadout.Model loadout, CancellationToken cancellationToken)
     {
         _logger.LogInformation("Starting {Name}", Name);
+        throw new NotImplementedException();
```
erri120, line 67: I'm guessing this is a TODO?

src/Abstractions/NexusMods.Abstractions.Loadouts/IApplyService.cs
halgari: The idea is that we return the updated loadout, if we made any chages, but nothing in the code looks at it. Removed

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/IFileGenerator.cs
halgari: Yeah it's a static member on an interface and the interface *must* implement it in that case.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/GeneratedFile.cs
halgari: Because you can't provide a type parameter to a property. It's really dumb but what I want is TType.Guid...and there's no way to do that in a property, because `Value` in this case would be IFileGenerator. Maybe there's a way to fix this, because yes it is ugly.

src/Abstractions/NexusMods.Abstractions.Installers/ModInstallerResult.cs
```diff
@@ -13,13 +15,13 @@ public record ModInstallerResult
     /// <summary>
     /// Unique identifier of the mod.
     /// </summary>
-    public required ModId Id { get; init; }
+    public required ModId? Id { get; init; }
```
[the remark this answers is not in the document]
erri120: Removing the question mark and compiling doesn't result in an error, so this should be removed.

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
```diff
@@ -39,6 +39,8 @@ public static class AbsolutePathExtensions
     /// <returns>The xxHash64 hash of the file.</returns>
     public static Hash XxHash64MemoryMapped(this AbsolutePath input, IActivitySource<Size>? job = null)
     {
+        if (input.FileSystem is InMemoryFileSystem) 
+            return XxHash64MemoryMappedInMemory(input, job);
```
erri120, line 43: This shouldn't be needed. The `InMemoryFileSystem` supports memory mapped files. The entire point of having a memory mapped file API in `IFileSystem` was so that consumers **don't** have to check for the FS implementation.

@Sewer56 wrote that API after my nagging about this.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/GeneratedFile.cs
halgari: We can put it in separate files, but what's the reasoning for this? I like to keep these methods close to the things they are talking about/used with. Since attributes are all on a static class it makes sense to keep the remappings on the same classes as well. I ended up moving these into the same static class as GeneratedFile, but we can move them if we come up with a better idea. 

GeneratedFiles don't exist in the same project (due to dependency issues) as the other file models, so there's a bit of an issue there

src/NexusMods.CrossPlatform/NexusMods.CrossPlatform.csproj
```diff
@@ -42,4 +42,8 @@
       </Compile>
     </ItemGroup>
 
+    <ItemGroup>
+      <PackageVersion Update="CliWrap" Version="3.6.6" />
+    </ItemGroup>
```
erri120, line 47: This seems very random. Probably should be removed.

src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/BooleanAttribute.cs
halgari: hah! I knew I made one of those before I made the other. Got tunnel vision working on this stuff after awhile

src/Abstractions/NexusMods.Abstractions.Games/RunGameTool.cs
Al12rs: Attempting to run the game throws this so currently it doesn't work

src/NexusMods.DataModel/Repository/Repository.cs
halgari: ah good point, I didn't think about the fact that we could have a live stream to a non-existant or badly mapped entity. This is needed because sometimes the UI will ask for `EID:0` as a default value and I don't want that to throw, but there's better ways of getting this effect.

erri120:
After looking around the code a bit more, I feel like we're leaving a lot of performance on the table by using `Loadout.Mods` and `Loadout.Files` everywhere. This was probably the easiest way to adapt the old code to the new DB, but it doesn't leverage the new DB features or the new models at all at just delegates all filtering to LINQ.

Going forward, we want something similar to the extension methods I made previously for diagnostic emitters, like `GetModWithMetadata` that abstract very common operations and internally use the DB better instead of using LINQ for everything.

src/Abstractions/NexusMods.Abstractions.Games/RunGameTool.cs
halgari: Yep...I thought I missed this. I'll get it fixed up

src/Abstractions/NexusMods.Abstractions.Installers/ModInstallerResult.cs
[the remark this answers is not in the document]
halgari: The idea was at one point in testing that we wouldn't require installers to return the id, and we'd use the main id by default. Turns out I didn't go down that route, and its required, so there's no point.

src/NexusMods.App.UI/LeftMenu/Items/LaunchButtonViewModel.cs
```diff
@@ -20,26 +22,26 @@ public class LaunchButtonViewModel : AViewModel<ILaunchButtonViewModel>, ILaunch
...
 
         Command = ReactiveCommand.CreateFromObservable(() => Observable.StartAsync(LaunchGame, RxApp.TaskpoolScheduler));
     }
 
     private async Task LaunchGame(CancellationToken token)
     {
         Label = Language.LaunchButtonViewModel_LaunchGame_RUNNING;
-        var marker = _loadoutRegistry.GetMarker(LoadoutId);
-        var tool = _toolManager.GetTools(marker.Value).OfType<IRunGameTool>().First();
+        var marker = _conn.Db.Get(LoadoutId);
```
Al12rs, line 40: I saw a bunch of these `_conn.Db.Get(Id)` all throughout the code.
Should something like the VM of a UI button really interact directly with the database?
Shouldn't the implementation of this be deferred to something like the `LoadoutRegistry`?

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/GeneratedFile.cs
erri120: It's mostly about being consistent and predictable. Keeping this extension class in the same file is fine for now, but the other `IsSomething` and `TryGetAsSomething` methods should all be extension methods instead of member methods.

src/Games/NexusMods.Games.StardewValley/Installers/SMAPIInstaller.cs
```diff
@@ -226,17 +229,10 @@ await using (var stream = await item.OpenAsync())
                 Name = "SMAPI",
                 Id = info.BaseModId,
                 Files = modFiles,
-                SortRules = new[]
+                Metadata = new TempEntity
                 {
-                    new SMAPISorter(),
+                    {SMAPIMarker.Version, version},
                 },
```
erri120, line 232: Do we deploy SMAPI correctly without this? I had to explicitly add this sorted because sorting is otherwise undefined and we wouldn't be able to override the game executable.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/NexusMods.Games.MountAndBlade2Bannerlord.csproj
```diff
@@ -26,4 +26,9 @@
     <InternalsVisibleTo Include="NexusMods.Games.MountAndBlade2Bannerlord.Tests" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Remove="MnemonicDB\SubModuleInfoExtended.cs" />
+    <Compile Remove="MnemonicDB\ModuleInfoExtendedWithPath.cs" />
+  </ItemGroup>
```
erri120, line 32: This seems wrong.

src/NexusMods.DataModel/Loadouts/DiskStateRegistry.cs
```diff
@@ -93,43 +101,46 @@ public async Task SaveInitialState(GameInstallation installation, DiskStateTree
...
-            .FirstOrDefault();
-        
-        if (initialDiskState == null)
-            return;
-
-        var tx = _connection.BeginTransaction();
-        initialDiskState.Tx = tx;
-        initialDiskState.AddRetractToCurrentTx();
-        await tx.Commit();
+        throw new NotImplementedException();
```
erri120, line 118: Is this a TODO?

src/NexusMods.DataModel/Loadouts/GroupMetadata.cs
erri120, line 1: This is not used anymore and can probably be deleted.

src/NexusMods.DataModel/ApplyService.cs
```diff
@@ -102,56 +85,49 @@ public ValueTask<FileDiffTree> GetApplyDiffTree(LoadoutId loadoutId)
...
+            last = new LoadoutRevisionId(LoadoutId.From(EntityId.From(0)), TxId.From(0));
+        
         // Return a deferred observable that computes the starting value only on first subscription
         return Observable.Defer(() => _diskStateRegistry.LastAppliedRevisionObservable
-            .Where(x => x.gameInstallation.Equals(gameInstallation))
-            .Select(x => x.loadoutRevision)
-            .StartWith(_diskStateRegistry.GetLastAppliedLoadout(gameInstallation) ?? IdEmpty.Empty)
+            .Where(x => x.Install.Equals(gameInstallation))
+            .Select(x => x.LoadoutRevisionId)
+            .StartWith(last)
```
erri120, line 127: Does this still have to be a deferred observable? @Al12rs do you have more info on this, I think you wrote this initially.

src/NexusMods.App.UI/Pages/LoadoutGrid/LoadoutGridViewModel.cs
```diff
@@ -129,23 +135,16 @@ public class LoadoutGridViewModel : APageViewModel<ILoadoutGridViewModel>, ILoad
...
             this.WhenAnyValue(vm => vm.LoadoutId)
-                .SelectMany(loadoutRegistry.RevisionsAsLoadouts)
-                .Select(loadout =>
-                {
-                    // NOTE(erri120): see https://github.com/Nexus-Mods/NexusMods.App/issues/1195 for details
-                    var showGameFiles = settingsManager.Get<LoadoutGridSettings>().ShowGameFiles;
-                    return loadout.Mods.Values
-                        .Where(mod => showGameFiles || mod.ModCategory != Mod.GameFilesCategory)
-                        .Select(mod => new ModCursor(loadout.LoadoutId, mod.Id));
-                })
```
erri120, line 140: Very sadge

src/NexusMods.App.UI/Pages/LoadoutGrid/Columns/ModEnabled/ModEnabledViewModel.cs
```diff
@@ -24,57 +27,49 @@ public class ModEnabledViewModel : AViewModel<IModEnabledViewModel>, IModEnabled
     public ModStatus Status { get; set; } = ModStatus.Installed;
 
     [Reactive]
-    public ReactiveCommand<bool, Unit> ToggleEnabledCommand { get; set; }
+    public ICommand ToggleEnabledCommand { get; set; }
 
     [Reactive]
-    public ReactiveCommand<Unit, Unit> DeleteModCommand { get; set; }
```
erri120, line 30: Should stay as reactive commands.

src/NexusMods.App.UI/Controls/Trees/ModFileTreeViewModel.cs
```diff
@@ -43,16 +45,17 @@ public ModFileTreeViewModel(LoadoutId loadoutId, ModId modId, ILoadoutRegistry r
         // Store GamePaths to dedupe the strings. No unsafe API in .NET to access the keys directly, but we need parent anyway, so it's ok.
         var folderToSize = new Dictionary<GamePath, (ulong size, uint numFileChildren, GamePath folder, GamePath parent, bool isLeaf)>();
 
-        var loadout = _registry.Get(loadoutId);
+        var db = _conn.Db;
+        var loadout = db.Get(loadoutId);
         var locationsRegister = loadout!.Installation.LocationsRegister;
-        var mod = _registry.Get(loadoutId, modId)!; // <= suppressed because this throws on error, and we should always have valid mod if we made it here.
+        var mod = db.Get(modId); // <= suppressed because this throws on error, and we should always have valid mod if we made it here.
```
erri120, line 51: Comment can be removed.

erri120 (changes requested):

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
halgari: I never actually got that code to work, because of how half-implemented our in memory filesystem is. I'd love to execute tests completely in memory, but the InMemoryFilesystem has a lot of edge cases currently.

Al12rs (commented):
### Things to fix:
These should be ticked off if implemented in this PR. Remaining items should be made into issues.

- [ ] `GameRegistry` isn't initialized (missing `host.StartAsync()` in `Program.cs`)
- [ ] MyGames page should only show one GameWidget per game install if you have multiple loadouts of the same game install
- [ ] Unmanage Game code has been removed and needs to be implemented back in
- [ ] Remove loadout code has been removed and needs to be implemented back in
- [ ] Reified deletions file entries show up as normal files in ModContents View
- [ ] Reified deletions file show up as additions in Apply diff view
- [ ] Applying when there are changes to ingest keeps showing the Apply button after completion. Current applied revision doesn't seem to update correctly
- [ ] Running the game through Launch button throws not implemented exception
- [ ] Fix CI test failures

src/NexusMods.App.UI/Pages/LoadoutGrid/Columns/ModEnabled/ModEnabledViewModel.cs
Al12rs: I guess this was also something that was reverted with merge conflicts

halgari: @erri120 how do we get that to work with .BindCommand? I couldn't get that to typecheck with ReactiveCommands because it complained that ReactiveCommand<> wasn't a ICommand so the call to `.BindCommand` failed.

src/NexusMods.DataModel/ApplyService.cs
Al12rs: The idea was to only start observing while there are actually interested parties. 
I think there might be better ways to do this but it should be fine for now.

src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/Operations.cs
halgari: Oh right, copied comment and I was looking at the wrong method

src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/ModCategory.cs
halgari: https://github.com/Nexus-Mods/NexusMods.App/issues/1315

src/Extensions/NexusMods.Extensions.Hashing/AbsolutePathExtensions.cs
halgari: removed

src/NexusMods.CrossPlatform/NexusMods.CrossPlatform.csproj
halgari: Yep was added through debugging some of the compile time failures, removed.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/NexusMods.Games.MountAndBlade2Bannerlord.csproj
halgari: removed, Rider likes to add these sometimes instead of deleting files

src/NexusMods.DataModel/Loadouts/GroupMetadata.cs
halgari: removed

src/NexusMods.App.UI/Pages/LoadoutGrid/Columns/ModEnabled/ModEnabledViewModel.cs
halgari: Used ReactiveCommand vs IReactiveCommand as Al suggested, and that worked

src/NexusMods.DataModel/Loadouts/DiskStateRegistry.cs
halgari: we'll add this in as a separate PR

src/NexusMods.App.UI/Pages/LoadoutGrid/Columns/ModEnabled/ModEnabledViewModel.cs
erri120: In the VM interface and class you should use the concrete type `ReactiveCommand<TInput, TOutput>` (replace `TInput` and `TOuput` with what makes sense or `System.Reactive.Unit` if it doesn't have an input/output).

src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/Mod.cs
halgari: Fixed this by removing one overload of "revision Id", LoadoutRevisionId is now "LoadoutWithTxId". We still need to improve this a bit, but what we have now is a good starting point.

src/NexusMods.App/Program.cs
```diff
@@ -43,6 +43,9 @@ public static int Main(string[] args)
             loggingSettings
         );
 
+        // Okay to do wait here, as we are in the main process thread.
+        host.StartAsync().Wait();
```
Al12rs, line 47: Shouldn't main just be async instead? Or are there reasons we can't do that?

Al12rs: I see, it appears to be running fine on my end, but I found the `[STAThread]` problems online, that it doesn't get applied correctly to main function that is generated out of the async one.

src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs
```diff
@@ -148,7 +134,16 @@ public GameInstallation Installation
...
         /// </summary>
-        public Mod.Model this[ModId idx] => Mods.First(m => m.ModId == idx);
+        public Mod.Model this[ModId idx]
+        {
+            get
+            {
+                var mod = Db.Get<Mod.Model>(idx.Value);
+                if (mod is null) 
+                    throw new KeyNotFoundException($"Mod with id {idx} not found in loadout {Id}");
+                return mod;
```
erri120, line 144:
```suggestion
                var mod = Db.Get<Mod.Model>(idx.Value);
                if (mod is null) 
                    throw new KeyNotFoundException($"Mod with id {idx} not found in database");
                if (mod.LoadoutId != LoadoutId)
                    throw new KeyNotFoundException($"Mod with id {idx} is not part of Loadout {LoadoutId} but of Loadout {mod.LoadoutId}");
                return mod;
```

src/NexusMods.App.UI/Pages/LoadoutGrid/Columns/ModEnabled/ModEnabledViewModel.cs
```diff
@@ -2,6 +2,7 @@
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using System.Windows.Input;
+using ExCSS;
```
erri120, line 5:
```suggestion
```

Common mistake, `ExCSS` also contains a `Unit` type but we want the one from `System.Reactive`.

src/NexusMods.App/Program.cs
erri120, line 47:
```suggestion
        // Okay to do wait here, as we are in the main process thread.
        host.StartAsync().Wait(timeout: TimeSpan.FromMinutes(5));
```

Timeout so we don't wait forever.

erri120 (changes requested):
See comments and suggestions.

src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs
erri120: Was missing the loadout check.

PR #1308, opened 2024-05-06
Use system extractor executable

### LINKED ISSUE #1306: Avoid bundling dynamically linked 7zz binary
# Improvement

## Currently 

The 7zz binary bundled with the Linux version of this package will not work on platforms like NixOS without additional tweaks. It would be good to have the option to not copy this executable to the output directory during the build.

## We want

A flag to exclude 7zz from the output when building.

## Supporting Information

See https://github.com/NixOS/nixpkgs/pull/270372/files for a workaround, using the 7zz package from nixpkgs instead.

@erri120

erri120:
Fixes #1306.

Usage:

`NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR=1 dotnet build`


Sewer56:
I'll handle this when I'm back to work tomorrow.
Just gotta remember that there's always the fact we also have to patch the code to invoke system `7zz`. i.e. just remove path up to 7zz itself.

Sewer56:
Looks perfectly reasonable to me.
We should probably maybe also check AppImage just in case, but I doubt that'd be an issue even if used there.
AppImage users probably won't want this flag as by design the AppImage should ship the deps.

PR #1309, opened 2024-05-06
Fix/1301 workspace breaks

### LINKED ISSUE #1301: Workspace panels resizing breaks when passing from fullscreen to half screen
image
If the major orientation is switched while 4 panels are open and the splitters are moved a bit things break apart.

Steps to reproduce:
- Make 4 panes in landscape mode
- Resize panels a bit
- Go in portrait mode (snap to half screen for example)
- Try resizing again , things should start to break
- Go back to landscape and resize, should break more

erri120:
Might fix #1301.

This needs testing. The PR also fixed the broken `Distinct` call when reacting to changes made to `IsHorizontal`. `Distinct` uses an internal `HashSet<T>` to check if the current values is new, however, since `IsHorizontal` is a boolean, we'll stop getting new values after `true` and `false` has been sent once. Instead, we'll use `DistinctUntilChanged` which the original implementation should've used.


Al12rs (approved):

erri120:
Panels still get re-ordered, haven't fixed that yet.

erri120:
Updated the code to keep the original order of the panels.

src/NexusMods.App.UI/WorkspaceSystem/GridUtils.cs
```diff
@@ -63,23 +63,31 @@ internal static bool IsPerfectGrid(WorkspaceGridState gridState, bool doThrow =
...
+        var panels = GC.AllocateUninitializedArray<PanelGridState>(length: count);
+        for (var i = 0; i < count; i++)
+        {
+            panels[i] = new PanelGridState(currentState[i].Id, fake[i].Rect);
+        }
+
+        var res = currentState.UnionById(panels);
         return res;
+
+        static PanelId MakeId(int i) => PanelId.From(new Guid(i, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0));
```
Al12rs, line 90: I'm not really sure what is happening here with the fake thing. Could you add some comments or change some var names to make what is happening clearer?

erri120: The code creates a "fake" WorkspaceGrid and fills it up with N panels. We then update the current WorkspaceGrid with those fake panel bounds. This resets the sizes and positions of the panels while keeping the same IDs.

Al12rs: Ah, clever, thanks for adding the comments.

Al12rs (approved):

PR #1317, opened 2024-05-07
Jtw auth to mnemonicdb

halgari:
Switches JWT and API key auth over to Mnemonic DB

Both of these need a singleton version of the IRepository. It might be worth looking into that at some point, but for now it's simple enough just to handcode it. We can replace this code in the future with some sort of singleton system if we need more of these.


erri120:
Closed in favor of #1318.

Issue #1284, opened 2024-04-29
Mod Library Refinements


Al12rs:
Todo items not implemented as part of initial PR: #1268 
- Mostly blocked on #1276 

### TODO:
- [x] Make list update in real-time (get observable stream from database) (better wait for #1276 )
- [x] Show "-" if we don't have Last installed date or Archive date
- [ ] Downloaded date/time data (put "Manually added" for mod archives not downloaded through the app) #1347
- [ ] Use the Nexus mod name for archives that come from nexus instead of the file name. #1347
- [ ] Version data if available (will not be for archives that don't come from nexus, but for manual archives that come from nexus   it should be possible to obtain the data with an API call, or by parsing the archive name for the embedded version number) #1347
- [ ] Added column data (not sure how to implement this yet, loadouts and mods probably need to be moved to Mnemonic first) #1347
- [ ] Filter out archives that are not manually added or downloaded (don't show nested archives like Installer.dat from SMAPI) #1347
- [ ] Filter entries by game for the current loadout #1347

PR #1332, opened 2024-05-09
Regression fixes and Mod Library improvements

### LINKED ISSUE #1299: Advanced Installer window is missing chrome and no longer has rounded borders
image

Should have a chrome titlebar and rounded borders

### LINKED ISSUE #1300: Mods installed with Advanced Installer have "<unknown>" name
image

Should use the archive name instead

### LINKED ISSUE #1310: Download throughput value is incorrect
I'm capped at 18MB/s and the SpineDownloadButton is showing much higher than that.

This happens when downloading multiple files, even one after another. It seems something isn't being reset after a download is paused/ cancelled or completed, and the new throughputs are added on top.

image

### LINKED ISSUE #1321: Regression: MyGames page should only show one widget for all loadouts of the same game install
In the image below there should only be one GameWidget instead of two.
image

Regression of #1302

### LINKED ISSUE #1327: Advanced Installer isn't being used as a fallback for installing mods (Stardew)
Advanced installer should be used as the fallback in case a download or archive isn't supported by any of the game installers.

This doesn't currently happen for Stardew when pressing the Install Mod button or when installing from Download Library view.

Al12rs:
- fixes #1300
- fixes #1310 
- fixes #1321 
- fixes #1299 
- fixes #1327
- Fixed exception in loadoutGrid Installed date column, value update was not happening on UI thread.
- Ticks a couple of items of #1284
    - Make list update in realtime
    - Show "-" for invalid dates
    - Show dates in humanized format (e.g. 2 minutes ago)


src/NexusMods.DataModel/ArchiveInstaller.cs
```diff
@@ -50,14 +51,18 @@ public async Task<ModId[]> AddMods(LoadoutId loadoutId, DownloadAnalysis.Model d
...
-        {
-            archiveName = download.Get(DownloadAnalysis.SuggestedName);
-        }
-        
-        string modName = defaultModName ?? archiveName;
+        var modName = defaultModName ?? (download.Contains(DownloadAnalysis.SuggestedName)
+            ? download.SuggestedName
+            : "<unknown>");
 
+        var dlTx = _conn.BeginTransaction();
```
erri120, line 58: This should be disposed.

erri120 (changes requested):

PR #1333, opened 2024-05-09
Fix Design Preview process accessing and locking the database

Al12rs:
This prevents the Design Preview process from locking the database, blocking actually running the app on windows after opening any xaml file in the IDE.

This makes it use the inMemoryDatabase.


Sewer56 (approved):
Looks reasonable to me.
In a way it feels like a hack, but at the same time, this will make designer start up faster, so I see it as a win.

erri120 (approved):

PR #1337, opened 2024-05-13
Regression: Hide Game Files

erri120:
Regressed in #1302. Same as #1286 for #1195.


Al12rs (approved):
The game files mod can still be disabled and deleted if is made visible in the settings.
Keep the #1195 open until we have that as well.

PR #1344, opened 2024-05-13
Don't manage saves

erri120:


Sewer56:
Not better to comment it out? Would be easier to re-add later.

Sewer56 (approved):

PR #1346, opened 2024-05-14
Edit text files

### LINKED ISSUE #916: As a user I want to open and edit Mod Files
For SDV we sometimes want to edit ini files, for this ticket we'll focus on delivering the following:

1. In the "View Mod Contents" panel, allow users to open a file for which we have registered a hangler
2. Add a handler for .ini and .json files that opens a new page, with a https://github.com/AvaloniaUI/AvaloniaEdit control
3. This panel should have a save button. If the user clicks "save" archive the text file, and replace the hash and size of the mod file with the new metadata. If they close the tab or do anything else discard the changes. 

For now this really is as simple as "Save" and a AvaloniaEdit control stacked together. We'll iterate on the UI once we're ready with some feedback.

Steps for updating the loadout item:

* get the LoadoutItemId from whatever creates the edit panel
* user edits the file
* on save, archive the new file data, and update the LoadoutItem to have the new Hash/Size of the edited file

erri120:
Resolves #916.

- Adds a Text Editor Page
- Adds a button to the mod contents page to open the text editor
- Saves the changes and updates the original file model

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,6 +1,7 @@
 <Project>
   <ItemGroup>
-    <!-- Custom Packages -->
+    <PackageVersion Include="Avalonia.AvaloniaEdit" Version="11.0.6" />
+    <PackageVersion Include="AvaloniaEdit.TextMate" Version="11.0.6" />
     <PackageVersion Include="FlatSharp.Compiler" Version="7.6.0" />
     <PackageVersion Include="FlatSharp.Runtime" Version="7.6.0" />
     <PackageVersion Include="LinqGen" Version="0.3.1" />
@@ -16,28 +17,20 @@
     <PackageVersion Include="NexusMods.Telemetry.OpenTelemetry" Version="1.0.0" />
     <PackageVersion Include="FomodInstaller.Interface" Version="1.2.0" />
     <PackageVersion Include="FomodInstaller.Scripting.XmlScript" Version="1.0.0" />
-    <!-- Bannerlord -->
     <PackageVersion Include="Bannerlord.LauncherManager" Version="1.0.76" />
     <PackageVersion Include="FetchBannerlordVersion" Version="1.0.6.45" />
     <PackageVersion Include="System.Linq" Version="4.3.0" />
     <PackageVersion Include="System.Text.Json" Version="8.0.3" />
-  </ItemGroup>
-  <ItemGroup>
-    <!-- Open Telementry -->
+    <PackageVersion Include="TextMateSharp.Grammars" Version="1.0.56" />
     <PackageVersion Include="OpenTelemetry" Version="1.8.1" />
     <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.8.1" />
     <PackageVersion Include="OpenTelemetry.Extensions.Hosting" Version="1.8.1" />
-  </ItemGroup>
-  <ItemGroup>
     <PackageVersion Include="JetBrains.Annotations" Version="2023.3.0" PrivateAssets="all" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="8.0.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageVersion>
     <PackageVersion Include="NetEscapades.EnumGenerators" Version="1.0.0-beta07" PrivateAssets="all" ExcludeAssets="compile;runtime" />
-  </ItemGroup>
-  <ItemGroup>
-    <!-- Avalonia -->
     <PackageVersion Include="Avalonia" Version="11.1.0-beta2" />
     <PackageVersion Include="Avalonia.Controls.DataGrid" Version="11.1.0-beta2" />
     <PackageVersion Include="Avalonia.Controls.TreeDataGrid" Version="11.0.2" />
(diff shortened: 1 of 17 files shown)


Directory.Packages.props
```diff
@@ -16,28 +17,20 @@
...
     <PackageVersion Include="FomodInstaller.Interface" Version="1.2.0" />
     <PackageVersion Include="FomodInstaller.Scripting.XmlScript" Version="1.0.0" />
-    <!-- Bannerlord -->
     <PackageVersion Include="Bannerlord.LauncherManager" Version="1.0.76" />
     <PackageVersion Include="FetchBannerlordVersion" Version="1.0.6.45" />
     <PackageVersion Include="System.Linq" Version="4.3.0" />
     <PackageVersion Include="System.Text.Json" Version="8.0.3" />
-  </ItemGroup>
-  <ItemGroup>
-    <!-- Open Telementry -->
```
Al12rs, line 26: Any reasons these were removed?

erri120: No point in organizing the file.

src/NexusMods.App.UI/Pages/TextEdit/TextEditorPage.cs
```diff
@@ -0,0 +1,31 @@
...
+using NexusMods.Hashing.xxHash64;
+using NexusMods.Paths;
+
+namespace NexusMods.App.UI.Pages.TextEdit;
+
+[JsonName("TextEditorPageContext")]
+public record TextEditorPageContext : IPageFactoryContext
+{
+    public required Hash FileHash { get; init; }
+    public required RelativePath FileName { get; init; }
```
Al12rs, line 14: Don't you also need the ModId ?

Al12rs (changes requested):

src/NexusMods.App.UI/Controls/ModInfo/ModFiles/ModFilesViewModel.cs
```diff
@@ -1,20 +1,111 @@
...
+
+            // NOTE(erri120): not a huge fan of this, but it works
+            var db = connection.Db;
+            var storedFile = db
+                .Find(StoredFile.Hash)
+                .Select(id => db.Get<StoredFile.Model>(id))
+                .FirstOrDefault(storedFile =>
+                {
+                    if (!storedFile.To.Equals(key)) return false;
+                    return targetModId.Equals(storedFile.ModId);
```
Al12rs, line 54: Check if modId is the same first, should be faster I think if you have lots of files.

src/NexusMods.App.UI/Controls/ModInfo/ModFiles/ModFilesViewModel.cs
```diff
@@ -1,20 +1,111 @@
...
+
+            if (storedFile is null) return;
+
+            var pageData = new PageData
+            {
+                FactoryId = TextEditorPageFactory.StaticId,
+                Context = new TextEditorPageContext
+                {
+                    FileId = storedFile.FileId,
+                    FileName = key.FileName,
```
Al12rs, line 65: I would pass the entire relative path, not just the name.

src/NexusMods.App.UI/Pages/TextEdit/TextEditorPageViewModel.cs
```diff
@@ -0,0 +1,122 @@
...
+            var db = connection.Db;
+            var storedFile = db.Get<StoredFile.Model>(fileId.Value);
+
+            using (var tx = connection.BeginTransaction())
+            {
+                tx.Add(storedFile.Id, StoredFile.Hash, hash);
+                tx.Add(storedFile.Id, StoredFile.Size, size);
+                storedFile.Remap<File.Model>().Mod.Revise(tx);
+                await tx.Commit();
+            }
```
Al12rs, line 93: You should probably reset `IsModified` somewhere here, button stays enabled after first save.

Al12rs (changes requested):

src/NexusMods.App.UI/Pages/TextEdit/TextEditorPageViewModel.cs
```diff
@@ -0,0 +1,122 @@
...
+        }, canSaveObservable);
+
+        this.WhenActivated(disposables =>
+        {
```
Al12rs, line 97: Would it be an idea to listen to the File Revisions from DB and reloading the page based on that?

This would keep two pages of the same document in synch if one makes changes and saves.

src/NexusMods.App.UI/Controls/ModInfo/ModFiles/ModFilesViewModel.cs
erri120: Why?

Al12rs: Why do we need the filename? Isn't it in `File.To`?

src/NexusMods.App.UI/Pages/TextEdit/TextEditorPageViewModel.cs
erri120: Like this?

https://github.com/Nexus-Mods/NexusMods.App/assets/16577545/1244314e-c76f-4d32-b0b3-433730e74d26

Al12rs: Cool, I guess the main issue would be if someone has both editors with changes, you are going to lose the progress of one of the two, but that is going to happen regardless. At least now it is clearer what is the latest version of the file.

Al12rs: Works great with IsModified updates

src/NexusMods.App.UI/Pages/TextEdit/TextEditorPageViewModel.cs
```diff
@@ -0,0 +1,145 @@
...
+                .InvokeCommand(_loadFileCommand)
+                .DisposeWith(disposables);
+
+            this.WhenAnyObservable(vm => vm._loadFileCommand)
+                .OnUI()
+                .SubscribeWithErrorLogging(output =>
+                {
+                    var (context, contents) = output;
+                    var filePath = context.FilePath;
+                    TabTitle = filePath.Path.ToString();
```
Al12rs, line 133: Tab name should still be FileName. It is convenient to have access to the full for any future changes.

Al12rs: If you hover over it you get the full path, so I guess it isn't that bad
image

Al12rs (approved):

PR #1347, opened 2024-05-14
Finish Mod Library

### LINKED ISSUE #1155: Download Library - Download list
As a user
I want to see a list of downloaded mods
So that I can see all the mods I have downloaded

Its important for the users to be able to see all the things they have downloaded and it be displayed in a why they can see which mods relates to which game and if they are associated with any loadouts.

A/C

Can be sorted by version
Can be sorted/filtered by size

- #1154 

Design:
- #1210

### LINKED ISSUE #1210: Download Library - UI
Based on the discussion we had last week the design will take into account the 3 states that a mod can be in. 

- Downloaded - Not installed
- Installed and disabled
- Installed and enabled

Design will take the approach of having a download library page utilising the existing download page, an additional button to add mods to a loadout at which point they install. History of downloads will also be available.

Deleting mods does not remove from disk but disables them and removes from view. 

The discussion about how to remove from disk is part of the garbage collection epic. 

### Designs

https://www.figma.com/file/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1Mod-library?type=design&node-id=2%3A1825&mode=design&t=zZbzsbWvLGDOXKsb-1

image

Create one version with add and the one version with the additional add duplicate button.

Epic:
- https://github.com/Nexus-Mods/NexusMods.App/issues/1154

### LINKED ISSUE #1284: Mod Library Refinements

Al12rs:
- Closes #1284
- Closes #1210 
- Closes #1155

This implements the remaining missing features of the Mod Library as current designs.

### Summary:
- Filter entries by Game, if entry has no Game, it is shown.
- Compute and show "Downloaded" DateTime column which is either when a manual archive was first analyzed, or when a download was started.
- Show Version data if available (it is only present on NXM downloads)
- Compute and update "Added" DateTime column by querying the loadout mods for matches and getting the most recent CreatedAt time
- Excluded Nested archives (e.g. SMAPI internal archive) from showing up
- Used Nexus Mod page name if available instead of file name

I'm not completely happy with the subscriptions and the all the attributes shenanigans, but at least everything seems to work now.

image

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.FileStore/Downloads/DownloadAnalysis.cs
+++ b/src/Abstractions/NexusMods.Abstractions.FileStore/Downloads/DownloadAnalysis.cs
@@ -6,6 +6,7 @@ using NexusMods.MnemonicDB.Abstractions;
 using NexusMods.MnemonicDB.Abstractions.Attributes;
 using NexusMods.MnemonicDB.Abstractions.IndexSegments;
 using NexusMods.MnemonicDB.Abstractions.Models;
+using NexusMods.MnemonicDB.Storage;
 using NexusMods.Paths;
 using ModFileTreeNode = NexusMods.Paths.Trees.KeyedBox<NexusMods.Paths.RelativePath, NexusMods.Abstractions.FileStore.Trees.ModFileTree>;
 
@@ -100,6 +101,21 @@ public static class DownloadAnalysis
         public ModFileTreeNode GetFileTree(IFileStore? fs = null)
         {
             return TreeCreator.Create(Contents, fs);
+        }
+        
+        /// <summary>
+        /// The timestamp of the creation of this <see cref="DownloadAnalysis"/> entity
+        /// Download start time or first manual installation time
+        /// </summary>
+        public DateTime CreatedAt
+        {
+            get
+            {
+                // Get the lowest transaction id, then get the timestamp of that transaction
+                var t = this.Select(d => d.T).Min();
+                var txEntity = Db.Get<Entity>(EntityId.From(t.Value));
+                return BuiltInAttributes.TxTimestanp.Get(txEntity);
+            }
         }
     }
 }
--- a/src/Abstractions/NexusMods.Abstractions.FileStore/Downloads/DownloadAnalysis.cs
+++ b/src/Abstractions/NexusMods.Abstractions.FileStore/Downloads/DownloadAnalysis.cs
@@ -102,20 +102,17 @@ public static class DownloadAnalysis
         {
             return TreeCreator.Create(Contents, fs);
         }
-        
+
         /// <summary>
         /// The timestamp of the creation of this <see cref="DownloadAnalysis"/> entity
         /// Download start time or first manual installation time
         /// </summary>
-        public DateTime CreatedAt
+        public DateTime GetCreatedAt()
         {
-            get
-            {
-                // Get the lowest transaction id, then get the timestamp of that transaction
-                var t = this.Select(d => d.T).Min();
-                var txEntity = Db.Get<Entity>(EntityId.From(t.Value));
-                return BuiltInAttributes.TxTimestanp.Get(txEntity);
-            }
+            // Get the lowest transaction id, then get the timestamp of that transaction
+            var t = this.Select(d => d.T).Min();
+            var txEntity = Db.Get<Entity>(EntityId.From(t.Value));
+            return BuiltInAttributes.TxTimestanp.Get(txEntity);
         }
     }
 }
--- a/src/Abstractions/NexusMods.Abstractions.FileStore/TypeFinder.cs
+++ b/src/Abstractions/NexusMods.Abstractions.FileStore/TypeFinder.cs
@@ -17,6 +17,6 @@ internal class TypeFinder : ITypeFinder
         typeof(FilePathMetadata),
         typeof(GameArchiveMetadata),
         typeof(DownloadAnalysis),
-        typeof(NestedArchiveMetadata),
+        typeof(StreamBasedFileOriginMetadata),
     };
 }
--- a/src/Abstractions/NexusMods.Abstractions.FileStore/TypeFinder.cs
+++ b/src/Abstractions/NexusMods.Abstractions.FileStore/TypeFinder.cs
@@ -17,5 +17,6 @@ internal class TypeFinder : ITypeFinder
         typeof(FilePathMetadata),
         typeof(GameArchiveMetadata),
         typeof(DownloadAnalysis),
+        typeof(NestedArchiveMetadata),
     };
 }
--- a/src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
@@ -1,12 +1,11 @@
 using System.Reactive;
 using System.Reactive.Linq;
-using DynamicData;
-using DynamicData.Aggregation;
 using Humanizer;
 using NexusMods.Abstractions.FileStore.Downloads;
 using NexusMods.Abstractions.Installers;
 using NexusMods.Abstractions.Loadouts;
 using NexusMods.Abstractions.Loadouts.Ids;
+using NexusMods.Abstractions.Loadouts.Mods;
 using NexusMods.MnemonicDB.Abstractions;
 using NexusMods.Networking.Downloaders.Tasks.State;
 using NexusMods.Paths;
@@ -24,9 +23,8 @@ public class FileOriginEntryViewModel : AViewModel<IFileOriginEntryViewModel>, I
     
     [ObservableAsProperty]
     public string DisplayArchiveDate { get; } = "-";
-
     
-    [ObservableAsProperty]
+    [Reactive]
     public DateTime LastInstalledDate { get; set; } 
     
     [ObservableAsProperty]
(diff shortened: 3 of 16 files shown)


src/Abstractions/NexusMods.Abstractions.FileStore/TypeFinder.cs
erri120, line 1: This entire file can probably be removed as it's not used anymore. @halgari.

src/Abstractions/NexusMods.Abstractions.FileStore/Downloads/DownloadAnalysis.cs
```diff
@@ -101,6 +102,21 @@ public ModFileTreeNode GetFileTree(IFileStore? fs = null)
...
+        public DateTime CreatedAt
+        {
+            get
+            {
+                // Get the lowest transaction id, then get the timestamp of that transaction
+                var t = this.Select(d => d.T).Min();
+                var txEntity = Db.Get<Entity>(EntityId.From(t.Value));
+                return BuiltInAttributes.TxTimestanp.Get(txEntity);
+            }
+        }
```
erri120, line 119: Should be a method `GetCreatedAt` instead of a property.

src/NexusMods.DataModel/FileOriginRegistry.cs
```diff
@@ -60,7 +62,14 @@ public async ValueTask<DownloadId> RegisterDownload(IStreamFactory factory, Meta
...
         await _extractor.ExtractAllAsync(factory, tmpFolder.Path, token);
-        return await RegisterFolderInternal(tmpFolder.Path, metaData, null, _fileStore.GetFileHashes(), archiveHash.Value, archiveSize, token);
+
+        return await RegisterFolderInternal(tmpFolder.Path, AppendNestedArchiveMetadata, null, _fileStore.GetFileHashes(), archiveHash.Value, archiveSize, token);
+
+        void AppendNestedArchiveMetadata(ITransaction tx, EntityId id)
+        {
+            metaData?.Invoke(tx, id);
+            tx.Add(id, NestedArchiveMetadata.NestedArchive, new Null());
+        }
```
erri120, line 72: Not sure I like this solution. You're calling this "Nested Archive" because the only usage of this is from trying to use a nested archive, however, the method is for registering a generic stream. It's not guaranteed to come from a nested archive.

You should rename this method to something more representative.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/IFileOriginsPageViewModel.cs
```diff
@@ -8,7 +7,9 @@ namespace NexusMods.App.UI.Pages.ModLibrary;
 
 public interface IFileOriginsPageViewModel : IPageViewModelInterface
 {
-    LoadoutId LoadoutId { get; set; }
+    LoadoutId LoadoutId { get; }
     
     ReadOnlyObservableCollection<IFileOriginEntryViewModel> FileOrigins { get; }
+    
+    void Initialize(LoadoutId loadoutId);
```
erri120, line 14: Don't do this. Just make `LoadoutId` reactive and move the implementation into a `WhenAnyValue` statement.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -15,40 +13,59 @@
...
+                                         || domain == game.Domain)
+                                        && !downloadAnalysis.Contains(NestedArchiveMetadata.NestedArchive)
+            )
+            .Transform(fileOrigin => (IFileOriginEntryViewModel)new FileOriginEntryViewModel(
+                    _conn,
+                    _archiveInstaller,
+                    LoadoutId,
+                    fileOrigin
+                )
+            )
```
erri120, line 68: This has some whack formatting.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -15,40 +13,59 @@
...
-                    {
-                        await archiveInstaller.AddMods(LoadoutId, fileOrigin);
-                    }
-                ),
-            })
+            .Filter(downloadAnalysis => (!downloadAnalysis.TryGet(DownloaderState.GameDomain, out var domain) 
+                                         || domain == game.Domain)
+                                        && !downloadAnalysis.Contains(NestedArchiveMetadata.NestedArchive)
+            )
+            .Transform(fileOrigin => (IFileOriginEntryViewModel)new FileOriginEntryViewModel(
```
erri120, line 62: `OnUI` should be before the transform to prevent weirdness.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
```diff
@@ -9,33 +16,60 @@ namespace NexusMods.App.UI.Pages.ModLibrary.FileOriginEntry;
...
+    public string Version { get; }
+    public Size Size { get; }
+    public DateTime ArchiveDate { get; }
+    public ReactiveCommand<Unit, Unit> AddToLoadoutCommand { get; init; }
+
+    [ObservableAsProperty] public string DisplayArchiveDate { get; } = "-";
+
+    [ObservableAsProperty] public DateTime LastInstalledDate { get; set; }
+
+    [ObservableAsProperty] public string DisplayLastInstalledDate { get; } = "-";
```
erri120, line 29: I really hate this attribute. It doesn't do what you expect it to do. Reflection is used at runtime to assign to the backing field. This attribute is pure evil, and shouldn't be used.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
```diff
@@ -9,33 +16,60 @@ namespace NexusMods.App.UI.Pages.ModLibrary.FileOriginEntry;
...
+        Name = fileOrigin.TryGet(DownloaderState.FriendlyName, out var friendlyName) && friendlyName != "Unknown"
+            ? friendlyName
+            : fileOrigin.SuggestedName;
+        Size = fileOrigin.TryGet(DownloaderState.Size, out var size) ? size : Size.Zero;
+        AddToLoadoutCommand = ReactiveCommand.CreateFromTask(async () => { await archiveInstaller.AddMods(loadoutId, fileOrigin); }
+        );
+        Version = fileOrigin.TryGet(DownloaderState.Version, out var version) && version != "Unknown"
+            ? version
+            : "-";
+        ArchiveDate = fileOrigin.CreatedAt;
```
erri120, line 46: Formatting is all over the place.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
```diff
@@ -9,33 +16,60 @@ namespace NexusMods.App.UI.Pages.ModLibrary.FileOriginEntry;
...
+            .Select(date => date.Equals(DateTime.MinValue) ? "-" : date.Humanize())
             .ToPropertyEx(this, vm => vm.DisplayArchiveDate);
 
+        // Update the LastInstalledDate every time the loadout is updated
+        conn.Revisions(loadoutId)
+            .StartWith(loadout)
+            .Select(rev => rev.Mods.Where(mod => mod.Contains(Mod.Source)
+                                                 && mod.SourceId.Equals(fileOrigin.Id)
+                )
+                .Select(mod => mod.CreatedAt)
```
erri120, line 63: Weird formatting.

erri120 (changes requested):

src/NexusMods.DataModel/FileOriginRegistry.cs
Al12rs: Do you have suggestions, I couldn't think of anything so I used that.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/IFileOriginsPageViewModel.cs
Al12rs: I don't see the advantage. `LoadoutId` is only ever going to be set once, and that should only happen at initialization.
This code better represents what is going to happen and what is expected. 

WhenAnyValue would add the overhead of a subscription, and would be less expressive, so I thought this approach would be better.

src/Abstractions/NexusMods.Abstractions.FileStore/TypeFinder.cs
Al12rs: Without adding this it wouldn't work

src/NexusMods.DataModel/FileOriginRegistry.cs
erri120: Call it what it is, a stream-based download. It's not from a file, but from a generic stream.

src/Abstractions/NexusMods.Abstractions.FileStore/Downloads/DownloadAnalysis.cs
Al12rs: I copied what was already on the Mod entity:
https://github.com/Nexus-Mods/NexusMods.App/blob/db8503381d4e3b1946cb8df8a09db77b6a396ba1/src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/Mod.cs#L165

Should I change both or leave as is?

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/IFileOriginsPageViewModel.cs
erri120: Then put `LoadoutId` as a parameter into the constructor. That's what the constructor is for, it initializes the class. Having a method called `Initialize` that does class initialization is literally the job of a constructor. Obviously wouldn't be able to use DI, and have to manually constructor the class, but it's still better than this evil.

src/Abstractions/NexusMods.Abstractions.FileStore/Downloads/DownloadAnalysis.cs
erri120: Change the other as well.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/IFileOriginsPageViewModel.cs
Al12rs: Not using DI isn't nice either. This pattern isn't uncommon for DI scenarios and for Async initialization. 
It's a tradeoff, I don't know if this is so bad to be worth changing.

erri120: It's an antipattern.

- Constructor: no overhead, can't create VM using DI, usage is similar to other VMs
- Reactive Property: tiny overhead, create VM using DI, usage is the same as other VMs
- Initialize Method: tiny overhead, create VM using DI, usage is different to other VMs

With a reactive property, you can take this VM everywhere, and it follows the same patterns as everything else we have in our code (we do use `Initialize` in some places and I really want to make those disappear as well).

Al12rs: Done

src/Abstractions/NexusMods.Abstractions.FileStore/TypeFinder.cs
erri120: Huh, weird.

erri120 (approved):

PR #1351, opened 2024-05-15
Add a setting to not log to the console, just to log files

halgari:
I've intended to add this for months now. Adds a setting that removes the log to console feature of the app so that logs only go into the log files. Reduces spam on the console and in the pop-up window. The console window still appears on Windows as that's a strange windows thing we need to look into at some point, but likely after we merge #1345

### DIFF
--- a/src/NexusMods.App/LoggingSettings.cs
+++ b/src/NexusMods.App/LoggingSettings.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using NexusMods.Abstractions.Settings;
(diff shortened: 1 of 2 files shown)


src/NexusMods.App/Program.cs
```diff
@@ -222,13 +222,21 @@ private static void AddLogging(ILoggingBuilder loggingBuilder, LoggingSettings s
...
+            var consoleTarget = new ConsoleTarget("console")
+            {
+                Layout = "${processtime} [${level:uppercase=true}] ${message:withexception=true}",
+            };
+            config.AddRuleForAllLevels(consoleTarget);
+        }
+        else
+        {
+            
+        }
```
erri120, line 236:
```suggestion
```

src/NexusMods.App/LoggingSettings.cs
```diff
@@ -40,6 +40,11 @@ public record LoggingSettings : ISettings
     /// </summary>
     public LogLevel MinimumLevel { get; init; } = LogLevel.Debug;
 
+    /// <summary>
+    /// When enabled, logs will be written to the console as well as the log file.
+    /// </summary>
+    public bool LogToConsole { get; init; } = false;
```
erri120, line 46: Can we keep this on `true` for debug builds?

erri120 (changes requested):

src/NexusMods.App/LoggingSettings.cs
halgari: Sure, I'll add that

erri120 (approved):

PR #1358, opened 2024-05-15
Ingest mod files

### LINKED ISSUE #1343: Add config files to mods for SDV (really their own loadout item)

erri120:
Resolves #1343.

Changes:

- Fixes an issue with the Apply Service and Disk State Registry where they use a default value for the last applied loadout
- Ingests new files into the correct SMAPI mod instead of "Override"

### DIFF
--- a/src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
+++ b/src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
@@ -71,8 +71,16 @@ public class StardewValleyLoadoutSynchronizer : ALoadoutSynchronizer
         var manifestFile = db
             .Find(File.To)
             .Select(db.Get<File.Model>)
-            .Where(model => model.Contains(File.Loadout))
-            .FirstOrDefault(file => file.LoadoutId.Equals(loadout.LoadoutId) && file.To.Equals(manifestFilePath));
+            .FirstOrDefault(file =>
+            {
+                if (!file.Contains(File.Loadout)) return false;
+                if (!file.LoadoutId.Equals(loadout.LoadoutId)) return false;
+
+                if (!file.To.Equals(manifestFilePath)) return false;
+
+                if (!file.Contains(File.Mod)) return false;
+                return file.Mod.Enabled;
+            });
 
         return manifestFile?.Mod;
     }
--- a/src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
+++ b/src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
@@ -19,6 +19,8 @@ public class StardewValleyLoadoutSynchronizer : ALoadoutSynchronizer
         using var tx = Connection.BeginTransaction();
         var overridesMod = GetOrCreateOverridesMod(loadout, tx);
 
+        var smapiModDirectoryNameToModel = new Dictionary<RelativePath, Mod.Model>();
+
         foreach (var newFile in newFiles)
         {
             newFile.Add(File.Loadout, loadout.Id);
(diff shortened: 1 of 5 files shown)


src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
```diff
@@ -0,0 +1,94 @@
...
+            newFile.AddTo(tx);
+        }
+    }
+
+    private static Mod.Model? GetSMAPIMod(RelativePath modDirectoryName, Loadout.Model loadout, IDb db)
+    {
+        var manifestFilePath = new GamePath(LocationId.Game, Constants.ModsFolder.Join(modDirectoryName).Join(Constants.ManifestFile));
+
+        var manifestFile = db
+            .Find(File.To)
```
halgari, line 72: The syntax here is odd, but if this is `.FindIndexed(manifestFilePath, File.To)` that will return only files where `To == manifestFilePath`. That should get rid of the need to scan everything. Still not optimal we'll get some better query logic for this at some poitn.

halgari (changes requested):

src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
```diff
@@ -0,0 +1,94 @@
...
+                if (smapiMod is null)
+                {
+                    AddToOverride(newFile);
+                    continue;
+                }
+
+                smapiModDirectoryNameToModel[modDirectoryName] = smapiMod;
+            }
+
+            newFile.Add(File.Mod, smapiMod.Id);
```
Al12rs, line 53: Just a heads up that this will include Deletions in the mods as well.

Al12rs (changes requested):

src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
erri120: I thought `FindIndexed` only works if the attribute has been marked with `IsIndexed`, which `File.To` is not.

src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
```diff
@@ -0,0 +1,94 @@
...
+
+    private static Mod.Model? GetSMAPIMod(RelativePath modDirectoryName, Loadout.Model loadout, IDb db)
+    {
+        var manifestFilePath = new GamePath(LocationId.Game, Constants.ModsFolder.Join(modDirectoryName).Join(Constants.ManifestFile));
+
+        var manifestFile = db
+            .Find(File.To)
+            .Select(db.Get<File.Model>)
+            .Where(model => model.Contains(File.Loadout))
+            .FirstOrDefault(file => file.LoadoutId.Equals(loadout.LoadoutId) && file.To.Equals(manifestFilePath));
```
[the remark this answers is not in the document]
erri120: Does it?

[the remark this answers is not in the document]
Al12rs: Yes, otherwise you are adding stuff in disabled mods.

Al12rs (approved):
ok from me

src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
halgari: Ah, we should certainly add IsIndexed to .To then, we'll want that for this attribute.

erri120: Is this a trivial change? I want to keep this PR focused.

halgari: Yeah, no big deal we can fix this later

halgari (approved):

PR #1362, opened 2024-05-15
Upgrade (almost) all the deps

halgari:
Upgrades almost all the dependencies in the app that are behind. Does not upgrade a SMAPI dependency (so we stay in sync with SMAPI) or the Bannerlord launcer code (that has a breaking change in a minor release)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -40,13 +40,13 @@
     <PackageVersion Include="NetEscapades.EnumGenerators" Version="1.0.0-beta07" PrivateAssets="all" ExcludeAssets="compile;runtime" />
     <PackageVersion Include="Avalonia" Version="11.1.0-beta2" />
     <PackageVersion Include="Avalonia.Controls.DataGrid" Version="11.1.0-beta2" />
-    <PackageVersion Include="Avalonia.Controls.TreeDataGrid" Version="11.0.2" />
+    <PackageVersion Include="Avalonia.Controls.TreeDataGrid" Version="11.0.10" />
     <PackageVersion Include="Avalonia.Desktop" Version="11.1.0-beta2" />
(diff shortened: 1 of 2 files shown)


Directory.Packages.props
```diff
@@ -57,21 +57,21 @@
...
     <PackageVersion Include="Microsoft.Extensions.Logging" Version="8.0.0" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.1" />
-    <PackageVersion Include="Microsoft.Extensions.ObjectPool" Version="8.0.4" />
+    <PackageVersion Include="Microsoft.Extensions.ObjectPool" Version="8.0.5" />
     <PackageVersion Include="Microsoft.IO.RecyclableMemoryStream" Version="3.0.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.8.0" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.8.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.9.2" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.9.2" />
```
erri120, line 70: These package versions are tied to the roslyn compiler version

erri120: As seen here: https://github.com/Nexus-Mods/NexusMods.App/pull/1064

halgari: yep, good point

erri120 (approved):

erri120:
This is what you usually do with those packages: https://github.com/meziantou/Meziantou.Analyzer/blob/main/Directory.Build.targets

PR #1364, opened 2024-05-15
Fix mod contents showing deletions as normal files and other fixes

### LINKED ISSUE #1323: Reified deletions show up as normal files in Mod Contents View
#1302 added implemented reified deletes, so now mods can contain files that represent a deletion.

These files are currently shown as normal files in the Mod Contents viewer (file tree).

How should we handle them?

- [ ] Skip them (we get mods that appear to be empty)
- [ ] Have special icon/styling for those entries
- [ ] Show deletions in a separate tree

### LINKED ISSUE #1339: TabHeader icons foreground doesn't seem to update in some cases
Non selected tabs should use a darker color for the foreground icons. It works for some icons and not for others.
image

### LINKED ISSUE #1340: ModContents Page Header Icon is missing
image
Should use the Icon on the View Mod Files button:
image

Al12rs:
- fixes #1323 
- fixes #1340
- fixes #1339
- fixes tab header text and icon vertical alignment
- fixes Advanced installer window missing a title

### DIFF
--- a/src/NexusMods.Icons/UnifiedIcon.cs
+++ b/src/NexusMods.Icons/UnifiedIcon.cs
@@ -81,6 +81,10 @@ public sealed class UnifiedIcon : ContentControl
 
             MaxHeight = value;
             MaxWidth = value;
+        } else if (change.Property == ForegroundProperty && Content is PathIcon)
+        {
+            // Note(Al12rs): workaround to update PathIcon foreground colors
+            ((PathIcon)Content).Foreground = (IBrush)change.NewValue!;
         }
     }
 
@@ -117,6 +121,7 @@ public sealed class UnifiedIcon : ContentControl
                 [WidthProperty] = this[WidthProperty],
                 [MaxHeightProperty] = this[MaxHeightProperty],
                 [MaxWidthProperty] = this[MaxWidthProperty],
+                [ForegroundProperty] = this[ForegroundProperty],
             }
         );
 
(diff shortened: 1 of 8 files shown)


src/NexusMods.Icons/UnifiedIcon.cs
```diff
@@ -81,6 +81,10 @@ protected override void OnPropertyChanged(AvaloniaPropertyChangedEventArgs chang
 
             MaxHeight = value;
             MaxWidth = value;
+        } else if (change.Property == ForegroundProperty && Content is PathIcon)
+        {
+            // Note(Al12rs): workaround to update PathIcon foreground colors
+            ((PathIcon)Content).Foreground = (IBrush)change.NewValue!;
```
erri120, line 87: This shouldn't be necessary.

Al12rs: Yeah, that is what I thought as well, but without this the icon only takes the initial foreground and doesn't respond to updates. So this is the workaround

erri120, line 87:
```suggestion
        } else if (change.Property == ForegroundProperty && Content is PathIcon pathIcon)
        {
            // Note(Al12rs): workaround to update PathIcon foreground colors
            pathIcon.Foreground = (IBrush?)change.NewValue;
```

erri120:
Seems to work fine. The strike-through looks very thin tho, but we can fix that later.

erri120:
Shouldn't this have a different icon:

Al12rs:
> 
> 
> Shouldn't this have a different icon:
> 
> 

I took the icon from the design. This is going to be the Mod Information page in the future, not just filetree.

src/NexusMods.Icons/UnifiedIcon.cs
Al12rs: Ran the changes locally and work fine👍

erri120 (approved):

PR #1381, opened 2024-05-20
Fix mod library styling

### LINKED ISSUE #1378: Mod Library is using the wrong font
# Bug Report

## Summary
The Mod Library and Loadout are using different fonts for some reason

halgari:
Fixes #1378

Old Mod Library text looked like this:
bad_font

When it should look like this:
compare

Updated the grid to use template cells with custom styles on the text blocks

Now looks like this:

good_font

Also fixed the strange padding on the actions buttons by widening the column a bit:

actions space


Sewer56 (approved):
Visually it looks fine to me.
I didn't validate with FIGMA though, I assume this PR brings it up to that.

PR #1382, opened 2024-05-20
Cancel any orphaned tasks on startup.

### LINKED ISSUE #1311: Closing and restarting the App while a Download is in progress doesn't automatically resume the download
Download appears stuck, no progress is shown.
User has to pause and resume the download manually to make it actually resume downloading.
image

halgari:
Fixes #1311. On a normal shutdown, all in-progress downloads are canceled. But in the case of a hard crash they are not cleaned up on app restart. This adds a cleanup of orphaned download tasks on the startup of the app.

### DIFF
--- a/src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
+++ b/src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
@@ -38,7 +38,7 @@ public class DownloadService : IDownloadService, IAsyncDisposable
         _conn = conn;
         _disposables = new CompositeDisposable();
         _fileStore = fileStore;
-
+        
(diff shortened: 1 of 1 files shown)


Sewer56 (approved):
No issues with the code, but I do have a question.

Is the plan to no longer resume downloads in the long run, or is this a temporary measure because free accounts can't resume `nxm://` downloads?

In the older versions of the App, last year, we'd auto resume downloads on startup.

halgari:
It's not really a decision one way or the other. Today the code auto-cancels on a clean shutdown this just brings the crashing behavior in-line with that. We'll certainly revisit this and/or make it a settings option in the future

PR #1383, opened 2024-05-20
Added: Ability to Uninstall the App

### LINKED ISSUE #377: As a user, I want to uninstall the Nexus Mods App
# User story

**As a** user

**I want to** uninstall the Nexus Mods App

**So that** it is no longer on my system

## Requirements

Consider how users uninstall the app, e.g. any wizard/splash screen we may provide.

We'll want to add an uninstallation survey, which will capture reasons for removing the app.

## Design

## DOD

## Scenario 1
**Given** _______

**When** _______

**Then** _______

### LINKED ISSUE #1161: Loadouts - Unmanage Game
Placeholder

As a user 
I want to be able to remove a game from being managed by the app
So that I know the app will not affect it anymore

Users need the ability to remove a game from being managed by the app. This would be part of the core user journey, if you can add a game to the app you must then be able to remove the game from the app. 
What does this mean? 
Should it remove loadouts and any app settings for that game? 
Mods will remain in the download library

### LINKED ISSUE #1200: Garbage Collection - Uninstall
When uninstalling the App what files does the user expect to be removed? 

Do we need to revert the game to a vanilla state?

Do any mods get deleted from disk?

Epic:
- https://github.com/Nexus-Mods/NexusMods.App/issues/1174

Sewer56:
This PR migrates the remaining code related to removing loadouts and app uninstallation.

- ✅ Uninstall script runs on Windows.
- ✅ Uninstalling with App Running on Windows (files in use) runs as intended.
    - On Windows, it kills all app instances and performs a manual cleanup via dynamically generated `powershell` script.
    - The `powershell` code written should be trivial enough that even 15+ year old versions of it should run it fine.
- ✅ Adds missing install and uninstall icon.
- ✅ Uninstalling deletes whole DataModel directory and...
    - Configured Logging Location
    - MnemonicDB Path
    - File Extractor Temp Folder
    - Logger base folder
    - Default Config folder for JSON Settings Storage backend
    - All archive locations where user has decided to store `.nx` archives.

I've tried to be comprehensive with the locations deleted, removing our defaults as well as any paths that are configurable. (***Even if they are not yet exposed via the Settings UI!!***). After deletion, no files of notable size should hopefully be left over on disk.

Note(s)
--------

- Still need to wait for upstream `https://github.com/kuiperzone/PupNet-Deploy/pull/36` to merge the necessary changes to PupNet.

fixes #377
fixes #1161
fixes #1200


Sewer56:
Oh, and of course.

fixes #377
fixes #1161

halgari (approved):

src/NexusMods.App/Commandline/CleanupVerbs.cs
```diff
@@ -0,0 +1,180 @@
...
+            scriptContent.AppendLine($"Remove-Item -Path \"{directory}\" -Recurse -Force -ErrorAction SilentlyContinue");
+
+        // Remove the script itself
+        scriptContent.AppendLine("Remove-Item -Path $MyInvocation.MyCommand.Path -Force -ErrorAction SilentlyContinue");
+
+        // Save the script to a temporary file
+        var tempFile = Path.GetTempFileName();
+        var tempScriptPath = tempFile.Replace(".tmp", ".ps1");
+        File.Delete(tempFile);
+        File.WriteAllText(tempScriptPath, scriptContent.ToString());
```
erri120, line 138: This should just be a file we ship, no need to generate this at runtime.

src/NexusMods.App/Commandline/CleanupVerbs.cs
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    private static async Task DeleteDirectoriesUnix(IRenderer renderer, IEnumerable<AbsolutePath> appDirectories)
+    {
+        foreach (var directory in appDirectories)
+        {
+            if (!directory.DirectoryExists())
+                continue;
+
+            directory.Delete();
```
erri120, line 157: `AbsolutePath.Delete` calls `IFileSystem.DeleteFile`. This should be `IFileSystem.DeleteDirectory`.

This should also be wrapped in a try/catch block.

erri120 (changes requested):

src/NexusMods.App/Commandline/CleanupVerbs.cs
Sewer56: There's actually a reason I wrote it this way.

The parameters `appFiles` and `appDirectories` are non-deterministic. Many of them can be customized by the user, or will be customizable by the user. For example, locations of `.nx` archives.

Therefore, if I were to ship this as a static file, I'd still need to pass locations to the script via parameters.

The edge case you get here is that the NT API uses [UNICODE_STRING](https://learn.microsoft.com/en-us/windows/win32/api/ntdef/ns-ntdef-_unicode_string) to start processes. The max length of that is technically 32K characters, however that is shared with some things such as the `Environment` block. In Microsoft's own software, such as `cmd`, they often therefore restrict the command length to `8191`.

Assuming `8191` as a 'safe' limit, passing all the paths as parameters yields a non-zero chance that the sum of all the paths is long enough that the length limit is exceeded. Because:

- The user could set a very long path to one of their locations. 
- We add more configurable locations to the App.
- The user sets a lot of locations where the limit is not bounded.
    - e.g. `Nx` archives can be set to an arbitrary number of locations.

If it happens that the command length is too long, the failure would (most likely) be silent, and thus no cleanup would be performed. That's problematic, as we'd then possibly never get a bug report for it.

That's why I chose to inline the deletes into the script.

Yes. The probability of this of course is all very low; a 'super edge case', but it still exists.
I can add this as a note to the code, however this is the reason why I chose to do it this way.

erri120: Make a static script and write the paths to a file in that case. Let the script iterate over each line in the generate text file to delete the directories.

The reason I'm so pedantic about this, is because generating powershell scripts at runtime is a big no no in terms of security. We have no idea what weird antivirus programs will do when they see this.

Sewer56: Pretty valid concern. Historically it hasn't caused me issues, but I might aswell.

I'll get that sorted after PR review.

erri120 (approved):

PR #1397, opened 2024-05-21
Completed downloads section on Downloads page

### LINKED ISSUE #888: Download Library - Add "Completed" section in InProgress download view
Add a "Completed (x)" section at the top of the Download view. This should show the completed downloads.

The completed downloads should have a Clear option to remove them from this view. Clear will not delete the downloads files, just hide it from the view.

This is to avoid users getting confused when a download disappears from the `InProgressView` after it finishes downloading.

### Design

[Figma designs](https://www.figma.com/file/Ntr9PttR5VrqpTRwwZHzLu/%F0%9F%93%B1-Mod-list---adding%2Fremoving-mods?type=design&node-id=3297%3A9207&mode=design&t=ks8dGvxaDXR4lzrM-1)

image

Al12rs:
- closes #888

### Breakdown:
- [x] Columns:
	- [x] Mod Name
	- [x] Version
	- [x] Game
	- [x] Size
	- [x] Completed DateTime
	- [x] Actions:
		- [x] View in library
		- [x] Clear (hide)
- [x] ToolBar:
	- [x] Clear button
	    - [x] CanExecute
	- [x] Clear completed button
	    - [x] CanExecute
	- [x] Fix Cancel button being enabled even with no selection
- [x] Functionality:
	- [x] Filter out hidden downloads
	- [x] Clear selected
	    - [x] CanExecute
	- [x] Clear all
	    - [x] CanExecute
	- [x] Only show Completed section if there are completed downloads
	- [x] View in library
	- [x] Counter of completed downloads
- [x] Backend data:
	- [x] Completed DateTime
	- [x] Hidden status
	- [x] Humanized size
	- [x] Humanized Date
- [x] Bindings
- [x] Use template rows to be able to set text styling
- [x] Don't use active AutoRefresh
- [x] Move all ICommands to ReactiveCommands
- [x] Fix bad async usages for existing methods
- [x] Translations
- [x] UI layout
	- [x] Put in scroller
	- [x] Fix layout 

### Skipped:
- Add collapsible sections (`Expander` needs to be restyled to match designs, which would have taken time)

image


src/Networking/NexusMods.Networking.Downloaders/Tasks/State/CompletedDownloadState.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        
+        /// <summary>
+        /// The timestamp the download was completed at
+        /// </summary>
+        public DateTime CompletedAt
+        {
+            get => CompletedDateTime.Get(this, default(DateTime));
+            set => CompletedDateTime.Add(this, value);
+        }
```
erri120, line 37: What's the reasoning behind not using the transaction timestamp?

src/Networking/NexusMods.Networking.Downloaders/Tasks/ADownloadTask.cs
```diff
@@ -236,6 +245,17 @@ private void UpdateActivity()
...
+    /// <inheritdoc />
+    public async Task SetIsHidden(bool isHidden, ITransaction? tx = null)
+    {
+        if (PersistentState.Status != DownloadTaskStatus.Completed)
+            return;
+        var transaction = tx ?? Connection.BeginTransaction();
+        transaction.Add(PersistentState.Id, CompletedDownloadState.Hidden, isHidden);
+        if (tx == null)
+            await transaction.Commit();
+    }
```
erri120, line 258: Why is `tx` nullable and why is this method async? Wouldn't this suffice?

```suggestion
    /// <inheritdoc />
    public void SetIsHidden(bool isHidden, ITransaction tx)
    {
        if (PersistentState.Status != DownloadTaskStatus.Completed) return;
        tx.Add(PersistentState.Id, CompletedDownloadState.Hidden, isHidden);
    }
```

src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
```diff
@@ -143,6 +143,14 @@ public Optional<Percent> GetTotalProgress()
         return Optional<Percent>.Create(Percent.CreateClamped(total / tasks.Length));
     }
 
+    /// <inheritdoc />
+    public async Task SetIsHidden(bool isHidden, IDownloadTask[] targets)
+    {
+        var tx = _conn.BeginTransaction();
```
erri120, line 149:
```suggestion
        using var tx = _conn.BeginTransaction();
```

src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
```diff
@@ -143,6 +143,14 @@ public Optional<Percent> GetTotalProgress()
...
+        await Task.WhenAll(targets.Select(x => x.SetIsHidden(isHidden, tx)));
```
erri120, line 150: See suggestion in `ADownloadTask`.

```suggestion
        foreach (var downloadTask in targets)
        {
            downloadTask.SetIsHidden(isHidden, tx);
        }
```

src/NexusMods.App.UI/Pages/Downloads/IInProgressViewModel.cs
```diff
@@ -30,6 +36,7 @@ public interface IInProgressViewModel : IPageViewModelInterface
     bool HasDownloads { get; }
 
     int ActiveDownloadCount { get; }
+    int CompletedDownloadCount { get; }
```
erri120, line 39: This doesn't have to exist, it's just `CompletedTasks.Count` which is already reactive.

src/NexusMods.App.UI/Pages/Downloads/InProgressView.axaml.cs
```diff
@@ -40,18 +42,34 @@ public InProgressView()
...
                 .DisposeWith(d);
 
             this.WhenAnyValue(view => view.ViewModel!.Columns)
-                .GenerateColumns(ModsDataGrid)
+                .GenerateColumns(InprogressDataGrid)
                 .DisposeWith(d);
 
+            // Fix the CompletedDataGrid Width when number of items changes
+            this.WhenAnyValue(view => view.ViewModel!.CompletedDownloadCount)
+                .OnUI()
```
erri120, line 66: `OnUI` statements should be in the VM, the View shouldn't have to do thread switching.

src/NexusMods.App.UI/Pages/Downloads/InProgressView.axaml.cs
```diff
@@ -108,26 +126,60 @@ public InProgressView()
                     }
                 })
                 .DisposeWith(d);
+            
+            // Dynamically Update Completed Download Count
+            this.WhenAnyValue(view => view.ViewModel!.CompletedDownloadCount)
+                .OnUI()
```
erri120, line 132: `OnUI` statements should be in the VM, the View shouldn't have to do thread switching.

src/NexusMods.App.UI/Pages/Downloads/InProgressView.axaml.cs
```diff
@@ -108,26 +126,60 @@ public InProgressView()
...
                 {
-                    ViewModel!.SelectedTasks.Edit(updater =>
+                    ViewModel!.SelectedInProgressTasks.Edit(updater =>
                     {
                         updater.Clear();
-                        foreach (var item in ModsDataGrid.SelectedItems)
+                        foreach (var item in InprogressDataGrid.SelectedItems)
+                        {
+                            if (item is IDownloadTaskViewModel task)
+                                updater.Add(task);
```
erri120, line 157: Can items be anything else? If not, this should have a `Debug.Assert`.

src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
erri120, line 1: This entire file has some interesting formatting choices.

src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
```diff
@@ -127,37 +142,103 @@ public class InProgressViewModel : APageViewModel<IInProgressViewModel>, IInProg
...
+                        {
+                            FactoryId = FileOriginsPageFactory.StaticId,
+                            Context = new FileOriginsPageContext { LoadoutId = workspace.Context.LoadoutId },
+                        };
+                        var behavior = GetWorkspaceController().GetOpenPageBehavior(pageData, navInfo, Optional<PageIdBundle>.None);
+                        
+                        controller.OpenPage(workspace.Id, pageData, behavior);
+                        controller.ChangeActiveWorkspace(workspace.Id);
+                    });
+                    vm.Activator.Activate();
```
erri120, line 201: This is a very bold statement, why is this here?

src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
```diff
@@ -127,37 +142,103 @@ public class InProgressViewModel : APageViewModel<IInProgressViewModel>, IInProg
...
+                        var pageData = new PageData
+                        {
+                            FactoryId = FileOriginsPageFactory.StaticId,
+                            Context = new FileOriginsPageContext { LoadoutId = workspace.Context.LoadoutId },
+                        };
+                        var behavior = GetWorkspaceController().GetOpenPageBehavior(pageData, navInfo, Optional<PageIdBundle>.None);
+                        
+                        controller.OpenPage(workspace.Id, pageData, behavior);
+                        controller.ChangeActiveWorkspace(workspace.Id);
+                    });
```
erri120, line 200: Why did you decide to use this approach instead of having empty commands where you listen to command execution to run this logic?

erri120 (changes requested):
We'd definitely wanna rework the downloads page completely in the near future. That thing is a horrifying mess of tech debt. For now, just check the comments I wrote.

src/Networking/NexusMods.Networking.Downloaders/Tasks/State/CompletedDownloadState.cs
Al12rs: Since the entity in question is created at download start, we would need to use the timestamp of the transaction that sets the Status to `Completed`. 

This means that if the status changes to something else and then back to Completed, the timestamp would be wrong. 
For example I considered implementing Hidden as a new Status, which would have complicated this.
Other potential status could be QueryingMetadata or other things. Rather than trying to guess the correct timestamp, I thought this was just less complicated and future proof.

src/Networking/NexusMods.Networking.Downloaders/Tasks/ADownloadTask.cs
Al12rs: Initial implementation allowed setting single tasks to hidden, without caller having to manage the transaction. Now everything uses the multi element version so I can change this

src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
Al12rs: Because I needed the task data and the Navigation info data for the logic and this approach made it much simpler to have that data available.
It also made async easier to deal with since it all goes into the `CreateFromTask`.

src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
Al12rs: Do you mean the Activator thing? That was there before. From my understanding `IDownloadTaskViewModel` is a fake VM that doesn't actually have a view, the individual columns take data from it, but not as the VM. Their VM is either `InProgressViewModel` or the column definition VM. 
Calling Activate there triggers the `WhenActivated()` of the VM. Personally I

src/NexusMods.App.UI/Pages/Downloads/InProgressView.axaml.cs
Al12rs: I think it can, I think headers and such.

src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
Al12rs: Maybe we should have a standard formatter.

erri120 (approved):

Issue #1367, opened 2024-05-16
Supported Games - SDV


LukeNexusMods:
The supported games section is currently showing games we don't support. 

For the Alpha please make sure its only showing Stardew Valley.

erri120:
Resolved in #1401.

PR #1401, opened 2024-05-21
Added: Ability to restrict which games we ship the Application with for Release Builds

Sewer56:
This PR introduces the ability to selectively enable or disable games from our Release builds.

We do this by straight up not compiling them into the App with the use of compile-time constants.

## Changes (Summary)

- Added a new `Directory.Build.targets` file to define compile-time constants across projects. 
    - Currently only Stardew Valley support is enabled in Release builds.
    - In Debug builds, all games are enabled by default. (Requested by @halgari)

- Updated the `NexusMods.App.csproj` file to only include project references when the relevant constants are defined.

- Moved conditionally enabled games to `AddSupportedGames` in `Services` class to keep things cleaner.

- Removed some unused `using` directives in startup project.

- Resolved cases where NuGet package(s) of now conditionally enabled projects were being imported transitively in tests, rather than directly. (Fix Build Errors)

- Added missing `DEBUG` conditional in `MainThreadData.cs` which would cause compile errors if unused using(s) were removed.

## Note

We couldn't use `Directory.Build.props`, it seems the constants we define there are overwritten further down the road, so I instead added a `Directory.Build.targets` file as that comes later down the build logic.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,7 +10,7 @@
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 
         <!-- 
-            Listing of all 
+            Listing of all build flags.
         
             Functionality:
 
--- a/src/NexusMods.App/ExperimentalSettings.cs
+++ b/src/NexusMods.App/ExperimentalSettings.cs
@@ -2,11 +2,10 @@ using NexusMods.Abstractions.Settings;
 
 namespace NexusMods.App;
 
-#if DEBUG
 /// <summary>
-/// Settings that are only available in debug releases.
+/// Settings that give access to experimental features.
 /// </summary>
-public record DebugSettings : ISettings
+public record ExperimentalSettings : ISettings
 {
     /// <summary>
     /// Enables games that are not enabled by default.
@@ -19,16 +18,15 @@ public record DebugSettings : ISettings
         var sectionId = SectionId.DefaultValue;
 
         return settingsBuilder
-            .ConfigureStorageBackend<DebugSettings>(builder => builder.UseJson())
-            .AddToUI<DebugSettings>(builder => builder
+            .ConfigureStorageBackend<ExperimentalSettings>(builder => builder.UseJson())
+            .AddToUI<ExperimentalSettings>(builder => builder
                 .AddPropertyToUI(x => x.EnableAllGames, propertybuilder => propertybuilder
                     .AddToSection(sectionId)
-                    .WithDisplayName("Enable all Games")
-                    .WithDescription("When set, all games will be enabled in the debug UI.")
+                    .WithDisplayName("Enable Unsupported Games")
+                    .WithDescription("When set, 'work-in-progress' games that are not yet fully supported will be enabled in the UI.")
                     .UseBooleanContainer()
                     .RequiresRestart()
                 )
             );
     }
 }
-#endif
--- a/src/NexusMods.App/Program.cs
+++ b/src/NexusMods.App/Program.cs
@@ -35,13 +35,13 @@ public class Program
 
         TelemetrySettings telemetrySettings;
         LoggingSettings loggingSettings;
-        DebugSettings debugSettings;
+        ExperimentalSettings experimentalSettings;
         using (var settingsHost = BuildSettingsHost())
         {
             var settingsManager = settingsHost.Services.GetRequiredService<ISettingsManager>();
             telemetrySettings = settingsManager.Get<TelemetrySettings>();
             loggingSettings = settingsManager.Get<LoggingSettings>();
-            debugSettings = settingsManager.Get<DebugSettings>();
+            experimentalSettings = settingsManager.Get<ExperimentalSettings>();
         }
 
         var startupMode = StartupMode.Parse(args);
@@ -50,7 +50,7 @@ public class Program
             startupMode,
             telemetrySettings,
             loggingSettings,
-            debugSettings
+            experimentalSettings
         );
         var services = host.Services;
 
@@ -176,7 +176,7 @@ public class Program
                 .AddSettingsStorageBackend<JsonStorageBackend>()
                 .AddSettings<TelemetrySettings>()
                 .AddSettings<LoggingSettings>()
-                .AddSettings<DebugSettings>()
+                .AddSettings<ExperimentalSettings>()
             )
             .ConfigureLogging((_, builder) => builder
                 .ClearProviders()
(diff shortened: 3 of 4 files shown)


Sewer56:
fixes https://github.com/Nexus-Mods/NexusMods.App/issues/1367

halgari:
> instead this is a hacky and unmaintainable solution which doesn't just exclude games from the UI, but from the entire build process.

@erri120 Can you go into more detail on the problems this will cause? Compile time switches for disabling modules is fairly common in a lot of applications. In the future we may want more options than just "Debug" or "Release" but that also is rather maintainable. Looking at this code, adding a new game will require adding abour 4-5 lines of code. I guess I'm not understanding where the hacks or unmaintainability comes into play here.

erri120:
> Can you go into more detail on the problems this will cause? Compile time switches for disabling modules is fairly common in a lot of applications. In the future we may want more options than just "Debug" or "Release" but that also is rather maintainable. Looking at this code, adding a new game will require adding abour 4-5 lines of code. I guess I'm not understanding where the hacks or unmaintainability comes into play here.

The `#if` preprocessor directive makes the code conditional, while the C# `if` statement makes the behavior conditional. Using conditional compilation means we now have multiple compilation configurations that result it different binaries. When we develop new features, we're using the `DEBUG` configuration, which enables all `#if DEBUG` blocks. Our IDE won't be able to report issues inside `#if RELEASE` blocks, because those have been conditionally excluded from compilation. Not only will our IDE not report issues inside these conditionally excluded blocks of code, it will also issue new warnings and maybe errors because of unused imports, classes, or other statements. @Sewer56 encountered this issue in this PR, he had to add a missing `#if DEBUG` directive to the imports in `MainThreadData`, because the compiler would otherwise complain in `RELEASE` mode.

Our IDE isn't the only place where these issues will pop up. Our build pipelines all use different configurations. The CI that runs on each PR uses `DEBUG`, our Git Builds pipeline uses `RELEASE` while our release pipeline uses `RELEASE` as well as our custom compile constants for `INSTALLATION_METHOD_*`. The binaries we get out of the Git Builds pipeline don't even match the ones we get out of the release pipeline because of conditional compilation.

I'm calling conditional compilation unmaintainable because you have to test each and every compilation configuration to determine whether the code compiles correctly and does what it's supposed to do. In contrast, using C# `if` statement means we don't have to worry about compilation and can focus on the behavior of our code.

As for this being "hacky", preprocessor directives are rarely a good choice, and never the only one. I regret adding `NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR` in #1308, that should've been a setting or something else, similar to the `INSTALLATION_METHOD_*` compile constants from #488. I want to avoid going deeper with preprocessor directives due to the reasons I wrote above.

Let's talk about alternatives. As mentioned before, this PR is trying to resolve #1367:

> The supported games section is currently showing games we don't support.
> For the Alpha please make sure its only showing Stardew Valley.

For the alpha release, we want an "on rails"-experience, meaning users can only interact and use the things we want them to test. In this case, we want users to manage Stardew Valley from the UI, not Skyrim, or Cyberpunk, or Sifu.

To remove other games from the UI, it's a one-line change:

https://github.com/Nexus-Mods/NexusMods.App/blob/0c8f1f9c2400a76af9b31ea9c706e76701c082f4/src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs#L59-L60

```csharp
var foundGames = gameRegistry.InstalledGames
                    .ToObservableChangeSet()
                    .Filter(x => x.Game.Domain != "stardewvalley");
```

Of course there should be a comment here linking to #1367, and similar with #1286, this could also have a setting. Either way, we only want Stardew Valley to appear in the UI, @Sewer56 mentioned that this affect the CLI, and I honestly don't think we care. If a user uses the CLI, then they're off the rails anyways. No point in adding restrictions there.

If you really wanna keep conditional compilation, then I might be amenable to these changes if we remove all the custom `NEXUSMODS_APP_ENABLE_*` constants and just use `DEBUG`. In `DEBUG` you get all games, in `RELEASE` you only get Stardew Valley. This reduces the amount of possible compilation configurations, while having the same effect.

Lastly, I believe the correct approach should be either settings, or feature flags using tools the other teams are already using, namely [Unleash](https://github.com/Unleash/unleash). Although, feature flags don't really make a ton of sense in non-web applications, we should definitely consider them if we run into something like this again.

Sewer56:
> he had to add a missing #if DEBUG directive to the imports in MainThreadData, because the compiler would otherwise complain in RELEASE mode.

I have to note, that this was actually unrelated to the PR.
On the main branch, `MainThreadData` has 

```
public static readonly bool IsDebugMode = Debugger.IsAttached;
```

But at the top of the file, you have only
```
using System.Diagnostics;
```

If you had your project set to `Release` mode and ran the cleanup task `Remove unused using statements` (either via Rider or by using `dotnet format`), that would generate a compile error when you switched back to Debug because the statement had to be eliminated. I noticed that error while cleaning up, so I fixed it.

It wasn't something I had to do, I just noticed it and fixed it.

Not to detract from the statement made, because it's a good example of how improper use of conditional compilation can break stuff.

-------------

In any case, the idea I had in mind when writing this code was that:

- Debug (`DEBUG`) has all of the constants enabled
- Constants only add extra code, not remove any code

That means that Debug is a superset of Release.
When we compile for Debug, we build all of the Release unaltered, *and then extra*.
That means our IDEs, build warnings, etc. pick up all the code during the development phase.

I would also recommend that any conditional code should also inline the namespaces if it introduces additional imports:
e.g.
```csharp
// Namespace prevents invalid cleanup on 'remove unused using statements' 
Games.BethesdaGameStudios.Services.AddBethesdaGameStudios()
```

What we must ***not do*** is disable code using directives, because that creates additional combinations of flags that would have to be tested in CI. And doing that *is in fact not fun to maintain* as @erri120 stated, but I am not doing that here. I really do not want to be doing that. The number of combinations remains 2; and one is a superset of the other.

I could also change the directives in the code to be in the form `#if (NEXUSMODS_APP_ENABLE_BETHESDA || DEBUG)`, and remove the debug block from the `Build.targets` instead. It may perhaps be better as it would clarify intent more explicitly.

erri120:
> I could also change the directives in the code to be in the form #if (NEXUSMODS_APP_ENABLE_BETHESDA || DEBUG), and remove the debug block from the Build.targets instead. It may perhaps be better as it would clarify intent more explicitly.

Just remove the `NEXUSMODS_APP_ENABLE_*` constants completely and only have `DEBUG` and `RELEASE`. Less is better when it comes to preprocessor directives, we gain nothing from having a more complex build system. It might be more flexible to pick and choose, but realistically, we don't need that. We have a list of games we support, that currently only contains Stardew Valley, and that list will only grow.

Sewer56:
Eh, sure. Gimme a moment.

halgari:
Thanks for the details @erri120, and I agree, stuff like this was the reason why we ditched Vogen. In that vein, we should proabably do the following:

@Sewer56 so we're all on the same page

1) Setup some sort of way to mark games as "officially supported"
2) Only show "officially supported" games in the "manage game UI", we won't worry about the rest
3) Add a config option so we can turn off that fiter in testing/development

erri120 (approved):

PR #1418, opened 2024-05-22
Docs

Patriot99:
I added a couple of changes to the documentation to make it slightly clearer and more concise. You may notice that in my version the difference is punctuation and the use of a semicolon, but the documentation for the user should be smooth to read and written correctly grammatically. Consider these as suggestions and let me know what you think.


halgari:
Only feedback here is I'd rather not use unicode quotes. Aside from that looks good

Patriot99:
> Only feedback here is I'd rather not use unicode quotes. Aside from that looks good

Done.

docs/development-guidelines/AddingAGame.md
```diff
@@ -2,7 +2,7 @@
 
 !!! info "Describes how to add a game to the Nexus Mods App."
 
-To make a game automatically recognised by the Nexus Mods App, make a class that inherits from `AGame`.
+To have a game automatically recognised by the Nexus Mods App, create a class that inherits from `AGame`.
```
Al12rs, line 5:
```suggestion
To have a game be automatically recognised by the Nexus Mods App, create a class that inherits from `AGame`.
```

docs/misc/DrawbacksOfNexusApproach.md
```diff
@@ -11,21 +11,21 @@ This takes considerable CPU and Disk resources.
 
 This typically results in an increase in disk usage of 60-100% compared to regular game size.
 
-!!! tip "For a 100GB game, an additional 60-100GB of space will be used once 'managed' with the Nexus Mods App."
+!!! tip "For a 100 GB game, an additional 60-100 GB of space will be used once 'managed' with the Nexus Mods App."
```
Al12rs, line 14:
```suggestion
!!! tip "For a 10 GB game, an additional 6-10 GB of space will be used once 'managed' with the Nexus Mods App."
```
The games we are actually targeting are not that big. We will likely never backup 100 GB of game files.

Al12rs (changes requested):

Al12rs (approved):

PR #1425, opened 2024-05-23
Fix saving overriden values

### LINKED ISSUE #1422: Settings manager is persisting setting override values (app uses InMemoryDB after running tests or Deisgner preview)

erri120:
Fixes #1422. Also finally adds tests for the settings manager.

### DIFF
--- a/NexusMods.App.sln
+++ b/NexusMods.App.sln
@@ -236,6 +236,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "NexusMods.ProxyConsole.Test
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Interprocess", "Interprocess", "{92AC7C89-5265-4D51-AB15-0EB9F3C168B5}"
 EndProject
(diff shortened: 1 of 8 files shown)


Al12rs:
Could you explain in a few words what the issue was and how you solved it? 
It wasn't too clear to me and I didn't want to look through all the manager code

erri120:
> Could you explain in a few words what the issue was and how you solved it? It wasn't too clear to me and I didn't want to look through all the manager code

This is the old code:

```csharp
        var settingsType = typeof(T);
        if (_values.TryGetValue(typeof(T), out var obj)) return (obj as T)!;

        var savedValue = Load<T>();
        if (savedValue is not null)
        {
            savedValue = Override(savedValue);
            CoreSet(savedValue, notify: false);
            return savedValue;
        }

        var defaultValue = GetDefaultValue<T>();
        defaultValue = Override(defaultValue);
        Set(defaultValue);
```

If we haven't cached the setting, we're first trying to load the setting from the storage backend. If we can't load the setting, we use the default value. The issue comes from the usage of `Set(defaultValue)`. The `Set` method will call `Save`, which isn't what we want. Instead, this PR uses `CoreSet` and uses the `notify` parameter for values that haven't been overridden:

```csharp
        var defaultValue = GetDefaultValue<T>();
        defaultValue = Override(defaultValue, out var didOverride);
        CoreSet(defaultValue, notify: !didOverride);
```

Al12rs:
Ah, thank you very much for the explanation! Makes sense 👍

PR #1429, opened 2024-05-23
Fix page discovery

### LINKED ISSUE #1428: Update page discovery section and page names
The following areas refer to the "My Mods" mods list page as "Loadout". We want the title of this page to be consistent with the left hand menu name and avoid confusion with the concept of a loadout.

**New tab/panel area:**

This heading for this section should be labelled "Mods" and the page should be labelled "My Mods"

Image

**Tab title:**

Should be labelled "My Mods"

Image

**The downloads page on the new tab/panel page is labelled as "In-progress downloads":**

Should be labelled "Downloads" and be in the "Mods" section of the new tab/panel page mentioned above.

Image

Lastly, the Library page in new tab/panel page should also be under the "Mods" section mentioned above (this leaves us with an empty "Downloads" section which can be removed from the new tab/panel page)

Note: The items in this new tabs/panels page also have the incorrect icons and their background fill should be surface/mid.

erri120:
Resolves #1428.

### DIFF
--- a/src/NexusMods.App.UI/WorkspaceSystem/NewTabPage/NewTabPageView.axaml
+++ b/src/NexusMods.App.UI/WorkspaceSystem/NewTabPage/NewTabPageView.axaml
@@ -44,7 +44,7 @@
                                             Command="{CompiledBinding SelectItemCommand}">
 
                                         <StackPanel Margin="16" Orientation="Horizontal" Classes="Spacing-1">
-                                            <icons:UnifiedIcon Classes="ProgressDownload" Size="24"/>
+                                            <icons:UnifiedIcon Size="24" Value="{CompiledBinding Icon}"/>
                                             <TextBlock x:Name="NameTextBlock"
                                                        Text="{CompiledBinding Name}"
                                                        Classes="BodyLGBold" />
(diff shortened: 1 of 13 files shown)


src/NexusMods.App.UI/WorkspaceSystem/NewTabPage/NewTabPageView.axaml
```diff
@@ -38,13 +38,15 @@
                             <DataTemplate DataType="{x:Type workspace:INewTabPageSectionItemViewModel}">
 
                                 <Border Classes="Mid Rounded">
-                                    <navigation:NavigationControl Width="375"
+                                    <navigation:NavigationControl
                                             Height="56"
                                             Padding="0"
+                                            Width="375"
+                                            Background="{StaticResource SurfaceMidBrush}"
```
Al12rs, line 45: The class should already be setting Mid background, you shouldn't need to set the background explicitly.
Did it not work for some reason?

erri120: Class `Mid` is for `Border.Mid` only.

Al12rs: Ah sorry, I read that wrong. 
I checked and I didn't see a specific button style on figma for these so I guess this is fine then. Might want to make a proper class if we have other places using similar buttons in the future.

Al12rs (approved):
I didn't test

PR #1431, opened 2024-05-23
Bugfix: Directory Cleanup on Unix should be Recursive

Sewer56:
This deletion should be recursive, or else it may fail depending on the order of directories obtained from code.

I forgot to set this to recursive when applying PR feedback.
And it slipped past review too.

### DIFF
--- a/src/NexusMods.App/Commandline/CleanupVerbs.cs
+++ b/src/NexusMods.App/Commandline/CleanupVerbs.cs
@@ -149,7 +149,7 @@ internal static class CleanupVerbs
 
             try
             {
-                directory.DeleteDirectory();
+                directory.DeleteDirectory(true);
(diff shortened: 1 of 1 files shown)


src/NexusMods.App/Commandline/CleanupVerbs.cs
```diff
@@ -149,7 +149,7 @@ private static async Task DeleteDirectoriesUnix(IRenderer renderer, IEnumerable<
 
             try
             {
-                directory.DeleteDirectory();
+                directory.DeleteDirectory(true);
```
Al12rs, line 152:
```suggestion
                directory.DeleteDirectory(recursive: true);
```

Al12rs (approved):

PR #1419, opened 2024-05-22
Fixes up the stress test CLI command

halgari:
Fixes the rather broken stress-test CLI command. Also fixes a lot of the flakey test issues we have in our test system. 

Previously: 

`IGame` would point to `IEnumerable<GameInstallation>` but `GameInstallation` also would point to `IGame`. In order to speed up the process of linking games to installations, `AGame` would cache Game installations. GameRegistry would be used by parts of the code to look up installations. Our manually added game locators would then also attempt to inject data into all these classes causing issues with caching. 

Now:

All these classes are condensed and their logic exists in `IGameRegistry`. Creating a new `GameInstallation` is as simple as registering it with `IGameRegistry` and then everything flows from there. `IGameRegistry` caches results and provides observable interfaces on all of these components. 

This means that in our tests we can say `StubbedGame.Create(...)` and get a brand-new stubbed game install, no clearing of caches required. Part of this PR also includes a `.zip` file that contains the contents of the `StubbedGame` install so that we have a ground-truth for all our game tests against the stubbed game. 

Fixing up the `stress-test` command required reworking some of the rendering code, so that's why the PR contains some of that data.

PR #1432, opened 2024-05-23
Fixed: Race Condition in early App Logic & CLI Commands

### LINKED ISSUE #1396: Race Condition may Lead to CLI Failures when App not Running
# Bug Report

## Summary

If you start a CLI command without the app already running there is a small possibility of a race condition occurring.

## Steps to reproduce

Remove the temporary sleep from `uninstall-app` command so it looks like so

```csharp
[Verb("uninstall-app", "Uninstall the application and revert games to their original state")]
private static async Task<int> UninstallApp(
    [Injected] IRenderer renderer,
    [Injected] IConnection conn,
    [Injected] ISettingsManager settingsManager,
    [Injected] IFileSystem fileSystem)
{
    // Step 1: Revert the managed games to their original state
    using var db = conn.Db;
```

If you then try running `uninstall-app` without it already running, there is a medium chance of failure

Running `uninstall-app` and other CLI commands which access some parts of the DB early should not fail.

## Other information

```
00:00:01.559 [ERROR] Encountered an unobserved task exception in the Task Scheduler from sender of type `System.Threading.Tasks.UnwrapPromise`1[System.Threading.Tasks.VoidTaskResult]`: `System.Threading.Tasks.UnwrapPromise`1[System.Threading.Tasks.VoidTaskResult]`|System.Collections.Generic.KeyNotFoundException: The given key 'EId:200000000000002' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at NexusMods.DataModel.Registry.Get(EntityId id) in C:\Shared\Code\NexusMods.App\src\NexusMods.DataModel\GameRegistry\Registry.cs:line 116
   at NexusMods.Abstractions.Loadouts.Loadout.Model.get_Installation() in C:\Shared\Code\NexusMods.App\src\Abstractions\NexusMods.Abstractions.Loadouts\Loadout.cs:line 140
   at NexusMods.App.Commandline.CleanupVerbs.<>c.<UninstallApp>b__1_0(Model loadout) in C:\Shared\Code\NexusMods.App\src\NexusMods.App\Commandline\CleanupVerbs.cs:line 40
   at System.Linq.Utilities.<>c__DisplayClass2_0`3.<CombineSelectors>b__0(TSource x)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.DistinctIterator`1.MoveNext()
   at NexusMods.App.Commandline.CleanupVerbs.UninstallApp(IRenderer renderer, IConnection conn, ISettingsManager settingsManager, IFileSystem fileSystem) in C:\Shared\Code\NexusMods.App\src\NexusMods.App\Commandline\CleanupVerbs.cs:line 43
```

Registry's `_byId` is not populated by the time CLI commands execute.

Sewer56:
fixes #1396

This PR is trivial code wise but warrants an explanation.

This fixes a race condition that can be experienced when running certain CLI commands and App startup logic that accesses the current set of game installs.

To explain concisely:

- App starts `IGameRegistry` in the background thread as an `IHostedService`.
- `IGameRegistry` requires initialization via `StartAsync` method (part of `IHostedService`).
- Access to some parts of `IGameRegistry` before `StartAsync` terminates is invalid.
    - Because `IGameRegistry` is initialized asynchronously.

Some parts of the app may request `IGameRegistry` directly or indirectly via DI.
For example running `db.Loadouts().Select(loadout => loadout.Installation)` is one easy way to trigger the error.

------

## Fundamental Issue

Fundamentally the issue is that items that inherit from `IHostedService` are started in the background.
Therefore it was crucial to evaluate whether other parts of the app may also exhibit from the same issues.

I was considering looking for a more general purpose solution. However, after I had a look around at all the current implementations of `IHostedService`, I concluded that this is the only component that needed patching.

The other `IHostedService` components don't rely on state that's silently initialized in the background, so they are not affected. Most of them just access the DB or something else that's already initialized which is totally ok.

Outside of that, performance was also crucial to consider. The API we're touching here thankfully is not one that is a super hot path. Therefore adding a single additional almost never taken branch to 2 `get` calls has no measurable impact.

This is why I'm comfortable with going for a simple solution.

### DIFF
--- a/src/NexusMods.DataModel/GameRegistry/Registry.cs
+++ b/src/NexusMods.DataModel/GameRegistry/Registry.cs
@@ -142,9 +142,23 @@ public class Registry : IGameRegistry, IHostedService
     
     private void WaitUntilInitialized()
     {
-        // This is an IHostedService.
-        // It will be initialized in a background threadpool thread.
+        /*
+            Note(sewer)
+            
+            This is an IHostedService.
+            It will be initialized in a background threadpool thread.
+            
+            Sleeping current thread is OK as this can't be called from
+            the thread initializing this.
+            
+            Extra Note:
+            
+            On Windows the Timer Resolution means we can only sleep
+            at minimum 15.6ms increments unless we request a lower one.
+            
+            So expect ~16ms sleeps there.
+        */
         while (!_isInitialized)
-            Thread.Sleep(100);
+            Thread.Sleep(8);
     }
 }
--- a/src/NexusMods.DataModel/GameRegistry/Registry.cs
+++ b/src/NexusMods.DataModel/GameRegistry/Registry.cs
@@ -133,7 +133,11 @@ public class Registry : IGameRegistry, IHostedService
     }
 
     /// <inheritdoc />
-    public EntityId GetId(GameInstallation installation) => _byInstall[GetKey(installation)];
+    public EntityId GetId(GameInstallation installation)
+    {
+        WaitUntilInitialized();
+        return _byInstall[GetKey(installation)];
+    }
 
     /// <inheritdoc />
     public async Task StartAsync(CancellationToken cancellationToken)
--- a/src/NexusMods.DataModel/GameRegistry/Registry.cs
+++ b/src/NexusMods.DataModel/GameRegistry/Registry.cs
@@ -1,6 +1,4 @@
 using System.Collections.ObjectModel;
-using System.Reactive.Linq;
-using System.Reactive.Subjects;
 using DynamicData;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
@@ -8,7 +6,6 @@ using NexusMods.Abstractions.GameLocators;
 using NexusMods.Abstractions.Games;
 using NexusMods.Abstractions.Games.DTO;
 using NexusMods.DataModel.GameRegistry;
-using NexusMods.MnemonicDB;
 using NexusMods.MnemonicDB.Abstractions;
 
 namespace NexusMods.DataModel;
@@ -27,6 +24,7 @@ public class Registry : IGameRegistry, IHostedService
     private readonly ReadOnlyObservableCollection<GameInstallation> _installedGames;
     private readonly ILogger<Registry> _logger;
     private readonly IEnumerable<ILocatableGame> _games;
+    private bool _isInitialized;
 
     /// <inheritdoc />
     public ReadOnlyObservableCollection<GameInstallation> InstalledGames => _installedGames;
@@ -45,7 +43,6 @@ public class Registry : IGameRegistry, IHostedService
             .Transform(g => g.Game)
             .Bind(out _installedGames)
             .Subscribe();
-        
     }
 
     private async Task Startup(IEnumerable<ILocatableGame> games)
@@ -102,6 +99,7 @@ public class Registry : IGameRegistry, IHostedService
             foreach (var (id, install) in results)
                 _cache.AddOrUpdate((install, id));
         });
+        _isInitialized = true;
     }
     
     private (GameDomain Domain, Version Version, GameStore Store) GetKey(GameInstallation installation)
@@ -110,10 +108,21 @@ public class Registry : IGameRegistry, IHostedService
     }
 
     /// <inheritdoc />
-    public IEnumerable<GameInstallation> AllInstalledGames => _byId.Values;
+    public IEnumerable<GameInstallation> AllInstalledGames
+    {
+        get
+        {
+            WaitUntilInitialized();
+            return _byId.Values;
+        }
+    }
 
     /// <inheritdoc />
-    public GameInstallation Get(EntityId id) => _byId[id];
+    public GameInstallation Get(EntityId id)
+    {
+        WaitUntilInitialized();
+        return _byId[id];
+    }
 
     /// <inheritdoc />
     public EntityId GetId(GameInstallation installation) => _byInstall[GetKey(installation)];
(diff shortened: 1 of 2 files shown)


src/NexusMods.DataModel/GameRegistry/Registry.cs
```diff
@@ -27,6 +25,11 @@ public class Registry : IGameRegistry, IHostedService
     private readonly ReadOnlyObservableCollection<GameInstallation> _installedGames;
     private readonly ILogger<Registry> _logger;
     private readonly IEnumerable<ILocatableGame> _games;
+    private bool _isInitialized;
+    
+#if DEBUG
+    private int _initThreadId;
+#endif
```
Al12rs, line 32: Remove #if DEBUG . Either leave all the code and do a Debug assert at the end or remove all the debug code.

src/NexusMods.DataModel/GameRegistry/Registry.cs
```diff
@@ -130,4 +147,29 @@ public Task StopAsync(CancellationToken cancellationToken)
...
+            On Windows the Timer Resolution means we can only sleep
+            at minimum 15.6ms increments unless we request a lower one.
+            
+            So expect ~16ms sleeps there.
+        */
+#if DEBUG
+        Debug.Assert(_initThreadId != Environment.CurrentManagedThreadId, "WaitUntilInitialized called from initialization thread");
+#endif
+        while (!_isInitialized)
+            Thread.Sleep(8);
```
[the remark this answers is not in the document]
Sewer56: > Any chance we are sleeping on the Main thread?

Practically speaking, no chance. The service starts in the background as the App's main process is booted up.

By the time the Avalonia window is even displayed, the background initialization should already have finished. The operation being waited for takes around as long as the sleep is set to.

Currently this is only reproducible in CLI commands when a new main process has to be started.

> Should we use an Async Lock?
> Should we rethink how this registry works?

The perf difference is negligible enough that it shouldn't really matter.

Technically speaking though, I could spin a ThreadPool thread (Task.Run) in the constructor and just wait for that to finish. That could delay app startup by a few millis then, not sure if worth it since the APIs here aren't very hot.

src/NexusMods.DataModel/GameRegistry/Registry.cs
Al12rs: >  waste 8 bytes of RAM
Insert shocked expression here.

I think we can spare 8 bytes in release.

Sewer56: 😭 

Think of all of the collective seconds that would be saved from all the hundreds of thousands of people people shaving nanoseconds from App startup on the daily.

*sobbbbb*

Yeah, don't worry, I'll get it sorted when I get back to my machine later.

halgari:
I had something in place like this in the past, and it kindof trickles all through the code, we have to make sure we call WaitForInitialized on every call, and since that's a sync call it can block a thread. There's 3 other services inside MnemonicDB that have to be started before this is started. That all means that this becomes a bit brittle. 

I'd prefer to find something that's a bit more robust for this.

Sewer56:
Ah, I think this situation may be a bit misunderstood.
Let me provide some further clarification on why I wrote it this way.
Please let me know what you think afterward.

- The `IGameRegistry` is a singleton.
- The initialization method `Startup` is therefore only called once in the Application's runtime.
    - It is called in the background.
    - It takes around 400ms. Mostly spent finding games in GameFinder.
- During asynchronous `Startup` we cache data built from GameFinder and DB and store it in the fields `_byId` and `_byInstall`.
- Thus we risk this local cache (`_byId` and `_byInstall`) not being initialized by the time it's accessed.

This problem is unique to this `IHostedService`; the other implementations of this interface do not require that `StartAsync` finishes by the time the service is usable. Therefore, they are not affected. This is why I have done further investigation and ultimately concluded that this is not a more systematic issue; it is isolated to this class.

The condition `while (!_isInitialized)` has only ever been observed to be `true` so far when running commands from the CLI.
This is because the CLI starts running code which interacts with `IGameRegistry` before it has finished initializing in the background. When starting the App UI normally, this has never been true for me because setting up the rest of the App, takes much longer than that aforementioned 400ms. (It's ~1.5 seconds after init completes on my machine when restoring a workspace)

---------------

So I had the following choices:

- Go for something robust that:
    1. Ensures IHostedService(s) are sorted by which service depends on which service.
        - For example, `CliService` depends on `IGameRegistry`.
    2. And all services are ready before main App logic starts.
- Initialize synchronously in constructor, delaying startup by up to 400ms.
- Wait for initialization completion.

Because the issue was isolated to 1 class, and because because the APIs that use `_byId` and `_byInstall` are not very frequently used in performance critical code; the last option felt like a no-brainer to me.

With that in mind, there is one improvement I could make here.

I could wrap `_byId` and _byInstall` in a struct which provides a getter. That getter would enforce that we wait if they're not yet initialized. That would mitigate the risk of someone writing new public APIs that access  `_byId` or `_byInstall` and forgetting to call `WaitUntilInitialized`.

That would be more reusable and normally I would do something like that. However I've just been trying to be more minimal today today; given the constructive criticism I received earlier recently for doing things way overkill. I wanted to gauge the feedback.

(NB. The risk here is being criticized of writing a redundant struct that has only 1 place of use within the App.)

Al12rs:
@halgari What do you want to do with this? 

Should we have the `CLIServer` only start listening to IPC commands after all `IHostedService`s have been successfully started with `StartAsync`? 
Effectively moving out the `StartTcpListenerAsync()` call from `CLIServer.StartAsync()` into somewhere in `Program.cs`

https://github.com/Nexus-Mods/NexusMods.App/blob/3fb9d63c6551e5509f9c42cf3517c04df3ec4e4c/src/NexusMods.SingleProcess/CliServer.cs#L138-L145

Or do you have a more general solution for `IHostedServices` dependencies in mind?

halgari:
Should be fixed by additions in #1419. See the `StartAsync` changes in that PR

Sewer56:
I'm closing this in favour of #1419, the rework done to the `IGameRegistry` there just happen to alleviate the issue.
In turn, that makes this PR redundant.

PR #1436, opened 2024-05-23
Apply Ingest fixes

### LINKED ISSUE #1175: Apply/Ingest -UX - Don't ask users to Apply after a loadout for a new game completed loading.
# Bug Report

### LINKED ISSUE #1325: Regression: Applying while there are changes to ingest doesn't always remove the Apply button from view
Since #1302 PR pressing apply sometimes doesn't remove the apply button, indicating that there are more things to apply.

It is unclear if this is just a UI bug or there is a bug in how the loadout is applied.

### LINKED ISSUE #1433: Remove Ingest button

### LINKED ISSUE #1434: Disable Play button while there are things to apply
Add tooltip on disabled play button hover: `Apply changes to enable Play`

### LINKED ISSUE #1435: Fix Apply and Apply diff button styling

### LINKED ISSUE #1442: Prevent apply action while the game is running
# User story

**As a** user

**I want to** be prevented from applying my changes while the game is running 

**So that** I don't hit permission issues or cause problems with my active game session 

## Requirements
_(Any other information related to but not covered by the story)_

## Design

## DOD

## Scenario 1
**Given** _______

**When** _______

**Then** _______

Al12rs:
- fixes #1433
- fixes #1434
- fixes #1325 
- fixes #1175
- fixes #1435
- fixes #1442

Main issue fixed here was that we were storing and then comparing the latest Db Transaction Id, instead of storing the latest Transaction that actually changed the loadout.

I also added a couple of `mod.Revise()` and `loadout.Revise()` calls during ingest that we were missing, without them the loadout Tx wouldn't be updated and we would store the incorrect revision in the LastAppliedSavedState.

This also has a bunch of various Style changes that were done with lausandy in call.


src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
```diff
@@ -52,6 +54,22 @@ protected override async Task<Loadout.Model> AddChangedFilesToLoadout(Loadout.Mo
...
+            // We need to manually revise the loadout in that case
+            // ReSharper disable once ConditionIsAlwaysTrueOrFalseAccordingToNullableAPIContract
+            if (mod.Db != null)
+            {
+                mod.Revise(tx);
+            }
+            else
+            {
+                loadout.Revise(tx);
+            }
```
erri120, line 72: What do you mean with

> If we created the mod in this transaction

What "mod" are you referring to? This synchronizer doesn't create any mods.

src/NexusMods.App.UI/LeftMenu/Items/ApplyControlViewModel.cs
```diff
@@ -110,25 +99,30 @@ public ApplyControlViewModel(LoadoutId loadoutId, IServiceProvider serviceProvid
...
-                            CanIngest = !isApplying && !isIngesting &&
-                                        LastAppliedLoadoutId.Equals(_loadoutId);
+                            var isToolRunning = data.Second;
+                            var lastAppliedWithTxId = data.First.First.Item2;
+                            CanApply = !isApplying &&
+                                       !isToolRunning &&
+                                       !NewestLoadout.GetLoadoutWithTxId().Equals(lastAppliedWithTxId);
+                            IsLaunchButtonEnabled = !isApplying && !CanApply;
                         }
                     ).DisposeWith(disposables);
```
erri120, line 127: `data.First.First.Item2` is a great indicator that the code is unreadable:

```csharp
                // CanApply
                // ReSharper disable once InvokeAsExtensionMethod
                Observable.CombineLatest(
                    loadoutOrLastAppliedTxObservable,
                    ApplyCommand.IsExecuting,
                    LaunchButtonViewModel.Command.IsExecuting,
                    (tuple, isApplying, isToolRunning) => (IsApplying: isApplying, IsToolRunning: isToolRunning, LastAppliedWithTxId: tuple.Item2)
                )
                    .OnUI()
                    .Subscribe(tuple =>
                    {
                        var (isApplying, isToolRunning, lastAppliedWithTxId) = tuple;
                        CanApply = !isApplying &&
                                   !isToolRunning &&
                                   !NewestLoadout.GetLoadoutWithTxId().Equals(lastAppliedWithTxId);
                        IsLaunchButtonEnabled = !isApplying && !CanApply;
                    }).DisposeWith(disposables);
```

erri120 (changes requested):

src/Games/NexusMods.Games.StardewValley/StardewValleyLoadoutSynchronizer.cs
Al12rs: If this is the first time you have outside changes that end up in Override, then Override is created by `GetOrCreateOverridesMod`

erri120: If that's the case, then you should change this code to reflect that. This code suggests that the comment applies to every mod we updated.

Al12rs: I updated the comment, did you want a more fundamental change to how the code works here or is that fine?

I could try to track Override specifically, but this way if we decide in the future to instead add Config files to new mods and attach those mods to existing mods or something, then this still works or at least it will be clear that any created mod needs this, not just Override.

Al12rs: We don't need special handling for override mod, we need special handling for any created mod. We just happen to create override. If you want to create a another mod to put files in, then you would need to call revise on that as well.

erri120 (approved):

PR #1448, opened 2024-05-27
Add double click handler

### LINKED ISSUE #1441: Add generic double click handler

erri120:
Resolves #1441.

### DIFF
--- a/src/NexusMods.App.UI/Helpers/DoubleClickHelper.cs
+++ b/src/NexusMods.App.UI/Helpers/DoubleClickHelper.cs
@@ -11,8 +11,9 @@ namespace NexusMods.App.UI.Helpers;
 [PublicAPI]
 public static class DoubleClickHelper
 {
-    // NOTE(erri120): Windows uses 500ms (https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setdoubleclicktime?redirectedfrom=MSDN#parameters)
-    // This can be fine-tuned later.
+    // NOTE(erri120):
+    // Windows uses 500ms (https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setdoubleclicktime?redirectedfrom=MSDN#parameters)
+    // Avalonia default is 500ms (https://github.com/AvaloniaUI/Avalonia/blob/b904ded9ab32022b3ce037c8b63d21887fbd85a0/src/Avalonia.Base/Platform/DefaultPlatformSettings.cs#L35)
     public static readonly TimeSpan DefaultTimeout = TimeSpan.FromMilliseconds(500);
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/NexusMods.App.UI/Helpers/DoubleClickHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+        private void EventHandler(object? sender, TRoutedEventArgs args)
+        {
+            var currentDateTime = DateTime.Now;
+            var isDoubleClick = currentDateTime - _lastDateTime <= DefaultTimeout;
+            _lastDateTime = currentDateTime;
+
+            if (!isDoubleClick) return;
+            OnNext(Unit.Default);
+            args.Handled = true;
+        }
```
Al12rs, line 65: I found this: https://github.com/AvaloniaUI/Avalonia/discussions/11922
Where they suggest using the `ClickCount` property of the event to handle double clicks.

There also appears to be a `DoubleTapped` event, though I don't know if all controls support that:
https://docs.avaloniaui.net/docs/concepts/input/pointer#more-information

Have you already considered those approaches?

erri120: `ClickCount` is handled not on the event, but on the input device itself:

https://github.com/AvaloniaUI/Avalonia/blob/4fdf4fa36cfc9b1e353a4ca79e5c8e8137bd9ea2/src/Avalonia.Base/Input/MouseDevice.cs#L133-L148

`PointerPressedEventArgs` are also the only event args that have the `ClickCount` property. My approach works for any event, as it just waits for the same event to occur twice within a given interval.

Al12rs: I see, this approach is more generalized 👍

src/NexusMods.App.UI/Helpers/DoubleClickHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+        private readonly List<IObserver<Unit>> _observers = new(capacity: 1);
+        private readonly IDisposable _handlerDisposable;
+
+        public DoubleClickSubject(Control control, RoutedEvent<TRoutedEventArgs> routedEvent)
+        {
+            _handlerDisposable = control.AddDisposableHandler(
+                routedEvent,
+                EventHandler,
+                routes: RoutingStrategies.Bubble,
+                handledEventsToo: true
```
Al12rs, line 51: What are the use cases for handling already handled events?
I thought this would help prevent double clicks from propagating up from nested controls that have their own click handlers?

erri120: The `PointerPressedEvent` is often already handled by something else. Think about the item selection in a list. The event propagating is done with the event routes, which I set to `Bubble`, meaning events from child controls will bubble up to the parent.

I'll add parameters to the API, so consumers can choose whether they want events from child controls to appear as well.

Al12rs: Does item selection actually mark the click event as handled? That doesn't sound like a great idea, I didn't know it did that.
Api exposing this to users sounds good. 

If we want to add a double click handler to items in a DataGrid, but these items also have a toggle button, we would want double click to work except when pressing on the toggle button. If the first click is already eaten by Selection then we are a bit in trouble, as we would have to handle Handled events to avoid Selection eating the first click, but would end up also handling the Toggle button events that way.

erri120: > Does item selection actually mark the click event as handled? That doesn't sound like a great idea, I didn't know it did that.

https://github.com/AvaloniaUI/Avalonia/blob/4fdf4fa36cfc9b1e353a4ca79e5c8e8137bd9ea2/src/Avalonia.Controls/ListBoxItem.cs#L68

> If we want to add a double click handler to items in a DataGrid, but these items also have a toggle button, we would want double click to work except when pressing on the toggle button.

Toggling the button is a single-click operation. Unless the user actually double-clicks, the double-click observable won't trigger.

Al12rs: The toggle button is the child in this scenario, so it isn't a problem that we could set the event as handled, the problem is that we want to ignore events handled by the button.

If the user double clicks on a child button, like a three state checkbox or whatever, I wouldn't expect the parent control to behave as it if were double clicked, because the events should have been consumed/handled by the child control already.

But if we need to listen to Handled events as well due to Selection otherwise eating those, then we end up listening to the child button clicks as well. 

I think we might be able to resolve this by choosing exactly which control we attach the handler to, e.g. a border that is inside the DataGrid row or cell, rather than on the row or cell itself.

tests/NexusMods.UI.Tests/Helpers/DoubleClickHelperTests.cs
```diff
@@ -0,0 +1,53 @@
...
+            numTriggered.Should().Be(0, because: "button was clicked twice, but the time difference was too great");
+
+            automationPeer.Invoke();
+
+            numPressed.Should().Be(3);
+            numTriggered.Should().Be(1, because: "button was double-clicked");
+        });
+
+        return;
+        static void Sleep() => Thread.Sleep(DoubleClickHelper.DefaultTimeout + TimeSpan.FromMilliseconds(10));
```
Al12rs, line 51: I would use `Task.Delay()` instead of `Thread.Sleep`. Add async in `OnUi async () =>` at the top and await it.

src/NexusMods.App.UI/Helpers/DoubleClickHelper.cs
```diff
@@ -0,0 +1,128 @@
...
+    public static readonly TimeSpan DefaultTimeout = TimeSpan.FromMilliseconds(500);
+
+    /// <summary>
+    /// Registers a double click handler and returns an observable that triggers whenever
+    /// a double click occurs.
+    /// </summary>
+    public static IObservable<Unit> AddDoubleClickHandler(
+        this Control control,
+        RoutingStrategies routingStrategy = RoutingStrategies.Bubble,
+        bool ignoreHandledEvents = false)
```
Al12rs, line 25: I would have this set to true by default as that is the default behavior for handlers

Al12rs (changes requested):

tests/NexusMods.UI.Tests/Helpers/DoubleClickHelperTests.cs
erri120: The code inside `OnUi` is supposed to run synchronously to prevent weirdness.

src/NexusMods.App.UI/Helpers/DoubleClickHelper.cs
erri120: I'd rather have this default to `false` so it works 99% of the time, instead of setting it to `true` for it works only 30% of the time.

Al12rs: I expect this to work 80% of the time or more, unless avalonia is really crap, but I haven't tried. I guess you speak from experience?

tests/NexusMods.UI.Tests/Helpers/DoubleClickHelperTests.cs
Al12rs: From my understanding the async version should work fine, I also tested it locally. The continuation after the `await Sleep()` call should be scheduled on the starting thread, because we don't use `ConfigureAwait(false)`, and the starting thread should be the UI one thanks to the dispatcher inside the `OnUI(Func<Task> action)` overload of `OnUI`.

erri120: I don't really see any benefit in doing this. Adding async isn't really an improvement.

Al12rs (approved):

Al12rs (approved):
I can't reproduce the CI failure locally

PR #1452, opened 2024-05-27
Add left menu badges

### LINKED ISSUE #1424: Add new item/s counter badge to "Library" left menu option
Add a counter badge to the Library item in the left hand menu when a mod is added to the Library list. This will indicate to users where their downloaded mods can be found. The badge should disappear once the Library page is opened.

## Design

[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1Mod-library?node-id=148%3A2124&t=GTNnAbzBa0CCknmH-1)

Image

erri120:
Resolves #1424.

Also adds badges for diagnostics instead of using `(count)` as before. Since the left menu doesn't know if the mod library page is currently open and visible, the badge will only be removed if the user actually clicks on the item.

### DIFF
--- a/src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
@@ -73,8 +73,12 @@ public class FileOriginsPageViewModel : APageViewModel<IFileOriginsPageViewModel
 
     public static bool FilterDownloadAnalysisModel(DownloadAnalysis.Model model, GameDomain currentGameDomain)
     {
-        if (!model.TryGet(DownloaderState.GameDomain, out var domain)) return false;
-        if (domain != currentGameDomain) return false;
+        // NOTE(erri120, Al12rs): Manually added archives don't have a game domain.
+        if (model.TryGet(DownloaderState.GameDomain, out var domain))
+        {
+            if (domain != currentGameDomain) return false;
+        }
+
         if (model.Contains(StreamBasedFileOriginMetadata.StreamBasedOrigin)) return false;
         return true;
     }
--- a/src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
@@ -4,6 +4,7 @@ using DynamicData;
 using DynamicData.Binding;
 using NexusMods.Abstractions.FileStore.ArchiveMetadata;
 using NexusMods.Abstractions.FileStore.Downloads;
+using NexusMods.Abstractions.Games.DTO;
 using NexusMods.Abstractions.Installers;
 using NexusMods.Abstractions.Loadouts;
 using NexusMods.Abstractions.Loadouts.Ids;
@@ -16,6 +17,7 @@ using NexusMods.Icons;
 using NexusMods.MnemonicDB.Abstractions;
 using NexusMods.Networking.Downloaders.Tasks.State;
 using ReactiveUI;
+using RocksDbSharp;
 
 namespace NexusMods.App.UI.Pages.ModLibrary;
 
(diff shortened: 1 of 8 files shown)


src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -41,35 +43,39 @@ public class FileOriginsPageViewModel : APageViewModel<IFileOriginsPageViewModel
...
         {
-            entriesObservable.Subscribe()
-                .DisposeWith(d);
+            entriesObservable.SubscribeWithErrorLogging().DisposeWith(d);
         });
     }
+
+    public static bool FilterDownloadAnalysisModel(DownloadAnalysis.Model model, GameDomain currentGameDomain)
+    {
+        if (!model.TryGet(DownloaderState.GameDomain, out var domain)) return false;
```
Al12rs, line 76: This should not return false
If we don't have a GameDomain on the DownloadState it is likely because it was installed from archive on disk, but we want to show those, not filter them out.

You can move the StreamBasedFileOrigin check first, then return true if has no game, then false if game doesn't match

Al12rs (changes requested):

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
erri120: Why do archives from disk not have a game domain? The button to do so is in the Loadout Grid, so we have access to the game domain.

Al12rs (approved):

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
Al12rs: The problem with that is that you are assuming that the user will add a mod for this game, which isn't necessarily true.
So someone on SSE could add a mod for Legendary edition and we would auto label it as SSE mod. Leaving it unknown is preferrable in that case. 
I guess we will have to figure something out when we add support for another game.

erri120: Updated the code to add `GameDomain` to the manually added archive.

erri120: We don't support SSE or LE yet. For Stardew Valley, modders would expect their Stardew Valley archives to appear for Stardew Valley.

Al12rs: You can add a random archive and it would show up as a StardewValley mod, which doesn't make too much sense.
We probably need a user facing value and an internal value.

erri120: I don't see why we'd want to be so complex about this. If the user is on a Stardew Valley loadout and manually adds an archive, then we should accept that as a Stardew Valley archive.

Al12rs: The file would show up even if installation failed because it wasn't actually a StardewMod. We might want to ask @captainsandypants

erri120: Mod installation isn't relevant to this, tho. Adding a mod manually is an escape hatch in the first place. Again, this shouldn't be complex.

Al12rs (changes requested):
Test fail locally as well: https://github.com/Nexus-Mods/NexusMods.App/actions/runs/9256864195?pr=1452

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
captainsandypants: I agree that if the user adds a mod manually to Stardew we should register is as Stardew, although where would the game domain show up in the UI? And do we need to consider multiple game store version of Stardew being managed for this?

erri120: > although where would the game domain show up in the UI?

We don't have a column that shows the game domain, but the code above filters by game domain because that's the only way to check if the download is for the current game.

> And do we need to consider multiple game store version of Stardew being managed for this?

No. The "game domain" is the identifier used by Nexus Mods `stardewvalley` and applies to all versions of the game. It's our way of uniquely identifying a game across installations.

Al12rs (approved):

PR #1468, opened 2024-05-28
Change mod name when installation fails

### LINKED ISSUE #1334: Cancelling mod installation leaves failed mod entry in LoadoutGrid
image

If user cancelled the installation explicitly then it should not leave the entry.

erri120:
Resolves #1334.

### DIFF
--- a/src/NexusMods.DataModel/ArchiveInstaller.cs
+++ b/src/NexusMods.DataModel/ArchiveInstaller.cs
@@ -54,20 +54,18 @@ public class ArchiveInstaller : IArchiveInstaller
         // Get the loadout and create the mod, so we can use it in the job.
         var useCustomInstaller = installer != null;
         var loadout = _conn.Db.Get<Loadout.Model>(loadoutId.Value);
-        
+
+        var archiveName = download.TryGet(FilePathMetadata.OriginalName, out var originalName) ? originalName.ToString() : null;
+
         var modName = defaultModName ?? (download.Contains(DownloadAnalysis.SuggestedName)
             ? download.SuggestedName
-            : "<unknown>");
+            : archiveName ?? "<unknown>");
 
         {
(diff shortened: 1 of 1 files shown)


src/NexusMods.DataModel/ArchiveInstaller.cs
```diff
@@ -218,6 +214,7 @@ private async Task SetStatus(ModId modId, ModStatus status)
         _logger.LogInformation("Setting status of ModId:{ModId}({Name}) to {Status}", modId, mod.Name, status);
         
         using var tx = _conn.BeginTransaction();
+        tx.Add(modId.Value, Mod.Name, $"Installation failed: {mod.Name}");
```
Al12rs, line 217: This method was meant to support setting any status, not just failed status.
Either change the method to be SetFailedStatus, or check if status==ModStatus.Failed before replacing the name

Al12rs (changes requested):

src/NexusMods.DataModel/ArchiveInstaller.cs
```diff
@@ -54,20 +54,18 @@ public async Task<ModId[]> AddMods(LoadoutId loadoutId, DownloadAnalysis.Model d
...
         var useCustomInstaller = installer != null;
         var loadout = _conn.Db.Get<Loadout.Model>(loadoutId.Value);
-        
+
+        var archiveName = download.TryGet(FilePathMetadata.OriginalName, out var originalName) ? originalName.ToString() : null;
+
         var modName = defaultModName ?? (download.Contains(DownloadAnalysis.SuggestedName)
             ? download.SuggestedName
-            : "<unknown>");
+            : archiveName ?? "<unknown>");
```
Al12rs, line 62: Do the same thing for SuggestedName to make the code a bit cleaner

Al12rs (approved):

PR #1470, opened 2024-05-28
Run Ingest on startup if current loadout is the last applied

### LINKED ISSUE #1380: Empty state for Mod List
## Designs

When no mods have been added from the library to the mod list it will need an empty state.

"Library" button opens the "Library" page in app.

"browse new mods on Nexus Mods" link opens external browser to the mods page for that game (Stardew valley):
`https://www.nexusmods.com/stardewvalley/mods/`

[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1Mod-library?node-id=110%3A561&t=NHnLOBOOXWAfHJ2w-1)

image

Al12rs:
- closes #1380

@erri120 I know there must be a better way to run the Async method than what I did, let me know what you think would be best

### DIFF
--- a/src/NexusMods.App.UI/LeftMenu/Items/ApplyControlViewModel.cs
+++ b/src/NexusMods.App.UI/LeftMenu/Items/ApplyControlViewModel.cs
@@ -145,16 +145,7 @@ public class ApplyControlViewModel : AViewModel<IApplyControlViewModel>, IApplyC
                     .DisposeWith(disposables);
                 
                 // Perform an ingest on first load:
-                Observable.Return(0)
-                    .Select(async  _ =>
-                        {
-                            if (_isFirstLoad && LastAppliedWithTxId.Id.Equals(_loadoutId))
-                                await Ingest();
-                            _isFirstLoad = false;
-                        }
-                    )
-                    .SubscribeWithErrorLogging()
-                    .DisposeWith(disposables);
+                Task.Run(FirstLoadIngest);
             }
         );
     }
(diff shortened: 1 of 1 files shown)


src/NexusMods.App.UI/LeftMenu/Items/ApplyControlViewModel.cs
```diff
@@ -142,6 +143,18 @@ public ApplyControlViewModel(LoadoutId loadoutId, IServiceProvider serviceProvid
...
+                Observable.Return(0)
+                    .Select(async  _ =>
+                        {
+                            if (_isFirstLoad && LastAppliedWithTxId.Id.Equals(_loadoutId))
+                                await Ingest();
+                            _isFirstLoad = false;
+                        }
+                    )
+                    .SubscribeWithErrorLogging()
+                    .DisposeWith(disposables);
```
erri120, line 157: Just do a `Task.Run`.

erri120:
```suggestion
Task.Run(Apply);
```

src/NexusMods.App.UI/LeftMenu/Items/ApplyControlViewModel.cs
```diff
@@ -155,4 +168,13 @@ private async Task Apply()
...
     }
     
+    private async Task Ingest()
+    {
+        await Task.Run(async () =>
+        {
+            var loadout = _conn.Db.Get(_loadoutId);
+            await _applyService.Ingest(loadout.Installation);
+        });
+    }
```
erri120, line 178: No point in double wrapping `Task.Run` in `Task.Run`.

src/NexusMods.App.UI/LeftMenu/Items/ApplyControlViewModel.cs
```diff
@@ -168,13 +159,17 @@ private async Task Apply()
...
-            await _applyService.Ingest(loadout.Installation);
-        });
+            _isFirstLoad = false;
+
+            if (LastAppliedWithTxId.Id.Equals(_loadoutId))
+            {
+                var loadout = _conn.Db.Get(_loadoutId);
+                await _applyService.Ingest(loadout.Installation);
+            }
+        }
```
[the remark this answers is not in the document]
Al12rs: I tested this and WhenActivated gets called every time you switch to the loadout, not just once on first load. 
We only want to do it once, not every time you switch to a loadout, so simple flag.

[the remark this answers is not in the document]
erri120: In that case, you can move the `Task.Run` out of the `WhenActivated` block so that it only runs once.

PR #1480, opened 2024-05-28
Added: Use Pinned PupNet Commit

Sewer56:
With the upstream maintainer going on hiatus, thus meaning we can't get our changes available in an actual release build for the time being; this PR changes our existing build script to use a self-built version of PupNet directly from upstream's latest commit on `main` (`dabed0cc2063c5a2d2c4f780bb6718f4b90cfd16`).

I've made the changes in such a way such that they're as non-destructive as possible, i.e. as much of the existing build script is unaltered.

That's about it.
I (of course) forked the app and tested it works.

-----

With this, hopefully Windows can finally have proper uninstalls for real.

### DIFF
--- a/.github/workflows/build-linux-pupnet.yaml
+++ b/.github/workflows/build-linux-pupnet.yaml
@@ -8,11 +8,16 @@ on:
         required: false
         description: "The version of the application to build"
         default: "1.0.0"
-      PupNetVersion:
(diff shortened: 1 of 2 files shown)


halgari (approved):
LGTM

erri120 (approved):

PR #1491, opened 2024-05-29
Update metrics

### LINKED ISSUE #1487: User Metrics: Questions
Is it possible for telemetry to answer the following questions:

How many games is a user managing through NMA?
How many mods does a user have installed? will need this per game. 
How many mods has the user downloaded?
How long does it take for a game to be managed? per game
What system specs are being used?
How many loadouts are being created per game?
How many people have undo turned on to turned off?
Can we track bugs or problems with the app?

Is there anything we can add that would tell us if there is a problem with the app?

- #1184

erri120:
Resolves #1487.

- [x] Users per Version
- [x] Managed Games Count
- [x] Mods Per Game (avg Loadout count)
- [x] Global Download Size

### DIFF
--- a/src/NexusMods.App/TelemetryProvider.cs
+++ b/src/NexusMods.App/TelemetryProvider.cs
@@ -1,11 +1,15 @@
 using System.Reactive.Disposables;
+using DynamicData.Aggregation;
+using DynamicData.Binding;
 using Microsoft.Extensions.DependencyInjection;
+using NexusMods.Abstractions.FileStore.Downloads;
 using NexusMods.Abstractions.Loadouts;
 using NexusMods.Abstractions.Loadouts.Mods;
 using NexusMods.Abstractions.MnemonicDB.Attributes;
 using NexusMods.Abstractions.Telemetry;
 using NexusMods.App.UI;
 using NexusMods.Networking.NexusWebApi;
+using NexusMods.Paths;
 using OneOf;
 
 namespace NexusMods.App;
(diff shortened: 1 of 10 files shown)


Al12rs:
Is average mod count more useful than max? Should we also add Max?

Al12rs (approved):

src/NexusMods.App/TelemetryProvider.cs
```diff
@@ -1,30 +1,55 @@
...
+            .ToObservableChangeSet()
+            .Sum(x => (double)x.Size.Value)
+            .SubscribeWithErrorLogging(totalDownloadSize => _downloadSize = Size.From((ulong)totalDownloadSize))
+            .DisposeWith(_disposable);
     }
 
     public void ConfigureMetrics(IMeterConfig meterConfig)
     {
+#if RELEASE
+        meterConfig.CreateActiveUsersCounter(BuildInfo.ApplicationConstants.Version);
```
Al12rs, line 43: Ugh, this is actually broken I think

Al12rs: @erri120 Is that correct?

erri120: Right, another case why preprocessor definitions suck.

Al12rs: Agreed, I will fix it in my pr since I need to fix conflicts in this file anyways

Al12rs: Pushed the fix as well

PR #1495, opened 2024-05-29
Cleanup: Toolbars in App no longer include unused nesting.

Sewer56:
image

This PR slightly improves how we present toolbars in the various user interface elements.

Previously, inner contents of every single toolbar was declared as something like:

```xaml
<Grid ColumnDefinitions="Auto, *">
    <StackPanel Grid.Column="0" Orientation="Horizontal">
```

(Note: Some views had `Orientation == Horizontal` and some had `Alignment == Right`, both are redundant with our styles)

The first column of the Grid being reserved to represent the title.
However, ***none of our current views*** actually have a title on the toolbar, therefore this can be simplified to:

```xaml
<StackPanel>
```

Greatly reducing nesting, and making UI work easier.

When I was working with moving some of the buttons over from My Mods to Mod Library, I noticed this, and thus have cleaned up this unneeded nesting.

### DIFF
--- a/src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml
+++ b/src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml
@@ -7,11 +7,41 @@
                                 xmlns:icons="clr-namespace:NexusMods.Icons;assembly=NexusMods.Icons"
                                 xmlns:resources="clr-namespace:NexusMods.App.UI.Resources"
                                 xmlns:fileOriginEntry="clr-namespace:NexusMods.App.UI.Pages.ModLibrary.FileOriginEntry"
+                                xmlns:navigation="clr-namespace:NexusMods.App.UI.Controls.Navigation"
                                 mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
                                 x:Class="NexusMods.App.UI.Pages.ModLibrary.FileOriginsPageView"
                                 x:CompileBindings="True">
     <Grid RowDefinitions="Auto, *">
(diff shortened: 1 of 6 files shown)


src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml
```diff
@@ -7,11 +7,41 @@
...
+                        <TextBlock Text="{x:Static resources:Language.LoadoutGridView__Add_Mod}"/>
+                    </StackPanel>
+                </Button>
+                <Button x:Name="AddModAdvancedButton">
+                    <StackPanel>
+                        <icons:UnifiedIcon Classes="PlusCircleOutline" />
+                        <TextBlock Text="{x:Static resources:Language.LoadoutGridView__Add_Mod_Advanced}"/>
+                    </StackPanel>
+                </Button>
+            </StackPanel>
```
Al12rs, line 44: This should not be here, this is the FileOriginsPageView, and these are from loadout. Some of thse might be moved or redone later but not in this PR.

Al12rs (changes requested):

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml
Sewer56: Seems I squashed at some point and missed this. My bad.

This was my WIP stuff on another branch, I was just cherrypicking the Loadout branch to make the 2 PRs we have.

Al12rs (approved):

Al12rs:
Closing and reopening to rebase CI tests onto latest main

PR #1496, opened 2024-05-29
Re-added: A minimal dummy Design VM for LoadoutGrid Toolbar Support

Sewer56:
This re-adds a very minimal design VM for LoadoutGrid

image

The intent here is to have a preview of the toolbar.
In its current state, on the main branch the preview is completely broken, which makes edits more tedious.

----------

Note: When adding this, I kinda thought that there ought to be a better way to do this.

We still haven't had that talk about redundancy of Design ViewModels after standup, but when working on this, I kinda started to think to myself: `Can't we just DI Inject and use the standard ViewModel?`. I haven't yet tried to find out; but maybe that's worth investigating.

### DIFF
--- a/src/NexusMods.App.UI/NexusMods.App.UI.csproj
+++ b/src/NexusMods.App.UI/NexusMods.App.UI.csproj
@@ -556,6 +556,9 @@
         <Compile Update="Overlays\AlphaWarning\AlphaWarningViewModel.cs">
           <DependentUpon>IAlphaWarningViewModel.cs</DependentUpon>
         </Compile>
+        <Compile Update="Pages\LoadoutGrid\LoadoutGridDesignViewModel.cs">
+          <DependentUpon>ILoadoutGridViewModel.cs</DependentUpon>
+        </Compile>
     </ItemGroup>
 
     <ItemGroup>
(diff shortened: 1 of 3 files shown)


halgari:
We can DI inject with a bit of hackery, The trick is making sure that when we start the app we don't start up all the extra services. For example we don't want the CLI server, we want a in-memory DB, etc.

Sewer56:
I do think it would be worth investigating, trying and adding into docs.
Running production/actual code in previewer sounds very valuable.

Sewer56:
I remember having to deal with some of that when setting up one of the example projects. So some of this is familiar, also from standup talks.

What I was thinking if it there's a way of doing it cleanly without introducing dead code. WPF has something called `Markup Extensions` which can be used to effectively inject custom XAML syntax to the files. Although I've never really written one, I reckon something like this could potentially be used to avoid boilerplate. 

For instance, you could make something like `d:DataContext="{local:DesignDataContext {x:Type local:ISomeViewModel}}"` a possibility by having the following in our codebase.

```csharp
public class DesignDataContextExtension : MarkupExtension
{
    public Type InterfaceType { get; set; }

    public DesignDataContextExtension(Type interfaceType)
    {
        InterfaceType = interfaceType;
    }

    public override object ProvideValue(IServiceProvider serviceProvider)
    {
        if (InterfaceType == null)
            throw new InvalidOperationException("InterfaceType property must be set.");

        return serviceProvider.GetRequiredService(InterfaceType);
    }
}
```

Note: Above is pseudocode, I wrote this without an IDE, and of course, I haven't tried it.

Al12rs (approved):
This is fine for now

Al12rs:
Closing and reopening to rebase the CI tests on latest main

Issue #1336, opened 2024-05-13
Loadout Items, Library Items, and Downloads


erri120:
Ever since the beginning of the project, a Loadout contained "Mods":

https://github.com/Nexus-Mods/NexusMods.App/blob/6354819d181ea8f3d6eb75f1d35529fbebec31e4/src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs#L20-L23

Which the new DB and PR #1302, this changed to "Mods" specifying the Loadout they are in:

https://github.com/Nexus-Mods/NexusMods.App/blob/afcab9601d6fda78ea6f0ec2f33f6eec6f1305fe/src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/Mod.cs#L132-L136

https://github.com/Nexus-Mods/NexusMods.App/blob/afcab9601d6fda78ea6f0ec2f33f6eec6f1305fe/src/Abstractions/NexusMods.Abstractions.Loadouts/Loadout.cs#L123-L126

We now have a reversed relationship between Loadouts and "Mods". Instead of getting a Loadout and then having access to its "Mods", we can now just get every Entity that is a Mod and has a `LoadoutId` property that matches the value of the Loadout we're trying to look up.

This issue is about removing the `Mod` model, and replacing it with a generic `LoadoutItem` model.

We've been trying to shoehorn everything into our `Mod` model. This is best illustrated by looking at our `ModCategory` enum that describes what the `Mod` is:

https://github.com/Nexus-Mods/NexusMods.App/blob/afcab9601d6fda78ea6f0ec2f33f6eec6f1305fe/src/Abstractions/NexusMods.Abstractions.Loadouts/Mods/ModCategory.cs#L3-L13

Translating this into English, we're saying that a `Mod` can be "Game Files", it can be "Saves", it can be a "Mod"...

The abstraction that we have, of Loadouts containing Mods, is woefully inadequate. The effects of this can already be seeing by reading #1195 and #1146. We don't even have two Mods installed and already our system is not doing the things we want it to do.

Other mod managers also have an issue with shoehorning everything into a "Mod". In Vortex (@insomnious), a Nexus Mods Collection is represented as a Mod, in Mod Organizer 2 (@Al12rs) a "Mod" is literally just a subdirectory in the `mods` directory.

Continuing to use "Mods" to represent the content of a Loadout will just lead to more and more issues, that'll require weird hacks, and that'll incur a lot of tech debt.

[...]

PR #1509, opened 2024-05-30
Moved: Setting `SuggestedName` into Analyze Step instead of Add Mod Step

Sewer56:
This is a change required for the new library design

https://www.figma.com/design/Ntr9PttR5VrqpTRwwZHzLu/%F0%9F%93%B1-Mod-list---adding%2Fremoving-mods?node-id=3474-29727&t=ac9JhBPo7ATUddi9-1

Previously, a Downloaded File would only become valid once it has been added to a loadout. This is because its `SuggestedName` property has not yet been set, thus the `Repository`'s `IsValid` method would fail, as not all attributes are available.

In simpler words: `Registering a file would not show it in Mod Library`.

This PR moves the logic to set the `SuggestedName` from the `Add Mod to Loadout` step to the `Register Download` step.
This way, registering a file shows it in mod library.

### DIFF
--- a/src/Games/NexusMods.Games.StardewValley/Installers/SMAPIInstaller.cs
+++ b/src/Games/NexusMods.Games.StardewValley/Installers/SMAPIInstaller.cs
@@ -120,6 +120,7 @@ public class SMAPIInstaller : AModInstaller
                 installDataFile.Item.StreamFactory!,
                 (tx, id) => 
                     tx.Add(id, FilePathMetadata.OriginalName, installDataFile.Item.FileName),
+                "SMAPI: Stardew Modding API",
                 cancellationToken);
         }
         else
--- a/src/NexusMods.App.Cli/ProtocolVerbs.cs
+++ b/src/NexusMods.App.Cli/ProtocolVerbs.cs
@@ -65,13 +65,13 @@ public static class ProtocolVerbs
             await httpDownloader.DownloadAsync(new[] { new HttpRequestMessage(HttpMethod.Get, uri) },
                 temporaryPath, null, null, token);
 
+            var nameToSet = string.IsNullOrWhiteSpace(modName) ? "Unknown" : modName;
             var downloadId = await fileOriginRegistry.RegisterDownload(temporaryPath,
                 (tx, id) =>
             {
                 tx.Add(id, FilePathMetadata.OriginalName, temporaryPath.Path.Name);
-            }, token);
-            await archiveInstaller.AddMods(loadout.LoadoutId, downloadId,
-                string.IsNullOrWhiteSpace(modName) ? null : modName, token: token);
+            }, nameToSet, token);
+            await archiveInstaller.AddMods(loadout.LoadoutId, downloadId, token: token);
             return 0;
         });
     }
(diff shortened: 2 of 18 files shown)


Al12rs (approved):
Looks good to me.
@erri120 Did you also want to take a look?

src/NexusMods.DataModel/ArchiveInstaller.cs
```diff
@@ -49,21 +47,15 @@ public class ArchiveInstaller : IArchiveInstaller
...
         var archiveName = download.TryGet(FilePathMetadata.OriginalName, out var originalName) ? originalName.ToString() : null;
         var suggestedName = download.TryGet(DownloadAnalysis.SuggestedName, out var outSuggestedName) ? outSuggestedName : null;
-
-        var modName = defaultModName ?? suggestedName ?? archiveName ?? "<unknown>";
-        {
-            using var dlTx = _conn.BeginTransaction();
-            dlTx.Add(download.Id, DownloadAnalysis.SuggestedName, modName);
-            await dlTx.Commit();
-        }
+        var modName = suggestedName ?? archiveName ?? "Unknown";
```
erri120, line 58: When will `suggestedName` be non-null? Can it be non-null after this PR?

src/NexusMods.App.Cli/ProtocolVerbs.cs
```diff
@@ -65,13 +65,13 @@ private static Task<int> AssociateNxm([Injected] IProtocolRegistration protocolR
             await httpDownloader.DownloadAsync(new[] { new HttpRequestMessage(HttpMethod.Get, uri) },
                 temporaryPath, null, null, token);
 
+            var nameToSet = string.IsNullOrWhiteSpace(modName) ? "Unknown" : modName;
```
erri120, line 68: Just use `name` instead of `modName` since it's set before this.

src/Games/NexusMods.Games.StardewValley/Installers/SMAPIInstaller.cs
```diff
@@ -120,6 +120,7 @@ public class SMAPIInstaller : AModInstaller
                 installDataFile.Item.StreamFactory!,
                 (tx, id) => 
                     tx.Add(id, FilePathMetadata.OriginalName, installDataFile.Item.FileName),
+                "SMAPI: Stardew Modding API",
                 cancellationToken);
```
erri120, line 124: This shouldn't really have a mod name. This isn't a mod, this is just a generic archive that we need to extract.

erri120 (commented):
I think our download API and models should be reworked at some point, probably with #1336. The changes to the SMAPI installer is a good indicator that the API is lacking. The `install.dat` file isn't a mod, it's just a generic archive. Putting `modName` as a **required** parameter will result in further shoehorning, trying to put everything into a "mod".

If you're blocked by this PR, then go ahead and merge it, but we'll be going back to this at some point, hopefully sooner rather than later.

src/NexusMods.DataModel/ArchiveInstaller.cs
Sewer56: > When will `suggestedName` be null?

Seems like you're onto something, yeah, this can be improved.

Originally I was under the impression downloads from 0.5 (e.g. HTTP Downloads) would have exhibited the edge case, if they were downloaded but not added. But after doublechecking, that's not true, let me improve this. 

> Can it be null after this PR?

It cannot

PR #1510, opened 2024-05-30
Remove the single-threaded limits on tests

halgari:
Now that our tests no longer trip over eachother with the stubbed game, we can *attempt* to run the tests in parallel.


halgari:
Did a bunch of work on this, but it's still not where I want it to be, closing for now as we'll likely go a different route

PR #1522, opened 2024-06-03
Update Git Builds pipeline

### LINKED ISSUE #1511: Use proper release builds in git builds
Git builds workflow uses a custom build script which is different from the actual release pipeline.

erri120:
Fixes #1511.

### DIFF
--- a/.github/workflows/git-builds.yaml
+++ b/.github/workflows/git-builds.yaml
@@ -7,38 +7,24 @@ on:
     - cron: '17 08,14 * * 1-5'
   workflow_dispatch:
 
+env:
+  ProjectFile: "src/NexusMods.App/NexusMods.App.csproj"
+
 jobs:
-  build:
-    name: Build (${{ matrix.os }})
-    runs-on: ${{ matrix.os }}
+  build-linux:
+    uses: ./.github/workflows/build-linux-pupnet.yaml
+    with:
+      AppVersion: "0.0.1"
+      # this is a bit of a hack, you can't use environment variables here
+      # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
+      ProjectFile: ${{ needs.transformInputs.outputs.ProjectFile }}
+      RetentionDays: 7
 
-    strategy:
-      matrix:
-        os: [ ubuntu-latest, windows-latest ]
-
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          submodules: "recursive"
-
-      - name: Info
-        run: dotnet --info
-
-      - name: Build (Linux)
-        if: runner.os == 'Linux'
-        shell: sh
-        run: ./scripts/build-linux-x64.sh
-
-      - name: Build (Windows)
-        if: runner.os == 'Windows'
-        shell: pwsh
-        run: scripts\build-win-x64.ps1
-
-      - name: Upload artifact
-        uses: actions/upload-artifact@v3
-        with:
-          # example: "NexusMods.App-Windows-279a5092173a6f687f89dc3c6a2b7c5898e289f6"
-          name: NexusMods.App-${{ runner.os }}-${{ github.sha }}
-          path: ${{ github.workspace }}/bin
-          if-no-files-found: error
-          retention-days: 7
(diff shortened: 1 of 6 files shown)


Sewer56 (approved):

.github/workflows/git-builds.yaml
```diff
@@ -7,38 +7,24 @@ on:
...
-          # example: "NexusMods.App-Windows-279a5092173a6f687f89dc3c6a2b7c5898e289f6"
-          name: NexusMods.App-${{ runner.os }}-${{ github.sha }}
-          path: ${{ github.workspace }}/bin
-          if-no-files-found: error
-          retention-days: 7
+jobs:
+  build-linux:
+    uses: ./.github/workflows/build-linux-pupnet.yaml
+    with:
+      AppVersion: "0.0.1"
```
Al12rs, line 17: Does this get replaced later on or is this going to be the final version?

Sewer56: It should be what it shows as, we use this to set the version during compilation (or rather, PupNet does).
And the App itself gets the version from its assembly, thus this propagates.

I was just considering the issue to be of a different kind than what is being solved here, so I figured I'd make a ticket for improving this instead and be fine with it for the time being.

Sewer56: Namely we would want to discuss what format we want first. So I was considering raising that during standup.

I would prefer something like commit ID (`${{ github.sha }}`) but without testing it, I'm not sure if the current build script would accept that. It might expect a semver, so something like `0.0.0-${{ github.sha }}`.

Al12rs: In MO2 we used Appveyor before, transitioning to GitHub actions now. 
There we would use the current app version and add a monotonically increasing number to it.

I would like to get the current app version from somewhere and append the sha of the commit.

erri120: The artifact names already use the SHA:

erri120: What would be gain by putting the SHA into the assembly version?

Al12rs: The main issue is distinguishing versions from user logs. E.g. someone tests one of these in the office, and a few days later we don't remember to which build those logs referred to.

Sewer56: It's mostly about being to determine which build we're looking at if someone, like @LukeNexusMods or @Pickysaurus  takes a screenshot.

erri120: We show the SHA in the bottom banner of the UI, and it should also be in the logs, no?

Al12rs: I guess we already have that information then. 
Do we still want to have these output an actual version rather than 0.0.1?

erri120: We don't have an actual version. This CI runs twice a day, every work day. These builds are between versions.

erri120: We use [sourcelink](https://github.com/dotnet/sourcelink/tree/main) that automatically injects the SHA into `AssemblyInformationalVersionAttribute`.

Al12rs: I guess it be beneficial to know between which versions though. 
Maybe we can revisit this later if we see it is an issue.

Al12rs (approved):

Pickysaurus:
I'll grab an updated link, don't merge yet

Sewer56:
Do we also want to remove the old build scripts in the `scripts` folder now? They should be unused.

erri120:
> I'll grab an updated link, don't merge yet

Are you going to create a new page?

erri120:
> I'll grab an updated link, don't merge yet

Did you want to post this in #1523?

erri120:
Updated the code to parse the SHA set from SourceLink.

Pickysaurus:
https://help.nexusmods.com/article/132-diagnostics-and-usage-nexus-mods-app

Al12rs (approved):

PR #1525, opened 2024-06-03
Fix missing variable

erri120:


.github/workflows/git-builds.yaml
```diff
@@ -11,7 +11,20 @@ env:
...
+    runs-on: ubuntu-latest
+    outputs:
+      ProjectFile: ${{ steps.setOutputs.outputs.projectFile }}
+
+    steps:
+      - name: Set Outputs
+        id: setOutputs
+        shell: pwsh
+        run: |
+          echo "projectFile=${{ env.ProjectFile }}" >> $env:GITHUB_OUTPUT
```
Al12rs, line 24: I'm a bit confused, why is it using `pwsh` on ubutu-latest?

erri120: I just prefer this syntax over bash. It makes it clearer that we're writing to an environment variable.

Al12rs (approved):

PR #1526, opened 2024-06-03
Add ephemeral page context

### LINKED ISSUE #1255: Support Pages with Ephemeral Context that should not be serialized
For some pages saving and reloading them after an app restart doesn't make sense, because the context is ephemeral.
Examples:
- `DiagnosticDetailsPage`: context is a specific diagnostic entry, that becomes invalid as soon as the loadout changes. It isn't possible to serialize this, nor we particularly desire to do so.

We need the ability to define certain pages as not intended for serialization.
We also need to decide what to do with these pages when the application is closed. Close them, replace them with newPage, show errorPage?

Requires:
- [ ] Define what to do with Ephemeral pages on application close

erri120:
Resolves #1255.


Al12rs (approved):
Didn't test this but code seems sound

PR #1529, opened 2024-06-03
Renamed: Key for AdvancedManualInstaller

### LINKED ISSUE #1527: Advanced installer button just calls the normal installer rather than the advanced one

Sewer56:
This fixes the key from `AdvancedInstaller` to `AdvancedManualInstaller`.
The key got renamed alongside the class at some point, but it seems not all cases were renamed.

Fixes
- #1527

In simpler terms, AdvancedInstaller now shows up when you hit `Add Advanced` in the UI

### DIFF
--- a/src/NexusMods.App.UI/Pages/LoadoutGrid/LoadoutGridViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/LoadoutGrid/LoadoutGridViewModel.cs
@@ -197,7 +197,7 @@ public class LoadoutGridViewModel : APageViewModel<ILoadoutGridViewModel>, ILoad
 
     public Task AddModAdvanced(string path)
     {
-        var installer = _provider.GetKeyedService<IModInstaller>("AdvancedInstaller");
+        var installer = _provider.GetKeyedService<IModInstaller>("AdvancedManualInstaller");
         return AddMod(path, installer);
     }
 


src/NexusMods.App.UI/Pages/LoadoutGrid/LoadoutGridViewModel.cs
```diff
@@ -197,7 +197,7 @@ public LoadoutGridViewModel() : base(null!)
 
     public Task AddModAdvanced(string path)
     {
-        var installer = _provider.GetKeyedService<IModInstaller>("AdvancedInstaller");
+        var installer = _provider.GetKeyedService<IModInstaller>("AdvancedManualInstaller");
```
halgari, line 200: we should probably make this a constant and use that on both ends (registration and usage) so we don't get this again.

halgari (changes requested):

src/NexusMods.App.UI/Pages/LoadoutGrid/LoadoutGridViewModel.cs
Sewer56: I would prefer to do that too. I actually tried doing that.

The issue is, the `DataModel` refers to `AdvancedManualInstaller` key.
Using a constant name requires that I add a dependency to the DataModel.

The alternative is I define the constant in the DataModel.

Sewer56: Edit: We discussed this in meeting

Al12rs (approved):

halgari (approved):

PR #1530, opened 2024-06-04
Installation docs

### LINKED ISSUE #1006: Add "Installation" section docs
Add a section to the docs explaining how to install the App. It should have subsections for each OS type we support: Windows, Linux, and macOS. These subsections should contain a list of requirements, like the .NET runtime, as well as notes on OS-specific functionality.

erri120:
Resolves #1006.

I'll probably rework the updater overlay separately.

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -1,33 +1,28 @@
-﻿!!! info "Describes how to compile and run the Nexus Mods App."
+﻿## Code Contributions
 
-The Nexus Mods App is programmed in C# with latest .NET Runtime. Setup should be easy.
+### Prerequisites
 
-
-## Prerequisites
-
-- Download the Latest [.NET SDK](https://dotnet.microsoft.com/en-us/download/dotnet/8.0).
-- Download an IDE, recommended options below:
-    - [Visual Studio](https://visualstudio.microsoft.com/downloads/)
-    - [JetBrains Rider](https://www.jetbrains.com/rider/)
-
-Our development team uses Rider; however, Visual Studio is a great free offering on Windows.
-Other code editors, like [VS Code](https://code.visualstudio.com/), may work for you with [extensions](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.csdevkit);
-however, they may be a bit more complicated to use.
-
-That's all. Open the project (`.sln`), build, and you're done.
-
-## Tools for UI Development
-
-Visual Studio and JetBrains Rider both have avalonia extensions that allow previewing ui designs (`.axaml` files).
-
-- [Visual Studio Extension](https://marketplace.visualstudio.com/items?itemName=AvaloniaTeam.AvaloniaVS)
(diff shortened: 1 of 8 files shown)


docs/Contributing.md
```diff
@@ -0,0 +1,28 @@
...
+Translations are currently handled via the IDE. See [this issue](https://github.com/Nexus-Mods/NexusMods.App/issues/598) for more details.
+
+## For Package Maintainers
+
+If you want to create a package for your distribution, here are some helpful tips to get started:
+
+- If possible, use `nexusmods-app` for the package name.
+- We ship a build of `7zz` and use that executable unless you set `NEXUSMODS_APP_USE_SYSTEM_EXTRACTOR=1` when publishing. See [this issue](https://github.com/Nexus-Mods/NexusMods.App/issues/1306#issuecomment-2095755699) for details.
+- Set `INSTALLATION_METHOD_PACKAGE_MANAGER` when publishing. We have an integrated updater that will notify users about new versions. If you set this flag, we'll tell the user to update with their package manager.
+- Let us know if you have questions or if you published a new package by joining our Discord.
```
Al12rs, line 28: Add link to discord?

Al12rs:
Index.md changed a lot. Personally I agree that the previous was to wordy and had some unnecessary parts, but this removed the entire explanation about why the app exists and why you should try it.

Sewer56:
I was always under the impression that the documentation we have here was intended to be developer oriented. Especially given that the whole sidebar is ultimately dedicated to dev stuff.

In that vein, I feel like the massive oversimplification of the landing page is trying to cater to end users, when this really shouldn't be the place they should be visiting. This PR also removes much of the reasoning why we built the App in the first place. Yes, it was quite wordy, but it would be a waste to throw away.

Aside from that, this also removes Former Maintainers (Seb) from the readme. Although Seb might have only been responsible for OAuth and Advanced HTTP Downloader for the most part, they still did something. If an actor leaves the production during the filing of a movie, they shouldn't be removed from the credits if they still appear in certain scenes of the film.

Misc Note: The text of the Team section was previously sourced from the Blog Post announcing the App and sorted by team join date. Right now it's team lead and then alphabetical. At least put Tim in a separate section if we're going to do that.

Sewer56 (approved):
What's here works for me, though I did like the older more wordy variation of `The Development Team`.

Run it by community. The old text there was basically what was written by @Pickysaurus in conjugation with @halgari for [the blog post](https://www.nexusmods.com/news/14874)

erri120:
> What's here works for me, though I did like the older more wordy variation of `The Development Team`.

I was originally just going to link to https://github.com/Nexus-Mods/NexusMods.App/graphs/contributors instead.

Sewer56:
> I was originally just going to link to https://github.com/Nexus-Mods/NexusMods.App/graphs/contributors instead.

I think that's a good idea, I'd put that in a admonition

```
!!! tip "For a full list of contributors, see ..."
```

etc.

l0b0 (commented):
Love it!

PR #1543, opened 2024-06-04
Added: Requested Mod Library Refinements

### LINKED ISSUE #1456: LoadoutGrid: change "Installed" column header to "Added"

### LINKED ISSUE #1479: Mod list and Library toolbar/button improvements
Currently the buttons for adding mods are in the incorrect place and adding a mod has the incorrect action.

**Mod list/My Mods page**

- Add mod and Add mod advanced buttons should be removed from the My Mods toolbar.
- Remove icon needs to be updated and aligned to the right (see designs)

**Library page**

- "Add" and "Advanced add" should be added to the Library toolbar, they should only be enabled when an existing mod in the library list is selected, and should only add the mod, either the basic or advanced way, to the mod list.
- A "Get mods" section to be added to the toolbar with 2 buttons:
"From drive" - Opens a window for users to browse their disk and select an archive.
"From Nexus Mods" - Opens a browser to the Stardew Valley mod page on Nexus Mods
- Add icon to be updated on the "Add" button.
- Add button changes to a view button with a check icon once a mod has been added. This button takes the user to the mod list page (Open as a new panel with the mod list page, or if easier/quicker just open the page in the current panel).

## Designs

[Figma link](https://www.figma.com/design/Ntr9PttR5VrqpTRwwZHzLu/%F0%9F%93%B1-Mod-list---adding%2Fremoving-mods?node-id=3474-29727&t=ac9JhBPo7ATUddi9-1)

Sewer56:
- fixes #1479
- fixes #1456

Moves registering and adding a mod to the `Mod Library` view.
This also accounts for edge cases, such as `Mod Library` updating when you perform actions on another panel.

For now installing follows the behaviour that a mod can be ran via Advanced Installer at any time. However, a regular installer may not install a mod twice (the button will be disabled). The design for that is not set in stone (currently undefined). We may change it depending on feedback. For now, I double checked and confirmed with @captainsandypants who said that the current behaviour in the PR is ok.

Misc Note:

I need to figure out if I can get Rider to output [src/NexusMods.App.UI/Resources/Language.Designer.cs](https://github.com/Nexus-Mods/NexusMods.App/compare/mod-library-refinement-v2?expand=1#diff-4151d89cfa8e39e569ed1d73e507bc177d2cd84c7dc80d735d618b2d822edf6c) in the same format as VS. Because that huge diff is unnecessary. I have to reboot to Windows every time I add some strings.


src/NexusMods.App.UI/Resources/Language.pl.resx
```diff
@@ -473,4 +473,16 @@
     <data name="LoadoutGridView__Remove" xml:space="preserve">
         <value>Usuń</value>
     </data>
+    <data name="FileOriginsPage_GetMods" xml:space="preserve">
+        <value>Pobraj mody:</value>
```
Patriot99, line 477:
```suggestion
        <value>Pobierz mody:</value>
```

Patriot99 (changes requested):
The word `Pobraj` does not exist in the Polish language.

src/NexusMods.App.UI/Pages/LoadoutGrid/LoadoutGridViewModel.cs
```diff
@@ -83,13 +74,8 @@ public LoadoutGridViewModel() : base(null!)
         IWindowManager windowManager,
         ISettingsManager settingsManager) : base(windowManager)
     {
-        _logger = logger;
-        _fileSystem = fileSystem;
         _conn = conn;
-        _archiveInstaller = archiveInstaller;
-        _fileOriginRegistry = fileOriginRegistry;
-        _provider = provider;
-
```
Al12rs, line 92: You can remove fileSystem, archiveInstaller and fileOriginRegistry parameters from the constructor as well

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPage.cs
```diff
@@ -37,6 +40,11 @@ public override IFileOriginsPageViewModel CreateViewModel(FileOriginsPageContext
             ServiceProvider.GetRequiredService<IArchiveInstaller>(),
             ServiceProvider.GetRequiredService<IRepository<DownloadAnalysis.Model>>(),
             ServiceProvider.GetRequiredService<IConnection>(),
+            ServiceProvider.GetRequiredService<IFileSystem>(),
+            ServiceProvider.GetRequiredService<ILogger<FileOriginsPageViewModel>>(),
+            ServiceProvider,
+            ServiceProvider.GetRequiredService<IFileOriginRegistry>(),
+            ServiceProvider.GetRequiredService<IOSInterop>(),
             ServiceProvider.GetRequiredService<IWindowManager>()
```
erri120, line 48: This is getting ridiculous. Just pass the `LoadoutId` and the `IServiceProvider` and let the constructor fetch the other stuff.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml.cs
```diff
@@ -16,8 +23,37 @@ public FileOriginsPageView()
...
+                .Subscribe(UpdateAddButtonState)
+                .DisposeWith(d);
+
             DataGrid.Width = Double.NaN;
+            
+            AddModButton.Command = ReactiveCommand.CreateFromTask(ViewModel!.AddMod);
+            AddModAdvancedButton.Command = ReactiveCommand.CreateFromTask(ViewModel!.AddModAdvanced);
+            GetModsFromNexusButton.Command = ReactiveCommand.CreateFromTask(ViewModel!.OpenNexusModPage);
+            GetModsFromDriveButton.Command = ReactiveCommand.CreateFromTask(
+                async () => await ViewModel!.RegisterFromDisk(TopLevel.GetTopLevel(this)!.StorageProvider));
```
Al12rs, line 56: Why are these commands in the View? Shouldn't these all be in the ViewModel and shouldn't this be using `BindCommand()` ?

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml.cs
```diff
@@ -16,8 +23,37 @@ public FileOriginsPageView()
...
+                .Subscribe(_ => { UpdateAddButtonState(false); })
+                .DisposeWith(d);
+            ViewModel!.SelectedModsObservable.WhenValueChanged(model => model.IsModAddedToLoadout)
+                .Subscribe(UpdateAddButtonState)
+                .DisposeWith(d);
+
             DataGrid.Width = Double.NaN;
+            
+            AddModButton.Command = ReactiveCommand.CreateFromTask(ViewModel!.AddMod);
+            AddModAdvancedButton.Command = ReactiveCommand.CreateFromTask(ViewModel!.AddModAdvanced);
```
Al12rs, line 53: This is enabled even if you have no item selected, which doesn't make sense.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
```diff
@@ -83,5 +98,23 @@ public class FileOriginEntryViewModel : AViewModel<IFileOriginEntryViewModel>, I
...
+            var pageData = new PageData()
+            {
+                FactoryId = LoadoutGridPageFactory.StaticId,
+                Context = new LoadoutGridContext()
+                {
+                    LoadoutId = loadoutId,
+                }
+            };
+
+            var behavior = workspaceController.GetOpenPageBehavior(pageData, info, Optional<PageIdBundle>.None);
```
Al12rs, line 116: You should pass in the `IdBundle` of the parent FileOriginsPage

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml.cs
```diff
@@ -16,8 +23,37 @@ public FileOriginsPageView()
...
+                AddModButton.IsEnabled = enable;
+                // Note Add(Advanced) should always be available.
+            }
+
+            ViewModel!.SelectedModsCollection.ObserveCollectionChanges()
+                .Subscribe(_ => { UpdateAddButtonState(false); })
+                .DisposeWith(d);
+            ViewModel!.SelectedModsObservable.WhenValueChanged(model => model.IsModAddedToLoadout)
+                .Subscribe(UpdateAddButtonState)
+                .DisposeWith(d);
```
erri120, line 48: I don't really understand this code. The Add buttons should only be enabled when the user has selected an item, no? This code will do that:

```csharp
            var serialDisposable = new SerialDisposable();
            serialDisposable.DisposeWith(d);

            this.WhenAnyValue(view => view.ViewModel)
                .Select(vm =>
                {
                    if (vm is null) return Disposable.Empty;

                    return vm
                        .WhenAnyObservable(x => x.SelectedModsObservable)
                        .Select(_ => vm.SelectedModsCollection.Count > 0)
                        .SubscribeWithErrorLogging(hasSelection =>
                        {
                            AddModButton.IsEnabled = hasSelection;
                            AddModAdvancedButton.IsEnabled = hasSelection;
                        });
                }).SubscribeWithErrorLogging(disposable =>
                {
                    serialDisposable.Disposable = disposable;
                }).DisposeWith(d);
```

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -13,34 +20,70 @@
...
+    private readonly IFileOriginRegistry _fileOriginRegistry;
+    private readonly IOSInterop _osInterop;
     private readonly IRepository<DownloadAnalysis.Model> _dlAnalysisRepo;
     private readonly IArchiveInstaller _archiveInstaller;
 
     public ReadOnlyObservableCollection<IFileOriginEntryViewModel> FileOrigins => _fileOrigins;
-    private ReadOnlyObservableCollection<IFileOriginEntryViewModel> _fileOrigins = new([]);
+    private ReadOnlyObservableCollection<IFileOriginEntryViewModel> _fileOrigins;
+
+    // Note(sewer). Adding a [Reactive] attribute breaks the setter, don't do it please.
```
erri120, line 48: Wdym with "breaks the setter"?

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -51,22 +94,34 @@ public class FileOriginsPageViewModel : APageViewModel<IFileOriginsPageViewModel
...
         var loadout = _conn.Db.Get<Loadout.Model>(LoadoutId.Value);
         var game = loadout.Installation.Game;
+        try
+        {
+            _gameDomain = loadout.Installation.Game.Domain;
+        }
+        catch (Exception)
+        {
+            _gameDomain = GameDomain.DefaultValue;
+        }
```
erri120, line 104: Why try/catch?

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -78,4 +133,68 @@ public static bool FilterDownloadAnalysisModel(DownloadAnalysis.Model model, Gam
         if (model.Contains(StreamBasedFileOriginMetadata.StreamBasedOrigin)) return false;
         return true;
     }
+
+    [UsedImplicitly]
```
erri120, line 137: Why `UsedImplicitly`?

erri120 (changes requested):

Al12rs (changes requested):

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
```diff
@@ -83,5 +98,23 @@ public class FileOriginEntryViewModel : AViewModel<IFileOriginEntryViewModel>, I
             .Merge(interval.Select(_ => LastInstalledDate))
             .Select(date => date.Equals(DateTime.MinValue) ? "-" : date.Humanize())
             .ToProperty(this, vm => vm.DisplayLastInstalledDate, scheduler: RxApp.MainThreadScheduler);
+
+        ViewModCommand = ReactiveCommand.Create<NavigationInformation>(info =>
```
Al12rs, line 102: You should make the commands settable and set them from `FileOriginsPageViewModel.cs` when FileOriginEntryVM is created in the transform. This allows having access to everything in FileOriginsPageVM and de duplicating functionality such as Add

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
```diff
@@ -20,7 +23,11 @@ public class FileOriginEntryViewModel : AViewModel<IFileOriginEntryViewModel>, I
     public string Version { get; }
     public Size Size { get; }
     public DateTime ArchiveDate { get; }
-    public ReactiveCommand<Unit, Unit> AddToLoadoutCommand { get; init; }
+    public ReactiveCommand<IModInstaller?, Unit> AddToLoadoutCommand { get; init; }
```
Al12rs, line 26: Don't pass IModInstaller as a parameter, this is never passed from UI, this is a hack to reuse the command for multi selection. This just makes the code confusing. 
You can use a method if you really want to reuse code.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml.cs
Al12rs: Even then, the button enabled state should be controlled by the command `CanExecute` observable, not by manually updating the IsEnabled state

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml.cs
```diff
@@ -16,8 +23,37 @@ public FileOriginsPageView()
...
+            void UpdateAddButtonState(bool isModAdded)
+            {
+                var enable = false;
+                if (!isModAdded)
+                    enable = ViewModel!.SelectedModsCollection.Count > 0 && 
+                             !ViewModel!.SelectedModsCollection.Any(x => x.IsModAddedToLoadout);
+
+                AddModButton.IsEnabled = enable;
+                // Note Add(Advanced) should always be available.
+            }
```
Al12rs, line 41: Use CanExecute of the Commands, put commands in VM. 
Both Add and Add advanced should be enabled if there is any selection. Doesn't matter if that mod was already installed.
That is overcomplicating the code and is isn't the desired behavior, please just get rid of it.

Al12rs (changes requested):

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
Sewer56: It discards the code in the `set` section.

This normally doesn't happen with the common Fody PropertyChanged weaver, I think it's something specific to this weaver. I amended the code to note that.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
Sewer56: No specific reason. 
It was part of the old code which was moved over.
I'll remove it.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
Sewer56: Rider wasn't picking up on this in an earlier code iteration for some reason; and requested to remove it.
VS was.

That's no longer the case thankfully.

src/NexusMods.App.UI/Resources/Language.pl.resx
```diff
@@ -473,6 +473,18 @@
     <data name="LoadoutGridView__Remove" xml:space="preserve">
         <value>Usuń</value>
     </data>
+    <data name="FileOriginsPage_GetMods" xml:space="preserve">
+        <value>Pobierz mody:</value>
```
Patriot99, line 477: I suggest changing `Pobierz` (Download) to `Uzyskaj` (Get), because `Pobierz` (Download) does not match to `Z dyksu` (From Drive).

```suggestion
          <value>Uzyskaj mody:</value>
```

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
erri120: Well yeah, you wouldn't wanna have custom code inside the `set` method.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml.cs
erri120: Yeah, with command bindings this should be an `canExecute` observable.

src/NexusMods.App.UI/Helpers/DataGridSelectedItemsAsObservableHelper.cs
```diff
@@ -0,0 +1,125 @@
...
+        // derived from the SourceCache will have a brief period of time where it shows
+        // an item which should just have been removed.
+        _selectedItemsCache.Edit(updater =>
+        {
+            foreach (var removedItem in e.RemovedItems)
+                updater.RemoveKey((T)removedItem);
+
+            foreach (var addedItem in e.AddedItems)
+                updater.AddOrUpdate((T)addedItem);
+        });
```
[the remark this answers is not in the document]
Sewer56: I added these for safety, just in case.
Though a failure here means an issue on Avalonia's part, as they manage this.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
Sewer56: Can we have some docs somewhere on what this provides us? I think it's for identifying the current view so we replace the right tab, panel etc. I only figured from reading source, it's not documented anywhere.

Anyway sorted.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
Sewer56: I moved out initializing `viewModCommand` to parent for now, since it's the same for all items, but can you elaborate a bit more on this tomorrow?

I can technically pass an `Add` method to the entry, but then that's just inverting control, so rather than VM calling Entry, the Entry calls VM. I'm not sure what's gained in that situation, since it's just swapping the two. The actual code to add mod is a single method call, so there's not much logic there.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml.cs
Sewer56: The behaviour in the proposed code is a bit different.
It only checks for selection. If you delete a mod from the `My Mods` page, the mod could still be re-added.

In any case, I did rewrite it so it's very similar to this, and moved it to the ViewModel, so it uses the `canExecute` on the commands instead of touching the buttons.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageView.axaml.cs
```diff
@@ -16,8 +23,21 @@ public FileOriginsPageView()
...
-            
+
+            DataGrid.SelectedItemsToProperty(ViewModel!, vm => vm.SelectedModsObservable)
+                .DisposeWith(d);
+
             DataGrid.Width = Double.NaN;
+
+            AddModButton.Command = ViewModel!.AddMod;
+            AddModAdvancedButton.Command = ViewModel!.AddModAdvanced;
+            GetModsFromNexusButton.Command = ViewModel!.OpenNexusModPage;
```
erri120, line 34: Use `BindCommand`.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -1,73 +1,169 @@
...
 
-        var loadout = _conn.Db.Get<Loadout.Model>(LoadoutId.Value);
+        var loadout = conn.Db.Get<Loadout.Model>(LoadoutId.Value);
         var game = loadout.Installation.Game;
+        _gameDomain = loadout.Installation.Game.Domain;
+
+        _fileOrigins = new ReadOnlyObservableCollection<IFileOriginEntryViewModel>([]);
+        this.WhenActivated(d =>
+        {
+            var workspaceController = GetWorkspaceController();
```
erri120, line 98: Don't have two `WhenActivated` blocks.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -1,73 +1,169 @@
...
+                })
+                .Subscribe()
+                .DisposeWith(d);
+            
+            AddMod = ReactiveCommand.CreateFromTask(async (token) => await DoAddModImpl(null, token), canAddMod);
+            AddModAdvanced = ReactiveCommand.CreateFromTask(async (token) =>
+            {
+                await DoAddModImpl(_provider.GetKeyedService<IModInstaller>("AdvancedManualInstaller"), token);
+            }, canAddAdvancedMod);
         });
```
erri120, line 167:
```csharp
        var canAddMod = new Subject<bool>();
        var canAddAdvancedMod = new Subject<bool>();

        AddMod = ReactiveCommand.CreateFromTask(async cancellationToken => await DoAddModImpl(null, cancellationToken), canAddMod);
        AddModAdvanced = ReactiveCommand.CreateFromTask(async cancellationToken =>
        {
            await DoAddModImpl(_provider.GetKeyedService<IModInstaller>("AdvancedManualInstaller"), cancellationToken);
        }, canAddAdvancedMod);

        this.WhenActivated(d =>
        {
            var serialDisposable = new SerialDisposable();
            serialDisposable.DisposeWith(d);

            this.WhenAnyValue(vm => vm.SelectedModsObservable)
                .Select(observable =>
                {
                    return observable
                        .WhenValueChanged(x => x.IsModAddedToLoadout)
                        .Select(_ => SelectedModsCollection.Count > 0)
                        .SubscribeWithErrorLogging(hasSelection =>
                        {
                            // Add (Advanced) is always available.
                            canAddAdvancedMod.OnNext(hasSelection);

                            // Add is only available if no mod is already added.
                            canAddMod.OnNext(hasSelection && SelectedModsCollection.All(x => !x.IsModAddedToLoadout));
                        });
                })
                .SubscribeWithErrorLogging(disposable => serialDisposable.Disposable = disposable)
                .DisposeWith(d);
        });
```

erri120 (changes requested):

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
Sewer56: Question: Is there a reason the inner query can't also just do `DisposeWith(d)`?

I'm not exactly sure of the purpose of the `SerialDisposable` here. I can tell that the intent is to dispose the inner with the outer, as both produce observable sequences, but I don't see what we gain from moving it through `SerialDisposable`, which I believe is just a redirection layer.

Since the outer `SubscribeWithErrorLogging` has the result of the inner `SubscribeWithErrorLogging` as a parameter, which then gets `DisposeWith(d)`'d anyway/.

erri120: Whenever the property `SelectedModsObservable` changes, we create a new observable stream and subscribe to it. When `SelectedModsObservable` changes again, we again create a new observable stream and subscribe to it.

If you don't dispose of the previous subscription when `SelectedModsObservable` changes again, you now have a leaking subscription.

It doesn't matter if `SelectedModsObservable` will only change once, that implementation detail is lost in this context. The best practice is to always dispose of subscription, and that's what you should do.

Al12rs: All this discussion for functionality that wasn't part of the requirements and that we need to remove anyways. 
You should always be able to reinstall the mod. It should not matter if you have a copy installed already or not.

Sewer56: It shouldn't be a leaking subscription if you `DisposeWith` the inner, however it will remain alive until the view itself is deactivated. Unless there's some very special implementation detail I'm unaware of.

Either way, using SerialDisposable is desirable here because we can dispose of previous items immediately, rather than keeping them forever until the view deactivates. I did figure it out eventually.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
```diff
@@ -83,5 +88,14 @@ public class FileOriginEntryViewModel : AViewModel<IFileOriginEntryViewModel>, I
...
             .ToProperty(this, vm => vm.DisplayLastInstalledDate, scheduler: RxApp.MainThreadScheduler);
+
+        ViewModCommand = viewModCommand;
+    }
+    
+    public async Task AddToLoadout(CancellationToken token) => await AddUsingInstallerToLoadout(null, token);
+
+    public async Task AddUsingInstallerToLoadout(IModInstaller? installer, CancellationToken token)
+    {
+        await _archiveInstaller.AddMods(_loadoutId, FileOrigin, installer, token);
```
Al12rs, line 99: This stuff can be put in FileOriginsPageViewModel instead of putting it inside the entry VM.
Parent is the one doing multi selection operation, and it shouldn't invoke the commands of the children to do the operation, it should do it directly. So it should be in charge of all the add mod operations. 


In the future we might handle multi installs in a more sensible manner than calling AddMod() multiple times.

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginEntry/FileOriginEntryViewModel.cs
```diff
@@ -47,17 +56,13 @@ public class FileOriginEntryViewModel : AViewModel<IFileOriginEntryViewModel>, I
...
-            await archiveInstaller.AddMods(loadoutId, fileOrigin);
-        });
-        
         Version = fileOrigin.TryGet(DownloaderState.Version, out var version) && version != "Unknown"
             ? version
             : "-";
         
         ArchiveDate = fileOrigin.GetCreatedAt();
-
+        AddToLoadoutCommand = ReactiveCommand.CreateFromTask(async () => await AddToLoadout(CancellationToken.None));
```
Al12rs, line 64: I think this as well should be in FileOriginsPageVM rather than entryVM

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -1,73 +1,159 @@
...
-                        _conn,
-                        _archiveInstaller,
+                .Transform(fileOrigin => (IFileOriginEntryViewModel)
+                    new FileOriginEntryViewModel(
+                        conn,
+                        archiveInstaller,
                         LoadoutId,
-                        fileOrigin
+                        fileOrigin,
+                        viewModCommand
```
Al12rs, line 138: When I say to move the Command Creations in the PageVM I mean here:
```csharp
var vm = (IFileOriginEntryViewModel)
                    new FileOriginEntryViewModel(
                        conn,
                        archiveInstaller,
                        LoadoutId,
                        fileOrigin)
vm.ViewModCommand = viewModCommand;
vm.AddModCommand = ReactiveCommand.CreateFromTask(){..
return vm
```

Al12rs (changes requested):

src/NexusMods.App.UI/Pages/ModLibrary/FileOrigins/FileOriginsPageViewModel.cs
```diff
@@ -199,7 +199,7 @@ private Task RegisterFileFromDisk(string path)
 
     public async Task OpenNexusModPage()
     {
-        var url = $"https://www.nexusmods.com/{_gameDomain.Value}";
+        var url = $"https://www.nexusmods.com/{_gameDomain.Value}?pk_source=nexusmodsapp";
```
erri120, line 202: Depending on what gets merged first, either wait for #1557 or I'll include this in #1557.

erri120 (approved):

Al12rs (approved):

Al12rs:
Plz wait to merge, I want to run the code locally to check a couple of things

Al12rs:
All good

PR #1546, opened 2024-06-05
ComboBox styling

### LINKED ISSUE #1545: ComboBox styles
Currently dropdown buttons are using the default Avalonia styles.

Toolbar buttons (Theme select, Changelog version) should use: **Tertiary**

Settings should use: **Secondary**

Image

[Figma link](https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=537-6795&t=rkPjV2EuHyAo4dYM-1)

erri120:
Resolves #1545.


src/Themes/NexusMods.Themes.NexusFluentDark/NexusMods.Themes.NexusFluentDark.csproj
```diff
@@ -57,4 +57,19 @@
...
+      <AvaloniaXaml Update="Styles\Controls\ComboBox\ComboBox.Primary.axaml">
+        <DependentUpon>ComboBox.axaml</DependentUpon>
+      </AvaloniaXaml>
+      <AvaloniaXaml Update="Styles\Controls\ComboBox\ComboBox.Secondary.axaml">
+        <DependentUpon>ComboBox.axaml</DependentUpon>
+      </AvaloniaXaml>
+      <AvaloniaXaml Update="Styles\Controls\ComboBox\ComboBox.Tertiary.axaml">
+        <DependentUpon>ComboBox.axaml</DependentUpon>
+      </AvaloniaXaml>
+    </ItemGroup>
```
Al12rs, line 73: I don't like all these separate files, especially if most of them share the same style but change the color.

Style files should also contain "Style" in the file name so that they can be distinguished from user controls during quick file search.  

I would put Primary, Secondary, Tertiary and ComboBox in a single file, up to you if you want to get ComboBoxItem in there as well but at least for that one it makes sense to have it separate.

Al12rs (changes requested):
Regarding the na

Al12rs (approved):

PR #1549, opened 2024-06-05
Restyle of ToggleSwitch

### LINKED ISSUE #1550: Toggle switch design improvement
Design improvement for toggles, improves accessibility with clearer hover and pressed states as well as larger tap targets.:

Image

https://www.figma.com/design/4Kt4QcSv8rACPpo3y4T4lr/%F0%9F%93%B1-Apply-loadout-changes?node-id=753-32953&t=c8ArLNB3hGOX5EfU-1

insomnious:
Restyled ToggleSwitch to match Figma. Added the smaller variation by creating a `Compact` class. Updated the two places that the control is used to use this new class. Needed to set ClipToBounds=False in a couple of places to accommodate the new hover state.

Fixes #1550


src/NexusMods.App.UI/Controls/Settings/SettingEntries/SettingInteractionControls/Toggle/SettingToggleControl.axaml.cs
```diff
@@ -12,6 +12,8 @@ public SettingToggleControl()
     {
         InitializeComponent();
 
+        ClipToBounds = false;
```
Al12rs, line 15: Logically I would think that this should not be needed since it should already be set by the style, but Avalonia doesn't always behave logically

src/NexusMods.App.UI/Controls/Settings/SettingEntries/SettingInteractionControls/Toggle/SettingToggleControl.axaml
```diff
@@ -8,7 +8,7 @@
     mc:Ignorable="d"
     x:Class="NexusMods.App.UI.Controls.Settings.SettingEntries.SettingToggleControl">
 
-    <ToggleSwitch HorizontalAlignment="Center" x:Name="ToggleSwitch">
+    <ToggleSwitch Classes="Compact" HorizontalAlignment="Center" x:Name="ToggleSwitch" ClipToBounds="False">
```
Al12rs, line 11: Here as well I would not expect ClipToBounds to be required since it's already set in the style

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/ToggleSwitch/ToggleSwitchStyles.axaml
```diff
@@ -2,11 +2,89 @@
...
         <Border Padding="10" Classes="Low">
+ 
             <StackPanel Classes="Spacing-1" Margin="6">
-                <ToggleSwitch />
-                <ToggleSwitch IsEnabled="False" IsChecked="False"/>
-                <ToggleSwitch IsEnabled="False" IsChecked="True"/>
-            </StackPanel>
+                <TextBlock Classes="HeadingXSSemi">ToggleSwitch</TextBlock>    
+            
+            <Grid>
```
Al12rs, line 9: Indentation seems to be wrong here, try formatting selection

Al12rs (commented):
Couple of items above, check if those ClipToBounds are still needed, or if the other place fixed the issue and they are redundant now.

src/NexusMods.App.UI/Controls/Settings/SettingEntries/SettingInteractionControls/Toggle/SettingToggleControl.axaml.cs
insomnious: It felt a bit weird needing to put that there but this component was set to True by default 🤷

insomnious:
Removed one of the ClipToBounds now it was set in the style

src/NexusMods.App.UI/Controls/Settings/SettingEntries/SettingInteractionControls/Toggle/SettingToggleControl.axaml.cs
Al12rs: I tested this and you can move this into the `SettingToggleControl.axaml`, but on the control definition itself, not on the ToggleSwitch. 

```xml
<reactiveUi:ReactiveUserControl
    x:TypeArguments="settingEntries:ISettingToggleViewModel" xmlns="https://github.com/avaloniaui"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    xmlns:reactiveUi="http://reactiveui.net"
    xmlns:settingEntries="clr-namespace:NexusMods.App.UI.Controls.Settings.SettingEntries"
    mc:Ignorable="d"
    x:Class="NexusMods.App.UI.Controls.Settings.SettingEntries.SettingToggleControl"
    ClipToBounds="False">
```

Al12rs (commented):
See comment above for moving ClipToBounds to xaml.

Btw I tested this and it's glorious.

One thing that I can think we might want in the future would be adding transition speeds, to simulate an animation when switching the states, but we can see about that later.

src/NexusMods.App.UI/Controls/Settings/SettingEntries/SettingInteractionControls/Toggle/SettingToggleControl.axaml.cs
insomnious: Oh nice, yeah that's way better.

Al12rs (approved):

PR #1557, opened 2024-06-06
Update URLs

erri120:


Al12rs (commented):
Couple of comments would be helpful to clarify the differences and significance of the two functions and whether more will be needed

Al12rs (approved):

PR #1559, opened 2024-06-06
Add ability to set mod name in `ArchiveInstaller.AddMods()`

### LINKED ISSUE #1541: Flaky test: `CanCreateAndManageLists`: "`System.InvalidOperationException` : Sequence contains no matching element"
image

[NexusMods.CLI.Tests.VerbTests.ModManagementVerbs.CanCreateAndManageLists: tests/NexusMods.CLI.Tests/VerbTests/ModManagementVerbs.cs#L32](https://github.com/Nexus-Mods/NexusMods.App/actions/runs/9369311792?pr=1540)

This has been happening consistently on PRs lately, on MacOS

Al12rs:
This should hopefully
- fix #1541 

fingers crossed

## Details
There are two reasons that test is failing on the CI:
- If an archive is installed multiple times, we were using the previous name, so the test was expecting a mod with a specific name, but wouldn't find it.
- Other problem is that the archive is not supposed to be installed multiple times, so this test is sharing the DataStore with other concurrent tests, which are also installing the same file concurrently, which should not be happening.

This PR only fixes the first problem, which is also a UX problem for normal users.
The sharing db in the tests is an issue that we also need to fix but outside the scope for this.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Installers/IArchiveInstaller.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Installers/IArchiveInstaller.cs
@@ -21,11 +21,21 @@ public interface IArchiveInstaller
     /// standard way to "install" a mod. If the installer is not provided, the installers on the loadout's game
     /// will be tried, in the order they are defined, otherwise the provided installer will be used.
     /// </summary>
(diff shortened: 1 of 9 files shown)


Sewer56:
This is fine, for UX too.
If we're adding a duplicate from library, this would be very handy for distinguishing a duplicate; so it gets a thumbs up from me.

Sewer56 (approved):

PR #1560, opened 2024-06-06
Fix Test_Locators_Windows failing on xboxGamePass since it was disabled

Al12rs:
In Game Finder
@erri120 Just a heads up that this will need to be enabled back in once we add support for GamePass again


Sewer56:
Add a reminder to the ticket for GamePass support, please.
That way we're less likely to miss it.

Sewer56 (approved):
I forgot to check this

Al12rs:
Added the note in the ticket

PR #1566, opened 2024-06-10
ToolTip Styling and Designer preview fix

### LINKED ISSUE #952: Add ToolTip Styling
image
image

Look into fixing the styles for those tooltips and see if it can't be done for all tooltips

Al12rs:
- fix #952 
- Fixed DI related exception in designer preview

Regarding the designer preview, we were not calling `Host.StartAsync()`
After doing that though, I excluded SingleProcess (CLIServer and CLIClient) from being added to the Designer Service collection, to avoid the designer claiming single process lock.

For ToolTip styling, I added the Styles as defined on figma and fixed the SpineDownloadButton and IconButton Styles from bleeding into the ToolTips.

The ToolTip BoxShadow seems to work only in some cases though, there might be a problem with the outer padding value not being respected for some reason. The main issue was that `ClipToBounds=False` didn't work on `ToolTip` control itself, so I had to add the padding. 

image
image
image
image


Al12rs:
Found the issue for tooltip styling not applying. I have to fix the tooltip position now though..

erri120 (approved):

PR #1597, opened 2024-06-12
Hide download speed graph line after 5 ticks

erri120:
As requested by @captainsandypants.

### DIFF
--- a/src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
@@ -477,14 +477,9 @@ public class InProgressViewModel : APageViewModel<IInProgressViewModel>, IInProg
         }
 
         const int delay = 5;
-        if (_throughputValues.Count > delay)
-        {
-            _lineSeries.IsVisible = _throughputValues.TakeLast(delay).Sum(x => x.Value!.Value) > 0.0;
-        }
-        else
-        {
-            _lineSeries.IsVisible = false;
-        }
+        _lineSeries.IsVisible = _throughputValues
+            .TakeLast(Math.Min(delay, _throughputValues.Count))
+            .Sum(x => x.Value!.Value) > 0.0;
 
         _throughputValues.Add(new DateTimePoint(DateTime.Now, throughput));
     }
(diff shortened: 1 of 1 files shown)


src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
```diff
@@ -472,6 +477,12 @@ private void UpdateWindowInfoInternal()
         }
 
         _throughputValues.Add(new DateTimePoint(DateTime.Now, throughput));
+
+        const int delay = 5;
+        if (_throughputValues.Count > delay)
+        {
+            _lineSeries.IsVisible = _throughputValues.TakeLast(delay).Sum(x => x.Value!.Value) > 0.0;
+        }
```
Al12rs, line 485: If you open the page with no downloads, the line will only disappear after 5 seconds

Al12rs (changes requested):
If you open the page

src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
```diff
@@ -471,6 +476,16 @@ private void UpdateWindowInfoInternal()
...
         }
 
+        const int delay = 5;
+        if (_throughputValues.Count > delay)
+        {
+            _lineSeries.IsVisible = _throughputValues.TakeLast(delay).Sum(x => x.Value!.Value) > 0.0;
+        }
+        else
+        {
+            _lineSeries.IsVisible = false;
```
Al12rs, line 486:
```suggestion
            _lineSeries.IsVisible = _throughputValues.TakeLast(_throughputValues.Count).Sum(x => x.Value!.Value) > 0.0;
```

Otherwise if you start a DL or a dl is ongoing it will not be shown for 5 seconds

Al12rs (changes requested):

src/NexusMods.App.UI/Pages/Downloads/InProgressViewModel.cs
Al12rs: Or just change the thing to pass in 
`_throughputValues.TakeLast(_throughputValues.Count < 5 ? _throughputValues.Count : 5 )`

Al12rs (approved):

Al12rs:
This is consistently stalling every time, should check tests locally before proceeding

Al12rs (approved):

PR #1615, opened 2024-06-13
Remove transform scale on left menu button click

### LINKED ISSUE #1609: Transition transform effect removal on left hand menu items
Remove the transform transition on click. Replace with a "pressed" state (pressed state is currently the same as active state).

[Figma link](https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=971-5497&t=vTR45Pqi1V8KWQrh-1)

Image

insomnious:
@captainsandypants is happy with the global  removal of transform scale, and I've commented out the press state as it doesn't make sense until the active state is implemented

Fixes #1609

### DIFF
--- a/src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/BareIconStyles.axaml
+++ b/src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/BareIconStyles.axaml
@@ -43,11 +43,6 @@
             <Setter Property="Size" Value="24" />
         </Style>
 
-        <Style Selector="^:pressed">
-            <!-- Accentuate the pressed animation -->
-            <Setter Property="RenderTransform" Value="scale(0.9)" />
-        </Style>
-
         <Style Selector="^:disabled">
             <Setter Property="Opacity" Value="{StaticResource OpacityDisabledElement}" />
         </Style>
--- a/src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/BareIconStyles.axaml
+++ b/src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/BareIconStyles.axaml
@@ -41,8 +41,21 @@
 
         <Style Selector="^ /template/ icons|UnifiedIcon">
             <Setter Property="Size" Value="24" />
+            <Setter Property="Foreground" Value="{StaticResource NeutralTranslucentModerateBrush}" />
         </Style>
-
+                       
+        <Style Selector="^:pointerover">
+            <Style Selector="^ /template/ icons|UnifiedIcon">
+                <Setter Property="Foreground" Value="{StaticResource NeutralStrongBrush}" />
+            </Style>
+        </Style>
+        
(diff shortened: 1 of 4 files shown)


Sewer56 (commented):
This change will also affect the Spine. Is that intended?

We might want to go through uses of `Property="RenderTransform" Value="scale(`, and selectively remove the ones we don't want for now.

Rest is fine to me.

insomnious:
The spine buttons are set specifically in their style and so they aren't affected by the global `<Setter Property="RenderTransform" Value="none" />` anyway. 

Let me add a commit removing the other localised scale resets like in:

image

Sewer56:
> The spine buttons are set specifically in their style 

Yeah, seems like it.
I was looking at the download button earlier, which didn't have that.

Seems to be a bug, as it scaled differently than the rest. I fixed that on my branch, so we're good.

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/PillStyles.xaml
```diff
@@ -219,12 +219,8 @@
 
     <!-- Pill Label -->
     <!-- Labels should have all interactions disabled, so they don't look interactable -->
+    <!-- NOTE(insomnious): left this empty selector here to make it easier to find in future -->
     <Style Selector="Button.Pill.Label">
-
-        <!--  Disable the scaling found in Fluent buttons when pressing -->
-        <Style Selector="^:pressed">
```
Sewer56, line 225: This should probably stay.
The pills, like `Premium` should be non-interactable generally. 
By removing the 1.0, Avalonia applies the default, which is to shrink.

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/BareIconStyles.axaml
```diff
@@ -43,11 +43,6 @@
             <Setter Property="Size" Value="24" />
         </Style>
 
-        <Style Selector="^:pressed">
-            <!-- Accentuate the pressed animation -->
```
Sewer56, line 47: This is used in the downloads (In Progress) page; and should not be removed.

Sewer56 (changes requested):
(See Above)

Anyway I tested the rest and it's good to me.
So just get those two undone and we're good.

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Button/BareIconStyles.axaml
insomnious: Requested by @captainsandypants, Updated with color states so we don't have to use transform scale

https://github.com/Nexus-Mods/NexusMods.App/assets/26166085/3cc7ea3f-7b57-4273-a085-c055f3a9f6ab

Sewer56 (approved):

PR #1624, opened 2024-06-13
Restructured the docs website to accommodate both user and dev docs

Pickysaurus:
Added some very basic user-facing docs.


docs/users/SystemRequirements.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Windows 
+
+|| Minimum | Recommended |
+| --- | --- | --- |
+| OS | Windows 10 or higher (64-bit OS required) | Windows 10/11 with updates |
+| Processor | Intel®️ Core™️ i5-7500 | Intel®️ Core™️ i7-7500 |
+| Memory | 4 GB RAM | 16 GB RAM |
+| Graphics | 2GB VRAM | 4GB VRAM |
+| Storage | 10GB available space* | 200GB available space* |
```
erri120, line 14: Where did these values come from?

erri120: These "requirements" don't really make a lot of sense. The only thing we require on Windows is Windows 10 or newer. We don't care what hardware the PC has.

docs/users/SystemRequirements.md
```diff
@@ -0,0 +1,43 @@
...
+<sub>\* Storage space varies by the number of games managed, the size of the managed games and the mods installed.</sub>
+
+## SteamOS + Linux 
+|| Minimum | Recommended |
+| --- | --- | --- |
+| OS | Ubuntu 12.04 LTS | A currently supported Linux distribution |
+| Processor | Intel®️ Core™️ i5-7500 | Intel®️ Core™️ i7-7500 |
+| Memory | 4 GB RAM | 16 GB RAM |
+| Graphics | 2GB VRAM | 4GB VRAM |
+| Storage | 10GB available space* | 200GB available space* |
```
erri120, line 25: This is also just straight up wrong. Ubuntu 12.04 is ancient, and **not** supported. I don't know where you got these values from, but you should just use the original requirements I wrote down.

Pickysaurus: I can update this page to just have vague recommended specs but the app will not run well on a potato so it's important to set expectations.

erri120: Just don't put any specs in there. Instead, you can write down that the App is bottlenecked by disk speed, so an SSD or NVME is recommended. The App is also multithreaded, and will make use of all available cores.

docs/styles.css
```diff
@@ -0,0 +1,9 @@
+/* Extra styles for the Nexus Mods App docs */
+
+.md-tabs {
+    background-color: var(--nexus-background-tertiary);
+}
+
+[dir=ltr] .md-header__title {
+    margin-left: 0
+}
```
erri120, line 9: This should probably be added to https://github.com/Nexus-Mods/NexusMods.MkDocsMaterial.Themes.Next.

docs/users/Uninstall.md
```diff
@@ -0,0 +1,44 @@
...
+## Manual Uninstall (Fallback)
+
+!!! warning "Caution"
+    Manual uninstall should only be used as a fallback if the normal installer is not available. You will need to manually remove any mod files from your game folder(s).
+
+If something goes wrong during uninstall, it may be necessary to remove the app manually. The following steps must be performed manually by the user:
+
+1. Inside the App, ensure that all mods for your loadouts in all games are disabled and these changes have been applied. This will remove any files added to the game folder and get the game back to a "vanilla" state. (When possible, you should delete your loadouts).</li>
+2. Close the App. 
+3. Delete the contents of `%LOCALAPPDATA%\NexusMods.App` on Windows or `$XDG_STATE_HOME/NexusMods.App/Logs` on Linux to remove all downloaded mods and internal databases used by the app.
```
erri120, line 43: On Linux, it's `$XDG_STATE_HOME/NexusMods.App` and `$XDG_DATA_HOME/NexusMods.App`.

erri120 (changes requested):
See comments.

docs/developers/changelog-assets/README.md
erri120, line 1: The `changelog-assets` folder should be inside `docs`, not `docs/developers`.

Pickysaurus: This was not documented anywhere but is easy to fix.

docs/styles.css
Pickysaurus: You're welcome to do so.

docs/developers/index.md
```diff
@@ -0,0 +1,77 @@
...
+
+#### System Package Manager
+
+If you're want to create a package for a platform, see [Contributing](./Contributing.md#for-package-maintainers) for more details.
+
+[![Packaging status](https://repology.org/badge/vertical-allrepos/nexusmods-app.svg)](https://repology.org/project/nexusmods-app/versions)
+
+### macOS
+
+We currently don't publish releases for macOS.
```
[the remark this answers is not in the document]
Pickysaurus: I'll update it later. It's way to complex and developer-y for general users.

docs/developers/changelog-assets/README.md
erri120: The directory is referenced by the changelog and by scripts directly, which is why checks are failing on the PR.

docs/developers/index.md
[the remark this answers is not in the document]
erri120: What about this is developer-y? These are requirements for running the App.

erri120 (changes requested):
See open comments.

docs/users/SystemRequirements.md
halgari: We should either not have the requirements or track down actual values. I'm more in favor of the latter. I think we can remove all of the windows requirements except for Windows 10/11 for now, as that's about all we currently know.

docs/users/SystemRequirements.md
halgari: I have to agree here, saying we support Ubuntu 12 means we actually support it, and I seriously doubt we'll do that. I'm also not sure if we want to tie the minimum requirements to Ubuntu when Steam Deck runs arch. So we may just want to say "we support Linux" or something of that nature

docs/styles.css
halgari: Well no, we want to keep the styling consistent across the projects.

docs/developers/index.md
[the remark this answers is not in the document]
halgari: I'm fine with these staying here for now, we can move it later easily enough

docs/styles.css
halgari: Eh, scratch that, the perms on that repo are borked and need to be updated. I'll work on that sometime in the future

docs/users/SystemRequirements.md
halgari: fixed

docs/users/SystemRequirements.md
halgari: fixed

halgari (approved):

PR #1644, opened 2024-06-18
Fix nxm downloads not opening the App if app is closed

### LINKED ISSUE #515: App UI doesn't open on download with mod manager
Right now just a console briefly shows up and then closes, with no further indication that anything was successful.
Users more likely expect NMA to open and to show the new download being processed.

The current behaviour is likely due to the CLI origin of the app.

Al12rs:
- fixes #515

### Recap:
- Slim process was not spawning new main process due to missing `LoggingSetting` in DI for `ProcessFactory`
- Fixed that, the DL was being skipped due to login check returning false
- `LoginManger` wouldn't have data until DB was initialized
- Had to make the `LoginManger` into a `HostedService` and await in `StartAsync()` on the db to be ready.
- This initialized the UserData from DB, but `CliServer.StartAsync()` started listening before that.
- Changed `CliServer` to require explicit `StartCliServerAsync` call in `Program.cs`
- `CliServer` is only started after all the other `HostedServices` have initialized.
- Removed hack workaround where `CliServer` was waiting for `GameRegistry` to initialize in `StartAsync`.
- Fixed DI for all serviced that used either `PrcessFactory` or `CrossPlatform` that also required `LoggingSettings`

### DIFF
--- a/src/Networking/NexusMods.Networking.NexusWebApi/LoginManager.cs
+++ b/src/Networking/NexusMods.Networking.NexusWebApi/LoginManager.cs
@@ -3,6 +3,7 @@ using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using DynamicData.Binding;
 using JetBrains.Annotations;
+using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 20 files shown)


Al12rs:
Test

halgari (approved):

PR #1660, opened 2024-06-20
Mark tests as Flakey

Al12rs:
Failures are too random and slowing down PRs getting merged as well as neutering CI effectiveness.
Tried looking into both tests, but apart from weird timing shenanigans with UI threads I didn't find possible causes for the failures.


Al12rs:
Want to let CI finish and see if there are more tests that need to be handled

Al12rs:
Seems everything passed on first try

erri120 (approved):

PR #1675, opened 2024-06-24
Implement `OpenDirectory` and enable TopBar button

### LINKED ISSUE #1588: Implement `OpenDirectory` OS-Interop method
Required for the top bar as we now have a menu item to open the logs directory.

erri120:
Resolves #1588.

Re-uses the same code we have for opening URIs by created a `file://` URI. Tested on both Linux and Windows.

It's not the best solution, and I'll probably change this to just use `explorer.exe` and `xdg-open` directly, after doing #163.


Sewer56 (approved):
Works for me.
Got no complaints.

PR #1680, opened 2024-06-24
Loadout Items ADR

halgari:
Pretty simple stuff, just documenting the plans we've had for awhile about loadout items (#1336).

### DIFF
--- a/docs/developers/decisions/backend/0014-loadout-items.md
+++ b/docs/developers/decisions/backend/0014-loadout-items.md
@@ -49,7 +49,7 @@ classDiagram
     
     class ContainerItem {
         +string Name
-        +bool Enabled
+        +marker IsDisabled
     }
     
     class DownloadedMod {
@@ -86,4 +86,34 @@ to find these items mostly becaomes `ContainerItem.FindByParentId(db, loadout)`.
 ### Synchronizer
 
 The Synchronizer will need to be updated to handle the new model, flattening a loadout becomes a matter of finding all `File`
-items that point to the loadout, then walking all those items to make sure each parent is enabled.
+items that point to the loadout, then walking all those items to make sure each parent is enabled. Stated another way, the
+loadout is flatted down by grouping all loadout items by `To` that are not an ancestor of a item that has the `Disabled` marker
+attribute.
+
+Naturally this approach does not handle file conflicts, but it is assumed that these conflicts will be rare in most cases. When these
+conflicts do occur the Synchronzier will pass the options to a `ResolveWinningFile` method that will pick which of the conflicting
+files is the winner. Due to this approach it is not required that loadout items or mod be sorted, instead only the conflicting
+files need to be ordered, in which case the ordering is often trivially detected based on the files and mods in question
+
+### Mod Library
+
+The mod library grid will look in the loadout for any loadout items with the given `DownloadMetadata` and use this to drive
+the "installed" or "not installed" states in the grid. 
+
+### Diagnostics
+
+Diagnostics likely will need to be updated to handle specific types of loadout items, the biggest changing being moving 
+mod referencs to container/downloaded mod items. 
+
+### Stardew Valley Updates
+
+Stardew valley installers will need to be updated to emit a single parent container of the type DownloadedMod that contains
+other containers of "SMAPI Mod" which each contain the files of the mod. In future UI updates we can decide if we want
+to display all of these mods as a list, a tree, or only the top or lower levels of the tree. 
+
+## Rationale
+
+The primary motivation for this change is to move questions of what a specific object is from code-time to runtime. Displaying
+items in a grid or a tree, or projecting parts of the loadout item tree into a list all now become runtime concerns and 
+we are free to project the data in any way required. New file or container types can be added to this process and existing
+grids and lists should work as-is. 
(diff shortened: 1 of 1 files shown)


docs/developers/decisions/backend/0014-loadout-items.md
```diff
@@ -0,0 +1,120 @@
...
+        +ReferenceId DownloadMetadata
+    }
+    
+    class GameFiles {
+         +GameInstallation Installation
+    }
+
+    LoadoutItem <|-- File
+    File <|-- StoredFile
+    File <|-- GeneratedFile
```
Al12rs, line 66: I'm not sure I understood this part:
image

My thought was that we would have an another level between `LoadoutItem` and the Files, something like `FileBucket`. 

The idea being that not all `LoadoutItems` would necessarily have files attached to them. If it is a `FileBucket` then it would have files and it would also have any extra data or functionality that might be needed for Files.

If a particular LoadoutItem is not supposed to have Files, then it would not be a `FileBucket` or whatever name we choose, but it would be something else, like `ModGroup` or `Ini Edit`.

Maybe the arrows were supposed to represent a different kind of relationship with `LoadoutItem` than what expressed with `GameFiles` and `ContainerItem`? If they are meant to be the same `Include<LoadoutItem>` kind of relationship, then that confuses me.

halgari: The arrows represent a "IsA<Parent>" relationship, and that's not really clear in that graph. A StoredFile `isa` File which `isa` LoadoutItem. What you're suggesting is what I called `ContainerItem` which is something that groups files in some way, like a mod, game files container, etc.

docs/developers/decisions/backend/0014-loadout-items.md
```diff
@@ -0,0 +1,120 @@
...
+Stardew valley installers will need to be updated to emit a single parent container of the type DownloadedMod that contains
+other containers of "SMAPI Mod" which each contain the files of the mod. In future UI updates we can decide if we want
+to display all of these mods as a list, a tree, or only the top or lower levels of the tree. 
+
+## Rationale
+
+The primary motivation for this change is to move questions of what a specific object is from code-time to runtime. Displaying
+items in a grid or a tree, or projecting parts of the loadout item tree into a list all now become runtime concerns and 
+we are free to project the data in any way required. New file or container types can be added to this process and existing
+grids and lists should work as-is.
```
erri120, line 120: The section can be expanded to include one of the main aspects of this: optimized for reading. The goal is to easily grab any data in the correct view. The SMAPI diagnostic emitters are a great example, most of them don't care about anything but SMAPI manifest files and SMAPI mods.

erri120 (commented):
I think we should clearly define some models in this ADR. Specifically, what you refer to as `ContainerItem`. This model is going to replace what we currently refer to as a "Mod" and I want us to get it right, as it's going to be the centerpiece for almost everything else.

I'm not married to the term `ContainerItem` yet, but for now, here's a shortlist of things we need to somehow deal with (this is not a complete list):

- Game Files
- "Override"
- Nexus Mods Collections
- SMAPI Mods, where each mod has its own manifest file. One archive might contain three SMAPI Mods, even though they came from the "remote mod" (using this generic term to represent a Nexus Mods mod page, open to change)
- SMAPI config files

One pain point I want to avoid with Loadout Items is figuring out what goes into the beginner-friendly Loadout grid. Sure, we can have some expert view that shows every single Loadout Item, but design wants a simple view that "shows mods", which is why the entry on the left menu says "Mods" and not "Loadout Grid".

As mentioned many times before, the term "Mod" is overloaded, and we need to find something else. I want us to find that term, and create a clear definition with this ADR.

docs/developers/decisions/backend/0014-loadout-items.md
Al12rs: This still seems a bit confusing to me, but maybe the idea is for Files to be LoadoutItems but they would usually have a parent that would be a File container.

I'm not sure we would ever have top level File LoadoutItems, as there would always need to be some context information attached to it, that we wouldn't want to have on all File Items.

I had not grasped from the original idea that LoadoutItem would scale down to File level.
I guess I'm not sure whether this would actually be beneficial as opposed to having a  FileBucket type entity that would represent a collection of Files, with any necessary metadata related to this collection of files.

I can see the symmetry that spawned the idea to just expand it lower, but I'm not sure whether doing so actually has tangible benefits other than having a system fully defined within itself. I found the concept a bit confusing, and I don't see Files being used directly without a dedicated FileBucket type container. 
I will read this through again tomorrow to see if I missed things.

docs/developers/decisions/backend/0014-loadout-items.md
```diff
@@ -0,0 +1,120 @@
...
+
+The Loadout View Model, in this model becomes a list of all ContainerItems, their names, their creation times, etc. The code
+to find these items mostly becaomes `ContainerItem.FindByParentId(db, loadout)`.
+
+### Synchronizer
+
+The Synchronizer will need to be updated to handle the new model, flattening a loadout becomes a matter of finding all `File`
+items that point to the loadout, then walking all those items to make sure each parent is enabled. Stated another way, the
+loadout is flatted down by grouping all loadout items by `To` that are not an ancestor of a item that has the `Disabled` marker
+attribute.
```
Al12rs, line 92: This bottom up approach, while more intuitively easy to parallelize, would require performing a walk up the tree for every single file to check if all parents are enabled.

Would it not be more efficient to walk down the tree of only the branches that are enabled? 
Skipping, in this manner, large chunks of files as top level items exclude them.

PR #1681, opened 2024-06-25
Added: Support for coloured vector icons.

### LINKED ISSUE #1666: Bug: SVG Icons in App Cannot be Recoloured
# Bug Report

## Summary

All icons that are imported via SVG files in the App cannot currently be recoloured.

As part of 
-  https://github.com/Nexus-Mods/NexusMods.App/issues/1551

My intent is to fix this, however, because this is a bit more involved; I'm lifting this out to separate PR/issue.
More technical info will be posted here.

## Steps to reproduce

- Toggle individual left menu items.

Items should change in colour, some of them do not; they happen to be SVGs.

## Other information

We need an improved mechanism for loading vectors.

Previously we loaded vectors only by the path, i.e. 

```csharp
Geometry.Parse(M12.46 17.9912L18.1722 13.5441L19.445 12.5584L12.46 7.12561L5.47498 12.5584L6.74004 13.5441L12.46 17.9912Z)
```

This has the flaw of not loading icons in the correct size, a.k.a.
- https://github.com/Nexus-Mods/NexusMods.App/issues/1551

The cause for this appears to be the lack of a ViewBox.

Many of the icons on Figma contain padding on the edges.
When a ViewBox is missing, this padding is not present, thus the icons are displayed larger than the design.

Consider the SVG version of the above icon

```xml
<svg width="25" height="25" viewBox="0 0 25 25" fill="none" xmlns="http://www.w3.org/2000/svg">
    <path fill-rule="evenodd" clip-rule="evenodd" d="M12.46 17.9912L18.1722 13.5441L19.445 12.5584L12.46 7.12561L5.47498 12.5584L6.74004 13.5441L12.46 17.9912Z"/>
</svg>
```

Note the `viewBox="0 0 25 25"` field here. We need to make a control which is `25x25`, and THEN render the path inside that control. This will in turn include the padding and the sizes should match.

To get this to working I believe we need a variation of `Avalonia.Controls.PathIcon`, or at least something that derives from `DrawingImage` and `IImage`.

The easiest approach would probably be making a variation of [IconImage.cs](https://github.com/Projektanker/Icons.Avalonia/blob/main/src/Projektanker.Icons.Avalonia/IconImage.cs) from Projectanker.

Sewer56:
Fixes #1666 

This adds support for coloured icons that can have a ViewBox and thus match design.

Not much more needs to be said.

I also wound up doing some additional research for more efficiently loading vectors in the future; including some alternative file formats, how to integrate, etc. but I decided to relegate carrying out that need into my own personal time; as for the purposes of the project, what we have is sufficiently good enough.

-----

Misc Note: Doing this also exposed to me just how inefficient our use of icons in some regards are.
Give for example Projectanker icons.

These:

1. Load SVG from embedded resource
2. Parse it using uncompiled regex
3. Create StreamGeometry etc. from the path from regex.

And we repeat this for every icon we load, for example, every individual tree row on AdvancedInstaller.

In addition, changing a colour on a Projectanker icon repeats the entire same process, from scratch. oof, when technically one could just change the brush.

In any case, my implementation reuses the data when possible, so the actual icon controls are cheap to create.


src/NexusMods.Icons/SimpleVector/Control/SimpleVectorIcon.cs
```diff
@@ -0,0 +1,39 @@
+using Avalonia;
+using Avalonia.Controls.Documents;
+using Avalonia.Media;
+
+namespace NexusMods.Icons.SimpleVector.Control;
+
+/// <summary>
+///     A wrapper for <see cref="Avalonia.Controls.Image"/> designed to work with <see cref="SimpleVectorIconImage"/>.
+/// </summary>
+public class SimpleVectorIcon : Avalonia.Controls.Image
```
erri120, line 10: Why do we need this wrapper?

Sewer56: It adds a Foreground property, so the control works with our existing styles.

erri120 (commented):
I'm not sure why we're doing DIY SVG rendering, the Avalonia `Svg` control does all of this already. I thought we'd just update our usage of the control with the viewbox and color.

Sewer56:
> I'm not sure why we're doing DIY SVG rendering, the Avalonia `Svg` control does all of this already. I thought we'd just update our usage of the control with the viewbox and color.

I tried variants of

```csharp
else if (change.Property == ForegroundProperty && Content is Avalonia.Svg.Skia.Svg svg)
{
    if (change.NewValue is not SolidColorBrush brush) return;
    var css = $".Black {{ fill: {brush.ToHexColour()}; }}";
    // ToHexColour() I made myself as an extension
    Avalonia.Svg.Skia.Svg.SetCss(svg, css);
}
```

Unfortunately I found that the CSS styling that's part of the Avalonia SVG package applied only to styling classes.

i.e. For every SVG we download, we'd need to manually add a class to each path so we can then style it.
And likewise, strip properties like `fill` for each path.

Then, at runtime, to change the colour once we got there, we'd convert the brush into hex, make a string out of it. Then the Svg library would pick up that string, have to do a class lookup, create a new brush, etc. It felt very silly and inefficient at that point. If we ever needed to animate with colour transitions; that would also be very far from ideal.

src/NexusMods.Icons/IconValue.cs
```diff
@@ -74,3 +82,5 @@ public record struct AvaloniaSvg(string? Path);
 [PublicAPI]
 public record struct AvaloniaPathIcon(Geometry? Geometry);
 
+[PublicAPI]
+public record struct SimpleVectorIcon(SimpleVectorIconImage? Image);
```
erri120, line 86: Can be made non-null. The others are nullable because they can be used from XAML.

```suggestion
public record struct SimpleVectorIcon(SimpleVectorIconImage Image);
```

src/NexusMods.Icons/UnifiedIcon.cs
```diff
@@ -122,7 +122,8 @@ private void UpdateControl(IconValue? value)
                 [MaxHeightProperty] = this[MaxHeightProperty],
                 [MaxWidthProperty] = this[MaxWidthProperty],
                 [ForegroundProperty] = this[ForegroundProperty],
-            }
+            },
+            f5: simpleVectorIcon => new NexusMods.Icons.SimpleVector.Control.SimpleVectorIcon(simpleVectorIcon.Image!.Clone())
```
erri120, line 126:
```suggestion
            f5: simpleVectorIcon => new NexusMods.Icons.SimpleVector.Control.SimpleVectorIcon(simpleVectorIcon.Image.Clone())
```

src/NexusMods.Icons/SimpleVector/SimpleVectorIconImage.cs
```diff
@@ -0,0 +1,190 @@
...
+    public IPen? Pen
+    {
+        get => GetValue(PenProperty);
+        set => SetValue(PenProperty, value);
+    }
+
+    /// <summary>
+    /// Gets the size of the icon.
+    /// </summary>
+    [PublicAPI]
```
erri120, line 123:
```suggestion
```

src/NexusMods.Icons/SimpleVector/SimpleVectorIconImage.cs
```diff
@@ -0,0 +1,190 @@
...
+    /// <param name="viewBox">The view box of the SVG icon.</param>
+    public SimpleVectorIconImage(string pathData, Rect viewBox)
+        : this(pathData, viewBox, new SolidColorBrush(0xFFFFFFFF))
+    {
+    }
+    
+    /// <summary>
+    /// Gets the view box of the SVG icon.
+    /// </summary>
+    [PublicAPI]
```
erri120, line 99:
```suggestion
```

src/NexusMods.Icons/SimpleVector/SimpleVectorIconImage.cs
```diff
@@ -0,0 +1,190 @@
+using Avalonia;
+using Avalonia.Media;
+using JetBrains.Annotations;
+
+namespace NexusMods.Icons.SimpleVector;
+
+/// <summary>
+/// Represents a simple SVG icon image that can be used in Avalonia applications.
+/// </summary>
+public class SimpleVectorIconImage : DrawingImage, IImage
```
erri120, line 10:
```suggestion
[PublicAPI]
public class SimpleVectorIconImage : DrawingImage, IImage
```

erri120 (changes requested):

Al12rs (approved):
No extra comments from me.
For development efficiency, it would be nice if we didn't have to cherry pick out parts of the SVG to put in the constructor, and just be able to paste the entire thing. But it is probably not worth the time investment to implement such a thing to save a couple of minutes on every icon add.

PR #1683, opened 2024-06-25
Simpler NXM link handling

### LINKED ISSUE #1677: Simplify protocol handler
Currently, we register NXM links to open `NexusMods.App --protocol-invoke %u`. This has major downsides when it comes to Desktop Entries on Linux, which are files that tell the desktop environment about the program and how to handle protocols.

With this separate command, we need two desktop entry files, one for launching the application itself using `NexusMods.App` and one for dealing with NXM links using `NexusMods.App --protocol-invoke %u`.

It'd be much easier if we only have one way: `NexusMods.App %u`. With this approach, if we have a NXM link, then it will be the first argument. If we don't have a link, we get no argument and launch the UI normally.

erri120:
I highjacked the `CommandLineConfigurator` to handle the case where we get a single URI argument that we can handle.

Resolves #1677.

### DIFF
--- a/src/NexusMods.SingleProcess/CommandLineConfigurator.cs
+++ b/src/NexusMods.SingleProcess/CommandLineConfigurator.cs
@@ -4,6 +4,8 @@ using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
 using System.Reflection;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using NexusMods.CLI.Types;
 using NexusMods.ProxyConsole.Abstractions;
 using NexusMods.ProxyConsole.Abstractions.VerbDefinitions;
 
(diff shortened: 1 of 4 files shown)


erri120:
This'll probably break some tests

Al12rs (changes requested):
Didn't test yet, did you already test on windows?

src/NexusMods.SingleProcess/CommandLineConfigurator.cs
```diff
@@ -91,6 +97,29 @@ private Option MakeOption<T>(OptionDefinition optionDefinition)
         return option;
     }
 
+    private async Task<bool> RunLink(string[] args, CancellationToken cancellationToken)
```
Al12rs, line 100: Add comment explaining why this is here and why it couldn't be handled like other command line options.
This will prevent us from invertedly reverting this change due to not knowing the cause.
Or allow us to revert in if we happen to find a better solution to the issue, since cause is readily available.

Al12rs (changes requested):
Sorry, comment was not submitted

src/NexusMods.SingleProcess/CommandLineConfigurator.cs
Al12rs: Also add this is only used on Linux

Al12rs (approved):

PR #1693, opened 2024-06-27
Disable Unsupported Games

halgari:
This may be a bit contraversial, and I appoligize to those who worked on the code I deleted here. 

This removes support for all games except SDV and Cyberpunk. This will allow us to focus on supporting these two games and add in features to the system as need. As an example nothing we currently are building needs `GeneratedFile`, we may need it for BG3 and certainly for Skyrim but not for SDV or Cyberpunk.

Once this PR is merged I'll tag the commit before this and we can save the existing code for when we return to these games.


Al12rs (changes requested):
Cyberpunk uses FOMODs. It shouldn't be removed.

src/NexusMods.App/Services.cs
```diff
@@ -107,12 +103,7 @@ private static IServiceCollection AddSupportedGames(this IServiceCollection serv
     {
         if (experimentalSettings is { EnableAllGames: true })
         {
-            Games.BethesdaGameStudios.Services.AddBethesdaGameStudios(services);
             Games.RedEngine.Services.AddRedEngineGames(services);
-            Games.DarkestDungeon.Services.AddDarkestDungeon(services);
-            Games.BladeAndSorcery.Services.AddBladeAndSorcery(services);
-            Games.Sifu.Services.AddSifu(services);
-            Games.MountAndBlade2Bannerlord.ServicesExtensions.AddMountAndBladeBannerlord(services);
         }
```
erri120, line 107: The setting `EnableAllGames` can be removed as well.

erri120 (changes requested):
I agree with @Al12rs that the FOMOD projects should stay. The FOMOD projects aren't really affected when it comes to DB changes, since they abstract almost everything away, and only interface using the `IModInstaller` implementation.

Aside from that, the FOMOD projects are worth keeping around for the UI alone. Even if we won't support Bethesda games for a while, the Guided Installer was designed to work for any kind of guided installation, not just FOMODs.

src/NexusMods.App/Services.cs
halgari: We don't officially have cyberpunk support in yet, and we have yet to find a person to drive that work this sprint, so I'll keep it around for now.

halgari: I'll get FOMODs added back in

Pickysaurus:
My question here is what happens if someone is on 0.5.2 and decides to manage an unsupported game,  then they update to 0.6.x and that game is gone?

erri120:
> My question here is what happens if someone is on 0.5.2 and decides to manage an unsupported game, then they update to 0.6.x and that game is gone?

All data will be reset between 0.5.2 and 0.6. There is no upgrade path except uninstalling 0.5.2 and installing 0.6.

Al12rs (approved):

Sewer56:
I do have one comment to ask before merging, gimme a minute, I'm replying elsewhere.

Sewer56:
There was just one question/comment I wanted to ask, but I didn't want to ask in super early morning/middle of night. 😅 

Would it not be easier to keep the code around but instead disable it by not registering it during DI?

My concern moreso is, as we start refactoring the codebase, for example, renaming things; then keeping the code around would make it so existing items also get updated in line, even if unused. If something is overhauled completely, such that the existing code is incompatible, then it could be removed from codebase if not yet-supported. (Unless this already happened and this is the response)

I think that way at least re-integrating the removed games would be easier & faster. Otherwise if the code is left dead for too long, things may be named or structured completely different and it would be very painful to rewrite, as you'd need to start making large mental maps of old name -> new name etc.

halgari:
I think this will certainly happen (bitrot of a sort), but that's the point here, to free up our dev time for a bit so we can get these new features in (Loadout Items, Apply/Ingest rework) without having to update code that we aren't even testing properly

PR #1704, opened 2024-07-01
Move downloads out of Temp and clean up after cancel and complete

### LINKED ISSUE #1663: Closing App cancels all ongoing downloads
# Bug Report

## Summary
Closing the App while a download is in `Downloading` state (not paused) makes it vanish. 
Download is no longer visible after starting the app again (likely cancelled). 

## Steps to reproduce
- Start app, 
- start a long download,
- close app while download is running
- open app
- download is missing

- Download should appear again on app restart
- Download should either be paused or be automatically restarted
- Download should resume from where it was when app was closed.

Al12rs:
- fixes #1663 

Ongoing download files are now stored next to DataModel folder, in Downloads/Ongoing.
Each download task will have a separate folder (guid name), which simplifies cleanup on download cancellation and download completion. 

The base folder for the Downloads is determined by a setting, but it can be overruled through an internal method for testing purposes (so that each test has a different download base folder).

### DIFF
--- a/src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
+++ b/src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
@@ -1,8 +1,8 @@
 using System.Collections.ObjectModel;
+using System.ComponentModel;
 using DynamicData;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
-using NexusMods.Abstractions.MnemonicDB.Attributes.Extensions;
 using NexusMods.Abstractions.NexusWebApi.Types;
 using NexusMods.MnemonicDB.Abstractions;
 using NexusMods.Networking.Downloaders.Interfaces;
@@ -14,8 +14,7 @@ using DynamicData.Kernel;
 using Microsoft.Extensions.Hosting;
 using NexusMods.Abstractions.Activities;
 using NexusMods.Abstractions.IO;
-using NexusMods.MnemonicDB.Abstractions.DatomIterators;
-using NexusMods.MnemonicDB.Abstractions.Query;
+using NexusMods.Abstractions.Settings;
 using NexusMods.Paths;
 
 namespace NexusMods.Networking.Downloaders;
@@ -25,6 +24,9 @@ public class DownloadService : IDownloadService, IDisposable, IHostedService
 {
     /// <inheritdoc />
     public ReadOnlyObservableCollection<IDownloadTask> Downloads => _downloadsCollection;
+    
+    /// <inheritdoc />
+    public AbsolutePath OngoingDownloadsDirectory => _downloadDirectory;
     private ReadOnlyObservableCollection<IDownloadTask> _downloadsCollection = ReadOnlyObservableCollection<IDownloadTask>.Empty;
 
     private readonly SourceCache<IDownloadTask, EntityId> _downloads = new(t => t.PersistentState.Id);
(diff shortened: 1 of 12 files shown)


src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
```diff
@@ -209,8 +223,17 @@ public async Task StopAsync(CancellationToken cancellationToken)
...
     }
+    
+    /// <summary>
+    /// Set a custom downloadDirectory, for tests only.
+    /// Directory should already exist.
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    internal void SetDownloadDirectory(AbsolutePath path)
+    {
+        _downloadDirectory = path;
+    }
```
erri120, line 238: Why is this needed?

src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
```diff
@@ -25,6 +24,9 @@ public class DownloadService : IDownloadService, IDisposable, IHostedService
 {
     /// <inheritdoc />
     public ReadOnlyObservableCollection<IDownloadTask> Downloads => _downloadsCollection;
+    
+    /// <inheritdoc />
+    public AbsolutePath OngoingDownloadsDirectory => _downloadDirectory;
```
erri120, line 29: No need to have a field for this.

```suggestion
    public AbsolutePath OngoingDownloadsDirectory { get; private set; }
```

src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
```diff
@@ -209,8 +223,17 @@ public async Task StopAsync(CancellationToken cancellationToken)
             // TODO(Al12rs): should Suspend() instead, but only after moving ongoing dl files outside Temp folder,
             // that is otherwise cleaned up on application close, causing exceptions due to file in use,
             // on top of discarding all the progress.
-            .Select(dl => dl.Cancel());
+            .Select(dl => dl.Suspend());
```
erri120, line 226: Is the comment above still relevant?

erri120 (changes requested):

src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
Al12rs: For tests, to allow each test to have a separate download location. 
This solution sucks. It would be much better if each test had its own `DownloadService` and entire service provider since we are already there.

src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
Al12rs: No, I should remove it

Al12rs:
Need to look into the test not passing on windows

src/Networking/NexusMods.Networking.Downloaders/DownloadService.cs
erri120: `OverrideForTests` doesn't fit this use-case?

Al12rs: It would be the same for methods in the same test class, since singletons injected in the constructors are shared, from what I understand.

erri120: Do these tests run in parallel? If not, is there an issue with using the default location?

Al12rs: I don't know tbh, they way we set it up with a separate guid named folder for each task makes it so that each task is pretty much isolated, but if we ever add logic to clean up orphaned folders or something like that then this could break down.

Al12rs: Ideally these tests would run all in parallel and they would each have a separate `DownloadService`, each with a different `DownloadSettings` override.

erri120: Not with how we currently run tests.

erri120 (approved):

Al12rs:
I will change the tests to use a Setting override, since it should actually be better for running tests concurrently

erri120 (approved):

PR #1715, opened 2024-07-03
User docs 0.6

Pickysaurus:
Docs for the next update


docs/users/games/StardewValley.md
```diff
@@ -1,2 +1,44 @@
...
+- SMAPI is recommended when no mods are installed yet.
+- The installed SMAPI version is too old to work with the game version.
+- The game version is too old to work with the installed SMAPI version.
+- An installed mod's dependencies are installed, but they are outdated.
+- An installed mod has missing dependencies.
+- An installed mod's dependencies are installed, but they are disabled.
+- A mod that requires SMAPI is installed but SMAPI is missing.
+- A mod that requires SMAPI is installed but SMAPI is disabled.
+- A mod is marked as obsolete or broken (according to SMAPI).
+- A mod is overwriting the game files (XNB mod).
```
erri120, line 25: We plan on generating documentation for all diagnostics based on the code definitions in the future. That would keep the docs up-to-date with the code.

For now, I think it's enough to say that we have diagnostics, and maybe a list of some examples instead of a trying to put a full list in the docs. This list is already outdated. See [the definitions](https://github.com/Nexus-Mods/NexusMods.App/blob/9712e30cc61f61ba69c881a9dd5b8586dc83a456/src/Games/NexusMods.Games.StardewValley/Diagnostics.cs) for a complete list.

Pickysaurus: Without this list the feature set looks pretty sparse so I'm fine updating this by hand for now.

docs/users/images/0.5.2/HealthCheckExample.png
erri120, line 1: Is this a screenshot from Figma?

Pickysaurus: Yes, it was the only way to get all three message types at once. The alt text on the article it's used in says it's a concept/design image.

erri120 (approved):

PR #1725, opened 2024-07-04
Library and Downloads

erri120:
Related to #1336.

This PR adds `ILibraryService` which is the new one-shop-stop for downloads, local files, and other things.

Eventually, we're going to completely remove the `IFileHashCache`, `IFileOriginRegistry` and change what `IFileStore` is supposed to be. The `IFileStore` is going to do one thing: take in a Hash, and return a Stream.


src/Abstractions/NexusMods.Abstractions.Downloads/IDownloadActivity.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// <summary>
+    /// Gets the downloader that handles this activity.
+    /// </summary>
+    IDownloader Downloader { get; }
+
+    /// <summary>
+    /// Gets the title of the download.
+    /// </summary>
+    string Title { get; }
```
Al12rs, line 28: Should this go in the persistent state?

erri120: Yes, I'll update the PR.

src/Abstractions/NexusMods.Abstractions.Downloads/ADownloadActivity.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// <inheritdoc/>
+    [Reactive] public Percent Progress { get; set; }
+
+    /// <inheritdoc/>
+    [Reactive] public Bandwidth Bandwidth { get; set; }
+
+    /// <summary>
+    /// Gets the current status.
+    /// </summary>
+    protected virtual PersistedDownloadStatus GetStatus() => _status;
```
Al12rs, line 75: Does this even return the latest value?

erri120: Yes.

erri120: Unless you mean the latest value from the DB. That's not really a concern, anything that mucks with the state outside this activity would be considered undefined behavior.

Al12rs: So this would be the sole and only place with where Status should ever be edited? Specifically through `DownloadActivity.Status` public property?

erri120: Yup

Al12rs: I guess this isn't the intuitive place I would use to update the status. 
Since this interface is public facing (e.g. we will end up using these in the UI) I don't think it should be possible to set the Status through the public interface at least.  

I would have this be implemented by observing the DB and not even allow setting it.


I would assume IDownloaders would be the ones in charge of updating this, inside the Pause(), Cancel(), Start() calls

erri120: Specifically, it would only be updated by the Downloader.

erri120: I can remove the setters from `IDownloadActivity`, you'd still have the setters in `ADownloadActivity` and the downloaders would have to cast to that.

src/Abstractions/NexusMods.Abstractions.Downloads/ADownloadActivity.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// <inheritdoc/>
+    public string Title { get; }
+
+    /// <inheritdoc/>
+    public AbsolutePath DownloadPath { get; }
+
+    /// <summary>
+    /// Constructor.
+    /// </summary>
+    protected ADownloadActivity(
```
Al12rs, line 33: Right, question I forgot to ask, who is supposed to create the `DownloadActivity` (e.g. NxmDownloadActivity())?
- Would it be the initiator of the download like the nxmHandler? 
- Or the Library with something like CreateNxmDL?
- Or the Downloader with something like NXMDownloader.CreateDL()?

erri120: The NXM protocol handler will create a `NXMDownloadActivity` and enqueue it using `LibraryService.EnqueueDownload`.

src/Abstractions/NexusMods.Abstractions.Downloads/ADownloadActivity.cs
Al12rs: Well the functionality doesn't strictly need to be on the activity, you just need the entity ID and anyone (the Downloader) can do it.

And the Getter would need to return the latest DB value, not the internal field value.

src/Abstractions/NexusMods.Abstractions.Downloads/ADownloadActivity.cs
Al12rs: I guess you would usually leave the creation of things like this to a dedicated service (be it the Downloader or the Library or something else), which would allow more decupling in case in the future you need to introduce further variations (subclasses).

src/NexusMods.Library/LibraryService.cs
```diff
@@ -0,0 +1,199 @@
...
+
+    /// <inheritdoc/>
+    public void EnqueueDownload(IDownloadActivity downloadActivity, bool addPaused = false)
+    {
+        _logger.LogInformation("Adding download `{Title}` to the library", downloadActivity.Title);
+
+        _downloadActivitySourceCache.AddOrUpdate(downloadActivity);
+
+        if (addPaused) return;
+        downloadActivity.Downloader.Start(downloadActivity);
```
Al12rs, line 71: I would add the utility methods on the activity tbh to just be 
```csharp
downloadActivity.Start();
```

src/Abstractions/NexusMods.Abstractions.Library/ILibraryService.cs
```diff
@@ -0,0 +1,35 @@
...
+/// <summary>
+/// Represents the library.
+/// </summary>
+[PublicAPI]
+public interface ILibraryService
+{
+    /// <summary>
+    /// Gets an observable collection containing all download activities.
+    /// </summary>
+    ReadOnlyObservableCollection<IDownloadActivity> DownloadActivities { get; }
```
Al12rs, line 18: Does this contain only the `Downloading` and `Paused` Activities? Or can this also contain `Cancelled` and `Completed` activities?

Currently the `DownloadService` exposes a similar thing but it contains all the downloads, including Completed and cancelled ones.

Comment should clarify what this actually contains.

src/NexusMods.Library/LibraryService.cs
```diff
@@ -0,0 +1,199 @@
...
+            .DisposeWith(_compositeDisposable);
+
+        _downloadActivitySourceCache
+            .Connect()
+            .WhenPropertyChanged(downloadActivity => downloadActivity.Status)
+            .Where(propertyValue => propertyValue.Value == PersistedDownloadStatus.Completed)
+            .Select(propertyValue => propertyValue.Sender)
+            .SelectMany(AddCompletedDownloadAsync)
+            .Subscribe()
+            .DisposeWith(_compositeDisposable);
```
Al12rs, line 60: I don't think there is anything here to handle cancelled activities like we do for Completed ones.
What is the plan for Cancelled downloads? I'm guessing they should no longer have an activity associated with them either.

src/Abstractions/NexusMods.Abstractions.Downloads/PersistedDownloadStatus.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// The download is cancelled.
+    /// </summary>
+    Cancelled = 2,
+
+    /// <summary>
+    /// The download is completed.
+    /// </summary>
+    Completed = 3,
+}
```
Al12rs, line 30: I don't think we are handling Download Failures well enough in the current system, and we are not doing anything to improve that in the new system.

What do you think would be the best way to handle a generic failure during download or during download start?
Should we just have a `PersistedDownloadStatus.Failed` state here?

src/NexusMods.Library/LibraryService.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            LibraryFile = libraryFile,
+            OriginalPath = absolutePath.GetFullPath(),
+        };
+
+        var result = await tx.Commit();
+        return result.Remap(localFile);
+    }
+
+    private async Task<LibraryFile.New> AddLibraryFile(
```
Al12rs, line 123:
```suggestion
    private async Task<LibraryFile.New> AddLibraryFileAsync(
```
For naming consistency

src/NexusMods.Library/LibraryService.cs
```diff
@@ -0,0 +1,199 @@
...
+        ITransaction tx,
+        EntityId entityId,
+        AbsolutePath filePath,
+        CancellationToken cancellationToken)
+    {
+        var (hash, isArchive) = await AnalyzeFileAsync(filePath, cancellationToken: cancellationToken);
```
Al12rs, line 129: This can take quite a while for larger files being analyzed for the first time, the old system had an `Analyzing` status on the `PersistedDownloadStatus` enum.

Do you still think it should not be in the status? If not, what do you think would be the best approach to keep track of this in the UI?

E.g. we want to show "Analizing" text and a spinner on the download entry in the UI while Analyzing work is being done.

erri120: The download will be marked as completed once it has finished downloading. Hashing and extracting the archive are two different activities that may be linked to the "parent" download activity. I really want to separate these concepts and make the process work more like a pipeline.

The only thing the download activity is concerned with is downloading.

src/Abstractions/NexusMods.Abstractions.Library/ILibraryService.cs
erri120: Still trying to figure out what we'd want to use this for. It'll be clearer once we start using this.

src/NexusMods.Library/LibraryService.cs
erri120: Still trying to figure out what we'd want to use this for. It'll be clearer once we start using this.

src/Abstractions/NexusMods.Abstractions.Library/ILibraryService.cs
Al12rs: I would guess we would use this in the UI of the InProgress page and later on in the LibraryPage as well, to show download progress on either new LibraryItems or on existing items for which we are redownloading missing files.

src/Abstractions/NexusMods.Abstractions.Downloads/ADownloadActivity.cs
```diff
@@ -0,0 +1,77 @@
...
+        PersistedStateId = persistedState.PersistedDownloadStateId;
+        Status = persistedState.Status;
+
+        Downloader = downloader;
+        Title = persistedState.Title;
+        Downloader = downloader;
+    }
+
+    /// <inheritdoc/>
+    public PersistedDownloadStatus Status { get; private set; }
```
Al12rs, line 45: This still assumes all status changes go through here. 
Currently the pattern we use is to instead listen to the DB for changes since db can be changed from anywhere in the code.

There is no good way to define this as the sole authority for Status, especially since the actual agents of the changes would be the IDownloaders.

I wouldn't mind having better separation of concerns and more controlled write access to DB data, but I don't think we have any way to express this or enforce it, so it would be easy for this to be forgotten and bugs to come from it.

src/Abstractions/NexusMods.Abstractions.Downloads/ADownloadActivity.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// <remarks>
+    /// This updates <see cref="PersistedDownloadState.Status"/> for <see cref="PersistedStateId"/>.
+    /// </remarks>
+    public void SetStatus(ITransaction tx, PersistedDownloadStatus status)
+    {
+        Status = status;
+        tx.Add(PersistedStateId, PersistedDownloadState.Status, status);
+
+        this.RaisePropertyChanged(nameof(Status));
+    }
```
Al12rs, line 74: Caller needs to do the commit?

Al12rs (approved):
I'm ok to merge this in favour of continuing work in future PRs, could you please mark down somewhere the following open issues?

#### LibraryService & Downloads open issues
- [ ] Figure out what should remain an activity and should not (completed, failed, cancelled downloads)
- [ ] Handle Cancelled downloads in LibraryService
- [ ] Handle download failures in LibraryService
- [ ] Add separate Activity for Analyzing
- [ ] Add `downloadActivity.Start()` convenience methods 
- [ ] Change `DownloadActivity` to return latest Db version of Status and notify of updates
	- [ ] Or find a way to make it clear and explicit that `DownloadActivity` should be the sole authority for setting Status

PR #1746, opened 2024-07-10
[Loadouts.Synchronizers][Actions] change action order

ZeeWanderer:
currently `DeleteFromDisk` action was ordered after `ExtractToDisk` action which in combination of the 2 resulted in a deleted file instead of a replaced one, it looks like actions need reordering but I am not too sure about the side effects so a draft. Also, bitshift, in C# should be endianness independent.


Al12rs:
From what I know this would seem to make sense, but we can wait for @halgari to check

halgari:
Agreed, this is an issue, but I think we need to take this further. Backup/Ingest should come before delete, the general order should be:

* Backup
* Ingest
* Reify
* Delete
* Extract

The idea being that backup needs the files to exist when it performs the backup, ingest and reify don't technically need the file to exist but it would be helpful in the case where the data needs to be referenced somehow

halgari (approved):

PR #1749, opened 2024-07-11
Add jobs

erri120:
Related to #1336.

I'm gonna stop working on this before this gets even more out of hand. Missing from this PR:

- Progress
- Groups and Child jobs
- More tests
- Global list of all running jobs
- Persistend jobs


Sewer56:
Quick question for @halgari (when you're around)

Why was the API originally renamed from Job -> Activity? Since we're going back to 'Job' here.

Am just curious.

src/Abstractions/NexusMods.Abstractions.Jobs/IJob.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// <param name="cancellationToken">
+    /// Optional cancellation token to stop the waiting. Note that this only
+    /// cancels the proxy task, it does not cancel the job.
+    /// </param>
+    Task<JobResult> WaitToFinishAsync(CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Starts the job.
+    /// </summary>
+    Task StartAsync(CancellationToken cancellationToken = default);
```
Al12rs, line 57: Would this return on job finish or when the operation has finished starting up?

src/Abstractions/NexusMods.Abstractions.Jobs/IJob.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// <summary>
+    /// Pauses the job.
+    /// </summary>
+    Task PauseAsync(CancellationToken cancellationToken = default);
```
Al12rs, line 62: What if a job does not support pausing?

src/Abstractions/NexusMods.Abstractions.Jobs/IJob.cs
erri120: This returns when the job has been started. If the job has some startup logic that requires async, then this might not return immediately.

Al12rs: Could you add this in the comment for clarification?

src/Abstractions/NexusMods.Abstractions.Jobs/AJobWorker.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    protected JobResult FailJob(string message)
+    {
+        throw new NotImplementedException();
+    }
+
+    protected JobResult CompleteJob<TData>(TData data)
+    {
+        throw new NotImplementedException();
+    }
```
Al12rs, line 48: Shouldn't these `NotImplementedException()` method just be marked as abstract instead?

erri120: Not really, I just haven't implemented them yet.

src/Abstractions/NexusMods.Abstractions.Jobs/JobStatusExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// Returns whether a transition from one status to another is valid.
+    /// </summary>
+    public static bool CanTransition(this JobStatus from, JobStatus to)
+    {
+        return (from, to) switch
+        {
+            (JobStatus.None, _) => true,
+
+            (JobStatus.Created, JobStatus.Running) => true,
+            (JobStatus.Created, _) => false,
```
Al12rs, line 56: You can't cancel a job without starting it then?

src/Abstractions/NexusMods.Abstractions.Jobs/JobStatusExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            (JobStatus.Running, JobStatus.Paused) => true,
+            (JobStatus.Running, JobStatus.Completed) => true,
+            (JobStatus.Running, JobStatus.Cancelled) => true,
+            (JobStatus.Running, JobStatus.Failed) => true,
+            (JobStatus.Running, _) => false,
+
+            (JobStatus.Paused, JobStatus.Running) => true,
+            (JobStatus.Paused, JobStatus.Cancelled) => true,
+            (JobStatus.Paused, _) => false,
```
Al12rs, line 66: Can't a paused job fail? Missing data on application startup, expired token etc?

erri120: A paused job is doing nothing. Only when you un-pause and the job goes back to running would it fail.

src/Abstractions/NexusMods.Abstractions.Jobs/ProgressRate.cs
```diff
@@ -0,0 +1,51 @@
...
+    public ProgressRate(double value, IProgressRateFormatter formatter)
+    {
+        Value = value;
+        Formatter = formatter;
+    }
+
+    public ProgressRate Add(double increase)
+    {
+        return new ProgressRate(Value + increase, Formatter!);
+    }
```
Al12rs, line 34: Mark as pure or non discard of results?

src/Abstractions/NexusMods.Abstractions.Jobs/AJob.cs
```diff
@@ -0,0 +1,103 @@
...
+        Worker = worker;
+
+        _subjectStatus = new Subject<JobStatus>();
+        _connectableObservableStatus = _subjectStatus.Replay(bufferSize: 1);
+        _disposable.Add(_connectableObservableStatus.Connect());
+    }
+
+    internal void SetStatus(JobStatus value)
+    {
+        if (Status.CanTransition(value)) _subjectStatus.OnNext(value);
```
Al12rs, line 49: Isn't setting the status

Al12rs (commented):
As we discussed, we are pretty happy with the Job interface itself, but are not happy yet with the JobWorker interface, as it requires a transient worker for each Job. 

Plan is to change the ExecuteAsync() to accept a Job as well, so that it may be implemented as a singleton service handling multiple jobs. 

Ideally a service implementation would be able to implement the interface for multiple JobWorkers if needed.

src/Abstractions/NexusMods.Abstractions.Jobs/AJobWorker.cs
```diff
@@ -1,69 +1,181 @@
...
+            else
+            {
+                job.SetStatus(JobStatus.Paused);
+            }
+
+            job.Task = null;
+            return ValueTask.CompletedTask;
+        }
+
+        task.Start();
```
Al12rs, line 106: From what I understand `Task.Start()` should be problematic here, since the task would count as already being in a started state

erri120: Why would it already be in the started state?

Al12rs: For this I think:
`task = ExecuteAsyncInner(job, job.CancellationTokenSource.Token);`
the task should already start executing until the first await is encountered.
Calling start again should cause invalidoperation exception from the docs

erri120: True, fixed.

src/Abstractions/NexusMods.Abstractions.Jobs/AJobWorker.cs
```diff
@@ -0,0 +1,176 @@
...
+
+        var tsc = new TaskCompletionSource();
+        using var disposable = job.ObservableStatus.Subscribe(status =>
+        {
+            if (status == JobStatus.Paused) tsc.SetResult();
+        });
+
+        job.IsRequestingPause = true;
+        await job.CancellationTokenSource.CancelAsync();
+        await tsc.Task.WaitAsync(cancellationToken: cancellationToken);
```
Al12rs, line 141: This code template to await the change of the status is used multiple times, I would suggest the following to simplify it:

```suggestion
        var statusUpdate = job.ObservableStatus
            .FirstAsync(status => status == JobStatus.Paused)
            .ToTask(cancellationToken);

        job.IsRequestingPause = true;
        await job.CancellationTokenSource.CancelAsync();
        await statusUpdate;
```

src/NexusMods.Library/AddLibraryFileJobWorker.cs
```diff
@@ -0,0 +1,218 @@
...
+        {
+            job.HashJobResult = await HashAsync(job.FilePath, cancellationToken);
+        }
+
+        cancellationToken.ThrowIfCancellationRequested();
+        if (!job.EntityId.HasValue)
+        {
+            job.EntityId = job.Transaction.TempId();
+        }
+
```
Al12rs, line 62: I don't really like this list of ifs to keep track of current resume state.
Have you considered using a switch and an internal status value to keep track of the step we are on?
Similar to async tasks state machines?

erri120: Can you provide an example of what you might want to see?

Al12rs: We are mixing data validation with resume logic, since there is a lot of overlap, but it does mean that attempting to resume the process after pausing would go through every step, requiring the keeping of all intermediate steps data, instead of just the data required for the next step. 

A switch would allow jumping directly to the last completed state.
This is actually made worse due to C# not doing cascading fall through of positive switch cases like other languages do.
```csharp
protected override async Task<JobResult> ExecuteAsync(AddLibraryFileJob job, CancellationToken cancellationToken)
{
    var absolutePath = job.FilePath;

    switch (job.LastCompletedStep)
    {
        case AddLibraryFileJobStep.Start:
            goto CheckIfArchive;
        case AddLibraryFileJobStep.CheckIfArchive:
            goto Hash;
        case AddLibraryFileJobStep.Hash:
            goto CreateLibraryFile;
        case AddLibraryFileJobStep.CreateLibraryFile:
            goto CreateLibraryArchive;
        case AddLibraryFileJobStep.CreateLibraryArchive:
            goto ExtractArchive;
        case AddLibraryFileJobStep.ExtractArchive:
            goto AddExtractedFiles;
        case AddLibraryFileJobStep.AddExtractedFiles:
            goto Commit;
        case AddLibraryFileJobStep.Commit:
        case AddLibraryFileJobStep.Completed:
            return JobResult.CreateCompleted(job.LibraryFile.Value);
    }

CheckIfArchive:
    // Your existing logic for checking if archive
    job.LastCompletedStep = AddLibraryFileJobStep.CheckIfArchive;
    cancellationToken.ThrowIfCancellationRequested();

Hash:
    // Your existing logic for hashing
    job.LastCompletedStep = AddLibraryFileJobStep.Hash;
    cancellationToken.ThrowIfCancellationRequested();

CreateLibraryFile:
    // Your existing logic for creating library file
    job.LastCompletedStep = AddLibraryFileJobStep.CreateLibraryFile;
    cancellationToken.ThrowIfCancellationRequested();

CreateLibraryArchive:
    // Your existing logic for creating library archive if necessary
    job.LastCompletedStep = AddLibraryFileJobStep.CreateLibraryArchive;
    cancellationToken.ThrowIfCancellationRequested();

ExtractArchive:
    // Your existing logic for extracting archive if necessary
    job.LastCompletedStep = AddLibraryFileJobStep.ExtractArchive;
    cancellationToken.ThrowIfCancellationRequested();
```

erri120: That doesn't really seem easier to use.

Al12rs: This is not supposed to be a persistent job right? If the job is paused, it would be lost on app close right?

src/NexusMods.Library/AddLibraryFileJobWorker.cs
```diff
@@ -0,0 +1,218 @@
...
+        //     var stream = job.Stream;
+        //     stream.Position = 0;
+        //
+        //     var hash = await stream.XxHash64Async(token: cancellationToken);
+        //     stream.Position = 0;
+        //
+        //     return hash;
+        // });
+        //
+        // return await AddJobAndWaitForResultAsync(hashJob);
```
Al12rs, line 180: This for when?

src/NexusMods.Library/AddLocalFileJobWorker.cs
```diff
@@ -0,0 +1,43 @@
...
+        var libraryFile = jobResult.RequireData<LibraryFile.New>();
+        var localFile = new LocalFile.New(job.Transaction, libraryFile.LibraryFileId)
+        {
+            LibraryFile = libraryFile,
+            OriginalPath = absolutePath.ToString(),
+        };
+
+        var transactionResult = await job.Transaction.Commit();
+        return JobResult.CreateCompleted(transactionResult.Remap(localFile));
+    }
```
Al12rs, line 42: Could `LibraryService` be made to implement `AJobWorker<AddLocalFileJob>` and `AJobWorker<AddLibraryFileJob>` at the seme time potentially?
Not saying it would be a good idea, just in theory?

src/NexusMods.Library/AddLocalFileJobWorker.cs
```diff
@@ -0,0 +1,43 @@
+using JetBrains.Annotations;
```
Al12rs, line 1: Consider adding a "Jobs" and potentially "JobWorkers" project subdirectories to organize these

Al12rs (commented):
Some questions above

src/NexusMods.Library/AddLocalFileJobWorker.cs
erri120: C# doesn't have multi-inheritance. You can't inherit multiple classes, but you can implement multiple interfaces. `AJobWorker<T>` doesn't really give you anything useful, so `ILibraryService` can just implement `IJobWorker` if you really want to.

src/NexusMods.Library/AddLibraryFileJobWorker.cs
erri120: When I add support for groups.

Al12rs (approved):

Issue #1685, opened 2024-06-25
Selected Panel


erri120:
- [x] #1768
- [x] #1780
- [ ] Update left menu to consider selected and open page status
  - Different design depending on whether the page is in a selected tab, or just open in any tab
  - Probably need to have some observable in `IWorkspaceController` or `IWorkspaceViewModel` that allows anyone to subscribe to page changes.

Notes from discussions:

- Add `IsSelected` to Panel
- need to figure how to know if a panel is "selected" (try focus, or events, or pointer enter)
- behavior:
  - default to replace tab in selected panel
  - if page is already open in any tab of any panel in active workspace we select the tab and don't open a new page
  - left menu highlights selected tab in any panel

Required design @captainsandypants:
- selected state of a panel
- selected state for left menu entry

# Design

**Single panel in workspace**
When there is only 1 panel in a workspace, any time a new page is opened in the same workspace, it will always replace the current page, in the current tab, in that panel. This is whether it's opened through the left hand menu (unless that page is already in another panel, this behaviour is described in the Interactions section), or through a button/link in the current opened page, or any other method. As there is only 1 panel, there is no need for any UI changes to show that it is the "selected" panel.

**Multiple panels in workspace**
When there are multiple panels opened will will have a new concept of a "Selected panel". There can only be 1 selected panel per workspace. This will be indicated in the UI with a change to the border style of that panel, as well as the border style of the left menu item (if applicable). The active panel behaviour will be the same as the single panel behaviour described above.

![image](https://github.com/Nexus-Mods/NexusMods.App/assets/118466382/5852d3dc-2fd4-4790-81ec-bd2ca1b0b714)
_Note that the left menu items will be in the active state (bg fill highlighted but no border) if a panel is open and the active tab is displaying that left menu item. The selected panel will also have its corresponding left hand menu item in the selected state (highlighted and border). There will only ever be one selected panel and left menu item, but there can be multiple active ones._

**Interactions that change which panel is selected**
Clicking anywhere inside a panel will make that panel selected.
When a new panel is created, that panel is automatically set as selected.
Changing to an open panel through the left hand menu.
Keyboard shortcuts, described below.

PR #1768, opened 2024-07-15
Selected Panel Part #1

erri120:
Part of #1685. This includes the styles and the concept of a "selected panel". This is still missing the behavior, that'll come in part 2.

https://github.com/user-attachments/assets/5e55808f-f916-46c9-b546-c5d92df24922

### DIFF
--- a/src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
@@ -35,7 +35,7 @@ public partial class PanelView : ReactiveUserControl<IPanelViewModel>
                 .DisposeWith(disposables);
 
             // panel selection
-            this.AddDisposableHandler(PointerEnteredEvent, (_, _) =>
+            this.AddDisposableHandler(PointerPressedEvent, (_, _) =>
             {
                 if (ViewModel is not null) ViewModel.IsSelected = true;
             }, routes: RoutingStrategies.Direct | RoutingStrategies.Bubble, handledEventsToo: true).DisposeWith(disposables);
--- a/src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
@@ -40,10 +40,12 @@ public partial class PanelView : ReactiveUserControl<IPanelViewModel>
                 if (ViewModel is not null) ViewModel.IsSelected = true;
             }, routes: RoutingStrategies.Direct | RoutingStrategies.Bubble, handledEventsToo: true).DisposeWith(disposables);
 
-            this.AddDisposableHandler(PointerExitedEvent, (_, _) =>
-            {
-                if (ViewModel is not null) ViewModel.IsSelected = false;
-            }, routes: RoutingStrategies.Direct | RoutingStrategies.Bubble, handledEventsToo: true).DisposeWith(disposables);
+            this.WhenAnyValue(view => view.IsKeyboardFocusWithin)
+                .Where(isFocused => isFocused)
+                .SubscribeWithErrorLogging(_ =>
+                {
+                    if (ViewModel is not null) ViewModel.IsSelected = true;
+                }).DisposeWith(disposables);
 
             // update scroll buttons and AddTab button (show left aligned or right aligned, depending on the scrollbar visibility)
             Observable.FromEventPattern<ScrollChangedEventArgs>(
--- a/src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
@@ -9,7 +9,7 @@ using ReactiveUI;
 
 namespace NexusMods.App.UI.WorkspaceSystem;
 
-[PseudoClasses(":one-tab", ":selected")]
+[PseudoClasses(":one-tab", ":selected", ":alone")]
 public partial class PanelView : ReactiveUserControl<IPanelViewModel>
 {
     private const double ScrollOffset = 250;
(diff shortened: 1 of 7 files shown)


src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
```diff
@@ -32,6 +34,19 @@ public PanelView()
                 .Subscribe()
                 .DisposeWith(disposables);
 
+            // panel selection
+            this.AddDisposableHandler(PointerEnteredEvent, (_, _) =>
+            {
+                if (ViewModel is not null) ViewModel.IsSelected = true;
+            }, routes: RoutingStrategies.Direct | RoutingStrategies.Bubble, handledEventsToo: true).DisposeWith(disposables);
```
Al12rs, line 41: This changes selection on pointer entered hover, the desired behavior is selection to change on interaction with contents.

src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
```diff
@@ -32,6 +34,19 @@ public PanelView()
...
+
+            this.WhenAnyValue(view => view.IsKeyboardFocusWithin)
+                .Where(isFocused => isFocused)
+                .SubscribeWithErrorLogging(_ =>
+                {
+                    if (ViewModel is not null) ViewModel.IsSelected = true;
+                }).DisposeWith(disposables);
```
Al12rs, line 48: This does not seem to be doing anything from my testing

src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
erri120: Right, forgot to change that.

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/UserControls/WorkspaceSystem/PanelViewStyles.axaml
```diff
@@ -45,14 +66,8 @@
...
 
-                    <!-- selected tab -->
-                    <Style Selector="^ Border#Container.Selected">
-                        <Setter Property="BorderThickness" Value="1 0 0 0" />
-                    </Style>
-
-                    <!-- not selected tab -->
-                    <Style Selector="^ Border#Container:not(.Selected)">
-                        <Setter Property="BorderThickness" Value="1 0 0 1" />
+                    <Style Selector="^ Border#Container">
```
Al12rs, line 69: Apparently this targets all Borders called "Container", even from child controls.
It seems we need to be more specific with with our naming or use more specific pathing

src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelView.axaml.cs
Al12rs: I tried using tab after disabling the pointer over version. I will try again if it is working on your end

Al12rs: Ah, I found the problem. If you close the selected tab or open a new Workspace without any selection, it tabbing will not select the panels. If one panel is already selected, tabbing does select panels correctly.

Al12rs: Actually pointer press doesn't seem to work either for new workspaces or if you close the selected panel

erri120: > Actually pointer press doesn't seem to work either for new workspaces or if you close the selected panel

Can you write down in detail how to repro? If you close a selected panel, then the workspace should select the first panel you have.

erri120: Closing selected panel, switching workspaces, and everything else works on my end.

Al12rs: I tried cleaning my datamodel folder but now I can't get any panel to be selected after app start. No panel is selected and nothing selects panels

Al12rs: Ah, I got it. If you have a single panel there is no selection highlighting, which was confusing me.
Is that by design?

erri120: Yeah

Al12rs (changes requested):
Closing and reopening the app while one of the panels contains a NewTab page causes exceptions on reopening and a workspace without panels.

Al12rs:
@captainsandypants in my testing of the feature I found the absence of the selected styling in the case of single panel very confusing. I was assuming a bug. 
The absence of the selected styling made me think something else must have been selected somewhere else or that the app was not responding correctly to interactions, as it does in multi panel scenarios.

Al12rs (approved):
Exception will be addressed in separate PR.

captainsandypants:
> I found the absence of the selected styling in the case of single panel very confusing. I was assuming a bug. The absence of the selected styling made me think something else must have been selected somewhere else or that the app was not responding correctly to interactions, as it does in multi panel scenarios.

@Al12rs I find it intuitive as there's only one panel so it has to be the selected one. I want to avoid unnecessary UI wherever possible. We'll keep an eye on further feedback on this though.

captainsandypants:
@erri120 when there's a single tab the UI doesn't match the designs. The right border shouldn't be edge to edge. See design here: 
image

erri120:
> @erri120 when there's a single tab the UI doesn't match the designs. The right border shouldn't be edge to edge.

This was a compromise I made when implementing the original design for tabs months ago. The line is edge to edge because I set the border on the right to be 1px. There might be a way to change the geometry of the border using a custom brush, but that'll take a lot of experimenting to figure out. Another option would be to have no border, and just put a "line" element in between the tab and the plus icon. That would also take more time than it's worth, as that would require code-behind changes.

I couldn't justify spending hours on turning this border into a floating line, so we have this for now.

erri120:
Added the separator

captainsandypants (approved):
All good.

PR #1780, opened 2024-07-16
Selected panel behavior

erri120:
Part of #1685. Missing is the left menu updates.

### DIFF
--- a/src/NexusMods.App.UI/WorkspaceSystem/Workspace/WorkspaceViewModel.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/Workspace/WorkspaceViewModel.cs
@@ -352,38 +352,33 @@ public class WorkspaceViewModel : AViewModel<IWorkspaceViewModel>, IWorkspaceVie
         return pageData;
     }
 
+    /// <summary>
+    /// Tries to select the first panel where the selected tab shows the same page.
+    /// </summary>
     private bool TrySelectPage(PageData pageData)
     {
-        var openTab = Panels
-            .Select(panel => (panel, tab: panel.Tabs.FirstOrDefault(tab => tab.Contents.PageData.Context.Equals(pageData.Context))))
-            .FirstOrOptional(tuple => tuple.tab is not null);
+        var panel = Panels.FirstOrDefault(panel => panel.SelectedTab.Contents.PageData.Context.Equals(pageData.Context));
+        if (panel is null) return false;
 
-        if (!openTab.HasValue) return false;
-        var (targetPanel, targetTab) = openTab.Value;
-        targetPanel.SelectTab(targetTab!.Id);
-        targetPanel.IsSelected = true;
-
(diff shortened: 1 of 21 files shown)


captainsandypants:
@erri120 when I open a Health Check diagnostic in a new tab, then go back to the Health Check tab and try to open the Diagnostic in a new panel, it instead selects the already created Diagnostic tab.

Same thing happens if I open the Diagnostic in a new panel first, when I then try and also open it in a new tab, it instead selects the already created panel.

src/NexusMods.App.UI/WorkspaceSystem/Workspace/WorkspaceViewModel.cs
```diff
@@ -352,8 +352,27 @@ private PageData GetDefaultPageData()
         return pageData;
     }
 
+    private bool TrySelectPage(PageData pageData)
+    {
+        var openTab = Panels
+            .Select(panel => (panel, tab: panel.Tabs.FirstOrDefault(tab => tab.Contents.PageData.Context.Equals(pageData.Context))))
+            .FirstOrOptional(tuple => tuple.tab is not null);
```
Al12rs, line 359: In case of multiple versions of the same page being already open, this is picking the first one in what I would guess be panel layout order and then tab order.

If one panel has a selected tab matching the pageData while another panel has a hidden tab matching the page data, I would think the user would expect the already selected tab to become the target.

If there are multiple tabs in different panels, current panel selection should perhaps also be considered when deciding on the tab to select.

@captainsandypants Do you have rules in mind for which tab should be selected in case of multiple already existing tabs across different panels that already contain the desired Page?

Al12rs (changes requested):

src/NexusMods.App.UI/WorkspaceSystem/Workspace/WorkspaceViewModel.cs
captainsandypants: If you open a page with a standard left click, it should replace the content in the selected panel unless there is another panel open in the workspace with an active tab already displaying that page, in this case that panel should be selected. If there's a hidden tab with the same page as the one that was left clicked, it still follows the default behaviour, it doesn't need to know about hidden tabs. The left hand menu pages that are open in an active tab will be given the active state, if they are open in a hidden tab then they will have their default rest state i.e. nothing happens.

If you open a page with a right click and select open in new tab or panel, then it should always open in a new tab or panel, it does not need to know if that page is already open in a panel or any tab.

Al12rs: Ah, I see. Yes, that would solve the hidden tab priority problem, thanks for the clarification!

captainsandypants (approved):
Working great.

Al12rs (approved):
Tested as well, and looked over the changes. Should work fine this way

PR #1789, opened 2024-07-18
Fix `Directory.package.props` line endings

Al12rs:


erri120:
We could also remove `*.props text eol=crlf` from `.gitattributes`

Al12rs:
If you prefer that I can do it

erri120 (approved):

PR #1799, opened 2024-07-22
Update cyberpunk installers to new LoadoutItem format

halgari:
This code does the following:

* Updates the Cyberpunk Installers to use the new LoadoutItem format.
* Path based installers dump their contents into a group with the filename as the archive they came from, probably want to do something smarter here
* RedMod installers will install themselves in the format of `Archive`->`RedMod1`->`Files..`. That is to say if you have 3 redmods in a single archive they are split out into several sub-mods.
* Tests are implemented for all installers.
  * Path based installers are tested together in one base test that takes input paths and uses `Verify` on the destination locations
  * Redmod installers are verified by looking at the transaction that writes the data to the database. This works fairly well to determine if all the metadata is making it into the database
* RedMods have their own subclass from `LoadoutItemGroup` and RedModInfo files `info.json` have their own subclass from `LoadoutFile` which records the version/name of the redmod metadata.

Redmod tests run off of a stripped-down archive I downloaded from the nexus. The script I wrote truncates all non-`info.json` files down to 128bytes, and leaves the info.json files alone. This allows us to run against actual data and verify the output of the installer.

Part of #1763.

Issue #1763, opened 2024-07-15
Road to Loadout Items


erri120:
## Loadout Items

- [x] #1830
- [x] #1750
- [ ] Remove almost all references of "mod" from code :)
- [x] Rework installers, should integrate with library items
  - [x] #1785
  - [x] #1799
  - [x] #1801
  - [x] #1802
  - [x] FOMOD Installer (#1823)
  - [x] Advanced Installer (#1823)
- [x] Rework diagnostic emitters (only implementations, nothing else has to change)
  - [ ] #1784
  - [x] #1807
  - [x] #1809
- [x] Rework synchronizer
  - [x] #1805
  - [x] #1806
  - [x] #1826
- [ ] Rework VMs
  - [ ] Rework "Library" page (#1864)
  - [ ] Rework "My Mods" page (#1864)
  - [x] #1880
  - [x] Implement delete command in Loadout Grid (#1854)
  - [x] Replace `IModInfoViewModel` with `ILoadoutItemViewModel` for `LoadoutItem` instead of `Mod` (#1810)
    - [x] Replace `IModFilesViewModel` and `ModFileTreeViewModel` to work with `LoadoutFile`. (#1810)
    - [x] Implement `IFileTreeViewModel` for loadout items (#1810)
    - [x] Update text editor to accept `LoadoutFileId` instead of `FileId` (#1810)
  - [x] `IFileOriginsPage` use `ILibraryService` directly (#1828)
    - [x] `IFileOriginEntryViewModel` use `LibraryItem` instead of `DownloadAnalysis` (#1828)
  - [x] Update `LoadoutCardViewModel` to not use `Loadout.Mods.Count` for the mod count (#1830)
  - [ ] Complete rework of #1405, requires downloads and jobs
  - [x] Rework `LoadoutGrid` (#1810)
    - [x] Remove `Category` and `Version` column for now
    - [x] List `LoadoutItem` instead of `Mod`
    - [x] Rework columns, mostly just renaming from `Mod` to `LoadoutItem` for now
- [ ] Rework all tests
  - [ ] #1455
- [ ] Rework CLI verbs (or remove)
  - [ ] `stress-test`: uses `IHttpDownloader`, `IFileOriginRegistry`, and `IArchiveInstaller`
  - [ ] `download-and-install-mod`: uses `IHttpDownloader`, `IFileOriginRegistry`, and `IArchiveInstaller`
[...]

PR #1801, opened 2024-07-23
Loadout Item API update

erri120:
Part of #1763, will conflict with #1799.

Changes:
- adds `Path` to `LibraryArchiveFileEntry`
- uses `MarkerAttribute` for `LoadoutItem.IsDisabled`
- renamed existing markers and updated usage to set them
- updated the installer API with a simpler `IsSupported`
- updated the job worker to skip unsupported installers

There is one issue here:

The job worker for installing a loadout item passes a transaction to each of the installer. If the installer doesn't support the item, we need to retract all entities it has added to the transaction before continuing. Not sure what's the best approach here @halgari.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts/LoadoutItem.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts/LoadoutItem.cs
@@ -20,13 +20,13 @@ public partial class LoadoutItem : IModelDefinition
     public static readonly StringAttribute Name = new(Namespace, nameof(Name));
 
     /// <summary>
-    /// Whether the item is disabled. A disabled loadout item will not participate
-    /// in actions.
+    /// Marker to signal that the item is disabled. A disabled loadout item will
+    /// not participate in actions.
     /// </summary>
     /// <remarks>
     /// The exact meaning of a "disabled" loadout item is up to the implementations.
     /// </remarks>
-    public static readonly BooleanAttribute IsDisabled = new(Namespace, nameof(IsDisabled));
+    public static readonly MarkerAttribute IsDisabledMarker = new(Namespace, nameof(IsDisabledMarker));
 
     /// <summary>
     /// Loadout that contains the item.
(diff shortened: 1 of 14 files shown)


src/Abstractions/NexusMods.Abstractions.Loadouts/LoadoutItem.cs
```diff
@@ -20,13 +20,13 @@ public partial class LoadoutItem : IModelDefinition
...
-    /// Whether the item is disabled. A disabled loadout item will not participate
-    /// in actions.
+    /// Marker to signal that the item is disabled. A disabled loadout item will
+    /// not participate in actions.
     /// </summary>
     /// <remarks>
     /// The exact meaning of a "disabled" loadout item is up to the implementations.
     /// </remarks>
-    public static readonly BooleanAttribute IsDisabled = new(Namespace, nameof(IsDisabled));
+    public static readonly MarkerAttribute IsDisabledMarker = new(Namespace, nameof(IsDisabledMarker));
```
Al12rs, line 29: In this case an optional Boolean attribute could actually make sense, if we wanted to consider things like the GameFiles.
You would have True or False set states for regular mods, but could also have null (attribute not set) state for special mods like GameFiles.

I don't know if that is actually a good idea, it feels hakish and easy to misinterpret

src/NexusMods.Library/InstallLoadoutItemJobWorker.cs
```diff
@@ -23,32 +21,38 @@ public InstallLoadoutItemJobWorker(IServiceProvider serviceProvider)
...
+        InstallLoadoutItemJob job,
+        ILibraryItemInstaller[] installers,
+        CancellationToken cancellationToken)
+    {
+        foreach (var installer in installers)
+        {
+            var isSupported = installer.IsSupportedLibraryItem(job.LibraryItem);
+            if (!isSupported) continue;
+
+            // TODO: we need to retract all added entities in the transaction if the installer doesn't support the item
```
halgari, line 49: Maybe we just restart the transaction in that case?

erri120: What do you mean with "restart"?

halgari:
Transactions are not final until they are committed. So if we try one mod
installer and it doesn’t work, we can just dispose of the transaction and
make a new one to start the next installer.

erri120:
> Transactions are not final until they are committed. So if we try one mod installer and it doesn’t work, we can just dispose of the transaction and make a new one to start the next installer.

Updated the code to create a new transaction. That works fine for this specific job, but we'd still run into issues with other jobs that can be paused/resumed.

PR #1805, opened 2024-07-24
Add `LoadoutGameFilesGroup`

erri120:
Part of #1763.

- enables `DeletedFile` model and fixes imports
- adds `LoadoutGameFilesGroup`
- updates the synchronizer to create groups per location

Planned for #1763 is putting game files in some capacity into the library as well. This would also remove the need for a "vanilla state loadout".


halgari (approved):
Pretty much the same as wehat I had in the synchronizer branch I was working on

PR #1807, opened 2024-07-24
Convert Cyberpunk Diagnostics to Loadout Items

halgari:
Part of #1763.

Not much to say here, pretty much a straight forward switch to the new format.


src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/Diagnostics.cs
erri120, line 1: The user-facing text shouldn't need to change. We still need to refer to things as "mods" for users.

erri120 (changes requested):

erri120 (approved):

PR #1808, opened 2024-07-25
Added: Core of the GC [Garbage Collection 1/?

Sewer56:
This PR contains the lightweight, independent 'core' of the Garbage Collection implementation.

This core is currently completely abstracted from the regular codebase of the App, in order to make long term maintenance easier as models, and other implementation details change. [Should also make review easy]

- Generic code lives `NexusMods.App.GarbageCollection`
- Nx specific code will live in `NexusMods.App.GarbageCollection.Nx`
- DataModel specific code will live in `NexusMods.App.GarbageCollection.DataModel`

Currently there isn't very much here, just what I got done yesterday after standup.
I figured splitting this into multiple PRs would be easier; so this is just the beginning.

In the next PR, I'll be doing the code that does the actual repacking using Nx.


erri120:
> Generic code lives NexusMods.App.GarbageCollection
> App specific code will live in NexusMods.App.GarbageCollection.NMA

Why the split? This is the App repo, you only have app-specific code.

src/NexusMods.App.GarbageCollection/Structs/Hash.cs
```diff
@@ -0,0 +1,5 @@
+using TransparentValueObjects;
+namespace NexusMods.App.GarbageCollection.Structs;
+
+[ValueObject<ulong>]
+public readonly partial struct Hash;
```
erri120, line 5: Why do we need another `Hash` type?

Sewer56:
> > Generic code lives NexusMods.App.GarbageCollection
> > App specific code will live in NexusMods.App.GarbageCollection.NMA
> 
> Why the split? This is the App repo, you only have app-specific code.

I can put them in the same project technically speaking.

I chose to abstract stuff away because it's easier to adapt when the App will feel like breaking the DataModel in two again.

The abstractions allow me to test easier; and I also wanted to avoid the situation where a large change in the DataModel led to tests being removed and not re-added.

src/NexusMods.App.GarbageCollection/Structs/Hash.cs
Sewer56: This was an accident and has been changed. I accidentally put this PR up before I was done with it.
I was typing the PR text, and accidentally hit `Enter` instead of `Shift+Enter`.

It's a leftover from when I was experimenting on an empty project while running the Nx stress tests in the background.

halgari:
I think this issue would do well to have some design behind it, perhaps with an ADR (you can forgo the trandeoffs in the ADR, mostly we want a design document here). 

All the hashes we track in the system are now referred to via `LoadoutFile.Hash`. So anything used in a loadout will be in one of those entities or a previous state of those entities. Likewise the new `LibraryArchiveEntry` has a `.Hash` that points to all files that are from downloaded mods.

So I think this does end up in a mark-and-sweep sort of operation: 

1) enumerate all current loadouts
1.1) get a history index (need to code this MnemoincDB) to find all hashes that were ever attached to this loadout
2) enumerate all `LoadoutArchiveEntry` models and collect the hashes from those
3) combine all the hashes into a "keep" hashset
4) enumerate all files in .NX and delete any that are not in the "keep" hashest

Since we store all .NX entries in MnemonicDB, we could perhaps store all the hashes in a sorted set and then do a merge join of those hashes against `ArchivedFile.Hash`. If you use `db.Datoms(ArchivedFile.Hash)` the resulting datoms will all be sorted by hash. From there you're essentially doing a merge join of two 64bit integer arrays, and that can be heavily optimized. 

This alone would give us the ability to reclaim any space from deleted loadouts and downloads. We can handle GC'ing history of loadouts in the future.

Sewer56:
I threw the parts that contain the Nx code into this PR; since this wasn't reviewed sorted.

What's left is:

- Writing an ADR explaining the process.
- Finding all used hashes in datastore.
- Updating datastore's metadata (saved `Nx` FileEntry items) on repack.

Sewer56:
I added the ADR to this PR today.
It's on the lengthier side, due to the nature of the topic; but I think I covered the important points without going too overkill into the details.

docs/developers/decisions/backend/0016-garbage-collector-design.md
```diff
@@ -0,0 +1,316 @@
...
+    - Find all files that are in use.
+    - This is any `LoadoutItem` which is a file (`LoadoutItemWithTargetPath`) that is not marked as deleted (retracted).
+    - This gives us ***UsedFileList (B)***.
+- [Remove all elements in ***AllFileList (A)*** that are *not* in ***UsedFileList (B)***](#removing-unused-items).
+    - The removal process involves `'repacking'` all archives which have had their files
+removed to exclude the files that are no longer needed.
+
+### Parsing Nx Archives
+
+!!! note "This process involves identifying all archives the App has available, and which files are contained within them."
```
halgari, line 38: Don't need to do this, we can instead do `db.Datoms(ArchivedFile.Hash)` and that will return a datom for each archived file. Bonus: the results will be sorted by hash

Sewer56: [Note: I went a bit overboard with the response here, but it may be good to include in ADR, because this is a good question; and I was feeling a little bit passionate]

--------

I technically don't need to do this step to find out which files are not in use.

*However*, it is more efficient for me to do so.
With `db.Datoms(ArchivedFile.Hash)`, I would need to do the following:

- Make 1 DB query to get the container name for every hash to determine the `Nx`.
  - And for every hash, check all locations in `DataModelSettings.ArchiveLocations` to see if the file exists.
      - Which basically means opening a file handle and seeing if it works.
      - In most cases there's only 1 `ArchiveLocation`, so usually 1 time per file.
  - i.e. Same as `NxFileStore.TryGetLocation`.

If I instead read the archive headers to get this information; the following applies:

- Scan a folder for all archives.
    - On Windows, around 450 file (archive) entries per millisecond. Uses `NtQueryDirectoryFileEx` under the hood.
    - Can scan multiple folders in parallel, and that does scale.
- Open each archive, parse the header. Usually just first 4K page for >90% of the mods.
- Grab the hashes of multiple files from the Nx header.

In this case, a handle is also opened for every archive file, to read the header.
The part that is slow on Windows is unfortunately opening that handle, and that is something we do in both situations.

In the datoms case, I technically would need to open a file multiple times; but I can avoid that with a `fileExistsCache` like I did in `NxFileStore`.

------

So what it really comes down to is asking the following:

1. Whether querying container name `entry.Container` for every datom `db.Datoms(ArchivedFile.Hash)` is faster than `NtQueryDirectoryFileEx`.

2. Whether reading Hash Datoms from DB is faster than reading Nx archive headers.

For both steps, I think they're about equal.

For step `1`, reading the MFT on an NTFS volume is mostly a sequential operation, with some occasional fragmentation leading to jumping around. Reading the DB on the other hand involves reading from at least 2 places for the raw data, and an index lookup (3). Both approaches also open file handles, so we can cancel that part out.

For `2`, they're also about equal; assuming the datom data is laid out sequentially in RocksDB; then both the `FileEntry` structure (20 bytes) and MnemonicDB's Hash Entry [16 byte KeyPrefix + 8 byte Value] are about equal in size; and sequential.

At which point, you may ask; `why not just use the DB?`. The answer is very simple.

Regardless of the approach, I need to parse the Nx header regardless in order to perform the repacking step. That is unavoidable. This is because I need metadata such as `Chunk Size`, `Block Sizes` (and thus offsets), `Block Compressions` etc. Because the archive headers also contain the hashes, simply pulling the hashes from the pre-parsed headers directly from RAM is faster than firing DB calls.

halgari: `Whether querying container name entry.Container for every datom db.Datoms(ArchivedFile.Hash) is faster than NtQueryDirectoryFileEx`

This is where set logic comes into play, you can have multiple sets of data in memory and link them up as needed. For example you could scan for all the file names (ignoring their size and contents) first that tells you how to filter the `ArchivedFileContainer` objects. From there you have a sorted set that you can use to probe into when figuring out what files reference what archives. 

In general though I think we should assume that all ArchivedFileContainers will exit if the DB says they do, going in and randomly deleting .nx archives is not supported behavior of the app. 

I think what I'm more concerned about here is that this ticket is dragging on, and we have some pretty complex logic here when the simpler approach could likely be nearing completion. I'd like to find ways of making this code at least testable soon and then go back and add microoptimizations as we figure out what's needed. Making any arguments about memory usage on something that isn't even testable yet is a bit out-of-order on the design. 

We can hook this into the UI to run as part of unmanaging a game, and then we have a "good enough" solution for now: whenever you delete the last loadout of a game, we run GC to delete any backed up files you no longer need.

Sewer56:
Closing in favour of the successor PR including library deletion.

PR #1826, opened 2024-07-30
Fix loadout status

halgari:
This PR does the following:

* Renames `IApplyService` to `ISynchronizerService`
* Moves all the loadout status (can apply, applied, outdated, etc) into IApplyService
* Leverages the new Analyzers from MnemonicDB to track tree updates
  * Updates to any entity cause the analyzer to collect all the entities that entity points to, and the ones it points to, recursively
  * Results in a `ISet<EntityId>` of all parents and children that were updated (by association)
  * That can then be used to drive any update in the UI that needs to know when the contents of loadout changes (or any tree-like structure)
* Updates the spine to drive off of this new analysis data


Sewer56 (commented):
This looks good to me so far, but I'll leave it for a second set of eyes to double confirm.

I did notice [NexusMods.Abstractions.MnemonicDB.Analyzers.csproj](https://github.com/Nexus-Mods/NexusMods.App/pull/1826/files#diff-9016b79bfea0024c6856085ead1026644201cebc463f62d8f59df4325b311b6f) is using `.NET 8` as the target framework. Although I haven't confirmed it, it may be problematic for external contributors using Visual Studio on Windows; since that still runs on .NET Framework, and would thus refuse to load the analyzer. It's probably still OK if it doesn't break build though.

halgari:
> This looks good to me so far, but I'll leave it for a second set of eyes to double confirm.
> 
> I did notice [NexusMods.Abstractions.MnemonicDB.Analyzers.csproj](https://github.com/Nexus-Mods/NexusMods.App/pull/1826/files#diff-9016b79bfea0024c6856085ead1026644201cebc463f62d8f59df4325b311b6f) is using `.NET 8` as the target framework. Although I haven't confirmed it, it may be problematic for external contributors using Visual Studio on Windows; since that still runs on .NET Framework, and would thus refuse to load the analyzer. It's probably still OK if it doesn't break build for them though. Needs confirmation.

Good catch, but this is a data analyzer that hooks into MnemonicDB not a source analyzer. English name overloading FTL again :|

PR #1828, opened 2024-07-31
FileOriginViewModel to Library

halgari:
Converts the internals of the FileOrigin Page/ViewModel over to `ILibraryService`. 

The loop of Download->Install->Sync->Disable->Sync has been hand verified by me. NXM links also work, but the name in the library is all whack. We'll handle that in a separate PR.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Library/ILibraryService.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Library/ILibraryService.cs
@@ -28,5 +28,6 @@ public interface ILibraryService
     /// </summary>
     /// <param name="libraryItem">The item to install.</param>
     /// <param name="targetLoadout">The target loadout.</param>
-    IJob InstallItem(LibraryItem.ReadOnly libraryItem, Loadout.ReadOnly targetLoadout);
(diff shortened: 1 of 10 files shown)


src/Abstractions/NexusMods.Abstractions.Library/ILibraryService.cs
```diff
@@ -28,5 +28,6 @@ public interface ILibraryService
     /// </summary>
     /// <param name="libraryItem">The item to install.</param>
     /// <param name="targetLoadout">The target loadout.</param>
-    IJob InstallItem(LibraryItem.ReadOnly libraryItem, Loadout.ReadOnly targetLoadout);
+    /// <param name="installer">If set, must be set to a ILibraryItemInstaller (untyped here as to not require a library reference). The Library will use this installer to install the item</param>
+    IJob InstallItem(LibraryItem.ReadOnly libraryItem, Loadout.ReadOnly targetLoadout, object? installer = null);
```
erri120, line 32: What's the issue with using `ILibraryItemInstaller`? I changed this on my branch and couldn't find an issue. The `Library*` abstraction projects were designed to prevent dependency issues by splitting up `ILibraryService`, the installers, and the models into separate projects, so there shouldn't really be any issues here.

PR #1829, opened 2024-07-31
Add user filters

erri120:
Part of #1763.

"user" filters in this context are filters for items to be shown to the user, not filters created by the user.


Sewer56 (approved):
LGTM

Just make sure `GameFiles` and/or `Override` filters are eventually accounted on.
Depending on what design wants in the long run.

PR #1830, opened 2024-07-31
Add Loadout Items ADR

erri120:
Also fixes the count on the loadout card.

Part of #1763.

### DIFF
--- a/docs/developers/decisions/backend/0015-loadout-items.md
+++ b/docs/developers/decisions/backend/0015-loadout-items.md
@@ -1,0 +1,34 @@
+# Loadout and Library Items
+
+## Context and Problem Statement
+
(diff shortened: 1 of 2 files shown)


Sewer56:
This works for me, but I'll leave it for @halgari and @Al12rs to add any addendums they wish to.

Previous Resources:

- https://github.com/Nexus-Mods/NexusMods.App/pull/1680
- https://github.com/Nexus-Mods/NexusMods.App/pull/1750

halgari (approved):

Sewer56:
I think what would help in understanding, is also giving an example of how a conventional 'mod' is represented with the new system.

I wasn't involved in original planning of LoadoutItems, but with the current system, as I understand it given around 1.5 hours of poking around for GC stuff:

- Each file is a `LoadoutFile` -> `LoadoutItemWithTargetPath` -> `LoadoutItem`
- An installed archive (formerly 'Mod') is a `LoadoutItemGroup` which does not contain a parent.
    - Each file from that archive has the `LoadoutItemGroup` set as the parent.
    - Any file with custom metadata should extend `LoadoutFile`, and contain a custom marker.
    - Any group (e.g. type of mod, `RedMod` etc.) should extend `LoadoutItemGroup` with custom marker.
- Items are stored in library by linking a `LoadoutItemGroup` with a `LibraryItem` in what's essentially a join table.

This works for me, but these kinds of relationships should be stated clearly somewhere in dev docs; in order to try better document the system.

Sewer56:
Aah, this also needs to update `mkdocs.yml`.

PR #1831, opened 2024-07-31
Cleanup and download types

erri120:
Part of #1763. Most of the changes come from renaming the marker attributes.

Besides that, I've added `DownloadedFile` as a `LibraryFile`.


Sewer56 (approved):

Sewer56:
(checks failed due to verify, since renamed property names no longer match)

PR #1832, opened 2024-07-31
Use `Downloader` package

erri120:
Part of #1763.


halgari (approved):

halgari:
I remember now, this doesn't support mirrors iirc, it supports multi part downloading, but not mirrors with load balancing, something our current advanced downloader supports. There's also something whacky with how it uses the temp folders that's breaking on Windows. We should look into this.

Sewer56:
Did you reply to the wrong place?

The discussion was here:

- https://github.com/Nexus-Mods/NexusMods.App/issues/1763#issuecomment-2228253745

erri120:
> I remember now, this doesn't support mirrors iirc, it supports multi part downloading, but not mirrors with load balancing, something our current advanced downloader supports.

The API allows passing multiple URLs:

> The download URLs of a file to download as parallel with mirror links.

https://github.com/bezzad/Downloader/blob/f0b8ebbc09ff11ffee07b7c46eb342fd3e8b9549/src/Downloader/IDownloadService.cs#L81

https://github.com/bezzad/Downloader/blob/f0b8ebbc09ff11ffee07b7c46eb342fd3e8b9549/src/Downloader.Test/IntegrationTests/DownloadIntegrationTest.cs#L775

Depending on the number of chunks to download, the library uses the mirrors like this:

https://github.com/bezzad/Downloader/blob/f0b8ebbc09ff11ffee07b7c46eb342fd3e8b9549/src/Downloader/DownloadService.cs#L163-L170

It does support mirrors with load balancing, trying to assign each chunk a different server if possible. The downside here is that the library doesn't load balance based on speed, which is something that our advanced downloader does on the fly. Our code makes sure only the fastest mirrors are selected while downloading chunks.

We should be able to use the library with mirrors by ranking the mirrors before initiating the download. Instead of passing all mirrors, even ones that are on the other side of the planet, we rank them beforehand and only pass the fastest.

Ranking mirrors beforehand also allows us to find good values for some other parameters like chunk size and parallel count. Instead of hardcoding values, these should obviously be settings, but we should do a speed test at some point and provide "suggested" values or similar.

erri120:
> There's also something whacky with how it uses the temp folders that's breaking on Windows. We should look into this.

I think that's just because the `SimpleHttpDownloader` doesn't support resuming. I'm not sure if it's worth investing that, since we're moving to the jobs system anyway.

halgari:
> > There's also something whacky with how it uses the temp folders that's breaking on Windows. We should look into this.
> 
> I think that's just because the `SimpleHttpDownloader` doesn't support resuming. I'm not sure if it's worth investing that, since we're moving to the jobs system anyway.

No I'm saying the `windows-latest` tests fail on this PR and they don't in main.

halgari:
What are we looking to fix by using the downloader library? The code we have in place is pretty stable and simple. I know we've talked about moving to a 3rd party library for this several times, but I'm not sure what that gains us as we'd be replacing a single file we own and does what we need current with a library that we don't control. I don't normally care if we swap out custom code with a 3rd party library, but in this cause it feels like we're doing the change on a whim.

erri120:
> What are we looking to fix by using the downloader library? The code we have in place is pretty stable and simple. I know we've talked about moving to a 3rd party library for this several times, but I'm not sure what that gains us as we'd be replacing a single file we own and does what we need current with a library that we don't control.

The `AdvancedHttpDownloader` implementation is far from simple, and the previous `SimpleHttpDownloader` was too simple. [`bezzad/Downloader`](https://github.com/bezzad/Downloader) is [feature rich](https://github.com/bezzad/Downloader?tab=readme-ov-file#features-at-a-glance) and provides an easy-to-use API that covers almost all of our needs. The library would do all the heavy lifting for downloading, while our new downloader code would provide the integration into our own internals like job reporting, persistence, and forwarding pause/resume requests to the library.

Downloading is somewhat core to the App, but replacing our hard-to-maintain DIY downloader, with a well-maintained and easy-to-use download library seems worth to me. I pitched the idea to @Al12rs as well, when designing the jobs' system, and he agreed with using the library.

Al12rs:
I agree with @erri120 that using a widely used and tested library for downloads would be a better approach than rolling our own and trying to add all the features users will want to it.

We had bugs and regressions in our download code already, and the code is not easy to work with.
Having a production tested and widely used API to rely on and less code to maintain would be preferrable.

PR #1837, opened 2024-07-31
Persisted restartable jobs

halgari:
Updates the job framework to support loading and persisting of jobs. 

* Implements `JobRestarter` a simple IHostedService that simply starts any non-finished jobs when the app starts, then exits
* Implements a test job for testing pause/resume
* Updates to MnemonicDB v0.9.69 - this version supports a way to get a IServiceProvider inside an attribute. It's not super fast, but it's needed for situations like this were we want to serialize workers via an ID. We'll use this same behavior on AbstractPaths (with need IFileSystem) later. 
* Implements a HttpDownloaderJob/Worker/State that just calls into IHttpDownloader for now. 

Note: `DownloadUri` is a single value and not an array at the moment because there's a bug in the source generators for MnemonicDB that causes them to not work with collection attributes. Once that's fixed we can add support for mirrors to the job. 

Part of #1763.


src/Abstractions/NexusMods.Abstractions.Jobs/APersistedJobWorker.cs
```diff
@@ -0,0 +1,13 @@
+namespace NexusMods.Abstractions.Jobs;
+
+public abstract class APersistedJobWorker<T> : AJobWorker<T>, IPersistedJobWorker
+    where T : AJob
+{
+    private PersistedJobStateId _stateId;
+
+    /// <inheritdoc />
+    public Guid Id => Guid.Parse("17DBF060-5A55-4960-81D6-F99E4CD24702");
```
erri120, line 9: Shouldn't this be abstract?

src/Networking/NexusMods.Networking.Downloaders/HttpDownloadJobWorker.cs
```diff
@@ -0,0 +1,51 @@
...
+    public HttpDownloadJobWorker(IConnection connection, IHttpDownloader downloader)
+    {
+        _connection = connection;
+        _downloader = downloader;
+    }
+
+    /// <inheritdoc />
+    protected override async Task<JobResult> ExecuteAsync(HttpDownloadJob job, CancellationToken cancellationToken)
+    {
+        var destination = job.Get(HttpDownloadJobPersistedState.Destination);
```
erri120, line 30: Should be `job.Destination`. The job would have a property `Destination` like this:

```csharp
AbsolutePath Destination => Get(HttpDownloadJobPersistedState.Destination);
```

(same for `Uri` property)

erri120 (changes requested):

src/Networking/NexusMods.Networking.Downloaders/HttpDownloadJob.cs
```diff
@@ -0,0 +1,33 @@
+using NexusMods.Abstractions.Jobs;
+using NexusMods.MnemonicDB.Abstractions;
+using NexusMods.Paths;
+
+namespace NexusMods.Networking.Downloaders;
+
+public class HttpDownloadJob : APersistedJob
```
erri120, line 7: This should also implement `IHttpDownloadJob` or at least `IDownloadJob`.

erri120 (changes requested):

erri120 (approved):

PR #1851, opened 2024-08-06
Revive RedMod Deployment

halgari:
Revives the RedMod Deployment. This is the first part of several PRs that will focus on improving Cyberpunk support in the App. What this PR does:

* Adds `IsIgnoredPath` to the Synchronizer. This is used to **completely** filter out paths from the sync process. Several folders in Cyberpunk are auto-generated by the game and/or redmod and so don't need to be tracked by us at all. 
* Move the indexing code into `ALoadoutSynchronizer` so that we can liverage the above filter
* Upgrade NX to .6 to get rid of a bug that happens when it tries to archive the empty `.stub` file. 
* Revive the `RedModDeployTool` class and auto-run it when the loadout contains redmods. Turns out it runs super fast if there is nothing to do, so we can run it fairly easily
* Perform the first index of the game files in parallel, I missed this before. On my box this means it indexes in half the time. 
* Fix several bugs with indexing that made the code on `main` rather broken.


src/Games/NexusMods.Games.RedEngine/RedModDeployTool.cs
```diff
@@ -23,18 +24,25 @@ public RedModDeployTool(ILogger<RedModDeployTool> logger)
...
         var result = await Cli.Wrap(exe.ToString())
             .WithArguments("deploy")
-            .WithWorkingDirectory(loadout.InstallationInstance.LocationsRegister[LocationId.Game].ToString())
+            .WithWorkingDirectory(exe.Parent.ToString())
             .WithStandardErrorPipe(PipeTarget.ToStringBuilder(stdErrBuffer))
             .WithStandardOutputPipe(PipeTarget.ToStringBuilder(stdOutBuffer))
-            .ExecuteAsync();
+            .ExecuteAsync(cancellationToken);
         _logger.LogInformation("Finished running {Program}", exe);
+        _logger.LogDebug("RedMod Deploy stdout: {StdOut}", stdOutBuffer.ToString());
```
erri120, line 45: You should use `IProcessFactory` for anything process related. Printing the output to the logs is also not necessary, since the process factory does all that for you.

erri120 (changes requested):

erri120 (approved):

Issue #1686, opened 2024-06-25
Tab History


erri120:
Requires #1685

- Each tab will have a history (limit of 10)
- History contains page data
- History is not persistent
- History skips ephemeral pages
- Linear history (similar to browser, at least Firefox)
- Back and Forward buttons in the TopBar
  - button will go-back one step in the selected tab of the selected panel

## Design

New Top bar design to accommodate back and forward buttons.

Changes:

- Dividers between top bar buttons removed to reduce visual clutter
- Premium indicator font style change to: Title/SM/Semi
- Addition of top bar buttons for back and forward

![Image](https://github.com/Nexus-Mods/NexusMods.App/assets/118466382/a5423893-c63c-4119-94fe-aa282c4bfd41)

![image](https://github.com/user-attachments/assets/9b252609-edb7-48cd-831c-0fbdd9081088)

[Figma link for component](https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=437-2192&t=sgNWQkkgxZ8cIYvr-1)

[Figma link showing in place](https://www.figma.com/design/VNqTTMrBsVaU2UZ9UlNsJj/%F0%9F%93%B1Panel-management?node-id=937-9342&t=fblIwhks6sUZkN7q-1) (Note alignment of back and forward button group with edge of workspace)

![Image](https://github.com/Nexus-Mods/NexusMods.App/assets/118466382/7f403386-b364-462a-948e-b9d87cf387bf)

PR #1853, opened 2024-08-07
Tab history fixes

erri120:
- fixes null exception
- fixes forward button being enabled for new panels
- fixes ephemeral pages messing with the history

Part of #1686.


Al12rs:
If you go back in history and then open an ephemeral page, it does not rest the future history, meaning you can still go forwards after that, which is weird:
image

erri120:
> If you go back in history and then open an ephemeral page, it does not rest the future history, meaning you can still go forwards after that, which is weird

Should be fixed with the latest commit.

Al12rs (approved):

PR #1854, opened 2024-08-07
Re-add deletion

erri120:
Part of #1763. I think using `tx.Delete(id, recursive: true)` is what we want here, seems to work.


halgari (approved):
Yep, should work.

Issue #401, opened 2023-07-05
UI refinements for Multiple Loadouts 


Greg-Nexus:
## EDIT (Al12rs): 
Converting this ticket to UI refinement ticket since the functionality was completed.
See https://github.com/Nexus-Mods/NexusMods.App/issues/401#issuecomment-2242416759 for remaining items.

# User story

**As a** user with multiple loadouts

**I want to** quickly switch between loadouts

**So that** I can apply different sets of mods to my game

## Requirements
Users need to be able to quickly switch between loadouts.

We need to consider whether switching a loadout _applies_ it, or whether that is a separate action that needs to be taken. 

This issue should consider how we clearly display which loadout is being viewed/edited vs. which is currently loaded.

This will require #982 to be completed first as we'll need to "unmod" a game in order to switch between loadouts

Al12rs:
Currently we show the loadouts on the Spine, but this can create confusion for users, specifically as to which loadout is active.
The active loadout is used to determine where Downloads should be installed and where Outside changed should be placed.

So we need a way to make it clear to users which loadout is currently active.

We should also keep in mind the ability of users to have multiple windows of the App open and being able to have two loaoduts open side by side. This should be allowed, so we still need a way to make it clear which of them is the active one.

Al12rs:
- Implement design #1180

Al12rs:
## Multiple Loadouts TODO:
- [ ] Loadout Cloning functionality
- [ ] Prevent multiple Synchronize operation at the same time on same install
- [ ] LoadoutBadge:
	- [ ] Show Progress spinner while Loadout is applying
- [ ] Remove "Enable multiple loadouts" setting
- [ ] MyLoadouts page UI and behaviors:
	- [ ] Show Skeleton Card while a new loadout is creating
	- [ ] Keep Deleting card state for at least 1500ms
	- [ ] Disable Hover state on card while child buttons are hovered
	- [ ] Only allow deleting Loadout if it isn't the last one
	- [ ] Add "Are you sure you want to delete the loadout" confirm message box
	- [ ] Implement LastApplied state
- [ ] Spine UI and behaviour changes:
	- [ ] Add button to open "My Loadouts" page
	- [ ] Add LoadoutBadge to ImageButtons
	- [ ] Grouping of loadout ImageButtons by game install
		- [ ] Make groups collapsible

LukeNexusMods:
@Al12rs Can this be closed now?

[...]

PR #1858, opened 2024-08-07
Multiple loadouts part6

### LINKED ISSUE #1748: Prevent users from Applying multiple loadouts at the same time
## Problem
With multiple loadouts for the same game installation, it becomes possible for users to attempt to apply different loadouts at the same time.
This should be avoided as it result in a mess inside the game folder. Only one loadout should be applying for a specific game installation.

## We want
Prevent Apply of multiple loadouts and communicate to the user the reason they can't apply another loadout.
UI/UX needs to be defined how this should happen in an intuitive way for the user.

We discussed in the past about showing Apply queues, so that could be a potential solution.

## Design

Apply button should be disabled in all other workspaces for the same game when an Apply process in progress.

#### Tooltips on hover of disabled Apply buttons
- The loadout currently being applied: "Applying changes to Game Name"
- Loadouts for the same game that aren't being applied: "Applying Loadout X"

### LINKED ISSUE #1804: Adding a 27th loadout results in a duplicate name
# Bug Report

## Summary
Added loadouts 1-26 correctly names them for each letter of the alphabet. Adding a 27th loadout results in a duplicate name.
image

## Steps to reproduce
1. Manage a game and keep creating loadouts until you have a duplicate.

Loadouts are uniquely named and/or capped at 26. 

## Other information

Log files can be found in these locations:

- Windows: `%LOCALAPPDATA%\NexusMods.App\Logs`
- Linux: `~/.local/state/NexusMods.App/Logs`

Please attach both `nexusmods.app.main` and `nexusmods.app.slim` log files with the most recent dates.

### LINKED ISSUE #1817: Settings: Remove "Enable multiple loadouts" and label changes
Remove the "Enable multiple loadouts" setting as this will always be enabled.

Rename "Cyberpunk 2077: Full game backup" and "Stardew Valley: Full game backup" to:

"Full game backup: Cyberpunk 2077" and "Full game backup: Stardew Valley 2077"

This is so they sit together in the settings page.

### LINKED ISSUE #1818: Loadouts should NOT be automatically applied when created
## Currently

When a loadout is created it is automatically applied. This is undesirable as we do not communicate to the user that this will happen and it could take some time to apply this empty loadout.

## We want

For the loadout to be created but not applied.

Al12rs:
- Part of #401 

## Completed:
- [x] closes #1817 
- [x] Update "Full Game Backup" setting names
- [x] closes #1804
- [x] Add badges to Spine loadout image buttons
- [x] Add scroll bar to spine and redo layout (fix loadouts going under Downloads button)
- [x] Add "Add" button to spine
- [x] Group loadouts on spine by install, sort by created date
- [x] Sort loadout cards by created date in MyLoadouts page
- [x] Improve Mod Count calculation performance for each card
- [x] Add Last Applied Date to Loadout Cards
- [x] Improved LoadoutNameProvider performance (ft @erri120)
- [x] Bumped MnDB version
- [x] Show progress spinner and Applied state on Loadout badges
- [x] Make `StatusFor` reuse the same observable for the same loadout
    - [x] Prevent Diffing logic from being run for each subscriber (every badge)
    - [x] Fix `StatusFor` handling of vanilla game state
- [x] Made switching loadouts pass through Vanilla state for now, to prevent switching issues
- [x] closes #1818 
- [x] Fix deleting active loadout
- [x] Have Badge spinner only spin when current loadout is being applied
- [x] closes #1748 
- [x] Fix adding mods not showing apply button
- [x] Make creation of first loadout apply it
- [x] Hide "Add" Button for now
- [x] Fix Spine flickering during Apply, Deletes etc..
- [x] Add tooltip to Delete button
- [x] Spine: remove separator border between home and loadout items
- [x] Hide Loadout Badge from spine if there is only one loadout per game install

See https://github.com/Nexus-Mods/NexusMods.App/issues/401#issuecomment-2242416759 for remaining work


src/NexusMods.DataModel/Synchronizer/SynchronizerState.cs
```diff
@@ -1,14 +1,21 @@
...
+public class SynchronizerState : ReactiveObject
 {
-    [Reactive] public bool Busy { get; set; } = false;
-    
+    private volatile bool _busy = false;
+
+    public bool Busy
+    {
+        get => _busy;
+        set => this.RaiseAndSetIfChanged(ref _busy, value);
```
halgari, line 16: Why did this swap to ReactiveObject?

src/NexusMods.App.UI/Controls/Spine/Spine.axaml
```diff
@@ -9,26 +9,37 @@
...
-            </ItemsControl>
-            <reactiveUi:ViewModelViewHost ViewContract="Add" x:Name="Add" />
-        </StackPanel>
-        <reactiveUi:ViewModelViewHost Grid.Row="1" x:Name="Download" />
-    </Grid>
+        </Border>
+
+        <Border x:Name="LoadoutSelectorBorder" >
+            <panels:FlexPanel x:Name="LoadoutSelector" >
+                <ScrollViewer x:Name="LoadoutSelectorSroller" >
```
Sewer56, line 25: Typo here, should say 'scroller'

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/UserControls/Spine/SpineStyles.axaml
```diff
@@ -0,0 +1,57 @@
...
+            <Style Selector="^ > Border#DownloadBorder">
+                <Setter Property="panels:Flex.Shrink" Value="0" />
+            </Style>
+            
+            <Style Selector="^ > Border#AddBorder">
+                <Setter Property="panels:Flex.Shrink" Value="0" />
+            </Style>
+        </Style>
+
+        <Style Selector="^ Border#LoadoutSelectorBorder">
```
Sewer56, line 37: Could you add a comment saying that this is for the region that hosts the loadout icons of the spine?

The name `Border` in Avalonia (and WPF for that matter) can sometimes be a misnomer, it's usually used to add an outline to a region of space like we are, however; the name `Border` itself implies it's an outline. Someone from outside of the core team could get confused, thinking that the actual border outline should be growing, when it rather instead is simply that the whole area should be shrinking.

src/NexusMods.DataModel/Synchronizer/SynchronizerState.cs
Sewer56: @Al12rs thought he had a threading issue while debugging last Thursday, so he wanted to check if making the structure volatile would have helped them.

In practice `volatile` is generally only used to:

1. Erect a memory barrier (cannot reorder operations before/after)
2. Ensure value is immediately written back to in loops

In any case, I've specified last week that `volatile` keyword here is redundant given how the variable is set and read.

I imagine the `ReactiveObject` comes from the fact they wanted to call `RaiseAndSetIfChanged` manually here since they made a private field which was volatile.

src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
```diff
@@ -114,44 +114,92 @@ public IObservable<LoadoutWithTxId> LastAppliedRevisionFor(GameInstallation game
...
+                        return LoadoutSynchronizerState.Pending;
+
+                    // Last DB revision is the same in the applied loadout
+                    if (last.Id == rev.LoadoutId && revDbTx == last.Tx)
+                        return LoadoutSynchronizerState.Current;
+
+                    if (last.Id != loadoutId)
+                        return LoadoutSynchronizerState.OtherLoadoutSynced;
+
+                    var diffFound = await Task.Run(() =>
```
Sewer56, line 187: Might be worth making a small note here.
We don't want anyone else touching this code to remove the `Task.Run`.

Technically the main thread isn't waiting for this directly, but a change to the loadout revision (e.g. disable/enable a mod) will fire a MMDB transaction which runs this via observable callback. The UI element whose receiving the result of this callback will the in turn end up waiting for the result.

So without `Task.Run`, the Main UI isn't actually stalled whatsoever, ***however*** the radio buttons which toggle a mod's enabled state will wait for completion, making the UI feel as responsive as it should be. Therefore the `Task.Run` should remain.

Sewer56 (commented):
Seems good to me, just the few notes I made.

src/NexusMods.DataModel/Synchronizer/SynchronizerState.cs
Al12rs: Right, this went through a number of iterations. Originally the Busy property was marked with the `[Reactive]` attribute without inheriting from `ReactiveObject`, which made it not notify of changes correctly.
`[Reactive]` requires in some way inheriting from `ReactiveObject`.

Due to the the value being set in a critical section, the reactive weaver wouldn't work even with `ReactiveObject`, I tried a few things and the working solution was to call `RaiseAndSetIfChanged` with a backing field.
`RaiseAndSetIfChanged` is a method from `ReactiveObject` so still inheriting from that.

I tested removing volatile, which was one of the experiments and it seems to work fine, so I'll remove that.

src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
Al12rs: My main concern here isn't threads waiting in this case, just making sure that the potentially long running operation of Diffing isn't being executed on either the UI thread or the DB thread. 

Since the operation could take a second or more and it isn't itself async (so not many opportunities to handle other tasks), whatever thread ends up executing this could stay busy for a second or more, which we want to avoid on the UI or DB thread.

Al12rs:
@Sewer56 good caches, thanks for the feedback. 
I addressed it with the commits above👍

Sewer56 (approved):

captainsandypants:
@Al12rs when there's only 1 loadout for the game  i.e. the user hasn't added an additional loadout, we should not be showing the letter A on the spine button or the "applied" tick. This stands if there's one game with multiple loadouts and another game with a single loadout, the single loadout game still shouldn't show those UI elements.

We can lose the "add" button in the spine for now.

When applying a loadout, the previously applied Loadouts spine button will flicker in and out of the UI (not a big deal for the time being).

Can we please add a tooltip to the delete button on the loadouts page with the copy "Delete Loadout". As there is currently no modal to confirm deletion we should at least make it as clear as possible what that button does.

We're missing the spinner for when the loadout gets created, either that or it's happening too fast for me to see it on my machine.

Al12rs:
@captainsandypants I'm skipping the Nexus Icon color fix, since we don't have an immediate solution. 

I addressed the following:
- [x] Hide "Add" Button for now
- [x] Fix Spine flickering during Apply, Deletes etc..
- [x] Add tooltip to Delete button
- [x] Spine: remove separator border between home and loadout items
- [x] Hide Loadout Badge if there is only one loadout per game install

Should be good to review again

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/UserControls/Spine/SpineStyles.axaml
```diff
@@ -0,0 +1,56 @@
...
+            <Setter Property="Direction" Value="Column" />
+        </Style>
+
+        <Style Selector="^ ScrollViewer#LoadoutSelectorSroller">
+            <Setter Property="VerticalScrollBarVisibility" Value="Hidden" />
+        </Style>
+
+        <Style Selector="^ ItemsControl#Loadouts">
+            <Setter Property="Margin" Value="0,12,0,0" />
+            <Setter Property="Padding" Value="0" />
```
Al12rs, line 52:
```suggestion
        <Style Selector="^ ItemsControl#Loadouts">
            <Setter Property="Padding" Value="0" />
```

Al12rs:
Passed review with design

Issue #1864, opened 2024-08-08
Mod table designs for Loadout items


captainsandypants:
## Currently
Our mod tables are only capable of showing a flat list meaning we do not have a good way of displaying different item types or showing their relationship to eachother.

## We want
A tree view in the table where mod pages can be expanded to show mod files, and mod files expanded to show mod components/sub files.

[Figma design](https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=2123-8867&t=WSClVYN4nuuKqj0c-1)

![Image](https://github.com/user-attachments/assets/56a14d02-687b-43ee-bad5-e58823259e62)

Al12rs:
@captainsandypants is there a figma example of an assembled loadout page with some rows expanded etc?
To get a better idea of how these will be used, spacing etc

captainsandypants:
> [@captainsandypants](https://github.com/captainsandypants) is there a figma example of an assembled loadout page with some rows expanded etc? To get a better idea of how these will be used, spacing etc

https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=2260-15881&t=cPqRYdQc7HYzV8NK-1

Al12rs:
> > [@captainsandypants](https://github.com/captainsandypants) is there a figma example of an assembled loadout page with some rows expanded etc? To get a better idea of how these will be used, spacing etc
> 
> https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=2260-15881&t=cPqRYdQc7HYzV8NK-1

Thank you!

PR #1872, opened 2024-08-13
Library Rework #1

erri120:
Part of #1864.


src/NexusMods.App.UI/Controls/Node.cs
```diff
@@ -0,0 +1,48 @@
...
+using ReactiveUI;
+using ReactiveUI.Fody.Helpers;
+
+namespace NexusMods.App.UI.Controls;
+
+/// <summary>
+/// Base class for models displayed in a <see cref="Avalonia.Controls.TreeDataGrid"/>.
+/// </summary>
+[PublicAPI]
+public class Node<TNode> : ReactiveObject, IDisposable
```
Al12rs, line 13: I would rename to be more descriptive, as node is a bit too generic. 
`TreeDataGridNode` maybe

src/NexusMods.App.UI/Pages/Library/LibraryViewModel.cs
```diff
@@ -0,0 +1,251 @@
...
+        {
+            this.WhenAnyValue(vm => vm.ViewHierarchical)
+                .Where(viewHierarchical => viewHierarchical && !Source.IsHierarchical)
+                .Select(viewHierarchical => CreateSource(nodes, createHierarchicalSource: viewHierarchical))
+                .BindToVM(this, vm => vm.Source)
+                .DisposeWith(disposables);
+        });
+    }
+
+    private IObservable<IChangeSet<LibraryNode>> ObserveModPages(ILocatableGame game)
```
Sewer56, line 135: Could this function be renamed when you get around to it?

It may be a leftover, but it messes with my brain, what we're really observing here seems to be `LibraryFile` instances. These are usually mods, so that's fine, but the wording `Pages` is really throwing me off, given that we're not dealing with UI pages here.

src/NexusMods.App.UI/Pages/Library/LibraryView.axaml.cs
```diff
@@ -0,0 +1,61 @@
...
+
+                    var cts = new CancellationTokenSource();
+
+                    // NOTE(erri120): this assumes that the row will not be assigned
+                    // a different model between now and when it gets detached
+                    return Observable.FromEventHandler<VisualTreeAttachmentEventArgs>(
+                        addHandler: handler => row.DetachedFromVisualTree += handler,
+                        removeHandler: handler => row.DetachedFromVisualTree -= handler,
+                        cancellationToken: cts.Token
+                    ).Select((node, cts), static (_, tuple) => tuple);
```
Al12rs, line 44: Could this be not be done with `RowClearing` at top level?
It would just copy the `RowPrepared` but call `node.Deactivate()`

erri120: Yeah, I'll get that in, thanks

Al12rs (approved):

Al12rs:
I tested this out locally, some parts are still rough but seems to work so far

PR #1875, opened 2024-08-14
Redscript diagnostics

halgari:
I want to rewrite this code a bit to give more diagnostic info (what paths are found to have what requirements, make it data driven instead of one class per dependency, etc.). But you can clearly see the dependency system in place here. In essence we're using regex matches to scan script files (pulled directly from .nx archives) to figure out what dependencies are required by what mods.

Node: none of the requirements here are hardcoded. That is to say, nothing in the system says that AMM requires Codeware. Instead we define what a codware dependency looks like and AMM matches this pattern.

NexusMods App_hOgyce6Tsk


src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/Emitters/Pattern.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    /// <summary>
+    /// A name for the required mod. Does not need to be unique across other patterns.
+    /// </summary>
+    public required string DependencyName { get; init; }
+    
+    /// <summary>
+    /// A documentation string that explains why we think the mod is required.
+    /// </summary>
+    public required string Explination { get; init; }
```
erri120, line 22: Typo

erri120 (approved):
I think this is fine for now. The cache I mentioned earlier is one improvement we can make here, and I'm not all too happy with how diagnostic details are structured yet. Might be something too look at again later.

PR #1889, opened 2024-08-19
Library and Loadout page #2

erri120:
Part of #1864.


src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridItemModel.cs
```diff
@@ -0,0 +1,205 @@
...
+
+                    if (isActivated)
+                    {
+                        serialDisposable.Disposable = model.ChildrenObservable
+                            .OnUI()
+                            .Bind(model._children)
+                            .SubscribeWithErrorLogging();
+                    }
+                })
+                .AddTo(disposables);
```
Al12rs, line 133: This breaks when the parent item is scrolled out of view, deactivating it, while the children are still in view.

The children subscription should not be disposed when the parent is deactivated.

It should only be disposed if the parent is no longer expanded, or potentially both the parent and all the children are deactivated.

Al12rs (changes requested):

halgari (approved):
I don't see anything crazy here, so LTGM.

halgari:
Ran some manual tests on this, I'm not seeing a scroll bar on the new library. I assume this will be in the next PR?

erri120:
> Ran some manual tests on this, I'm not seeing a scroll bar on the new library. I assume this will be in the next PR?

Already fixed.

Al12rs (changes requested):
Tested latest commit but Library isn't showing new mod page entries if you have the Library open and download a new mod

erri120:
> Tested latest commit but Library isn't showing new mod page entries if you have the Library open and download a new mod.

That's expected, still a WIP.

Al12rs (approved):

PR #1895, opened 2024-08-19
Postfix datamodel/config names with the release date of the next beta so we don't get so many complaints about failure to start

halgari:
Adds a constant value postfix to path names so that we don't get so many startup issues after a new release. We can remove this once we stop killing the datamodel every release


halgari:
We decided not to solve the problem this way

PR #1896, opened 2024-08-19
Update of the Polish UI

Patriot99:


src/NexusMods.App.UI/Resources/Language.pl.resx
```diff
@@ -40,13 +40,13 @@
 		<value>Później</value>
 	</data>
 	<data name="Updater_UPDATE" xml:space="preserve">
-		<value>Zaktualizuj</value>
+		<value>Zobacz aktualizację</value>
```
Sewer56, line 43: (Note: This changed in original English too, so the amendment here is correct)

Sewer56 (approved):

PR #1900, opened 2024-08-20
Copy Loadouts

### LINKED ISSUE #406: As a user, I want to clone a loadout
# User story

**As a** user

**I want to** clone a loadout

**So that** I can take an existing loadout and safely make changes to it

## Requirements
Users need to be able to take a stable loadout and clone it, so that they can safely make destructive changes.

This issue is to implement the GUI for cloning loadouts.

- Add a 'Clone' button when viewing a loadout
- On use, prompt the user to name their new loadout
  -  Default to the cloned loadouts name, with a suffix of "(COPY)"
- After successfully cloning, take the user to the new loadout and display a success message
  - E.g. "Loadout cloned" 

## Design

## DOD

## Scenario 1
**Given** _______

**When** _______

**Then** _______

halgari:
Allows users to clone loadouts. This is a shockingly fast operation. Even with 50k files the clone time is less than a second. 

*Note:* this PR also updates to the latest MnemonicDB which while it's faster also changes `ObserveDatoms` to return a `IChangeSet<Datom, DatomKey>` I fixed most of the code by using `.RemoveKey()` but we likely will want to go back later and remove these method calls and update the Observable chains to also use keyed changesets.

Example usage:

NexusMods App_9Umw9XfDmP

fixes #406

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,8 +9,8 @@
     <PackageVersion Include="LinqGen" Version="0.3.1" />
     <PackageVersion Include="Nerdbank.FullDuplexStream" Version="1.1.12" />
     <PackageVersion Include="Nerdbank.Streams" Version="2.11.74" />
-    <PackageVersion Include="NexusMods.MnemonicDB" Version="0.9.76" />
-    <PackageVersion Include="NexusMods.MnemonicDB.Abstractions" Version="0.9.76" />
+    <PackageVersion Include="NexusMods.MnemonicDB" Version="0.9.79" />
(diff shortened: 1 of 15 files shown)


erri120 (approved):

Al12rs (changes requested):

Al12rs:
@erri120 if you want we can merge this and I can quickly fix the names in a new PR so we can get this in

erri120:
> @erri120 if you want we can merge this and I can quickly fix the names in a new PR so we can get this in

Sure, this upgrades the DB, which is what I'd like to make us of.

Al12rs (approved):
Will fix the names in separate PR so we can get this in

PR #1903, opened 2024-08-20
Fix/1897 setting open page

### LINKED ISSUE #1897: Fix Open Setting Page behavior to always open on the Home workspace, but not in a new panel

Al12rs:
- closes #1897
- Remove custom page behavior for Settings page (was open in new panel, now uses default, but still opens in Home workspace).
- Fix `NavigationMenuItem` erroneously remembering last used context menu open behavior for normal clicks
- Fix PageTabIds not being restored correctly on app startup, making selected tab restore not work
- Fix nullpointer exception when opening new tab in non active workspace that wasn't opened yet in this app session
  - This required making sure that the `SelectedTab` would be updated even when the workspace and panel were not in view.


src/NexusMods.App.UI/Controls/Navigation/NavigationMenuItem.cs
```diff
@@ -41,8 +40,8 @@ public NavigationMenuItem()
 
         _contextMenuCommand = ReactiveCommand.Create<OpenPageBehaviorType>(openPageBehaviorType =>
         {
-            NavigationInformation = NavigationInformation.From(openPageBehaviorType);
-            OnClick(new RoutedEventArgs());
+            var navigationInformation = NavigationInformation.From(openPageBehaviorType);
+            NavigationCommand?.Execute(navigationInformation).Subscribe();
```
erri120, line 44: The proper solution would be to do what the `NavigationControl` does and hook into the pointer pressed/released events, but this is fine for now.

src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelViewModel.cs
```diff
@@ -65,6 +65,11 @@ public PanelViewModel(IWorkspaceController workspaceController, PageFactoryContr
             .Bind(out _tabs)
             .Subscribe();
 
+        this.WhenAnyValue(vm => vm.SelectedTabId)
+            .Select(selectedTabId => Tabs.FirstOrDefault(tab => tab.Id == selectedTabId))
+            .WhereNotNull()
+            .SubscribeWithErrorLogging(selectedTab => SelectedTab = selectedTab);
```
erri120, line 71: Not a fan of this. This isn't fixing the underlying issue.

erri120 (changes requested):

src/NexusMods.App.UI/WorkspaceSystem/Panel/PanelViewModel.cs
Al12rs: Which would be that it can still be null you mean? Or what problem are you referring to?

src/NexusMods.App.UI/Controls/Navigation/NavigationMenuItem.cs
Al12rs: Btw, should I be disposing that subscription?

PR #1913, opened 2024-08-21
Library and Loadout view styling 

Al12rs:
This isn't complete but has the view elements to then bind commands and stuff to

### DIFF
--- a/src/NexusMods.App.UI/Pages/LibraryPage/LibraryItemModel.cs
+++ b/src/NexusMods.App.UI/Pages/LibraryPage/LibraryItemModel.cs
@@ -1,6 +1,7 @@
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Reactive.Linq;
+using Avalonia.Controls;
 using Avalonia.Controls.Models.TreeDataGrid;
 using DynamicData;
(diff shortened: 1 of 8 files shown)


src/NexusMods.App.UI/Pages/LibraryPage/LibraryItemModel.cs
```diff
@@ -118,7 +120,8 @@ public static IColumn<LibraryItemModel> CreateNameColumn()
                 IsTextSearchEnabled = true,
                 CanUserResizeColumn = true,
                 CanUserSortColumn = true,
-            }
+            },
+            width: GridLength.Auto
```
erri120, line 124: Default is `Auto`, so we don't need this.

erri120 (approved):

src/NexusMods.App.UI/Pages/LibraryPage/LibraryItemModel.cs
Al12rs: Yeah, I was experimenting with star, which had the usual issue of column collapsing on itself. I'm going to investigate treeDataGrid sizing this weekend, there is definitely a bug there to fix

PR #1914, opened 2024-08-21
Update Library and Loadout functionality

erri120:
Binds the buttons of the toolbar to commands. Almost all commands have been implemented.

Part of #1864.


Al12rs (approved):
Yeah removing items from the Loadout is weird and even after removing stuff is remaining behind, but for now it is good enough.

The tests are failing on CLI `download-and-install-mod` and `install-mod` verbs, not sure if this is due to the changes in this PR or previous prs, but the verbs have not been updated to use the new pipelines Library and loadout items

erri120:
> Yeah removing items from the Loadout is weird and even after removing stuff is remaining behind, but for now it is good enough.

There's a TODO for it in the code. Haven't optimized the queries yet. The loadout view will list all mod pages, even if they don't have any files installed.

PR #1918, opened 2024-08-21
Fix some of the CLI commands and fix logic with the mod installer fallback to advanced installer

halgari:
Fixes a regression caused by incorrect boolean logic around how we fallback to advanced installers. It's a 3-state situation, not a 2 state

### DIFF
--- a/src/NexusMods.Library/InstallLoadoutItemJobWorker.cs
+++ b/src/NexusMods.Library/InstallLoadoutItemJobWorker.cs
@@ -34,8 +34,11 @@ internal class InstallLoadoutItemJobWorker : AJobWorker<InstallLoadoutItemJob>
 
         var result = await ExecuteInstallersAsync(job, installers, cancellationToken);
 
-        if (!result.HasValue && job.Installer is not AdvancedManualInstaller)
+        if (!result.HasValue)
         {
+            if (job.Installer is not AdvancedManualInstaller)
+                return JobResult.CreateFailed($"Advanced installer did not succeed for `{job.LibraryItem.Name}` (`{job.LibraryItem.Id}`)");
+            
             var manualInstaller = AdvancedManualInstaller.Create(_serviceProvider);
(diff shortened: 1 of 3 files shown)


src/NexusMods.Library/InstallLoadoutItemJobWorker.cs
```diff
@@ -34,8 +34,11 @@ protected override async Task<JobResult> ExecuteAsync(InstallLoadoutItemJob job,
 
         var result = await ExecuteInstallersAsync(job, installers, cancellationToken);
 
-        if (!result.HasValue && job.Installer is not AdvancedManualInstaller)
+        if (!result.HasValue)
         {
+            if (job.Installer is not AdvancedManualInstaller)
+                return JobResult.CreateFailed($"Advanced installer did not succeed for `{job.LibraryItem.Name}` (`{job.LibraryItem.Id}`)");
```
erri120, line 40:
```suggestion
            if (job.Installer is AdvancedManualInstaller)
                return JobResult.CreateFailed($"Advanced installer did not succeed for `{job.LibraryItem.Name}` (`{job.LibraryItem.Id}`)");
```

erri120 (changes requested):

src/NexusMods.Library/InstallLoadoutItemJobWorker.cs
erri120: If the provided installer was the advanced installer, then we don't want to run it again if it failed/user cancelled.

Al12rs, line 40: This doesn't sound right though, if the last installer is _not_ the Advanced one and the result is empty we should attempt using the Advanced installer.

Al12rs (changes requested):

src/NexusMods.Library/InstallLoadoutItemJobWorker.cs
Al12rs: erri beat me to it

PR #1926, opened 2024-08-22
Fix CleanDirectories deleting folders still containing other files

### LINKED ISSUE #1907: Replacing file deletes root folder of the path to it relative to LocationId
# Bug Report

## Summary
Since transitioning to Library Items when replacing an existing file Apply nukes the folder structure. Granted I only manage Pacific Drive with the App support for which is in my fork, #1747 so it's entirely possible my own code is at fault.

`{Game}` is `H:\SteamLibrary\steamapps\common\Pacific Drive`, `{GameMainUE}` is `H:\SteamLibrary\steamapps\common\Pacific Drive\PenDriverPro`, and the file game path is `{GameMainUE}\Plugins\DLSS\Binaries\ThirdParty\Win64\nvngx_dlss.dll`. I have gone over action groupings, and that part works as expected: the file gets backed up, deleted (just the file), then the new file gets extracted and then at some later step the entire relative root of the path to it gets nuked (`{GameMainUE}\Plugins`).

## Steps to reproduce
Manage `Pacific Drive`
Add [dlss DLL zip](https://www.techpowerup.com/download/nvidia-dlss-dll/) to library as `From Drive`. 
Enable the "mod"
Apply it
`{GameMainUE}\Plugins` disappears entirely

dll gets replaced with a new one

## Other information
Diff appears to be correct:
image
Unmanaging the game only puts back the backed up dll that was replaced, all the other files in Plugins remain gone.
[nexusmods.app.main.current.log](https://github.com/user-attachments/files/16677100/nexusmods.app.main.current.log)

### LINKED ISSUE #1925: Synchronize sometimes wrongly deletes folders that still contain files during `CleanDirectories` step
This is triggered when at least one file deletion happened during Apply, potentially deleting the parent folders even if they still had files.

This could be cause for #1907

Al12rs:
- fixes #1925
- Could maybe fix #1907  

To explain the fix, the function created a hashset of existing file paths and was using `hashSet.Contains(parentPath)` to check whether a parent folder had any children, this always returned false, since the hashset contained the file paths, not parent paths. This change builds a hashset of all the parent paths, so the check works correctly.

Probably not the most efficient way to do the entire directory check but this is the simplest fix.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -85,7 +85,7 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
         var seenDirectories = new HashSet<GamePath>();
         var directoriesToDelete = new HashSet<GamePath>();
 
-        var newStatePaths = newState.Select(e => (GamePath)e.Path).ToHashSet();
+        var newStatePaths = newState.SelectMany(e => ((GamePath)e.Path).GetAllParents()).ToHashSet();
 
         foreach (var entry in toDelete)
         {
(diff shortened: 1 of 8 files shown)


halgari:
Can we get tests for this in either `GeneralLoadoutManagmeentTests` or copy that class and make a new test class to containt these bug fixes?

ZeeWanderer:
Just checked, it does fix #1907

halgari:
I wrote the tests for this, and they almost instantly broke because we don't have DI isolation in our tests, so I implemented that via the new `AIsolatedGameTest` class. Works pretty well, we can swap our tests slowly over to this new system as we work through other issues.

halgari (approved):

PR #1929, opened 2024-08-22
Add `TreeDataGridAdapter`

erri120:
Consolidation and cleanup of pages with `TreeDataGrid` related functionality. This keeps the pages focused and slim.

Part of #1864.

### DIFF
--- a/src/NexusMods.App.UI/Controls/ReactiveR3Object.cs
+++ b/src/NexusMods.App.UI/Controls/ReactiveR3Object.cs
@@ -1,0 +1,34 @@
+using R3;
+
+namespace NexusMods.App.UI.Controls;
+
+public class ReactiveR3Object : ReactiveUI.ReactiveObject, IDisposable
+{
+    private readonly BehaviorSubject<bool> _activation = new(initialValue: false);
+    public Observable<bool> Activation => _activation;
+
+    internal void Activate() => _activation.OnNext(true);
+    internal void Deactivate() => _activation.OnNext(false);
+
+    /// <inheritdoc/>
+    public void Dispose()
+    {
+        Dispose(disposing: true);
+        GC.SuppressFinalize(this);
+    }
+
+    private bool _isDisposed;
+
+    /// <inheritdoc cref="IDisposable.Dispose"/>
+    protected virtual void Dispose(bool disposing)
+    {
+        if (_isDisposed) return;
+
+        if (disposing)
+        {
+            _activation.Dispose();
+        }
+
+        _isDisposed = true;
+    }
+}
--- a/src/NexusMods.App.UI/Controls/ReactiveR3Object.cs
+++ b/src/NexusMods.App.UI/Controls/ReactiveR3Object.cs
@@ -1,8 +1,15 @@
+using System.ComponentModel;
+using JetBrains.Annotations;
 using R3;
 
 namespace NexusMods.App.UI.Controls;
 
-public class ReactiveR3Object : ReactiveUI.ReactiveObject, IDisposable
+/// <summary>
+/// Base class using R3 with support for activation/deactivation,
+/// <see cref="INotifyPropertyChanged"/>, and <see cref="INotifyPropertyChanging"/>.
+/// </summary>
+[PublicAPI]
+public class ReactiveR3Object : ReactiveUI.IReactiveObject, IDisposable
 {
     private readonly BehaviorSubject<bool> _activation = new(initialValue: false);
     public Observable<bool> Activation => _activation;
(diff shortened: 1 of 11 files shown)


src/NexusMods.App.UI/Extensions/R3Extensions.cs
```diff
@@ -8,6 +9,37 @@ namespace NexusMods.App.UI.Extensions;
 [PublicAPI]
 public static class R3Extensions
 {
+    [MustDisposeResource] public static IDisposable WhenActivated<T>(
+        this T obj,
+        Action<T, CompositeDisposable> block)
+        where T : ReactiveR3Object
```
Al12rs, line 15: Just for curiosity, what is the reason to keep this separate from `ReactiveR3Object`?

erri120: What do you mean with "keep it separate"?

Al12rs: Why is this an extension method rather than being defined directly on ReactiveR3Object?

src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
```diff
@@ -0,0 +1,148 @@
...
+using Avalonia.Controls.Models.TreeDataGrid;
+using Avalonia.Controls.Selection;
+using DynamicData;
+using NexusMods.App.UI.Extensions;
+using ObservableCollections;
+using R3;
+
+namespace NexusMods.App.UI.Controls;
+
+public abstract class TreeDataGridAdapter<TModel> : ReactiveR3Object
```
Al12rs, line 11: Since this is new stuff, we should add comments to the new classes, to give some idea on their purpose and when they should be used.

src/NexusMods.App.UI/Controls/ReactiveR3Object.cs
```diff
@@ -0,0 +1,34 @@
+using R3;
+
+namespace NexusMods.App.UI.Controls;
+
+public class ReactiveR3Object : ReactiveUI.ReactiveObject, IDisposable
```
Al12rs, line 5: Add comment for purpose and when to be used.

src/NexusMods.App.UI/Extensions/R3Extensions.cs
```diff
@@ -8,6 +9,37 @@ namespace NexusMods.App.UI.Extensions;
 [PublicAPI]
 public static class R3Extensions
 {
+    [MustDisposeResource] public static IDisposable WhenActivated<T>(
```
Al12rs, line 12: Add comment for purpose

src/NexusMods.App.UI/Pages/LibraryPage/LibraryView.axaml.cs
```diff
@@ -55,13 +55,13 @@ public LibraryView()
             deactivate.Merge(activate)
                 .Where(static tuple => tuple.Model is LibraryItemModel)
                 .Select(static tuple => ((tuple.Model as LibraryItemModel)!, tuple.Item2))
-                .Subscribe(this, static (tuple, view) => view.ViewModel!.ActivationSubject.OnNext(tuple))
+                .Subscribe(this, static (tuple, view) => view.ViewModel!.Adapter.ModelActivationSubject.OnNext(tuple))
                 .AddTo(disposables);
```
Al12rs, line 59: Could this and the lines above (from L45), be moved into a method on the Adapter? 
It is repeated code on every view using the new TreeDataGrid with the Adapter

src/NexusMods.App.UI/Pages/LibraryPage/LibraryViewModel.cs
```diff
@@ -155,57 +139,28 @@ public LibraryViewModel(
...
                     {
-                        model.Ticker = null;
                         var didRemove = vm.InstallCommandDisposables.Remove(model, out var disposable);
                         Debug.Assert(didRemove, "subscription for the model should exist");
                         disposable?.Dispose();
-
-                        model.Deactivate();
                     }
                 })
                 .AddTo(disposables);
```
Al12rs, line 162: I don't particularly like this dictionary of disposables that we have to maintain for subscribing on child entries commands. It seems an overengineered solution.

I do agree that it is preferrable to deal with the commands from the parent VM. 
Maybe we can come up with a better approach

src/NexusMods.App.UI/Extensions/R3Extensions.cs
erri120: `this` is only available in static classes. Making this method an extension method means you can use `this.WhenActivated`. If I put this method inside `ReactiveR3Object` you have to use `WhenActivated(this)`. Slightly different usage, but I think `this.WhenActivated` is easier to adapt.

src/NexusMods.App.UI/Pages/LibraryPage/LibraryViewModel.cs
```diff
@@ -283,42 +238,50 @@ await Parallel.ForAsync(
...
+
+    protected override void BeforeModelActivationHook(LibraryItemModel model)
+    {
+        model.Ticker = _ticker;
+    }
+
+    protected override void BeforeModelDeactivationHook(LibraryItemModel model)
+    {
+        model.Ticker = null;
+    }
```
Al12rs, line 262: I know parent base methods are empty, but good practice would be to call the base methods, right? In case those are updated to do something in the future.

Al12rs (changes requested):
Some minor things above like adding more comments and some questions.

src/NexusMods.App.UI/Extensions/R3Extensions.cs
Al12rs: Ah, I see, makes sense. Thanks for explaining, I appreciate it!

Al12rs (approved):

PR #1931, opened 2024-08-23
Parallel Tests

halgari:
This reworks how DI is handled in our test suite. Now the tests that require a ton of local state get their own `IHost` that is constructed in a constructor and is therefore an isolated DI container. In previous work, we already had code for adding a manual game on-demand as part of a test, and for setting the paths of this test to isolated folders. That combined with this PR allows us to run tests seprate from eachother and therefore tehy can all run in parallel. 

This also kills the flakey View tests that were legacy code anyways. 

Had to do some troubleshooting to get get some of the tests to be fully isolated, but it should be working, at least it works on my box.


Sewer56 (approved):
No objections.
I saw a bit of this before I fell asleep yesterday and re-checked just now.

It's using the little class scoped DI trick I found when working on GC for isolation, good to me.

Issue #1924, opened 2024-08-22
Synchronize does not save changes to existing files


Al12rs:
If an existing file is modified, the change seems to be discarded on successive Synchronize operations. 
It seems like the `IngestFile` operation is performed, and a new LoadoutItem is created for the new version of the file, but then the new version is added to the mod, without removing the previous LoadoutItem version, resulting in two LoadoutItems with different contents pointing to the same file inside the same mod (either override or SMAPI mod where the changes have been placed).

PR #1932, opened 2024-08-23
Add Synchronizer GeneralModManagementTests

Al12rs:
Follow-up on 
- #1926

since it was already merged.

Added a test for adding and removing mods, using verify. 
Plan would be to add more tests in the future, like adding, removing and changing files, after #1924 is fixed for example.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -85,7 +85,7 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
         var seenDirectories = new HashSet<GamePath>();
         var directoriesToDelete = new HashSet<GamePath>();
 
-        var newStatePaths = newState.Select(e => (GamePath)e.Path).ToHashSet();
+        var newStatePaths = newState.SelectMany(e => ((GamePath)e.Path).GetAllParents()).ToHashSet();
 
         foreach (var entry in toDelete)
         {
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -85,7 +85,12 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
         var seenDirectories = new HashSet<GamePath>();
         var directoriesToDelete = new HashSet<GamePath>();
 
-        var newStatePaths = newState.SelectMany(e => ((GamePath)e.Path).GetAllParents().Skip(1)).ToHashSet();
+        var newStatePaths = newState.SelectMany(e =>
+            {
+                // We need folder paths, so skip first path as it is the file path itself
+                return ((GamePath)e.Path).GetAllParents().Skip(1);
+            }
+        ).ToHashSet();
 
         foreach (var entry in toDelete)
         {
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -85,7 +85,7 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
         var seenDirectories = new HashSet<GamePath>();
         var directoriesToDelete = new HashSet<GamePath>();
 
-        var newStatePaths = newState.SelectMany(e => ((GamePath)e.Path).GetAllParents()).ToHashSet();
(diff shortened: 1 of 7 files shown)


src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -85,7 +85,7 @@ private void CleanDirectories(IEnumerable<GamePath> toDelete, DiskState newState
         var seenDirectories = new HashSet<GamePath>();
         var directoriesToDelete = new HashSet<GamePath>();
 
-        var newStatePaths = newState.SelectMany(e => ((GamePath)e.Path).GetAllParents()).ToHashSet();
+        var newStatePaths = newState.SelectMany(e => ((GamePath)e.Path).GetAllParents().Skip(1)).ToHashSet();
```
erri120, line 88: Why are we skipping 1?

Al12rs: The first one is the full file path, which we don't need since we only care about the folders.

tests/Games/NexusMods.Games.TestFramework/AIsolatedGameTest.cs
```diff
@@ -377,6 +379,49 @@ private AbsolutePath GetArchivePath(Hash hash)
         store.TryGetLocation(Connection.Db, hash, null, out var archivePath, out _).Should().BeTrue("Archive should exist");
         return archivePath;
     }
+    
+            /// <summary>
+    /// Uses Verify to validate all the states of the game, previously applied states, etc.
+    /// </summary>
+    protected void LogDiskState(StringBuilder sb, string sectionName, string comments = "", Loadout.ReadOnly[]? loadouts = null)
```
erri120, line 386: Somewhat misleading description, this isn't using Verify at all, this method just prints the Loadouts as tables into the string builder.

tests/Games/NexusMods.Games.TestFramework/AIsolatedGameTest.cs
```diff
@@ -377,6 +379,49 @@ private AbsolutePath GetArchivePath(Hash hash)
...
+        var metadata = GameInstallation.GetMetadata(Connection);
+        sb.AppendLine($"{sectionName}:");
+        if (!string.IsNullOrEmpty(comments))
+            sb.AppendLine(comments);
+        
+        Section("### Initial State", metadata.InitialDiskStateTransaction);
+        if (metadata.Contains(GameInstallMetadata.LastSyncedLoadoutTransaction)) 
+            Section("### Last Synced State", metadata.LastSyncedLoadoutTransaction);
+        Section("### Current State", metadata.LastScannedDiskStateTransaction);
+        foreach (var loadout in loadouts ?? [])
```
erri120, line 399: When using Verify, it's important that the output is consistent. If the sorting is indeterminate, you might get inconsistent outputs. I'd suggest sorting by name or something.

tests/Games/NexusMods.Games.TestFramework/AIsolatedGameTest.cs
```diff
@@ -377,6 +379,49 @@ private AbsolutePath GetArchivePath(Hash hash)
...
+            if (!loadout.Items.Any())
+                continue;
+
+            var files = loadout.Items.OfTypeLoadoutItemWithTargetPath().OfTypeLoadoutFile()
+                .Where(item=> item.AsLoadoutItemWithTargetPath().AsLoadoutItem().IsEnabled()).ToArray();
+            
+            sb.AppendLine($"### Loadout {loadout.ShortName} - ({files.Length})");
+            sb.AppendLine("| Path | Hash | Size | TxId |");
+            sb.AppendLine("| --- | --- | --- | --- |");
+            foreach (var entry in files)
```
erri120, line 410: Same here, should be sorted.

tests/Games/NexusMods.Games.TestFramework/AIsolatedGameTest.cs
```diff
@@ -377,6 +379,49 @@ private AbsolutePath GetArchivePath(Hash hash)
...
+        }
+        sb.AppendLine("\n\n");
+        
+        void Section(string sectionName, Transaction.ReadOnly asOf)
+        {
+            var entries = metadata.DiskStateAsOf(asOf);
+            sb.AppendLine($"{sectionName} - ({entries.Count}) - {TxId.From(asOf.Id.Value)}");
+            sb.AppendLine("| Path | Hash | Size | TxId |");
+            sb.AppendLine("| --- | --- | --- | --- |");
+            foreach (var entry in entries)
```
erri120, line 421: Same here, should be sorted.

erri120 (changes requested):

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
erri120: You should put a comment in there.

erri120 (approved):

PR #1937, opened 2024-08-26
Don't manage the game twice

### LINKED ISSUE #1879: Stop automatically opening the game workspace when a game has been added
## Currently

When on the home workspace and adding a game, when that game has finished being added. the UI changes to open that games workspace.

This is unexpected and stops the user from learning how to navigate between workspaces themselves.

## We want

To stop changing the workspace and allow the user to navigate to the game workspace themselves.

### LINKED ISSUE #1928: Prevent Managing Games multiple times
It is possible to manage the same game multiple times by closing and reopening My Games page, as the loading state is reset.

Unmanaging a game has the same issue. Also continuously displays game removing, doesn't refresh to show the game was removed without changing pages.

erri120:
Fixes #1928, fixes #1879.

This just adds a job and a check, it doesn't change the UI.


src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
```diff
@@ -127,28 +127,32 @@ private async Task RemoveAllLoadouts(GameInstallation install)
...
-            }
-        );
+                    },
+                },
+                () => new LoadoutContext
+                {
+                    LoadoutId = install.LoadoutId,
+                }
+            );
+        });
```
Al12rs, line 155: You can remove this, see 
- #1879

Al12rs (commented):
It is still a bit broken since the UI isn't showing the Managing state while it is creating the loadout, if you move away and back. Like you can delete the loadout that doesn't even exist, and we are not showing any UI feedback that the app is still working on managing the game.

To fix that we would need to set the GameWidget VM.State to `AddingGame` when we detect the job to be running. Problem is that creating a Loadout and managing a game are the same operation, which we might just want to change.

Al12rs (changes requested):
image

After adding the game, state seems to be stuck in `AddingGame`

src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
```diff
@@ -41,114 +45,105 @@ public MyGamesViewModel(
...
-            }
-            
+                        if (GetJobRunningForGameInstallation(install).IsT1) return;
+
+                        vm.State = GameWidgetState.AddingGame;
+                        await Task.Run(async () => await ManageGame(install));
+                        vm.State = GameWidgetState.ManagedGame;
+                    });
+
+                    vm.State = ToState(GetJobRunningForGameInstallation(install), GameWidgetState.DetectedGame);
```
Al12rs, line 108: This should ignore `AddingGame`

src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
```diff
@@ -41,114 +45,105 @@ public MyGamesViewModel(
...
-                gameRegistry.InstalledGames
-                    .ToObservableChangeSet()
-                    .Except(_managedGames.ToObservableChangeSet().Transform(s => s.Installation))
-                    .OnUI()
-                    .Transform(install =>
+                    });
+
+                    vm.ViewGameCommand = ReactiveCommand.Create(() => { NavigateToLoadout(conn, loadout); });
+
+                    vm.State = ToState(GetJobRunningForGameInstallation(loadout.InstallationInstance), GameWidgetState.ManagedGame);
```
Al12rs, line 81: This should ignore `AddingGame`, if it is here (managed games) it means the transaction adding the game was already committed, and there is not going to be anything removing the state.

Al12rs (approved):
Yee it works

PR #1938, opened 2024-08-26
Rework downloader

### LINKED ISSUE #1927: Insane heap allocations when downloading
Running the project through Rider in Debug mode, with the Debugger attached, will slow performance down drastically. When downloading, the entire UI is frozen if debugger is attached.

This isn't related to exceptions being swallowed somewhere, I enabled every exception to trigger a breakpoint, and we only have one looping excepting in the `CliServer` by design.

It'd be interesting to know if debugging with Rider on Windows, and Visual Studio on Windows also have massive performance penalties, or if this is only a Rider on Linux issue. If this issue appears on any configuration of platform and IDE, then we have some issues.

erri120:
Resolves #1927.

Replaces the downloader with a resilient `HttpClient` that supports resuming using ranges.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -3,10 +3,10 @@
     <PackageVersion Include="Avalonia.AvaloniaEdit" Version="11.1.0" />
     <PackageVersion Include="Avalonia.Labs.Panels" Version="11.1.0" />
     <PackageVersion Include="AvaloniaEdit.TextMate" Version="11.1.0" />
-    <PackageVersion Include="Downloader" Version="3.1.2" />
     <PackageVersion Include="FlatSharp.Compiler" Version="7.6.0" />
     <PackageVersion Include="FlatSharp.Runtime" Version="7.6.0" />
     <PackageVersion Include="LinqGen" Version="0.3.1" />
+    <PackageVersion Include="Microsoft.Extensions.Http.Resilience" Version="8.8.0" />
     <PackageVersion Include="Nerdbank.FullDuplexStream" Version="1.1.12" />
     <PackageVersion Include="Nerdbank.Streams" Version="2.11.74" />
     <PackageVersion Include="NexusMods.Paths" Version="0.10.0" />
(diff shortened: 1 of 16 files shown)


halgari:
I haven't compared this to the old code we had, but I feel like this is going to quickly reach the complexity of our old code. I'm all for Polly (I'm a huge fan of chaos engineering), but we might be able to add it to the older code instead of re-implementing the downloader from scratch.

erri120:
> I haven't compared this to the old code we had, but I feel like this is going to quickly reach the complexity of our old code. I'm all for Polly (I'm a huge fan of chaos engineering), but we might be able to add it to the older code instead of re-implementing the downloader from scratch.

The new code behaves very different from the old `AdvancedHttpDownloader`. Besides not using Polly, and not doing range requests properly, the complexity in the old code came from downloading chunks in parallel from multiple sources. As discussed before, we don't need that functionality.

The complexity in the new code is in handling range requests properly, something that the old code didn't do (it had range requests, they just weren't done properly).

halgari:
I'd really love to have the parallel downloads howevera as it makes a major difference on my box, and was a somewhat major part of the design for that downloader. I'm just not sure why we're going down this route now. We originally went with a different downloader to save on code complexity by replacing our code with another library. That library ended up having performance issues, but now we're rewriting this code again when the previous code worked just fine.

Sewer56:
> I'd really love to have the parallel downloads howevera as it makes a major difference on my box, and was a somewhat major part of the design for that downloader. I'm just not sure why we're going down this route now. We originally went with a different downloader to save on code complexity by replacing our code with another library. That library ended up having performance issues, but now we're rewriting this code again when the previous code worked just fine.

I don't understand either; sure the old code needed some cleaning up, it wasn't written in the most ideal way, but it was functional.

I just didn't comment on it; I would have felt bad asking them to drop it when a bunch of effort was already put in place.

halgari (approved):

PR #1940, opened 2024-08-26
Remove IModInstaller.cs and associated code

halgari:
* Removes IModInstaller and associated uses
* Adds a extension method for `ILibraryService` that will in one line add a file to the library and add it to the loadout. The Job system is fairly verbose, it'd be nice sometime to make it easier to do 1 line job invocations like this.


src/Abstractions/NexusMods.Abstractions.Library/ILibraryServiceExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+        await installJob.WaitToFinishAsync(token);
+        
+        if (!installJob.Result!.TryGetCompleted(out var installCompleted))
+            throw new Exception("Failed to install the file");
+        
+        if (!installCompleted.TryGetData<LoadoutItemGroup.ReadOnly>(out var loadoutItemGroup))
+            throw new Exception("Failed to install the file");
+        
+        return loadoutItemGroup;
+    }
```
erri120, line 41: Instead of having this rather specialized extension method, if you want to cut down on the verbosity of the job system, we can add a method that starts, waits, and gets the result.

erri120 (dismissed):

src/Abstractions/NexusMods.Abstractions.Library/ILibraryServiceExtensions.cs
halgari: Split this into two methods

erri120 (approved):

Issue #1087, opened 2024-03-20
Bug: Vertically shrinking the App in a quick motion causes it to crash. 


Sewer56:
# Bug Report

## Summary

Vertically shrinking the App in a quick motion causes it to crash.

https://github.com/Nexus-Mods/NexusMods.App/assets/6697380/f5f53c8b-41f0-4cbb-8e9d-86a629c2a9ee

## Steps to reproduce

On Linux, with Hyprland.

1. Run the App
2. Shrink the App Vertically (very fast)
3. App Crashes

## What is the expected behaviour?

The App does not crash.

## Other information

This is an upstream Avalonia issue that replicates on any App (ours included), will open an issue on Avalonia repo soon.

The specifics of what causes this are currently unclear, i.e.
- Is it X11/Wayland Specific?
- Is it Window Manager Specific?

The App seems to crash when the height is roughly less than the height of the titlebar on a Windows system.

Al12rs:
We probably also need Minimum MainWindow sizes to avoid allowing users to reduce the size to unusable or invisible amounts.

Issue #1109, opened 2024-03-21
Add `MinWidth` and `MinHeight` to `MainWindow`


Al12rs:
# Improvement

## Currently 
It is possible to arbitrarily resize the mainwindow to nonsensical amounts, literally 0 height and width. This should be avoided.

## We want
Sensible minimum height and width that make working with the App ui possible without restricting the resizing options for the users.
Specifically it should be possible to snap the UI to half a 1080p screen, both horizontally and vertically (effectively up to 1/4 of 1080p area)

erri120:
As discussed previous on the Discord, using a minimum _area_ is better than a fixed minimum _resolution_. The difference between those two is that an area doesn't care whether the width is larger than the height or vise versa.

The Steam Deck has a resolution of `1280x800` which is an area of `1024000` pixel squared. Using this value as a minimum area, we can calculate the minimum width or height based on the other value.

As an example, if the user were to split a 1080p screen vertically in half, the two halves would have a width of `960` pixels each. Using the area calculation, we can get the minimum height for this scenario: `1024000 / 960 = 1067`.

`1280x800` and `960x1067` both share the same area, but they are completely different in terms of orientation. This also shows that we can't have fixed `MinWidth` or `MinHeight` values. A min height of `800` might be okay on a horizontal layout, but it's not enough for a vertical layout. Similarly, a min width of `1280` isn't even possible to achieve on a vertical 1080p monitor.

PR #1947, opened 2024-08-27
Added: Minimum Size to Main Window

Sewer56:
This adds a minimum size requirement to the window, which I've set approximately to the following:

20240827_17h55m35s_Nexus_Mods_App

The main motivation here is adding a temporary workaround for 
- #1087 

And to be consistent with other windows which have minimum sizes.

Though, after merging, I'm not sure if we should close the original issue.
I think it may be worthwhile to do a few more quick tests and report the bug upstream to Avalonia.


Al12rs (changes requested):
Please see our notes on minimum size as discussed here:
https://github.com/Nexus-Mods/NexusMods.App/issues/1109#issuecomment-2011695339

src/NexusMods.App.UI/Windows/MainWindow.axaml
```diff
@@ -3,6 +3,8 @@
     Title="Nexus Mods App"
     d:DesignHeight="450"
     d:DesignWidth="800"
+    MinHeight="380"
+    MinWidth="1030"
```
Al12rs, line 7:
```suggestion
    MinHeight="360"
    MinWidth="360"
```
Half of 720 to allow snapping to half size, independently from orientation.

Al12rs (changes requested):

Al12rs (approved):

PR #1960, opened 2024-08-29
change markdown hyperlink styles

insomnious:
I've been using [this](https://github.com/whistyun/Markdown.Avalonia/blob/master/Markdown.Avalonia.Tight/StyleCollections/MarkdownStyleDefaultTheme.axaml) as a reference for styles as official docs on the wiki haven't been updated since 2020 and I'm not convinced it still works in the same way. This now looks how @captainsandypants wants it but we will need a refinement task to add\tidy up the styles so the changelog (for instance) looks as good as it can due to lots of little spacing issues.

Honeycam 2024-08-29 16-00-47


insomnious:
image

Added Roboto Mono font to match Figma

erri120 (approved):

captainsandypants (approved):

PR #1967, opened 2024-09-02
Add ViewModFiles implementation to new Loadout View

Al12rs:
This will only show the option if the selection can resolve in a single `LoadoutItemGroup` of files, and then render only that group.
E.g. if a root entry with two children is selected, the ViewModFiles option will remain disabled, but if it has a single child, then it will work.

Some specific SMPI mod are internally divided into multiple modules containing files, these when selected will not show the option since there is no single group of files to show.

This PR also has some light refactoring and renaming of the ViewModFilesPage.

This only works for the new Loadout View, not the old LoadoutGrid.


src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutViewModel.cs
```diff
@@ -98,6 +128,56 @@ public LoadoutViewModel(IWindowManager windowManager, IServiceProvider servicePr
...
+
+    
+    /// <summary>
+    /// Returns the appropriate LoadoutItemGroup of files if the selection contains a LoadoutItemGroup containing files,
+    /// if the selection contains multiple LoadoutItemGroups of files, returns None.
+    /// </summary>
+    private async Task<Optional<LoadoutItemGroup.ReadOnly>> GetViewModFilesLoadoutItemGroup(CancellationToken cancellationToken)
+    {
+        return await Task.Run(() =>
+            {
```
erri120, line 140: Remove this `Task.Run`.

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutViewModel.cs
```diff
@@ -98,6 +128,56 @@ public LoadoutViewModel(IWindowManager windowManager, IServiceProvider servicePr
...
+                        }
+                        default:
+                            return children.First().IsLoadoutItemWithTargetPath() ? currentGroup : Optional<LoadoutItemGroup.ReadOnly>.None;
+                    }
+                }
+
+                return Optional<LoadoutItemGroup.ReadOnly>.None;
+            },
+            cancellationToken: cancellationToken
+        );
```
erri120, line 179: This can be replaced with the code used in the `LoadoutDataProviderHelper`. That code is much simpler and faster.

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutViewModel.cs
```diff
@@ -46,10 +50,36 @@ public LoadoutViewModel(IWindowManager windowManager, IServiceProvider servicePr
...
+
+                    var group = await GetViewModFilesLoadoutItemGroup(cancellationToken);
+                    return group.HasValue;
+                },
+                AwaitOperation.Drop
+            )
+            .ToReactiveCommand<NavigationInformation>(async (info, cancellationToken) =>
+                {
+                    var group = await GetViewModFilesLoadoutItemGroup(cancellationToken);
+                    if (!group.HasValue) return;
```
erri120, line 67: I don't like this approach, as we're doing the same thing twice. Instead, have a subscription to the selected items, that calculates whatever we need. The can-execute observable would be a second observable that checks whether we have a calculated value or not.

erri120 (changes requested):

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutViewModel.cs
```diff
@@ -112,32 +162,34 @@ public class LoadoutTreeDataGridAdapter : TreeDataGridAdapter<LoadoutItemModel,
...
-        {
-            Disposable.Create(adapter._commandDisposables, static commandDisposables =>
             {
-                foreach (var kv in commandDisposables)
-                {
-                    var (_, disposable) = kv;
-                    disposable.Dispose();
-                }
+                Disposable.Create(adapter._commandDisposables,
+                        static commandDisposables =>
```
erri120, line 174: What the hell happened with the indentation.

erri120 (approved):

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutViewModel.cs
Al12rs: The auto formatter must have kicked in when I moved the function.
Yeah, it wants to split it on two lines for some reason.

PR #1985, opened 2024-09-05
Remaining functionality for new library and loadout pages

erri120:
- adds the remove button to the new library page
- adds a filter to the new library and loadout pages
- fixed an issue with row selection

TODO:
- [x] filter by game in the library
- [x] created at for mod pages should show file created at

### DIFF
--- a/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridItemModel.cs
+++ b/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridItemModel.cs
@@ -1,9 +1,9 @@
+using System.Collections.ObjectModel;
 using System.Diagnostics;
 using Avalonia.Controls.Models.TreeDataGrid;
 using DynamicData;
 using JetBrains.Annotations;
 using NexusMods.App.UI.Extensions;
-using ObservableCollections;
 using R3;
 using Observable = System.Reactive.Linq.Observable;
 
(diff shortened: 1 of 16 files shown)


Al12rs (commented):
The parent here in the Loadout view should be using the most recent `Installed` value from the children:
image

Al12rs (commented):
This is also very strange:

Added column says 3days ago, I did have this installed and I removed it, but it should not be picking that up right?
image

src/NexusMods.App.UI/Pages/NexusModsDataProvider.cs
```diff
@@ -165,7 +175,7 @@ public IObservable<IChangeSet<LoadoutItemModel, EntityId>> ObserveNestedLoadoutI
 
                 var installedAtObservable = observable
                     .Transform((_, e) => LibraryLinkedLoadoutItem.Load(_connection.Db, e).GetCreatedAt())
-                    .QueryWhenChanged(query => query.Items.FirstOrDefault());
+                    .QueryWhenChanged(query => query.Items.Max());
```
Al12rs, line 178: This causes exception on deletes, as it can become an empty collection. Should check if Items is empty and return the default in that case.

src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridItemModel.cs
```diff
@@ -96,23 +102,88 @@ protected TreeDataGridItemModel()
...
+                                    UpdateChildren(model._children, changeSet);
+                                });
                         }
                     }, onCompleted: static (_, model) => CleanupChildren(model._children));
             }
         });
     }
 
-    private static void CleanupChildren(ObservableList<TModel> children)
+    private static void UpdateChildren(SourceCache<TModel, TKey> cache, IChangeSet<TModel, TKey> changeSet)
```
Al12rs, line 120: This isn't working in case one of the children is removed while the parent is collapsed.
We still have the child, but when we start processing updates again we only receive Adds for every element in the current collection, no Removes for our existing items since we had stopped listening.

So we would need to have special handling for when we make the subscription that checks the current SourceCache against the current contents and updates, then goes to listen changes from there.

Al12rs (approved):
Code looks good, need to wait for QA approval.

Al12rs (changes requested):
Somehow managed to break selection again. Will need a bit more testing.

Al12rs (approved):
I couldn't break it anymore

Al12rs (changes requested):
Never mind, deleting from Library stops working after a bit

erri120:
> Never mind, deleting from Library stops working after a bit

I'd say that's unrelated.

Al12rs (approved):
Yeah, haven't been able to reproduce, not for lack of trying.
We will keep an eye out to see if it happens again in the future.

Issue #1986, opened 2024-09-05
Research: Diagnostic Refresh Triggers


Sewer56:
# Spike

Not to be confused with:

- https://github.com/Nexus-Mods/NexusMods.App/issues/1852

## Currently 

We only run diagnostics when the state of the Loadout changes.
In some cases, there may be alternative times when you may want to refresh diagnostics.

### An UX Example

- User needs to install `REDmod` to Cyberpunk2077
- App shows an Error Diagnostic for `missing REDmod`
- `REDmod` is a DLC installed via your game store.

Because `REDmod` is installed externally, the diagnostic for `missing REDmod` continues to show
until the loadout is updated.

In practice this update will most likely happen when the user hits Sync to `Ingest` the new items
in the game folder. However, until that time, there is still a minor risk that the user will continue to
see the diagnostic, leading to confusion.

This time window however is very small. ***Therefore it is currently not that big of an issue.***

## Additional Information

Some questions arise:

- Should diagnostics opt-in to which events they should be updated on?
- Do we continue to just refresh the diagnostics entirely?
    - Only some diagnostics involve the current disk state, for example.
- Can this be turned into a caching system?
    - For example reuse a previous diagnostic result and check if it can be 'invalidated'.

erri120:
If you want to refresh on something else, you make a new emitter type. Previously, we had a mod and file emitter as well, it was just never used and didn't make a lot of sense. Since diagnostics are scoped to a Loadout, they'd change when the Loadout changes.

Al12rs:
You need 
- #1198  
Then a new emitter can be attached to that potentially. 
I think we can revisit this after we have detection and have a clearer idea how that system will all work.

PR #1989, opened 2024-09-05
Added: Support for running tools via Proton(tricks) and REDmod on Linux

### LINKED ISSUE #1774: Cyberpunk 2077 - REDMod Diagnostic
# User story

**As a** user

**I want to** be informed if I have not installed the REDMod DLC when it is required

**So that** my mods load correctly

## Requirements
The 1.6 update for Cyberpunk 2077 included an optional DLC called REDmod.  This DLC allows the game to load modded content created with the REDmod toolkit. [More info](https://wiki.redmodding.org/cyberpunk-2077-modding/for-mod-users/users-modding-cyberpunk-2077/redmod). [Even more info](https://www.cyberpunk.net/en/modding-support)

The app should detect when this is not installed on the user's system and prompt them to download it from the appropriate game store as required. 

### Detection Details
| Install Type | Detection | Custom Command |
| --- | --- | --- |
| Steam | Check for DLC with the store ID `2060310`. [Store Page](https://store.steampowered.com/app/2060310/Cyberpunk_2077_REDmod/)  | `steam://run/2060310` |
| GOG (Galaxy) | Check for the REDMod EXE file. [Store Page](https://www.gog.com/en/game/cyberpunk_2077_redmod)  | `goggalaxy://openStoreUrl/embed.gog.com/game/cyberpunk_2077_redmod` |
| Epic Games Store | Check for the REDMod EXE file. [Store Page](https://store.epicgames.com/en-US/p/cyberpunk-2077)  | `com.epicgames.launcher://store/p/cyberpunk-2077` |
| Manual or non-Galaxy GOG | Check for the REDMod EXE file. [Info Page](https://www.cyberpunk.net/en/modding-support) | n/a |

## Design

## DOD

## Scenario 1
**Given** _______

**When** _______

**Then** _______

### LINKED ISSUE #1915: Red Mod - Linux
As a user playing Cyberpunk on Linux
I want red mods to be compiled with red mod
So that I can use red mods on Linux and play the game

Red mod currently doesn't work in Linux, this work will be to get it working so that cyberpunk players can use red mods in Linux.

### LINKED ISSUE #1982: Cyberpunk 2077: Add Missing 'redmod' diagnostic.
# Improvement

## Currently 

It is possible to install a such as [Infinite Carrying Capacity (Weight)](https://www.nexusmods.com/cyberpunk2077/mods/9450/?tab=files) without the [REDmod] DLC installed inside the App.

This is problematic for the following reasons:

- A user new to modding Cyberpunk may be unaware of having to install [REDmod]
- A user may not be aware they just installed a [REDmod]
    - Someone new to the game might not know what a [REDmod] even is...
- The App does not report Apply Errors
    - https://github.com/Nexus-Mods/NexusMods.App/issues/1980

## We want

1. A diagnostic to tell the user that they need to install the [REDmod] DLC.
2. A way to block Apply if they try to apply a [REDmod] without a way to deploy it.
    - This requires design work.

## Design

CC. Feedback from @captainsandypants is requested.

Essentially, installing a [REDmod] without [REDmod] installed should (probably) block the Apply operation.
If that is not done, a user may hit `Apply`, and get an incorrectly 

[REDmod]: https://www.cyberpunk.net/en/modding-support

Sewer56:
This PR adds support for:

- Detection of protontricks
- Running tools via Proton(tricks)
- Running REDmod via Protontricks
- Diagnostic for missing protontricks
- Diagnostic for missing REDmod
    - Clicking the link takes you directly to relevant store page.
- Fixes incorrect casing of `redMod.exe` (previously `redmod.exe`)
- Overlay messagebox with only 'OK' button
    - This is no longer used, I've kept this in case anyone thinks it will be useful.
    - Don't request changes if you don't want this, just remove it and push.

https://github.com/user-attachments/assets/8f58394e-0874-4b4b-9ee5-c9c14b493bca

image

image

image

Design
--------

The screenshots above were already shown to design; with no further feedback requested for the time being.

It may be useful to print out the list of REDmod(s) that caused for a certain diagnostic to be emitted, however we've decided to wait for now.

QA Matrix
-----------

Check the following:

- Hyperlinks for `REDmod DLC` work on all stores, Windows and Linux
- Nexus App via AppImage (you may get 'missing protontricks' error)
- Cyberpunk 2077 without REDmod installed
- Anything else you can think of.

Issues I'm aware of:

- If you install REDmod without updating loadout, the diagnostic will still show up
    - https://github.com/Nexus-Mods/NexusMods.App/issues/1986

Tags
-----

fixes #1982
fixes #1774
fixes #1915


src/Games/NexusMods.Games.RedEngine/RedModDeployTool.cs
```diff
@@ -34,9 +26,19 @@ public async Task Execute(Loadout.ReadOnly loadout, CancellationToken cancellati
...
+        if (fs.OS.IsWindows)
+        {
+            var command = Cli.Wrap(exe.ToString())
+                .WithArguments("deploy")
+                .WithWorkingDirectory(exe.Parent.ToString());
+            await _toolRunner.ExecuteAsync(loadout, command, true, cancellationToken);
+        }
+        else
+        {
+            var batchPath = fs.GetKnownPath(KnownPath.EntryDirectory).Combine("Resources/Cyberpunk2077/deploy_redmod.bat");
```
erri120, line 39: This won't work. I'd suggest making the bat file an embedded resource and creating the file at runtime in the temp directory or similar.

erri120 (changes requested):

LukeNexusMods:
@Sewer56 Whats the best way to test this?

src/Games/NexusMods.Games.RedEngine/RedModDeployTool.cs
Sewer56: If I recall correctly, I brought this up at standup.

Generating a .bat and extracting to disk was the thing I originally wanted to do.

However the people there requested I don't do that and make it a static file that ships with the App pre-extracted instead.

So I'm rather confused, why am I being asked here to do what I originally intended to do? This shouldn't be a case of something like `Working Directory` vs `$OWD` for AppImage.

There is the fact I didn't make the `.bat` file get copied on Linux only, but that's a separate fixup.

Sewer56:
@LukeNexusMods Try to install a REDmod like
https://www.nexusmods.com/cyberpunk2077/mods/9450/ to your game. I listed the things to watch out for QA in the opening post.

src/Games/NexusMods.Games.RedEngine/RedModDeployTool.cs
erri120: Did you test this with an AppImage? I want to avoid using `CopyToOutputDirectory` if possible, as it can make packaging difficult.

Sewer56: Good point. Let me do that after I fixup CI

Sewer56: Took me a bit to respond. I just downloaded Cyberpunk 3 times, because Steam would give me errors after downloading the entire 80GB.

Works fine on AppImage, just need it to not copy to output on Windows.

Sewer56: I made the file not copy on Windows, however it depends on the Host OS the App was built in.
Details in the commit contents.

erri120: I still think we should go with the embedded route. `CopyToOutputDirectory` is just a hassle.

erri120: > Putting it as an embedded resource will make it embed in Windows too

How? What you have currently only copies for Linux. `EmbeddedResource` can have a condition as well.

Sewer56: Ah, I thought you wanted me to use the standard machinery here because the `.targets` file was unsightly to see.
If it's embedding via the targets file, then that would work.

erri120 (approved):

PR #1996, opened 2024-09-09
Collections Install Spike

halgari:
Initial spike on collections implementation. In this PR:

* Adds models for Collection metadata
* Adds GraphQL setup and a few basic queries for getting GQL data
* Adds methods for getting Collection metadata via GQL
* Based on a collection slug and revision number, allows for downloading a collection
* Downloaded collections are added to the library but tagged as collections not as downloaded archives
* Supports partial loading of the `collections.json` metadata file. We'll expand this as we add features
* Adds a job for installing a collection given the previously downloaded library item
  * The Installation job attempts first to download all the mods in the collection, then install the mods
  * "Replicated" mod types are supported, these involve on-the-fly MD5 hashing
  * Supports bastic mod installs
  * Skips any FOMOD collection members (for now, this will come in a future update)
* Adds support to the CLI for installing collections into a loadout
* Adds support for installing a collection via NXM links
  * **You can now install simple collections by clicking on the install button on the website**
* Fixes some bugs with the download meters broken by the job system
* All the above is wrapped up in jobs
* Deleted a few dead lines from the job system changeover
* Left menu shows collections after they are installed, and clicking on a collection in the left menu will show only mods in that collection

**Developer NOTE:** if you use Rider, install the GraphQL plugin, and with that you can get syntax highlighting and even run GQL queries directly inside the IDE. The endpoints are configured in the attached `graphqlrc` file.

Example installing the absolute best collection ever made:
    
chrome_D7EX9q9riw


Sewer56:
(I'm assigning myself because I want to know how they work under the hood. Chances are the others will get around to this before me)

src/NexusMods.Jobs/JobContext.cs
```diff
@@ -86,8 +87,9 @@ public void SetPercent<TVal>(TVal current, TVal max) where TVal : IDivisionOpera
         _progress.OnNext(percent);
     }
 
-    public void SetProgressRate(double rate)
+    public void SetRateOfProgress(double rate)
     {
+        Debug.Write($"Rate of progress: {rate}");
```
Sewer56, line 92: This may be an accidental leftover.
And if it isn't, it's missing a line ending.

src/NexusMods.App.UI/Controls/Spine/Buttons/Download/SpineDownloadButtonViewModel.cs
```diff
@@ -21,19 +21,22 @@ public SpineDownloadButtonViewModel(IJobMonitor jobMonitor, ILogger<SpineDownloa
...
             
             jobMonitor.ObserveActiveJobs<IHttpDownloadJob>()
-                .SumProgressRate()
+                .SumRateOfProgress()
                 .OnUI()
                 // Convert from bytes to MB/s
                 // TODO: use Humanizer
                 .Subscribe(rate => Number = rate / 1024 / 1024)
                 .DisposeWith(disposables);
+
```
Sewer56, line 38: Accidentally introduced redundant spacing here.

src/Abstractions/NexusMods.Abstractions.Collections/Json/ModSource.cs
```diff
@@ -0,0 +1,20 @@
+using System.Text.Json.Serialization;
+using NexusMods.Abstractions.NexusWebApi.Types;
+using NexusMods.Paths;
+
+namespace NexusMods.Abstractions.Collections.Json;
+
+public class ModSource
+{
+    [JsonPropertyName("type")]
+    public string Type { get; init; } = string.Empty;
```
Sewer56, line 10: This would probably be better represented as an enum for the purpose of better discovery (of available options)
and use of 'switch' statements.

System.Text.Json has a string to enum converter (`JsonStringEnumConverter`); which we already use in `MembershipRole`.

src/Networking/NexusMods.Networking.NexusWebApi/NexusModsLibrary.cs
```diff
@@ -147,4 +206,40 @@ public async Task<IJobTask<NexusModsDownloadJob, AbsolutePath>> CreateDownloadJo
 
         return nexusJob;
     }
+    
+    public async Task<IJobTask<NexusModsDownloadJob, AbsolutePath>> CreateDownloadJob(
```
Sewer56, line 210: Implementation of this method can be deduplicated with the other overload.

The difference here and the above overload is that the parameters are extracted from NXMModUrl, while here they are explicit. Aside from that, implementation doesn't differ. The above overload should be able to simply call this overload.

However to make calling more hassle free for public API users, since you don't want `nxmData` as additional parameter make a private method like `CreateDownloadJobImpl` and call that from both methods.

Sewer56:
Will finish up review tomorrow morning.

src/NexusMods.Collections/InstallCollectionJob.cs
```diff
@@ -0,0 +1,218 @@
...
+            TemporaryFileManager = provider.GetRequiredService<TemporaryFileManager>(),
+        };
+        return monitor.Begin<InstallCollectionJob, NexusCollectionLoadoutGroup.ReadOnly>(job);
+    }
+
+
+
+    public async ValueTask<NexusCollectionLoadoutGroup.ReadOnly> StartAsync(IJobContext<InstallCollectionJob> context)
+    {
+        _context = context;
```
erri120, line 64: I don't think we should be saving the context like this.

erri120 (dismissed):

src/Abstractions/NexusMods.Abstractions.Collections/NexusCollectionLoadoutGroup.cs
```diff
@@ -0,0 +1,17 @@
+using NexusMods.Abstractions.Loadouts;
+using NexusMods.Abstractions.NexusModsLibrary;
+using NexusMods.MnemonicDB.Abstractions.Attributes;
+using NexusMods.MnemonicDB.Abstractions.Models;
+
+namespace NexusMods.Abstractions.Collections;
+
+[Include<LoadoutItemGroup>]
+public partial class NexusCollectionLoadoutGroup : IModelDefinition
+{
```
erri120, line 10: We should have a chat about this today. There are two types of collections:

1) Downloaded from Nexus Mods, these are supposed to be read-only
2) Created in the App, these are writable since the user is the "owner"

When creating a loadout we want to create a default collection named "My Collection". Everything the user does will go into this collection by default.

Not sure how we want to represent this in the DB, we can talk about that later.

src/Abstractions/NexusMods.Abstractions.Collections/Json/Choices.cs
```diff
@@ -0,0 +1,9 @@
+using System.Text.Json.Serialization;
+
+namespace NexusMods.Abstractions.Collections.Json;
+
+public class Choices
+{
+    [JsonPropertyName("type")]
+    public string Type { get; init; } = string.Empty;
```
Sewer56, line 8: Another case of 'should probably be an enum'

src/Abstractions/NexusMods.Abstractions.Collections/NexusCollectionLoadoutGroup.cs
Sewer56: There are technically 2 ways to represent a collection.

1. Converting loadout to collection, i.e. Export
    - Same way you export PSD -> PNG
2. Treating a Loadout as a Collection (Suggested Here)

They have different semantics in terms of code.
The latter ensures loadouts will be easy to pack into collections, the former allows the App to remain more flexible.

Can go either way here.
Conceptually a loadout really is a collection, but what we have available in each format differs.
Maybe warrants a discussion.

src/NexusMods.Collections/InstallCollectionJob.cs
Sewer56: Does the context even need to be saved? It seems only the cancellation token is used out of it.
And you can pass that down to `EnsureDownloaded` via lambda.

docs/developers/misc/CollectionsOverview.md
```diff
@@ -0,0 +1,186 @@
...
+### Introduction
+Collections are, as the name suggests, mostly a collection (really a list) of mods. At their core they consist of a `.zip`
+file that contains a `collection.json` file and various other files bundled with it. Most of the mods are stored as a Nexus
+mod ID and fileId, and are therefore sourced from the normal download APIs. 
+
+### Collections Metadata
+Most of the collections metadata is stored and read from the Nexus Mods v2 GraphQL Api. The three main types of ids to be
+aware of are:
+
+- `slug` - a unique identifier for the collection, a 7 character alphanumeric string
```
Sewer56, line 12: Is this `a-ZA-Z0-9` specifically? Or `a-Z0-9` etc? Might be worth noting.

Weird it's 7 characters, it overflows a 32-bit int, unless it's `a-ZA-Z0-9` and converted to 64-bit int.

src/Abstractions/NexusMods.Abstractions.Collections/NexusCollectionLoadoutGroup.cs
Al12rs: We can discuss after standup, but I think we can't have the loadout itself be the "My Collection", because from design, "My collection" is supposed to only contain the loose mods, not other installed collections, which would be at the same level as "My collection", not nested to it.

docs/developers/misc/CollectionsOverview.md
halgari: I think it may actually be `a-z0-9` or at least that's the only examples I've seen

Sewer56: Probably is that in this case.
Given they're randomly generated, it's unlikely for them to be anything else.

src/Abstractions/NexusMods.Abstractions.Jobs/IJobMonitorExtensions.cs
```diff
@@ -47,16 +48,17 @@ public static IObservable<Percent> AverageProgressPercent<TJobType>(this IObserv
     /// <summary>
     /// Gets an observable of the sum of the progress rate of all given jobs.
     /// </summary>
-    public static IObservable<double> SumProgressRate<TJobType>(this IObservable<IChangeSet<TJobType, JobId>> jobs) 
+    public static IObservable<double> SumRateOfProgress<TJobType>(this IObservable<IChangeSet<TJobType, JobId>> jobs) 
         where TJobType : IJob
     {
         return jobs.TransformOnObservable(job => job.ObservableRateOfProgress)
-            .Filter(p => p.HasValue)
+            //.Filter(p => p.HasValue)
```
Al12rs, line 55: cleanup comment

src/Networking/NexusMods.Networking.NexusWebApi/GraphQL/schema.graphql
```diff
@@ -0,0 +1,3228 @@
+schema {
+  query: Query
+  mutation: Mutation
+}
+
+"Autogenerated return type of AcceptModerationFixMutation."
```
Al12rs, line 6: What is the best practice for generated `.graphql` files? Are they usually committed to source control?
This is 3k lines, and it might cause conflicts depending on how often it gets changed.

src/NexusMods.App.Cli/Types/IpcHandlers/NxmIpcProtocolHandler.cs
```diff
@@ -80,6 +86,39 @@ public async Task Handle(string url, CancellationToken cancel)
...
+                    var downloadJob = nexusModsLibrary.CreateCollectionDownloadJob(destination, collectionUrl.Collection.Slug, collectionUrl.Revision,
+                        CancellationToken.None
+                    );
+        
+                    var libraryFile = await library.AddDownload(downloadJob);
+        
+                    if (!libraryFile.TryGetAsNexusModsCollectionLibraryFile(out var collectionFile))
+                        throw new InvalidOperationException("The library file is not a NexusModsCollectionLibraryFile");
+
+                    var installJob = await InstallCollectionJob.Create(_serviceProvider, loadout, collectionFile);
```
Al12rs, line 115: Should this code be in here? I think the same for the normal nxm download above. I would prefer this just call a method of some service that is in charge of downloads or library.
At the least I would prefer this be extracted to a local method to keep the switch case more readable.

src/Networking/NexusMods.Networking.NexusWebApi/GraphQL/schema.graphql
Sewer56: I read the docs for the library used when looking at the PR.
Seems to be as intended. The generated files are generated manually using the CLI tool.
Unless the schema breaks on the site, or we need some new field/feature, this shouldn't need updating.

halgari: This part of Strawberry shake is also a bit jank. The initial setup has involved me doing `dotnet clean` deleting `bin` and `obj` folders, then doing a restore on all of that, so i don't want everyone to go through that hassle. And yes, adding the schema to git does allow us to do change tracking on the schema.

src/Abstractions/NexusMods.Abstractions.Collections/NexusCollectionLoadoutGroup.cs
halgari: Discussed at standup, I'll be adapting this post-merge to install in a format that fits into the UI and loadout system

src/NexusMods.Collections/InstallCollectionJob.cs
erri120: It just looks like a code smell, it feels wrong. There is this implicit assumption that we're only setting this field once, that we're never accessing the field before we set it, almost like a constructor. T

Al12rs (approved):

Sewer56 (approved):

PR #2000, opened 2024-09-10
Image store

erri120:
Image store to be used for mod page images, collection images, and other remote images we might want to store.

The implementation is a bit scuffed, but should be verified to work with tests. The store has two files: one file with only the image metadata and one file with only the data. It's a bit messy, especially the concurrency code, but we can improve this over time.


src/NexusMods.App.UI/IImageStore.cs
```diff
@@ -0,0 +1,16 @@
...
+
+/// <summary>
+/// Optimized storage for images.
+/// </summary>
+[PublicAPI]
+public interface IImageStore
+{
+    ValueTask Store(EntityId id, Bitmap bitmap);
+
+    ValueTask<Bitmap?> Retrieve(EntityId id);
```
Al12rs, line 15: Should we use Nominal Typing and make subclass for `EntityId` for this?

Al12rs: Ok I'm reading a bit more of the code and I see that Entity is just a unique identifier. This isn't using MnemonicDB in any way, it just needs a unique id.

Could you add some comments on Store and Retrieve to clarify this?

halgari:
What if we store this data in MnemonicDB? The images here likely aren't going to be huge (we should probably verify that) and I'd prefer that over writing new data formats, concurrency checks and the like. 

For reference, a mod thumbnail image as a webp image is ~14KB. Collection images are around 65KB.

halgari:
Storing this in MnemonicDB also allows us to have attributes that reference images. For example we could have `myMod.Image` be a `ReferenceAttibute<StoredImage>`. If we need other features, like pinned memory or streaming access to this data, that's fairly easy to get into MnemonicDB. Although for now it's likely that a `HashedBlobAttribute` would do the work

halgari:
There's reasons why images aren't traditionally put into SQL databases, but it's more common for KV stores or other NoSQL databased. 

Microsoft did some explination on this here: https://www.microsoft.com/en-us/research/wp-content/uploads/2006/04/tr-2006-45.pdf but in general it comes down to a question of speed, performance and fragmentation. In this case, assuming we are saving the images raw (it sounds like we are) the size of the images will result in fixed size blobs that are slightly over the NTFS cluster size (4KB). This ends up making it even better for storing in a database as on-disk they will each cause waste due to not matching the clusters, causing disk waste. 

But really the reason why tranditional databases recommend against blob storage because it fragments the tables with variable sized blobs, and blows out the cache with rarely used data. This isn't as big of a deal in our case where the images are small, and fixed sized. But if we want we can use a separate partition ID for these images, in which case the images will be stored in a separate part of the cache and only loaded as-needed. 

In general I think all this "just works" with MnemonicDB. We might want to re-evaluate performance if we ever start storing blobs over 1MB. But as the above doc says, there are some clear wins in general for database where the file size is small.

Sewer56:
Going to agree with Tim here, and add some extra context.
Storing in a DB or compacted in an archive would be better _**because we're dealing with Windows.**_

On Linux, FileSystem is fairly competitive for blobs of this size.  Here's a bit of a [more recent paper on the subject matter](https://knowledge.e.southern.edu/cgi/viewcontent.cgi?article=1110&context=crd) [Figure 5.2] [[raw results]](https://github.com/jesse-r-s-hines/KeyValueStoreBenchmark/blob/711a734e8fdec8088d2315406c615c03930f1ae4/results/benchmark.csv). However, that is, _on Linux._ On Windows, the FileSystem is unfortunately, notoriously slow for random access; and it's been getting worse over the years. 11 is slower than 10 which is slower than 7 for random access, and so forth.

The main cause in Windows' slowness unfortunately lies in ever increasing number of filter drivers. For example, a driver handling the UWP/Xbox Store DRM, the Antivirus filter driver etc. These process callbacks on every file open. Other reasons include FileSystem on Windows needing to handle either path separator, case insensitivity, etc. It's very much a 'death by a 1000 cuts` situation.

There's also this informative document [published by the SQLite authors in 2017](<https://www.sqlite.org/fasterthanfs.html>) showing this trend and [my own testing you may have seen on Discord](https://discord.com/channels/1134149061080002713/1134151290130268200/1270046252226314310). My `redb` fork can be easily cloned to change item sizes during bench.

In any case, use a separate partition ID or a separate DB instance. Given we're using RocksDB, the former should achieve the same performance characteristics, be easier to use, and is already there and ready.

I wouldn't recommend storing any blobs under 1MiB on Disk **_on Windows_**.

-------------------

> This ends up making it even better for storing in a database as on-disk they will each cause waste due to not matching the clusters, causing disk waste.

Misc note on this. 
If you are using storage manufactured in the last 10 years, you probably have a physical sector size of 4096 bytes.

```
fsutil fsinfo ntfsinfo c:
# Bytes Per Physical Sector: 4096
```

a.k.a. [Advanced Format](https://learn.microsoft.com/en-us/windows/win32/fileio/file-buffering#alignment-and-file-access-requirements). 

You do actually want your images to start at an offset aligned at 4096, in case the image size is in 6144-8192 byte range, or sub 4K; as this will likely save an additional sector read. Although it eats space, it is actually desirable for read performance; Nx actually aligns blocks to 4K for this reason.

Al12rs:
Isn't using the DB here heavily restricting the size of the images that we can store in the Image store? 
We are currently focused on Page thumbnails, but we have all the bigger images to store as well. Full size page images, game images, fomod images etc.

Do we want an image store that works exclusively for small images or a generic image store that is potentially slower for small images?

erri120:
> Isn't using the DB here heavily restricting the size of the images that we can store in the Image store? We are currently focused on Page thumbnails, but we have all the bigger images to store as well. Full size page images, game images, fomod images etc.
> 
> Do we want an image store that works exclusively for small images or a generic image store that is potentially a bit slower for small images?

I've set a limit of 16 MB per file, that's enough for a 2000x2000 image with 4 bytes per pixel. An image of 1920x1080 is around 8.3 MB.

This image store is only for remote images, which are usually optimized. We won't be putting 4K screenshots in this, but instead images that have already been resized. Mod page images will be around 5000 bytes, that's what this is optimized for.

tests/NexusMods.UI.Tests/ImageStoreTests.cs
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    [Fact]
+    public async Task SimpleTest()
+    {
+        var bitmap = new Bitmap(AssetLoader.Open(new Uri("avares://NexusMods.App.UI/Assets/DesignTime/cyberpunk_game.png")));
+        var storedImage = await _imageStore.PutAsync(bitmap);
+        using var lifetime = _imageStore.Get(storedImage);
+        lifetime.Should().NotBeNull();
+    }
```
Al12rs, line 25: Want to add a test to verify metadata is stored and loaded correctly as well?

Al12rs (commented):
Looks ok enough for me, some questions regarding deletion from DB if we start having lots of deleted mods.

Al12rs (approved):

halgari (approved):

Issue #1705, opened 2024-07-01
Collections Spike


LukeNexusMods:
Investigate the technical aspects of collections and how we will begin to implement them. 

Phase 1 will be downloading and installing collections.

- [x] Interpreting the nxm:// links containing a collection payload.
- [x] Requesting the collection bundle from the Graph API.
- [x] Inspecting the collection bundle to identify the manifest file.
- [x] Inspecting the collection bundle to identify any supplementary metadata.
- [x] Inspecting the collection bundle to identify any bundled mods.
- [x] ~~Show the pre-install instructions.~~ Removed as this is part of the UI side of things
- [x] ~~Check the user's game version against the one used to make the collection.~~ Also a UI constraint for most (all?) games?

LukeNexusMods:
Here are the notes from Mikes ticket around the current functionality of downloading and installing collections. 

Downloading Collection Metadata

 Interpreting the nxm:// links containing a collection payload.
 Requesting the collection bundle from the Graph API.
 Inspecting the collection bundle to identify the manifest file.
 Inspecting the collection bundle to identify any supplementary metadata.
 Inspecting the collection bundle to identify any bundled mods.
 Show the pre-install instructions.
 Check the user's game version against the one used to make the collection.

PR #2003, opened 2024-09-10
Loadout support for collections

halgari:
This probably needs to wait until #1705, but the structure here is helpful for what comes after that code

* Adds the concept of collections in the loadouts
* Makes a default "My Mods" collection when a loadout is created
* All mods that are not assigned to another group will be put into the "My Mods" collection
* Adds a "group" filter to the Loadout tree view that scopes it to "My Mods" by default

This should be a unoticable change to users, but internally all mods are now in a collection and the loadout should filter out any mods that are not part of the currently selected collection. There is no way to select a collection or add a new one, but the code is there. We can merge this or #1705 first, or merge them both. Once the code is in, we can update the modesl in #1705 to include the models in this PR and things should "just work".

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts/CollectionGroup.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts/CollectionGroup.cs
@@ -1,0 +1,42 @@
+using DynamicData.Kernel;
+using NexusMods.MnemonicDB.Abstractions.Attributes;
+using NexusMods.MnemonicDB.Abstractions.Models;
+
+namespace NexusMods.Abstractions.Loadouts;
+
+/// <summary>
+/// A definition of a collection. This could be a user's collection, a Nexus Mods collection or perhaps a collection from some other source.
+/// </summary>
+[Include<LoadoutItemGroup>]
+public partial class CollectionGroup : IModelDefinition
+{
+    private const string Namespace = "NexusMods.Abstractions.Loadouts.CollectionGroup";
+    
+    /// <summary>
+    /// Mostly a marker attribute to find all the collections on a given loadout. Likely faster than searching for all the loadout items
+    /// and filtering them by their attributes
+    /// </summary>
+    public static readonly ReferenceAttribute<Loadout> Loadout = new(Namespace, nameof(Loadout));
+
+    /// <summary>
+    /// In this UI this is reflected as "My Collection", it's a group of all mods that the user has installed outside of any
+    /// other collection.
+    /// </summary>
+    public static readonly MarkerAttribute UserCollection = new(Namespace, nameof(UserCollection));
+}
+
+public static partial class CollectionGroupLoaderExtensions
+{
+    /// <summary>
+    /// Find the user collection for a given loadout
+    /// </summary>
+    /// <param name="loadout"></param>
+    /// <returns></returns>
+    public static Optional<CollectionGroup.ReadOnly> FindUserCollection(this Loadout.ReadOnly loadout)
+    {
+        return loadout.Db
+            .Datoms(CollectionGroup.Loadout, loadout.Id)
+            .Select(d => CollectionGroup.Load(loadout.Db, d.E))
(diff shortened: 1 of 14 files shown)


src/Abstractions/NexusMods.Abstractions.Loadouts/CollectionGroup.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Mostly a marker attribute to find all the collections on a given loadout. Likely faster than searching for all the loadout items
+    /// and filtering them by their attributes
+    /// </summary>
+    public static readonly ReferenceAttribute<Loadout> Loadout = new(Namespace, nameof(Loadout));
+
+    /// <summary>
+    /// In this UI this is reflected as "My Collection", it's a group of all mods that the user has installed outside of any
+    /// other collection.
+    /// </summary>
+    public static readonly MarkerAttribute UserCollection = new(Namespace, nameof(UserCollection));
```
erri120, line 25: The semantics of this should be different. A collection is either read-only, because it was downloaded, or mutable, because the user created it. We can have more than 1 "user collection", the "My Collection" is just the default one we create.

There isn't anything special about the "My Collection", the only code that actually cares about that is the code that creates the Loadout. Everything else will either be passed a collection to use or use the first mutable collection as a default.

The comment here should be updated to reflect these semantics. Currently it suggests that there is only one user collection, when the user can actually create as many as they want.

erri120 (changes requested):

src/NexusMods.App.UI/Pages/ILoadoutDataProvider.cs
```diff
@@ -25,6 +28,15 @@ public static IObservable<IChangeSet<Datom, EntityId>> FilterInStaticLoadout(
...
     {
+        if (loadoutFilter.CollectionGroupId.HasValue)
+        {
+            return source.Filter(datom =>
+                {
+                    var item = LoadoutItem.Load(connection.Db, datom.E);
+                    return item.LoadoutId == loadoutFilter.LoadoutId && item.IsChildOf(loadoutFilter.CollectionGroupId.Value);
+                }
+            );
+        }
```
erri120, line 39:
```suggestion
        var filterByCollection = loadoutFilter.CollectionGroupId.HasValue;
        return source.Filter(datom =>
        {
            var item = LoadoutItem.Load(connection.Db, datom.E);
            if (!item.LoadoutId.Equals(loadoutFilter.LoadoutId)) return false;
            if (filterByCollection) return item.IsChildOf(loadoutFilter.CollectionGroupId.Value);
            return true;
        });
```

A bit more compact and readable.

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutPage.cs
```diff
@@ -21,17 +22,22 @@ public record LoadoutPageContext : IPageFactoryContext
...
     }
 
     public static readonly PageFactoryId StaticId = PageFactoryId.From(Guid.Parse("62fda6ce-e6b7-45d6-936f-a8f325bfc644"));
     public override PageFactoryId Id => StaticId;
 
     public override ILoadoutViewModel CreateViewModel(LoadoutPageContext context)
     {
-        var vm = new LoadoutViewModel(ServiceProvider.GetRequiredService<IWindowManager>(), ServiceProvider, context.LoadoutId);
+        // Default to the user group for now
+        var userGroup = Loadout.Load(_connection.Db, context.LoadoutId).MutableCollections().First().AsLoadoutItemGroup();
```
erri120, line 39: Group should be passed via the context.

erri120 (changes requested):

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutPage.cs
halgari: There's nothing to pass into this yet, I'll add that once I add the UI elements that will populate that value

erri120 (approved):

PR #2015, opened 2024-09-12
Fixed: Adding Library Files no longer Adds Duplicates

### LINKED ISSUE #1994: Exception when deleting an item from the library
```
System.Collections.Generic.KeyNotFoundException: Entity EId:20000000000077C does not have attribute NexusMods.DataModel.ArchivedFile/Container
   at NexusMods.MnemonicDB.Abstractions.Attributes.ScalarAttribute`2.ThrowKeyNotfoundException(IHasEntityIdAndDb entity)
   at NexusMods.MnemonicDB.Abstractions.Attributes.ScalarAttribute`2.Get(IHasIdAndIndexSegment entity)
   at NexusMods.Abstractions.FileStore.Nx.Models.ArchivedFile.ReadOnly.get_Container() in /mnt/redline/projects/NexusMods/app/main/src/Abstractions/NexusMods.Abstractions.FileStore.Nx/obj/Debug/net8.0/NexusMods.MnemonicDB.SourceGenerator/NexusMods.MnemonicDB.SourceGenerator.ModelGenerator/NexusMods_Abstractions_FileStore_Nx_Models_ArchivedFile.Generated.cs:line 297
   at NexusMods.App.GarbageCollection.DataModel.NxFileStoreUpdater.UpdateNxFileStore(List`1 toRetract, AbsolutePath newArchivePath) in /mnt/redline/projects/NexusMods/app/main/src/NexusMods.App.GarbageCollection.DataModel/NxFileStoreUpdater.cs:line 44
   at NexusMods.App.GarbageCollection.DataModel.RunGarbageCollector.<>c__DisplayClass0_0.<Do>b__0(IProgress`1 progress, List`1 toArchive, List`1 toRemove, ArchiveReference`1 archive) in /mnt/redline/projects/NexusMods/app/main/src/NexusMods.App.GarbageCollection.DataModel/RunGarbageCollector.cs:line 27
   at NexusMods.App.GarbageCollection.ArchiveGarbageCollector`2.CollectGarbage(IProgress`1 progress, RepackDelegate doRepack) in /mnt/redline/projects/NexusMods/app/main/src/NexusMods.App.GarbageCollection/ArchiveGarbageCollector.cs:line 133
   at NexusMods.App.GarbageCollection.DataModel.RunGarbageCollector.Do(Span`1 archiveLocations, IFileStore store, IConnection connection) in /mnt/redline/projects/NexusMods/app/main/src/NexusMods.App.GarbageCollection.DataModel/RunGarbageCollector.cs:line 24
   at NexusMods.DataModel.GarbageCollectorRunner.Run() in /mnt/redline/projects/NexusMods/app/main/src/NexusMods.DataModel/GarbageCollectorRunner.cs:line 17
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
```
### Edit:
From @Sewer56 

##### Reproduction

To understand the dangers involved, let's try reproducing the bug.

!!! note "The bug is fixed today."

    Today, the 'bug' is fixed as of commit ( `34de799623cf9688e6c3dcabca7fe029426583ed` ) , however prior to it;
    the bug could be reproduced in the way listed below.

We will make an assumption that we have an un-patched GC and faulty code that
creates duplicates.

Create 2 archives:

- dummy-1 (zip)
- dummy-2 (zip)

Inside `dummy-1`, add 1 file.
Inside `dummy-2`, add the file from `dummy-1`, and an additional file.

Add the `dummy-1` and `dummy-2` archives from disk in the following order:

- dummy-1
- dummy-2

Adding `dummy-2` creates a duplicate file due to an error by the programmer,
the `DataStore` entry for the duplicate hash will be re-routed to `dummy-2` when
it previously pointed to `dummy-1`.

If we now delete `dummy-1`; the error should have a 50-50 chance of reproducing;
depending on order archives are processed. There is a risk retracting the item in the
DataStore will retract the item in the wrong archive/container.

!!! note "There is some RNG involved."

    Reproduction is non-determinstic, due to nature of data.
    Bigger archives with duplicates are more likely to yield errors.

    So you can replace `dummy-1` with `SMAPI 4` and `dummy-2` with `SMAPI 3` 
    for a more reliable reproduction.

Sewer56:
## Summary

This is a fix for
- #1994

This fixes a bug where we unnecessarily back up items which may already be in the FileStore.

## Explanation

[Explanation Copied from Discord]

When the code for adding library items was rewritten to the new job system, someone forgot to add a check to make sure we're not packing duplicates.

[Which is a bug]

When adding an item we're only supposed to back up new files. Much of our previous code, including the GC was written under this assumption. We've broken this behaviour for adding library items; and now we've got a situation where adding mods with duplicates causes us to store compressed duplicates in separate archives on disk.

The general repro on this is:

- Install an archive for ModA
- Install an archive for ModB where some files in ModB existed in ModA
- Delete ModB

It's around 90% consistent, part of it depends on operation order.
[This was very not trivial to find/reproduce, given the amount of variables and bit of RNG required to even repro this]

So the exact reason this happens (I believe) is that there are multiple of the same file in different archives.

i.e. A file with a given hash can be in more than 1 container (Archive) due to this.

But when we query for files throughout the App (and in the GC included), we query them by *hash*. i.e. There should only be a single entry and single container for a hash. We have multiple.
 
So when an `ArchivedFile` is swooped up by the GC, it's retracted. Retract works on first element of the `Entity`. That first element happens to be container reference which is how #1994 shows up.

When the GC reassigns the container of ***the duplicate*** of that item that's been repacked, it errors because there is no container. It was retracted by accident due to the failed assertion above. Because the query by hash returned the wrong entity (there should be 1 entity for a given hash, but we have multiple).

Tags
-----
fixes #1994

### DIFF
--- a/benchmarks/NexusMods.Benchmarks/Benchmarks/Loadouts/Harness/DummyFileStore.cs
+++ b/benchmarks/NexusMods.Benchmarks/Benchmarks/Loadouts/Harness/DummyFileStore.cs
@@ -11,7 +11,7 @@ public class DummyFileStore : IFileStore
         return ValueTask.FromResult(false);
     }
 
-    public Task BackupFiles(IEnumerable<ArchivedFileEntry> backups, CancellationToken token = default)
+    public Task BackupFiles(IEnumerable<ArchivedFileEntry> backups, bool deduplicate = true, CancellationToken token = default)
     {
         return Task.CompletedTask;
     }
--- a/src/Abstractions/NexusMods.Abstractions.IO/IFileStore.cs
+++ b/src/Abstractions/NexusMods.Abstractions.IO/IFileStore.cs
@@ -17,12 +17,23 @@ public interface IFileStore
     public ValueTask<bool> HaveFile(Hash hash);
 
     /// <summary>
-    /// Backup the given files. If the size or hash do not match during the
-    /// backup process a exception may be thrown.
+    /// Backup the given set of NEW files.
+    ///
+    /// If the size or hash do not match during the
+    /// backup process an exception may be thrown.
     /// </summary>
-    /// <param name="backups"></param>
+    /// <param name="backups">
+    ///     The files to back up.
+    ///     These should not contain any already stored files unless <paramref name="deduplicate"/>
+    ///     is set to true.
+    /// </param>
+    /// <param name="deduplicate">
+    ///     Ensures no duplicate files are stored.
+    ///     Only set this to false if you are certain there are no duplicates
+    ///     with existing files in the input.
+    /// </param>
     /// <param name="token"></param>
-    Task BackupFiles(IEnumerable<ArchivedFileEntry> backups, CancellationToken token = default);
+    Task BackupFiles(IEnumerable<ArchivedFileEntry> backups, bool deduplicate = true, CancellationToken token = default);
 
     /// <summary>
     /// Extract the given files to the given disk locations, provide as a less-abstract interface incase
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -808,7 +808,8 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
             }
         );
 
-        await _fileStore.BackupFiles(archivedFiles);
+        // SAFETY: We deduplicate above with the HaveFile call.
+        await _fileStore.BackupFiles(archivedFiles, deduplicate: false);
     }
 
     private async Task<DiskState> GetOrCreateInitialDiskState(GameInstallation installation)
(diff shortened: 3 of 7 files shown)


halgari:
I really think the IFileStore should be doing this de-dupping. Otherwise we're going to keep getting this issue

Sewer56:
> I really think the IFileStore should be doing this de-dupping. Otherwise we're going to keep getting this issue

I don't see why not.
Could be enabled by default, with an opt-out in the future.

There will be cases where we will be doing the checks ourselves before calling backup, so in those cases opting out would be valuable.

halgari:
The more I think about this, the more I see situations were we may have race conditions or accidentally archive a file too many times. Is there a reason why we can't make this more resiliant to duplicate files?

Sewer56:
> The more I think about this, the more I see situations were we may have race conditions or accidentally archive a file too many times. Is there a reason why we can't make this more resiliant to duplicate files?

You're in luck, I had this concern back when working on the GC. There's a custom `ReaderWriterLock` in the NxFileStore. Custom because we needed something that works with Async.

When the synchronizer or some other writer is doing its hashing work, etc. we could lock for writing to prevent unwanted duplicates. Do let me improve that in a follow up PR however; ensuring this is in for Release is crucial.

erri120 (approved):

PR #2053, opened 2024-09-17
Improved: Correctly Handle Duplicate files in GC

Sewer56:
This is a patch for the Garbage Collector that ensures that deletions are correctly handled in the presence of duplicate files spanning multiple archives. This includes a test for the underlying condition that causes for this failure to ultimately occur in the long run.

Patching this up took a while; because:

- Testing this is non-deterministic.
- Testing requires fully resetting the DataStore (since the issue is affected by DB history)

In any case, the patch itself should be fairly trivial; I tested with a bunch of inputs.

## Documentation

I added a section named `Duplicate Items` to the GC docs.

The subsection, aptly named `Reproduction` should probably not be in the docs; if anyone has a better idea of where to move it (Gist?); be my guest. Don't block merging, move it yourself and commit a hyperlink.

## Tags

fixes #1994
fixes #2026

PR #2055, opened 2024-09-18
Fixed: Adding Mods to Library no longer Adds Duplicate Files

Sewer56:
This is a revised version of #2015 , with the following changes:

- `AddLibraryFileJob` is left unchanged, sans adding the duplicate check.
- `IFileStore.BackupFiles` has a `remarks` section with recommended usage.

And fixed any comments related to the code changes.

Quick Notes.

## Deduplication

It's enabled by default and opt-out.

Manually checking for duplicates and opting out has small minimal performance improvement;
so it is recommended that you do so by the remarks section. However it is not mandatory.

## Race Condition(s)

No change here.
Calling `BackupFiles` or `ExtractFiles` from multiple threads is safe, but may lead to duplicates.
This is a-ok as noted in the remarks section, GC will clean up duplicates after #2053

If a GC is happening, you can't update the FileStore info in DataStore or query the FileStore due to RwLock, same as before.

### DIFF
--- a/tests/NexusMods.Collections.Tests/CollectionInstallTests.CanInstallBasicCollection.verified.txt
+++ b/tests/NexusMods.Collections.Tests/CollectionInstallTests.CanInstallBasicCollection.verified.txt
@@ -2,7 +2,7 @@
   mods: [
     Game Files,
     Halgari's Helper,
-    My Collection
+    My Mods
   ],
   files: [
     {Game}/archive/pc/mod/_1_Ves_HanakoFixedBodyNaked.archive,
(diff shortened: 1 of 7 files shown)


tests/NexusMods.Collections.Tests/CollectionInstallTests.CanInstallBasicCollection.verified.txt
```diff
@@ -2,7 +2,7 @@
   mods: [
     Game Files,
     Halgari's Helper,
-    My Collection
```
Al12rs, line 5: How is this not breaking the tests before or after?

Al12rs (approved):

tests/NexusMods.Collections.Tests/CollectionInstallTests.CanInstallBasicCollection.verified.txt
Sewer56: It's a carryover from last PR, it was broken before in that case. Truth be told I'm curious myself.

Issue #2056, opened 2024-09-18
[User Reported] NXM Handler Not Working [Missing Working Dir?]


Sewer56:
# Bug Report

## Summary

Yesterday during load order meeting, a user [reported an issue](https://discord.com/channels/1134149061080002713/1134149062589960215/1285605726286516446) on the Discord server, whereby their login was not fully working.

After some back and forth, post meeting; it seems the following is true:

- `xdg-settings set default-url-scheme-handler nxm com.nexusmods.app.desktop` ran successfully
- `update-desktop-database /home/g/.local/share/applications` ran successfully

At least according to log.

In other words, all is good, but the user's web browser isn't booting the App.

## Steps to reproduce

I cannot currently reproduce this on my machine, although I run the same Linux distro as the user.
The user was also running archive install, not AppImage, which isn't as commonly used/tested.

However I have a possible *theory*; when I asked the user to run

```
/home/g/Desktop/Nexus\ mod\ app/mog\ manager/NexusMods.App-0.6.0-1.linux-x64/NexusMods.App
```

Which was derived from the user's `.desktop` file, the App failed to start.
When I asked them to change the current working directory to the App folder, i.e.

```
/home/g/Desktop/Nexus\ mod\ app/mog\ manager/NexusMods.App-0.6.0-1.linux-x64
```

The App successfully booted.

I suspect there's a possibility that the user's web browser is not setting the working directory of the App to the App folder. Since some browsers don't delegate to `xdg-open`.

The [XDG .desktop specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html) does not mention the ***default working directory*** anywhere in the specification. So if the `Path` field is not specified, there's a chance the working directory is defaulting to `~` instead.

Some more investigation is needed, I had limited time. I want to test my theory by providing a user with a patched build that explicitly sets working directory first, before doing additional investigation; just in case I'm right.

> [!NOTE]  
> I forgot to ask the user to test `xdg-open nxm://` at the time.

But I believe we should set a working directory nonetheless.

## What is the expected behaviour?

Login should work.

## Other information

User distro is CachyOS, same as mine.

The user's log file:

[nexusmods.app.main.current.log](https://github.com/user-attachments/files/17038646/nexusmods.app.main.current.2.log)

User's `.desktop` entry:

```ini
[Desktop Entry]
Type=Application
Version=1.0
Name=Nexus Mods App
GenericName=Mod Manager
[...]

PR #2057, opened 2024-09-18
Changed: Explicitly Set Working Directory for nxm handler on Linux

Sewer56:
For the details, please see:
- https://github.com/Nexus-Mods/NexusMods.App/issues/2056

Putting this PR up so I can get a build; want to know if this fixes the end user issue.

---------------

Because this may be brought up. Yes, ideally our App shouldn't break if executed from the wrong working directory, but that's a topic for another day.


erri120:
I'm not a fan of this, downstream would have to `sed` yet another key. Do we have logs of the failed attempt to start the app from the home directory? I'd much rather fix that.

Sewer56:
Closing this because it didn't solve the user end-issue. Turns out they're running AppImage, but our logs said 'Archive'. I'll talk and look into that a bit more.

PR #2058, opened 2024-09-18
Fix the file deletion error by removing the deadlock

halgari:
This fixes the file deletion deadlock, by removing the lock inside `HaveFile` this is a one line change

But won't this cause issues? Perhaps, but the lock was mostly worthless anyways. Look at the code, we lock and say "do we have this file" because someone may want to make sure a file exists before they open it. Okay, so what if they call `HaveFile` and then before they can call `GetFileStream` the GC deletes the file? 

A good rule of thumb here is no predicate can protect against race conditions unless the predicate is itself inside a lock. All the locking code in the GC needs to be re-considered and reworked to either be lock-free or fault tollerant. 

At any rate, this solves the issue at hand and won't introduce any new bugs.

### DIFF
--- a/src/NexusMods.DataModel/NxFileStore.cs
+++ b/src/NexusMods.DataModel/NxFileStore.cs
@@ -60,7 +60,6 @@ public class NxFileStore : IFileStore
     /// <inheritdoc />
     public ValueTask<bool> HaveFile(Hash hash)
     {
-        using var lck = _lock.ReadLock();
         var db = _conn.Db;
         var archivedFiles = ArchivedFile.FindByHash(db, hash).Any(x => x.IsValid());
         return ValueTask.FromResult(archivedFiles);


Al12rs (approved):

Sewer56 (approved):
This works for me, as a temporary measure.

More robust fault tolerance wasn't originally within the time budget; so I did what I could with the limited time I had (under a day) to try prevent faults with what we had.

We may need to discuss and document the long term strategy at some point however.
The `GetFileStream` -> `File Deleted while in Use` situation is not the end of the world; that can be handled.

But a file not being backed up when it should (we thought we have it, but it's scheduled for deletion) is a critical error on the other hand. It means failure on future apply operations.

A synchronisation mechanism around that needs to be present; and that mechanism should not be 'backup every file used or not', because that for instance means backing up 11GB (instead of e.g. 2GB) when installing an update to Skyrim's most downloaded mod, 202X.

The HaveFile lock currently prevents that specific edge case. An alternative solution should be introduced, as soon as possible, to handle that case.

Sewer56:
> An alternative solution should be introduced, as soon as possible, to handle that case.

Usually for this sort of design, I prefer having a writer and reader.

You have a factory which provides you with writer and reader, obtaining either takes the relevant lock; and the returned reader/writer holds onto that lock. When the reader/writer is disposed, the lock is released.

Creating derivative items from a given reader/writer also takes hold of a lock, e.g. opening a stream with `GetFileStream`, would increase the read lock counter by 1 and the returned stream struct/reader would then release the lock on dispose. (The lock is an atomic semaphore, essentially)

Unfortunately this isn't as good in C#, as it is in Rust for example. There's no way to enforce the user will dispose the item as soon as possible. There would need to be a way to enforce the API user makes use of the `using` statement; so they dispose to release the lock as soon as possible. We're trying to avoid situations where the user needs to be aware of these things, as that's error prone; so unless we can guarantee disposal, a solution like this is undesireable.

I'd be curious to know if anyone has better ideas here.

PR #2063, opened 2024-09-19
Changed: Prevent possible FileStore Deadlocks on GC By Preventing Yielding to external code. 

Sewer56:
Very quick patch.

Updating the DataBase's file copy of the FileStore objects interacts with external non-GC components,
such as MnemonicDB. This may cause us to yield to external code that could touch the FileStore lock.

To avoid deadlocks, we should prevent this from happening if possible. This is why we release `store.Lock()` early
and instead put another lock on the GC operation itself.

----------

Long term, we may want to update IFileStore a bit to be resilient when opening streams. For additional comments see:
https://github.com/Nexus-Mods/NexusMods.App/pull/2058#issuecomment-2359024216

In addition, we still need a non-blocking Commit in MMDB; I put the details surrounding that in the code itself.


Al12rs:
Seems to be working correctly with these changes

Al12rs (approved):

PR #2064, opened 2024-09-21
Update to the new version of MnemonicDB

halgari:
Pretty much a nothing PR. The recent internal changes to MnemonicDB split the AttributeRegistry into two distinct parts:

* AttributeCache - a cache of all the properties of attributes found within a specific datom store. This class is 100% driven by the datoms in the store itself. Is a given attribute indexed? Well that data is in the datom store, and this cache provides rapid access to the data. The backend store (RocksDB, in-memory DB) create this cache and various parts of the system refresh the cache when a new schema is inserted
* AttributeResolver - this is the DependencyInjection portion of the AttributeRegistry. About the only thing it does is allow for finding a `IAttribute` class based on an AttributeId. This class is created and maintained by the `IConnection`

This change makes most of MnemonicDB free from any sort of global context or DI. Datoms are inserted into transactions and the backend store serves up `Datom`, all the mapping to and from C# objects happens outside of the MnemonicDB internals

All that to say, about 40 places in the code needed a class name change, and few places needed some small changes.


src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ThreeWayMerger.cs
```diff
@@ -0,0 +1,21 @@
...
+    public abstract int CompareBC(TB b, TC c);
+    
+    public abstract TOut Combine(Optional<TA> a, Optional<TB> b, Optional<TC> c);
+
+    public void Merge(IEnumerable<TA> aColl, IEnumerable<TB> bColl, IEnumerable<TC> cColl)
+    {
+        
+    }
+    
+}
```
erri120, line 21: This seems out of place.

erri120 (approved):

PR #2071, opened 2024-09-23
Fix Library deletions 

Al12rs:
by ensuring that even never activated items have `LibraryItems` subscription set up.

Apparently it is possible to select rows while their models are never activated, if the Library is scrolled using the Scrollbar very quickly, as some rows will be skipped from rendering.
This means it is possible to select those rows and attempt to delete them, while their models were never activated, and so their data never properly populated.

This was unexpected, as it was assumed that for an item to be selected it would at some point have to appear into view at least once.

This moves the setting up of the subscription for `LibraryItems` property into the constructor instead of inside the `WhenActivated` block.


halgari (approved):

halgari:
Makes sense

PR #2080, opened 2024-09-24
Initial UI design for the collection install page

halgari:
Putting this here so the UI mages can get to work figuring out all the things I screwed up in the design. This is only the design components (view models and views) so that we can get the look and feel updated while I continue work on the backend bits. 

A few matters of note for UI implementation:

* The expanders are not styled at all yet. I'm not sure if we use these anywhere in the app, but we should update these to match the figma designs of streatching the full width of the panel
* The `...` button has a flyout menu that will need icons, but I like the way that menu is implemented (as a flyout static resource), so we're free to style that as-is
* The background image is mapped into the ViewModel and can be bound via un-commenting the lines in the view. We somehow have to add a gradient to this and set it as the unscaled background of the panel. I really love this feature but I think we'll need to research how to implement this. 
* The mods grids are not yet implemented, those will come in a future PR
* Nothing is wired into the app, nothing is hooked up, this is just a draft of the UI components.


Al12rs (approved):
It's all new stuff so this doesn't break anything and we can fix up later

PR #2081, opened 2024-09-24
Invert the Collections GQL call 

halgari:
This PR inverts the way we query GQL data for collections. Instead of starting at the collection node and getting *all* revisions, we start at the revision and query to the collection. Based on this we can also pull the mod metadata (merging results with metadata calls from other parts of the library). 

Using this data we can then do `revision.Files.Count` to see how many downloads are in the collection, and in general treat collection data more like a graph. This work is required for future work on the collections download page


src/Abstractions/NexusMods.Abstractions.NexusModsLibrary/Models/CollectionRevisionModFile.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace NexusMods.Abstractions.NexusModsLibrary.Models;
+
+public partial class CollectionRevisionModFile : IModelDefinition
+{
+    private const string Namespace = "NexusMods.Library.CollectionRevisionModFile";
+    
+    /// <summary>
+    /// The Nexus, globally unique id identifying a specific file of a collection revision.
+    /// </summary>
+    public static readonly ULongAttribute FileId = new(Namespace, nameof(FileId)) { IsIndexed = true };
```
Sewer56, line 14: This field is misleading.
It maps to the `Id` field of the `ModFiles` node in the GraphQL response; however the `ModFiles` node (and others) have a field called `FileId`.
Therefore, this field could very easily be mistaken for the `FileId` field present in the GraphQL API responses; when it is not the case.

------

**Potential Improvement**:

I think there's something better we can actually store here.
Storing the `Id` field doesn't do us much here, while it is a unique identifier, none of the GraphQL APIs actually support it as input.

What they do support however, is a field called `uid`.
The UID is a composite 8 byte key. The upper 4 bytes represent the item ID, and the lower 4 bytes represent the item group.

Example: [My branch for mod update cache stuff has a definition for Uid for a mod page.](https://github.com/Nexus-Mods/NexusMods.App/blob/71a131b273ca1600c23f6abc2c66f1df4d392cf0/src/Networking/NexusMods.Networking.ModUpdates/Structures/Uid.cs#L12) [I'll probably move this struct to main WebAPI project]

For files the definition is the same, the upper 4 bytes are the `FileId`, and the lower 4 bytes are `GameId`.
Since the API response gives you those fields back you can reconstruct the `uid`.

Which in turn, gives us more flexibility as we can use the V2 APIs to query these files (unlike the 'ID' field):

```graphql
query ModFiles {
    modFilesByUid(uids: ["14315126081750"]) {
        nodesCount
        nodesFilter
        totalCount
        nodes {
            category
            categoryId
            changelogText
            count
            date
            description
            fileId
            id
            manager
            modId
            name
            primary
            reportLink
            requirementsAlert
            scanned
            size
            sizeInBytes
            totalDownloads
            uCount
            uid
            uniqueDownloads
            uri
            version
        }
    }
}
```

Response:

```json
{
    "data": {
        "modFilesByUid": {
            "nodesCount": 1,
            "nodesFilter": null,
            "totalCount": 1,
            "nodes": [
                {
                    "category": "ARCHIVED",
                    "categoryId": "ARCHIVED",
                    "changelogText": [
                        "#<ModChangelog:0x00007fda4e6f3df0>"
                    ],
                    "count": 31009,
                    "date": 1722900467,
                    "description": "",
                    "fileId": 84182,
                    "id": "Z2lkOi8vYXBpL01vZEZpbGUvMTQzMTUxMjYwODE3NTA",
                    "manager": 0,
                    "modId": 4198,
                    "name": "ArchiveXL",
                    "primary": 0,
                    "reportLink": "https://www.virustotal.com/gui/file/cf5c14c6ad61839a6e2a69a026143bd134964028121ee29e6020ebc1081f5396/detection/f-cf5c14c6ad61839a6e2a69a026143bd134964028121ee29e6020ebc1081f5396-1722900543",
                    "requirementsAlert": 1,
                    "scanned": "VERIFIED",
                    "size": 774,
                    "sizeInBytes": "793064",
                    "totalDownloads": 31009,
                    "uCount": 26076,
                    "uid": "14315126081750",
                    "uniqueDownloads": 26076,
                    "uri": "ArchiveXL-4198-1-16-4-1722900467.zip",
                    "version": "1.16.4"
                }
            ]
        }
    }
}
```

etc.

(The UID input was reconstructed from `FileId: 84182`, `GameId: 3333`; from response in collection slug `iszwwe` revision `469`)

So I believe it's better to reconstruct `Uid` here; as this in turn allows us to make use of the V2 APIs for further querying (if we ever need it).

Since this is technically implementation detail, I did ask backend about this a while back when doing my own research for Reloaded3 stuff. They don't intend to change this, and in fact; it appears that it's now also sneakily documented in the GraphQL API site.

image

src/Networking/NexusMods.Networking.NexusWebApi/NexusModsLibrary.cs
```diff
@@ -71,33 +71,34 @@ public NexusModsLibrary(IServiceProvider serviceProvider)
...
+        if (revisions.TryGetFirst(r => r.RevisionNumber == revisionNumber, out var revision)) 
+            return revision;
         
+        var info = await _gqlClient.CollectionRevisionInfo.ExecuteAsync(slug.Value, (int)revisionNumber.Value, true, token);
         using var tx = _connection.BeginTransaction();
+        
         var db = _connection.Db;
+        var collectionInfo = info.Data!.CollectionRevision.Collection;
+        var collectionTileImage = DownloadImage(collectionInfo.TileImage?.ThumbnailUrl, token);
+        var avatarImage = DownloadImage(collectionInfo.User.Avatar, token);
```
Sewer56, line 91: There should be:

- A way to download the asset at a later date if it is missing.

And ideally also:

- A retry policy for blob (image) downloads.

In periods of high traffic to the site; for example, the Fallout TV release earlier this year, the website can become unresponsive if the load scales past what our regular infrastructure can handle.

In such cases, it's necessary that these assets can be acquired at a later date.
This isn't something that has to be done 'right now' of course, so just open an issue before merging if you'd prefer to do it at a later date.

Sewer56 (changes requested):
Just disambiguating the ID field (renaming and/or applying my suggestion).
After that feel free to merge. Make an issue if you don't want to error handle `DownloadImage` right now.

halgari:
I'm going to create issues for both of these. I agree that moving to UIds for files is a good thing, but that currently requires also rewriting how we read and load mod files as the V1 API doesn't return this information and V2 doesn't provide a way of pulling file data. So basically we need to rewrite all the query code in NexusModsLibrary. Not a ton of work, but I want to do that separately.

Sewer56:
I was napping while this got merged.

Anyway, can you please rename the field (to `Id` ideally) or change the description (to make unexpected relationship explicit) to make sure it's not misleading? Just push direct to main if needed, it's fine.

PR #2088, opened 2024-09-25
Collection Cards layout fixes

Al12rs:
Fixes layout and FlexPanel usage to better handle changes in description size and author name.
There will likely be more fixes needed. It seems that the starting layout taken from figma didn't really cover these cases well


halgari (approved):
LGTM, others may want to give it a once over though

PR #2092, opened 2024-09-26
[2/4] Added: Caching system for updates.

Sewer56:
This is 2nd PR in chain. Continues from:

- #2072

Should be merged after that PR, this will reduce the diff from main to only changed parts.

Next in chain:

- #2095

Or ***final PR*** if you want to review all parts as a collective at once:

- #2113 [TODO]

------------

# Summary

This PR implements a class called `PerFeedCacheUpdater` (handles single game feed) and `MultiFeedCacheUpdater` (handles multi game feeds at once), including associated logic.

These classes implement the section `1. Determine Updated Mod Pages` of the mod update documentation.
You can find the exact details of the XML implementation in the documentation attached to the code.

This code essentially determines the actions we need to take for each mod page with. 
Which is either: 
- `update last checked timestamp`
- `fetch updated data from API`
or 
- `default/unknown`

The last state should be treated with fetch from API, but as it's technically a possible error state, it is output separately for future logging purposes.

## Approach

Due to current uncertainty in terms of API availability, and backend's plans; this code has been very carefully implemented to be 'generic', as to make it easy to adapt between V1 (REST API) and V2 (GraphQL) API.

The entirety of the code is self contained, and makes minimal assumptions; because we simply don't know what backend's future plans are.

Rough requirements are:

1. We display updates in library.
2. Library shows mods for all games.
3. Therefore we must efficiently check for updates for all mods.

With that in mind, we take the following assumptions:

- We have a list of items (locally) which have a `last update time` and `uid` for mod (`GameId+ModId` tuple).
- We receive results from an external query that returns items with a `last updated timestamp` and `uid`.

These assumptions are as minimal as possible; and use traits (interfaces) to allow easy integration with external code.

This makes it easy to integrate both `V1`'s existing `updated.json` endpoint or in the future, a V2 API which may, for example return results for items from multiple feeds (games) at the same time. The code is able to handle results of cross-game queries without an issue.

Likewise, the code can work with both a V1 `updated.json` style result, or the result of asking a hypothetical V2 API the last update types of 5000 mods with arbitrary `uid`(s). Whatever backend does, this should be able to adjust under the assumption some API will return a `last updated timestamp` and `uid`.

For results from the V1 API, we would manually construct the `uid` by combining the `GameId+ModId`. For V2 we would have that info from the API request.

------------

Note: The successive PR integrates the `GameId`, `ModId` and `Uid` types into the abstractions library:
- #2095

Because doing that has effects on external code, I kept that out of this PR to make reviewing easy. Following PR uses types from abstractions.


halgari:
Why do we need this instead of coding against V2 (GQL) from the start? If we are ingesting from the APIs and inserting that data into our local DB, we don't need this, we'll just change how we ingest and process data before inserting it into the DB.

src/Networking/NexusMods.Networking.ModUpdates/PerFeedCacheUpdater.cs
```diff
@@ -0,0 +1,155 @@
...
+    ///     The max for Nexus Mods API is 1 month.
+    /// </param>
+    /// <remarks>
+    ///     In order to ensure accuracy, the age field should include the lifespan of
+    ///     the <see cref="PerFeedCacheUpdater{TUpdateableItem}"/> as well as the cache
+    ///     time on the server's end. That should prevent any technical possible
+    ///     eventual consistency errors due to race conditions. Although
+    /// </remarks>
+    public PerFeedCacheUpdater(TUpdateableItem[] items, TimeSpan expiry)
+    {
```
halgari, line 56: If we go off of MnemonicDB cache items, we can replace this class with a query over items that are older than a certain value. No reason to keep all this in-memory

src/Networking/NexusMods.Networking.ModUpdates/Structures/ModId.cs
```diff
@@ -0,0 +1,10 @@
+using TransparentValueObjects;
+namespace NexusMods.Networking.ModUpdates.Structures;
+
+/// <summary>
+/// Represents the unique ID of an individual mod in the Nexus Backend.
+/// This <see cref="ModId"/> is specific to the <see cref="GameId"/> it belongs to;
+/// forming a composite key. 
+/// </summary>
+[ValueObject<uint>] // Do not modify. Unsafe code relies on this.
+public readonly partial struct ModId { }
```
halgari, line 10: why not use the existing ModId?

halgari (changes requested):

src/Networking/NexusMods.Networking.ModUpdates/Structures/ModId.cs
Sewer56: I did that in the following PR. This was noted in the opening post.

I didn't do it here because some of the changes required would touch external code, meaning a lot of files would get changed in small ways.

src/Networking/NexusMods.Networking.ModUpdates/PerFeedCacheUpdater.cs
Sewer56: Can you link me to the relevant API for this? I'm curious actually.

Sewer56:
> Why do we need this instead of coding against V2 (GQL) from the start? If we are ingesting from the APIs and inserting that data into our local DB, we don't need this, we'll just change how we ingest and process data before inserting it into the DB.

The purpose of this code is to determine whether we need to fetch and store a newer version of a given mod page in a generic way. More specifically, so we can reduce the number of API calls required against the Nexus Web API.

V1 has an API to get the last updated timestamp for a given game feed within a timeframe (max 1 month) [`updated.json`].
It is very efficient server side because the results are cached for 5 minutes.

V2 does not have a similar API. We don't know what backend intends to replace it with.
The closest thing the V2 API currently has is `modsByUid`; which fetches an entire mod page and has a max of 50 results per request.

What this means is that if the user has 5000 mods, we would have to send 80 API calls to the Nexus Web API. Those API responses would be rather big, are not anywhere close as cacheable and thus be expensive. And I do mean expensive, if you think of all the API calls made by the App, most of them will consist of checking for updates on boot, and at intervals; without a caching system, it may aswell be over 95% of all API calls made over a longer lifespan. Massive load on API. If we shipped the App to millions of users doing that, backend would probably be upset, and I do not wish to be blamed.

This PR provides the mechanism that allows us to minimise the amount of calls we need to make externally to Nexus; by determining what we need to fetch with the `modsByUid` V2 endpoint.

The `Update` method accepts the results of an API call which returns timestamps for files en masse. How we get those timestamps is irrelevant; I'll work with whatever backend provides us. For now that's V1's existing `updated.json` endpoint.

The abstractions make that distinction minimal; if V2 API implements a method equivalent to `updated.json` that gives us timestamps for a lot of files in batch, I can simply plug it in here. With the abstractions it should give-take an hour to add the API call and plug it in. Whatever backend decides to do with V2, whether it be an equivalent endpoint, an optimized path in `modsByUid` for timestamps, a new API or otherwise; we'll be able to work with it without changing existing code. Get the result, wrap in struct, plop it in, use that as-is.

I could even use `modsByUid` for the timestamps themselves if we really-really want to be strict about V2 only, in this case the cache would end up skipped for the time being since we would have pulled a newer version 'just now' so all items would be within the expiry by definition.

The importance of this PR is this. I can now go to backend and say '*these are my exact requirements*' (i.e. efficient mass batch query on timestamps); and plug in whatever they decide to come up with within an hour. Saying 'I am ready to use what you give me, today, right here, right now' inspires confidence and is more likely to help pushing a needed API on backend towards implementation. Likewise, if backend wants to experiment on real workloads, updating on App end would be very easy.

src/Networking/NexusMods.Networking.ModUpdates/PerFeedCacheUpdater.cs
halgari: We already have `NexusModsFileMetadata` and Mod metadata, we can extend those models to include the last updated date. then do a `db.Datoms(NexusModPageMeatadata.LastUpdated)` and do a where on that to check for those that are older than a given date

Sewer56: Ah, the relevant research part for that is `1. Determine Updated Mod Pages`.
Let's break down the problem (concisely).

> [!NOTE]  
> V2 API fetches max 50 items at once.

# Core Issue

1. User has a lot of mods
2. Lot of mods == lot of API calls
3. We should reduce API calls.

## Proposed Alternative

Filter out (exclude) items more recent than a set TimeSpan, and check only the mod pages older than that.

### Effect

This throttles the update check rate; leading to a reduction of API calls.
Reduction is proportional to throttle rate.

### Cache Hit Rate

> [!NOTE]  
> Assuming 300 app launches in a month, i.e. 10 per day.
> And 1000 mods.

At 4 week throttle: (1000 / 50) API Calls
At 2 weeks throttle: (2000 / 50) API Calls
At 1 week throttle: (4000 / 50) API Calls

### Side Effects

Updates for mods get delayed for a while.

Is this a problem?

Normally no, BUT if a game gets updated, breaking mods, the user may require to see mod updates in a timely manner.

This is a concern for big games, for example Baldur's Gate 3. If a user obtains the game from a store that does not keep older versions (e.g. Epic, Xbox) and don't do a full game backup, they're cooked.

Alternatively if a mod author or server informs users of an update, the user should not be expected to use a 'fetch updates manually' button, which isn't part of the designs either.

## Current Solution

Obtain a list of last updated timestamps for installed mods via the API.
Process results via rules described in `1. Determine Updated Mod Pages`.

### Effect

- Updates can be checked on a more regular basis (e.g. every app boot).
    - Time to detected update may be 1 hour (if we refresh every hour), as opposed to 1/2/4 weeks.

- Requires a batch timestamp API call to backend. (Available in V1, not yet in V2)

- Switchover from V1 to V2 API is cheap (i.e. 1 hour to define the query, make wrapper struct, test code).

- Website can cache query for updated items; making computation cheap and cache hit rates high.
    - Currently V1's API caches for 5 minutes before recomputing.
    - Because everyone does same query, nothing needs recomputed on per-game basis until 5 minutes have elapsed.

### Side Effects

Currently only has V1 API. V2 API is a request (ticket) to backend.
However, with this implementation, constraints (requirements) for ticket to backend are well defined.

### Cache Hit Rate

> [!NOTE]  
> Assuming 300 app launches in a month, i.e. 10 per day.
> And 1000 mods.

***Initial Fetch***:

- (1000 / 50) API calls *if never checked before* or *all items are expired* (user hasn't launched App in 1 month).

***Subsequent Fetches***:

(Where `$NumGames` is the number of game feeds we downloaded mods from)
(Where `$NumToUpdate` is number of mod pages updated since we last checked)

- `$NumGames` API calls (typically 1 or 2) if last update check was ran less than `expiry` (currently 1 month).
- (`$NumToUpdate` / 50) API calls (typically 0/1) for any mods which have been updated since last check.

Assuming an extreme unrealistic scenario where we check for updates every hour, and the user has the App open for 12 hours a day, this means 336 API calls over a month. (+1 for Number Of Times any Mod Received an Update)

-------------

[Apologies for late response, was preoccupied, then quiz night happened]

halgari: The issue I have with all of this is that it's overkill for the current ticket. I would like to have a conversation with the backend team, and others before deciding on the proper way to pull and query updates. 

For now all we need is:

1) a way to "check for updates" we can make this a button, or a single API call, with the number of users/mods we currently work with this isn't an issue. 
2) as mentioned in the initial kickoff meeting it's completely fine to do this check at startup as we're still in the early development phase
3) we can then write a mnemonicDB query that finds possible update candiates 
4) we can display those updates to the user in the UI

The nice thing about this approach is that it segments out the different parts. We already store mod metadata in MDB, we already query it when building our datagrids, so the whole project breaks down into 3 clean parts:

1) looking for updates in MDB
2) performing an update given two `NexusLibraryFile` instances
3) updating the cached data in MDB using some critera

We don't need it to be generic, split off, or even worry about the number of API calls at this point. Once we get the above built it will be simple enough to talk to the backend and get a implementation for whatever update query logic we would like to have.

halgari (approved):

PR #2101, opened 2024-09-26
Non downloaded library items

halgari:
This delinks-the `NexusModsLibraryFile` models from `DownloadedFile` and instead links them to `LibraryItem`. A few reasons we discussed this morning for why we may want to do this:

1. Updates for mods will not yet be downloaded, but we still what to show them in the UI
2. Collections will reference files that are not yet downloaded. 

In general it would be helpful if the `Library` grid was a generic way to display any sort of file we know about, we could then filter this grid by downloaded, not downloaded, required, has update, game, etc. The alternative option is to re-implement parts of the grid for each page. But I think in general we'd like it if we could simply display the same UI control everywhere and use different filters

A few issues with the design in this file:

1. We do not currently have a way to add a Nexus file without also starting the download
2. If we somehow did add a nexus library item and not start the download there would be no way to start it (we need to implement the `download` control for these rows in the library
3. Should we then re-name `NexusModsLibraryFile` to `NexusModsLibraryItem`?

This PR exists mostly as a proof-of-concept, to show how simple it would be to swap over to a model where Nexus files don't have to exist on disk to be in the library.


src/Abstractions/NexusMods.Abstractions.NexusModsLibrary/NexusModsLibraryFile.cs
```diff
@@ -9,7 +9,7 @@ namespace NexusMods.Abstractions.NexusModsLibrary;
 /// Represented a <see cref="DownloadedFile"/> originating from Nexus Mods.
 /// </summary>
 [PublicAPI]
-[Include<DownloadedFile>]
+[Include<LibraryItem>]
 public partial class NexusModsLibraryFile : IModelDefinition
```
erri120, line 13: Comment should be updated. We can rename this to `NexusModsLibraryItem`.

erri120 (changes requested):
For `NexusModsLibraryFile`, which we can rename to `NexusModsLibraryItem`, we should be able to just manually add the `LibraryFile` attributes when we download the item.

halgari:
agreed, I'll refactor this next to break apart adding the LibraryFile attributes when the download completes

PR #2102, opened 2024-09-26
Switch GraphQL calls to fragments

halgari:
Switches GraphQL calls over to using fragments. 

With our GQL library (StrawberryShake), any fragments become C# interfaces. This then allows us to reuse code across various GQL calls by casting subsections of the restults to a fragment interface then calling methods to upsert the data. 

Nothing semantically changes here, just a reorganization of the code.


Sewer56 (approved):

Al12rs:
Tests seem to be needing fixing

Issue #1695, opened 2024-06-27
Support GOG and Heroic Launcher on Linux


erri120:
Related to #308.

- native GOG alternative: [Heroic Launcher](https://heroicgameslauncher.com/) which is cross-platform.
- GOG running in WINE
  - GameFinder already supports finding GOG inside default WINE prefix and Bottles, issue is launching the game

LukeNexusMods:
We will be doing this. Linux users like drm free games. 

Florian can work with game finder to achieve this work. 

Turned it into a spike to investigate new ways of playing games through linux. Could be a large amount of work but could be smaller depending on the outcome.

Sewer56:
[2025 Issue Cleanup]
Last part is dependent on https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/issues/4411 . 
Aside from that we're basically there.

PR #2103, opened 2024-09-28
Find GOG games installed with the Heroic launcher

erri120:
Part of #1695.

There is an issue with running the game through heroic, as heroic never exists. We'll probably want to track WINE.


Al12rs (approved):
Seems good to me, but I didn't try it out. If the changes in game finder are unlikely to introduce regressions we can go ahead an merge.

PR #2106, opened 2024-09-30
Skip running redmod on non-Steam installations on Linux

erri120:


Sewer56 (approved):
Presumably for development/temporary purposes, till we figure out Heroic's WINEPREFIXes, works for now

Issue #2091, opened 2024-09-25
Switch to GQL for all Nexus Mods queries (or the ones we can)


halgari:
# Improvement

## Currently 
Collections use the V2 API (GQL) and the rest of the mod library uses the V1 API. We should unify this behavior. This will allow us to reduce the calls we make by batching requests. But it will require reworking 2 core methods in NexusModsLibrary.cs

Sewer56:
Originally noted in this PR feedback: 
- https://github.com/Nexus-Mods/NexusMods.App/pull/2081#discussion_r1775624794

It's not directly correlated, but this issue was created as a response to that comment.

PR #2113, opened 2024-10-02
[4/4] Query for Updates from Nexus, and Fix Miscellaneous Bugs in DataStore

### LINKED ISSUE #2008: Refresh remote metadata
Need to figure when and how we refresh. Whether that's on command, periodically, or both.

Sewer56:
This is last PR in chain. Continues from:

- #2092

Review this if you want to review all parts at once, else review by parts sequentially.

----------

Also contributes a bit towards:

- https://github.com/Nexus-Mods/NexusMods.App/issues/2091#issue-2549044273

----------

# Summary

This PR adds the final puzzle pieces towards efficient checking for updates; and also
a little bit of work towards migration to V2 that was required as part of this.

## Changelog

- Marked various uses of V1 GameDomain(s) as 'obsolete' and 'for removal'.
- Fixed invalid property names in a certain existing API call.
- Added MnemonicDB Attributes for many of the V2 API keys. (`UidForFile`, `UidForMod`, `GameId`).
- Added two-way translations for `GameId` and `GameDomain`. This is temporary, and helps us iteratively
  migrate towards full usage of V2.
- Adds a simple set of methods for fetching update info.
    - 1. Check which mod pages are out of date.
    - 2. Update out of date mod pages.
    - 3. Show/filter newer files than installed.
- Adds a round trip test for converting back to V2 API `uid` strings.
- Adds a simple test (`UpdatingModPageMetadata_ViaWebApi_ShouldWork`) for fetching file updates, which:
    - Creates a Loadout
    - Downloads and Installs a Mod
    - Checks for updates.
    - Verifies expected minimum number of updates.

The last mentioned test will be further improved in the future; with mocks.
For now this is left out in order to reduce amount of work done in long run.
i.e. Waiting for new APIs from backend, and not mocking anything V1.

### Extra: Use `uid` as primary key of `NexusModsFileMetadata`

This makes it so the primary key uniquely identifies a file on Nexus.
Previously, the code used `FileId`, however `FileId` is *game specific*; 
multiple games may have a file with same `FileId`.

This actually fixes some bugs. To give an example, if you installed Mod 1000 in Stardew Valley,
installing Mod 1000 from Cyberpunk2077 as part of a collection will no longer assume you
have that specific mod installed.

## Some Notes

We currently use the time the mod page itself was last updated for determining if we should
refersh local mod page info. This is due to a missing field in the V2 API. Trivial 1 line change
once we do get it.

fixes #2095
fixes #2092
fixes #2008


src/Networking/NexusMods.Networking.ModUpdates/PerFeedCacheUpdater.cs
```diff
@@ -0,0 +1,155 @@
...
+/// Within the Nexus Mods App:
+///
+/// - 'Input' is our set of locally cached mod pages.
+/// - 'Update' is our results of `updated.json` for a given game domain.
+/// - 'Output' are the pages we need to update.
+///
+/// The 'Feed' in the context of the Nexus App is the individual game's 'updated.json' endpoint;
+/// i.e. a 'Game Mod Feed'
+/// </remarks>
+public class PerFeedCacheUpdater<TUpdateableItem> where TUpdateableItem : ICanGetLastUpdatedTimestamp, ICanGetUidForMod
```
halgari, line 34: This code project is overkill for the task at hand. Instead implement what we can on the V2 API and put that code inside the NexusWebApi project.

When we over-abstract like this we endup with a lot of code that has to be assembled to do something that is only ever used with one set of data. We have a "Per feed updater" that can update any number of things, mixins for Mod and Page metadata, `ICanGet...` interfaces. All that can be reduced to a few small method in the API class if we assume that what we're caching is mod info from the V2 API

halgari (changes requested):
Looks pretty solid aside from the above comment.

Sewer56:
I made a small improvement here.

This code now also closes:
- #2008

I also:

- Merged the two interfaces to simplify the code
- Moved `RunUpdateCheck.cs` to the `NexusWebApi` project.
- Migrated the API call to get mod info to use `GameId` rather than `GameDomain` (1 step closer to V2 integration)

halgari (approved):

PR #2119, opened 2024-10-02
Moved tooltip out of border

### LINKED ISSUE #1822: Double tooltip on "Add tab" button
When hovering on the add tab button you will see the correct tooltip with another lower opacity tooltip behind it. See screen shot:

Image

MistaOmega:
I didn't expect this to work, but here we are...

Closes #1822 

image
image


Sewer56 (approved):
I suppose this works for the time being.

Issue #1667, opened 2024-06-20
Collections - Feature List [WIP]


Pickysaurus:
This ticket is to keep a running list of features/considerations for adding Collections to the app. This may be broken up into separate tickets when work on this feature starts.

# Downloading Collection Metadata
- [ ] Interpreting the `nxm://` links containing a collection payload.
- [ ] Requesting the collection bundle from the Graph API.
- [ ] Inspecting the collection bundle to identify the manifest file.
- [ ] Inspecting the collection bundle to identify any supplementary metadata.
- [ ] Inspecting the collection bundle to identify any bundled mods. 
- [ ] Show the pre-install instructions.
- [ ] Check the user's game version against the one used to make the collection. 

# Downloading Collection Mods
- [ ] Resolve a downloadable file on Nexus Mods for "Exact only".
- [ ] Resolve a downloadable file on Nexus Mods for "Prefer Exact".
- [ ] Resolve a downloadable file on Nexus Mods for "Latest".
- [ ] Resolve a downloadable file from an external website based on a direct link.
- [ ] Resolve a downloadable file from an external website where the user must browse to start the download.
- [ ] Import mods bundled with the collection.
- [ ] Handle errors resolving downloads.

# Installing Collection Mods
- [ ] Understand the difference between a required and optional mod.
- [ ] Understand the concept of "Phases" when installing mods. (An apply action must be completed before any mods in the next phase can be installed).
- [ ] Auto-select FOMOD options based on the choices saved in the collection (where applicable).
- [ ] Replicate the file structure exported with the collection for a specific mod.
- [ ] Apply binary patches to edited mod files.
- [ ] Handle errors resolving customised mods.
- [ ] Present instructions for mods to the user.
- [ ] Allow the user to review optional mods and select the ones they want to use. 

# Installing Collection Metadata
- [ ] Check collection-specific functionality for all games supported by Vortex.
- [ ] Apply rules for mod asset priority.
- [ ] Apply plugin load order rules or static list for load order (varies by game).
- [ ] Apply INI Tweaks (Bethesda games).

# Updating Collections
- [ ] Check for updates.
- [ ] Download the new collection bundle.
- [ ] Compare the installed revision to the new revision.
- [ ] Display the Collection changelogs to the user.

# Social features
- [ ] Allow rating of a collection.
- [ ] Allow endorsing of a collection.
- [ ] Allow the user to browse the collection page to report a bug/leave a comment.

[...]

PR #2120, opened 2024-10-03
Binary patching and FOMOD Presets

halgari:
This all ended up being both harder and easier than I expected. 

**Binary patching** - it's implemented and works on the unit tests I included in the code. It's a bit harder than one would expect as often binary patching is paired with `replicate` that uses a MD5 hash to identify files to install to specific locations. Here's the catch, the MD5 in this case is the MD5 after the patch, *and*  no one ever thought to provide a link between MD5 hashes and patches. *And* the patch's hash is a CRC32, because why not?

This all means that you kindof have to hash everything in the mod, then apply all patches, then hash any patched files and *then* you have what you need to install the files. In our case we store the patched files and then link them up by xxHash like any other file. 

**FOMOD Presets** - this ended up being harder than I thought thanks to the utterly broken `fomod-installer` project. In the end I gave up on trying to pass the options into the installer. Instead we implement a new `IGuidedInstaller` and emit `UserChoice` objects based on the presets. The resulting solution is quite clean, just took me a good 4 hours to figure it out.

TL;DR - two of the biggest blockers with collection installs are now solved.


Pickysaurus:
Related https://github.com/Nexus-Mods/NexusMods.App/issues/1667

src/NexusMods.Collections/Verbs.cs
```diff
@@ -42,4 +48,35 @@ private static async Task<int> InstallCollection([Injected] IRenderer renderer,
 
         return 0;
     }
+    
+    [Verb("gather-collection-definitions", "Downloads all the collection definitions for a given game, and extracts them to a folder")]
```
Sewer56, line 52: Is this debug functionality? I don't see this used anywhere in the App or tests.
I think having this is fine, but it may be worth it to disable this from release builds; so the CLI is not cluttered for end users.

src/Games/NexusMods.Games.FOMOD/CoreDelegates/PresetGuidedInstaller.cs
```diff
@@ -0,0 +1,50 @@
...
+    public void CleanupInstaller()
+    {
+    }
+
+    public Task<UserChoice> RequestUserChoice(GuidedInstallationStep installationStep, Percent progress, CancellationToken cancellationToken)
+    {
+        var step = _steps[_currentStep];
+        
+        // This looks gross, but it's fairly simple we map through the two trees matching by name, and it's cleaner than 4 nested loops.
+        var choices =
```
Sewer56, line 38: Do we know if the names are case sensitive? I imagine the preset choices aren't set by humans, but if they can be modified by humans, a case insensitive comparison may be better here.

src/NexusMods.Collections/InstallCollectionJob.cs
```diff
@@ -196,6 +266,77 @@ await Parallel.ForEachAsync(archive.Children, async (child, token) =>
         return result.Remap(group);
     }
 
+    /// <summary>
+    /// This will go through and generate all the patch files for the given archive based on the mod's patches.
+    /// </summary>
+    private async Task CreatePatches(Mod modInfo, LibraryArchive.ReadOnly modArchive, LibraryArchive.ReadOnly collectionArchive,
```
Sewer56, line 272: Could this function please be renamed?
The description and title is misleading.

When it comes to binary diffing the common terminology is:

- **Input:** (Source/Original) file
- **Diff:** (Diff/Patch) File
- **Output:** Output/Target File

The existing name `CreatePatches` implies that we're creating the actual `bsdiff` patches themselves; this is not the case; we're applying an existing `bsdiff` patch, and storing the resulting patched files in the `FileStore`.

-------------

Misc note: The design sucks here.

> and no one ever thought to provide a link between MD5 hashes and patches. And the patch's hash is a CRC32, because why not?

I feel your pain; this sucks.

src/NexusMods.Collections/Verbs.cs
halgari: Just for testing, I'll add some comments

Sewer56 (changes requested):
The other feedback is mainly up to whatever you want to do with it; however this piece is blocking.

https://github.com/Nexus-Mods/NexusMods.App/blob/6444ae4562124c59bcd1751d0af2de03fca3c987/src/Abstractions/NexusMods.Abstractions.IO/IFileStore.cs#L79-L89

I have a small number of issues with this:

- Every chunked size read goes through the entire caching mechanism, which involves a O(N) lookup in `LightweightLRUCache` for every `ChunkSize` (usually 1MiB).
    - Because we're always reading forward (`ReadToEnd`), ***every read is a cache miss***, so we sequentially iterate 16 keys every 1MB read, none of the keys hit, and we rent a new set of bytes from the pool each time.
    - `LightweightLRUCache` (capacity 16) always rents from the `MemoryPool` when we add a new cached chunk.
    - Because we have `16 * NumThreads` active rentals, number of rentals from global is likely to run out on high core count processors.
    - I believe the rentals are per core, ([see runtime code](https://github.com/dotnet/corert/blob/c6af4cfc8b625851b91823d9be746c4f7abdc667/src/System.Private.CoreLib/shared/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs#L36)), we probably shouldn't use more than 8 at once if using the global pool otherwise every pool rental is a full allocation. (i.e. Reduce `LightweightLRUCache` capacity to 7, or to 8, and reuse rentals)
    - Because every read is a cache miss, every rental (1MB read) will become a full allocation. 
    - ***tl;dr 1MB read after 8MB of file is a full allocation; and every read is a cache miss that iterates over 16 keys.***

- This will always run on a single thread, when larger files could be extracted in parallel.
    - If we need control over thread count, that can be a parameter.

***I feel very uneasy throwing this specific part to the main branch.*** At least as part of the `IFileStore` interface.

-----

Misc Note:

Bsdiff patches can also technically be bigger than 2GiB in file size, but that point is very moot; most code implementations limit files to 2GiB even if format supports it; and nobody makes patches for files this big because memory usage is astronomical (17 times the input size). Technically with diff algorithms you should be using streams in general; however in this case a full read to RAM is unavoidable since we need to read the file twice to hash it. Without very custom code.

Sewer56 (approved):

Sewer56:
Ah, actually, before you merge, you may want to have a way to adjust thread count in parameter. Nx will default to number of physical cores.

PR #2128, opened 2024-10-04
Corrected tooltip foreground

### LINKED ISSUE #2123: Hover tooltip on Apply button does not match other styles
Image

MistaOmega:
Closes #2123 

Just a foreground mismatch.
Another issue with the style-inheritance, like seen on #1822 i think

image

### DIFF
--- a/src/NexusMods.App.UI/LeftMenu/Items/ApplyControlView.axaml
+++ b/src/NexusMods.App.UI/LeftMenu/Items/ApplyControlView.axaml
@@ -41,7 +41,7 @@
                     Height="44"
                     x:Name="ApplyButton">
                 <ToolTip.Tip>
-                    <TextBlock Text="{x:Static resources:Language.ApplyControlView_Apply_ToolTip}" />
+                    <TextBlock Text="{x:Static resources:Language.ApplyControlView_Apply_ToolTip}" Foreground="White" />
                 </ToolTip.Tip>
                 <StackPanel Orientation="Horizontal"
                             HorizontalAlignment="Center"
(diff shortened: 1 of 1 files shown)


src/NexusMods.App.UI/LeftMenu/Items/ApplyControlView.axaml
```diff
@@ -41,7 +41,7 @@
                     Height="44"
                     x:Name="ApplyButton">
                 <ToolTip.Tip>
-                    <TextBlock Text="{x:Static resources:Language.ApplyControlView_Apply_ToolTip}" />
+                    <TextBlock Text="{x:Static resources:Language.ApplyControlView_Apply_ToolTip}" Foreground="White" />
```
Al12rs, line 44: We are using a set of color aliases found in `src/Themes/NexusMods.Themes.NexusFluentDark/Resources/Palette/Colors/ElementColors.axaml` for the colors.
We want to avoid using hardcorded colors like "White" or "Red" to allow us to change them from the theme all throughout the app easily.

In this case the easiest way to get what we want is to just apply the "ForegroundStrong" style class to the `TextBlock`, which will in turn apply `NeutralStrong` as the foreground color.
```suggestion
                    <TextBlock Text="{x:Static resources:Language.ApplyControlView_Apply_ToolTip}" Classes="ForegroundStrong" />
```

Al12rs (changes requested):

src/NexusMods.App.UI/LeftMenu/Items/ApplyControlView.axaml
MistaOmega: Hmm, I tried this initially to no avail.

I'll have to look in to why that's happening, it's why I ended up settling for forcing it to white

Al12rs: As an alternative you can set it like this:
```suggestion
                    <TextBlock Text="{x:Static resources:Language.ApplyControlView_Apply_ToolTip}" Foreground="{StaticResource NutralStrong}" />
```

Al12rs (approved):

Issue #1733, opened 2024-07-08
Update Health check menu badge, icon and colours


captainsandypants:
### Currently

The badges on the left menu item "Health check" are all the same colour:

![Image](https://github.com/Nexus-Mods/NexusMods.App/assets/118466382/2e66d015-633e-4d0c-8b57-0ff8240f8052)

They also don't increase in width when needed for increased digits:

![Image](https://github.com/Nexus-Mods/NexusMods.App/assets/118466382/8ba0bcb9-d5fd-41d1-a8ff-2d90b31ebf56)

The "Suggestion" icon is currently a question mark, which isn't the best representation of what it's for. Change to info icon.

### We want

To show the badge in the same colour as the colour used for the corresponding severity level and to use the corresponding icon rather than a count.

Examples of how these would stack below:

(The highest severity will always be the furthest to the right)

![Image](https://github.com/user-attachments/assets/87957927-fc52-4aad-b163-0f329ea15b06)

Change the "Suggestion" question mark icon to the info icon.

[Figma link](https://www.figma.com/design/hsUiqro210T3tsK1F9l1fQ/%F0%9F%93%B1%F0%9F%A9%BAHealth-Check?node-id=1-2954&t=lGfuPVg9yBIUovdd-1)

PR #2133, opened 2024-10-07
Fix for badge width with larger numbers

insomnious:
Part of #1733 

- Removed badge fixed width
- Added quick design view model


Al12rs (approved):
Didn't test in action but code makes sense

Issue #2141, opened 2024-10-08
After Mass install, Library View does not update all installed entry statuses


LukeNexusMods:
When holding shift, scrolling down to select all items in the library and then clicking add to install them, the view does not update when the items are installed. You have to navigate away from that page to then get it to update. 
Some items appear to update that are in view but any out of view looking like they need to be installed as the install button is visible.

Issue #2140, opened 2024-10-08
Mass delete holding shift and scrolling to select all does not delete


LukeNexusMods:
When holding shift and selecting a large amount of mods to delete from the library, the popup appears, click delete, nothing is deleted.

LukeNexusMods:
Fixed in latest build

PR #2146, opened 2024-10-08
Fix off-screen row updates in the Library

halgari:
The install/installed status of a row is dependent on the number of connect loadout items. `ObserveCountChanged()` does not notify of the current count on a subscription (by default). So when a row was re-activated the installed status of the row defaulted to the status when it was last seen on screen, and would only update if the status changed again in the future. This PR adds `notifyCurrentCount:true` to these invocations to fix this problem. 

Also updates `GetCreatedAt` so it no longer crashes Rx chains during deletion of entities. 

fixes: #2141 and #2140

### DIFF
--- a/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
+++ b/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
@@ -54,7 +54,7 @@ public abstract class TreeDataGridAdapter<TModel, TKey> : ReactiveR3Object
                 .AddTo(disposables);
 
             self.Roots
-                .ObserveCountChanged(true)
+                .ObserveCountChanged(notifyCurrentCount: true)
                 .Subscribe(self, static (count, self) => self.IsSourceEmpty.Value = count == 0)
                 .AddTo(disposables);
 
--- a/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
+++ b/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
@@ -54,7 +54,7 @@ public abstract class TreeDataGridAdapter<TModel, TKey> : ReactiveR3Object
                 .AddTo(disposables);
 
             self.Roots
-                .ObserveCountChanged()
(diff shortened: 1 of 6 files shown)


src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
```diff
@@ -54,7 +54,7 @@ protected TreeDataGridAdapter()
                 .AddTo(disposables);
 
             self.Roots
-                .ObserveCountChanged()
+                .ObserveCountChanged(true)
```
erri120, line 57:
```suggestion
                .ObserveCountChanged(notifyCurrentCount: true)
```

src/NexusMods.App.UI/Pages/LibraryPage/FakeParentLibraryItemModel.cs
```diff
@@ -33,7 +33,7 @@ public FakeParentLibraryItemModel(LibraryItemId libraryItemId, IObservable<IChan
             model.NumInstalledObservable
                 .ToObservable()
                 .CombineLatest(
-                    source2: model.LibraryItems.ObserveCountChanged(),
+                    source2: model.LibraryItems.ObserveCountChanged(true),
```
erri120, line 36:
```suggestion
                    source2: model.LibraryItems.ObserveCountChanged(notifyCurrentCount: true),
```

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutViewModel.cs
```diff
@@ -164,7 +164,7 @@ public LoadoutViewModel(IWindowManager windowManager, IServiceProvider servicePr
                 configureAwait: false).AddTo(disposables);
 
                 // Compute the target group for the ViewFilesCommand
-                Adapter.SelectedModels.ObserveCountChanged()
+                Adapter.SelectedModels.ObserveCountChanged(true)
```
erri120, line 167:
```suggestion
                Adapter.SelectedModels.ObserveCountChanged(notifyCurrentCount: true)
```

erri120 (changes requested):

erri120 (approved):

PR #2147, opened 2024-10-09
Add support for bundled mods during collection installation

halgari:
While doing some other testing I noticed that some collections that have bundled mods were causing the collection installation to crash. This fixes that issue.

Bundled mods another one-off sort of thing: 

* mods can have a `source=bundled`
* mods from this source have a `fileExpression=<string>`
* the files for these mods are found in the collection `.zip` file at the location `bundled/<fileExpression>/...` and follow the standard path format, for example a file in Cyberpunk may be in `bundled/Bundled-mod/archive/pc/mod/foo.archive`


src/NexusMods.Collections/NexusCollectionBundledLoadoutGroup.cs
erri120, line 1: This file should probably go into `NexusMods.Abstractions.NexusModsLibrary`.

erri120 (approved):

src/NexusMods.Collections/NexusCollectionBundledLoadoutGroup.cs
halgari: I considered this, but it's not really part of the library, it links *to* the library, and it's part of the collection system.

PR #2148, opened 2024-10-09
New StandardButton component

### LINKED ISSUE #1876: DRAFT: Create a Button component that matches Figma
# Spike

## Proposal

In Figma, Design has now a matrix of button variations that include numerous styles, types, fills, sizes, icons etc. Designs for other parts of the app already use or will be using these buttons and there is a sort of undocumented conversion that is happening between what is in Figma and what is currently in the app whenever UI is being built. 

image

In [#1833](https://github.com/Nexus-Mods/NexusMods.App/issues/1833) was a quick stop gap to update the appearance of the buttons to match Figma, but not the structure or the long term benefit of doing this 'properly'. This was done by changing some of the original styles in the App without worrying about the naming and everything else to match Figma.

This is basically adding to the tech debt and we are better off staying inline with Design as quickly as we can without making things more hacky.

I'd like to propose that we build a Button component (we need a name, StandardButton?) that will that will account for all of the button variations that are in Figma through a mix of properties and classes. 

We could do this all with classes and we could also do it all with properties but I think there is a natural separation between what uses classes (more style related things like size, fill, type) and properties (show label, what icons are visible, what icon for what position etc.)

### Proof of concept

I worked backwards from what feels right when it comes to using a component, such as dropping `<NewButton>` into a axaml and without adding anything, it should look like the default button in Figma with a single line of code. We can then add properties and classes to get the necessary variation.

Please note: I've not matched paddings\margins\spacing etc to Figma.

#### Properties

| code | example |
| - | - |
| `<controls:NewButton />`  | image |
| `<controls:NewButton Text="Lovely" />` | image |
| `<controls:NewButton Text="Right Icon" ShowIcon="Right" />` | image |
| `<controls:NewButton Text="Left Icon" ShowIcon="Left" LeftIcon="{x:Static icons:IconValues.Steam}" />` | image |
| `<controls:NewButton Text="Both Icons" ShowIcon="Both" />` | image |
| `<controls:NewButton ShowIcon="Left" LeftIcon="{x:Static icons:IconValues.Visibility}" ShowLabel="False" />` | image |

#### Classes

Classes should be used for the more _obvious_ styles, such as colors, sizes, types etc.

We default to type `primary`, fill `none`, size `md`

| code | example |
| - | - |
| `<controls:NewButton Classes="secondary" />`  | image |
| `<controls:NewButton Classes="secondary strong" />` | image |
| `<controls:NewButton Classes="secondary strong sm" ShowIcon="Right" />` | image |
|  `<controls:NewButton Classes="tertiary" />`  | image|
| `<controls:NewButton Classes="tertiary" ShowIcon="Right" />` | image |

### How was it done

This proof of concept was completed by creating a new TemplatedControl (`NewButton.axaml`) with custom properties and default values in the code-behind. A new control theme was then created for this control (`NewButtonControlTheme.axaml`) which contains the actual change in template (see snippet below) as well as all the default styles that use our existing resources. This could then be overidden further using styles if we ever needed to.

```
<ControlTemplate>
    <Button>
        <panels:FlexPanel Direction="Row" AlignContent="Center">
            <icons:UnifiedIcon Name="PART_LeftIcon" />
            <TextBlock Text="{TemplateBinding Text}" Name="PART_Label" />
            <icons:UnifiedIcon Name="PART_RightIcon" />
        </panels:FlexPanel>
    </Button>
</ControlTemplate>
```

### What to do

- Test that the 'extended' button component actually works the same with the backend as the current implementation
- Modify the new ControlTheme to match current standards and expectations within the codebase
- Match the default styles to Figma - paddings, margins etc. and also decide on what defaults to use when no classes\properties are set. @captainsandypants. 
- Build a decent design view to be the single source of truth
- Remove legacy styles that were only temporary\created before Design was onboard. Button styles such as Standard, Rounded, Pill buttons etc.
- Replace these legacy styles throughout the codebase with the new component.

## Impact

- Reduce tech debt
- Reduce code repetition
- Easier to document and communicate
- Easier to use with specific properties and classes that match Figma

## Supporting Information

[Figma](https://www.figma.com/proto/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?node-id=1949-6839&t=Hb4D0IEYkH2ibCed-1)

insomnious:
Closes #1876

This new `StandardButton` component aims to replace our usage of the regular `Button` in the future. Please see the linked issue for a deeper breakdown. Summary of advantages are:

- Reduce tech debt
- Reduce code repetition
- Easier to document and communicate
- Easier to use with specific properties and classes that match Figma

This component maintains backwards compatibility with a regular Button and how it can be used with any type of Content if needed.

The next stage is for me to start using it when I'm building controls\views\styles within the app in order to iron out the kinks and check it's real work usage. This will happen before we start looking at the task of replacing the old Button usage and clearing out unwanted and unneeded legacy styles.


src/NexusMods.App.UI/Controls/StandardButton.axaml.cs
```diff
@@ -0,0 +1,285 @@
...
+                break;
+            case ShowIconOptions.Both:
+                _leftIcon!.IsVisible = true;
+                _rightIcon!.IsVisible = true;
+                break;
+            default:
+                _leftIcon!.IsVisible = false;
+                _rightIcon!.IsVisible = false;
+                break;
+        }
```
Al12rs, line 283: This code can be simplified a bit like this I think:
The idea is that you set the value directly rather than doing it in conditional branches.

```suggestion
        _content.IsVisible = Content is not null;
        _border.IsVisible = Content is null;
        
        _leftIcon.IsVisible = ShowIcon is ShowIconOptions.Left or ShowIconOptions.Both;
        _rightIcon.IsVisible = ShowIcon is ShowIconOptions.Right or ShowIconOptions.Both;
```

src/NexusMods.App.UI/Controls/StandardButton.axaml.cs
```diff
@@ -0,0 +1,285 @@
...
+    {
+        base.OnApplyTemplate(e);
+
+        _leftIcon = e.NameScope.Find<UnifiedIcon>("PART_LeftIcon");
+        _rightIcon = e.NameScope.Find<UnifiedIcon>("PART_RightIcon");
+        _label = e.NameScope.Find<TextBlock>("PART_Label");
+        _content = e.NameScope.Find<ContentPresenter>("PART_ContentPresenter");
+        _border = e.NameScope.Find<Border>("PART_Border");
+
+        if (_leftIcon == null || _rightIcon == null || _label == null || _content == null || _border == null) return;
```
Al12rs, line 242: At first I didn't understand why this was necessary, but I get it I think, because the Template is being set by the Theme or the style, so we can't just use the names directly like we would in `UserControls` code behinds

src/Themes/NexusMods.Themes.NexusFluentDark/Extensions/ThicknessExtension.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace NexusMods.Themes.NexusFluentDark.Extensions;
+
+/// <summary>
+/// Xaml extension for Thickness (used by Margin, Padding, BorderThickness).
+/// This allows setting a Thickness using a resource for each side.
+/// </summary>
+/// <example>
+///     <Border Padding="{extensions:Thickness Left={StaticResource Spacing-2}, Top=0, Right={StaticResource Spacing-2}, Bottom=0}"/>
+/// </example>
+public class ThicknessExtension: MarkupExtension
```
Al12rs, line 15: Uhh exciting. It is going to be a bit wordy to use fur sure, but still better than hardcoded values. 
Updating all occurrences to this might be another fun thing to do during winter holidays.

src/Themes/NexusMods.Themes.NexusFluentDark/Resources/ControlThemes/StandardButtonControlTheme.axaml
```diff
@@ -0,0 +1,457 @@
...
+        <!-- these can also be overriden in NexusMods.Themes.NexusFluentDark -->
+
+        <!-- all button states -->
+        <Style Selector="^">
+            <Setter Property="ClipToBounds" Value="False" />
+            <Setter Property="Margin" Value="0" />
+            <Setter Property="Padding" Value="{DynamicResource ButtonPaddingMedium}" />
+            <Setter Property="CornerRadius" Value="{StaticResource Rounded}" />
+            <!-- Set an invisible border thickness to 1 to avoid size differences with outlined variants. -->
+            <Setter Property="BorderThickness" Value="0" />
```
Al12rs, line 275: I think this might be an outdated comment? We are not doing it and I don't see a difference in size.

I think I saw this problem in rounded buttons specifically before, where the extra border could shift the contents a bit, but with more squared design it should not be a problem I think.

Al12rs (approved):
Just a couple of small suggestions above, but otherwise looks great.

src/NexusMods.App.UI/Controls/StandardButton.axaml.cs
insomnious: This was mimicking how it was done in the Avalonia source and yes, as you've said, it could change by another Theme and so I think Avalonia needs to rebuild the references.

Al12rs (approved):

Issue #1262, opened 2024-04-24
Epic - Game Support - Baldurs Gate 3


LukeNexusMods:

participle38:
"Unpacked pak loose files, folder structure will usually have root in Game/data
Should be discouraged, often used for texture replacements. Might cause file conflicts."

Then users will install them manually and it will be even worse.

PR #2149, opened 2024-10-09
BG3: Add support for serializing and deserializing `modsettings.lsx` Load Order file

Al12rs:
- Part of #1262 

- Add support for serializing a sorted array of `ModuleShortName` into modsettings.lsx format
- Added serialization test
- Add support for deserializing a modsettings.lsx string to sorted array of `ModuleShortName` structs.
- Added deserialization test


Sewer56 (commented):
I think this is mostly fine.
I haven't played BG3, I cannot verify the accuracy of the resulting files; so I'll leave it to a second set of eyes.

It's a bit unusual seeing XML being parsed by hand; the structure present, at least within
the Verify file seems to be representable statically with classes; and a single call to `Serialize` / `Deserialize` like you'd do with JSON. I'll assume you either didn't know, think of it, or something else of the kind.

Sewer56:
If you want to be really fancy about it though, you could use a `struct` and `IDisposable`. Creating the struct does a `WriteStartElement`, and disposing it does an `WriteEndElement`.

That would spare you having to label every `WriteEndElement` and the code structure would closely resemble the XML.

Sewer56 (approved):
I've marked this as 'you can merge it'; but please give it a while so another set of eyes can give their opinion too.

Al12rs:
> C# has some XML formatters for objects as well, probably a few dozen of them, that way we don't need to do the band serialization. That'd require some investigation into what library to use though, so I'm fine with looping back around later to change it.

Yeah I avoided that for now since they changed the format of `modsetting.lsx` a couple of times already, and I found it easier to match whatever they do this way, rather than trying to get a objects of the right shape and formatters for them to do what we need. 
E.g. before `modsettings.lsx` used to have two lists, one with module short names, and another with the load order, containing just a UUID per line. That kind of stuff is easy to change with something like this.

PR #2151, opened 2024-10-09
Library: Use new columns

erri120:
Due to the composition aspect of the new system, there is quite the amount of boilerplate code.

TODO:

- update style selectors


src/Abstractions/NexusMods.Abstractions.NexusWebApi/Types/V2/GameId.cs
```diff
@@ -23,7 +23,7 @@ public static GameId FromGameDomain(GameDomain domain)
     {
         return domain.Value switch
         {
-            "stardewvalley" => (GameId)1704,
+            "stardewvalley" => (GameId)1303,
```
Al12rs, line 26: What is this about?

Sewer56: It was an accident when I introduced the temporary mapping, during copy-pasting. My branch which throws out as much V1 as possible does proper resolution via the website.

I noticed this bug this morning, and @erri120 ran into it later too.

src/NexusMods.App.UI/Pages/LibraryPage/LocalFileLibraryItemModel.cs
```diff
@@ -0,0 +1,101 @@
...
+            FormattedSize,
+            DownloadedDate,
+            FormattedDownloadedDate,
+            InstalledDate,
+            FormattedInstalledDate,
+            InstallItemCommand,
+            IsInstalled,
+            InstallButtonText
+        );
+    }
```
Al12rs, line 57: I hate we have to do this. 
It seems way to easy to just forget something, or not add it here when something new is added.

erri120: Not much you can do about that.

Sewer56: :(

Al12rs: :(

src/NexusMods.App.UI/Pages/LibraryPage/LibraryViewModel.cs
```diff
@@ -167,10 +168,19 @@ public LibraryViewModel(
...
+                            var libraryItem = LibraryItem.Load(_connection.Db, id);
+                            if (!libraryItem.IsValid()) continue;
+                            await InstallLibraryItem(libraryItem, _loadout, cancellationToken);
+                        }
+                    }
+                    else
+                    {
+                        var libraryItem = LibraryItem.Load(_connection.Db, singleId);
+                        if (!libraryItem.IsValid()) return;
                         await InstallLibraryItem(libraryItem, _loadout, cancellationToken);
```
Al12rs, line 184: Was there no way to just always pass a collection, and just pas a collection of one item in the cases of a single item? 
Or is there an advantage of keeping the cases separate?

erri120: It just looked weird to return an array with a single item. Now that we're using composition, we can have these distinctions.

Al12rs (commented):
Code looks good, I ran into the
`[ERROR] While commiting transaction|System.InvalidOperationException: Invalid UTF-8 sequence.`
bug when testing mass deletions etc, that should be solved by #2154.

Would like to get that in first so this can be tested better.

Al12rs (approved):
I tested now with the fix from #2154 and it appears to be working.
Another pr seems to have been merged in the meantime causing another merge conflict but this PR should be good

Al12rs (approved):

PR #2153, opened 2024-10-10
Switch LocationId (and GamePaths) over to an ushort instead of a string

halgari:
A longstanding problem with efficiency with `LocationId`, is that internally it stores the location as a string (not even interned incidentally). With MnemonicDB this is even worse as we store GamePaths as tuples, which results in the stored value being a composite value of two variable sized strings. We can clean all this up fairly simply, and that's what this patch does:

* Stores LocationId now as a ushort
* The ushort is calculated via a FNV1a (32bit) hash with the upper and lower halves combined with an XOR
* When a string is converted to a LocationId we hash it, then make sure it doesn't collide with existing values
* Performance of creating LocationIds from strings will be slightly slower with this method but the most performance critical parts of the code use `ushorts` and not strings
* The LocationId struct overrides the ToString value to provide a nice conversion back to the original string value

All told this is a fairly small code change that I don't think has much of an impact on anything besides a sizable reduction memory usage and a massive improvement in serialization in MnemonicDB

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.GameLocators/LocationId.cs
+++ b/src/Abstractions/NexusMods.Abstractions.GameLocators/LocationId.cs
@@ -1,11 +1,48 @@
-﻿namespace NexusMods.Abstractions.GameLocators;
+﻿using System.Collections.Immutable;
+using TransparentValueObjects;
+
+namespace NexusMods.Abstractions.GameLocators;
 
 /// <summary>
 /// The base folder for the GamePath, more values can easily be added here as needed
 /// </summary>
-[TransparentValueObjects.ValueObject<string>]
+[ValueObject<ushort>]
 public readonly partial struct LocationId
 {
+    private static ImmutableDictionary<ushort, string> _cache = ImmutableDictionary<ushort, string>.Empty;
+    
+    /// <summary>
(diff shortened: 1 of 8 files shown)


Sewer56 (approved):
No objections.
I actually originally requested this when LocationId was being added to the codebase.

src/Abstractions/NexusMods.Abstractions.GameLocators/LocationId.cs
```diff
@@ -1,11 +1,48 @@
...
+            }
+            return From(mixed);
+        }
+        
+        while (true)
+        {
+            var newCache = _cache.Add(mixed, value);
+            if (ReferenceEquals(Interlocked.CompareExchange(ref _cache, newCache, _cache), _cache))
+                break;
+        }
```
erri120, line 39: I'm curious, what is the user supposed to do if they encounter this? Wait until we implement hash collision resolution? If this is highly unlikely to happen anyways, we can move the check into debug mode.

halgari: My thought is that whomever is implementing a new game, will hit this, then slightly alter the name of the location id, just enough to get it not to collide. So we're likely only going to get collisions during the initial steps of adding a new game.

PR #2154, opened 2024-10-10
Small fixes and Export CLI method

halgari:
Updates MnemonicDB to the latest version (fixes the UTF8 decode error), also adds an `Export` CLI method for exporting the MnemonicDB database to a file so that we can use the data for testing in MnemonicDB


Sewer56 (approved):

src/Abstractions/NexusMods.Abstractions.FileStore/Downloads/DownloadAnalysis.cs
```diff
@@ -10,6 +10,7 @@
 using NexusMods.MnemonicDB.Storage;
 using NexusMods.Paths;
 using ModFileTreeNode = NexusMods.Paths.Trees.KeyedBox<NexusMods.Paths.RelativePath, NexusMods.Abstractions.FileStore.Trees.ModFileTree>;
+using ULongAttribute = NexusMods.Abstractions.MnemonicDB.Attributes.ULongAttribute;
```
erri120, line 13: This attribute exists in two places, we should probably remove our custom one and just use the DB attribute.

PR #2170, opened 2024-10-14
Collection download page

erri120:
Adds a concrete VM for the collections download page and wires up all the properties. I'm going to add the trees in the next PR.


halgari (approved):
This is all pretty normal looking code, LGTM

PR #2174, opened 2024-10-16
Cleanup: remove activities

erri120:
This was mostly just deleting things, except for the login overlay, which was the only non-deprecated part of the app that was still using the activity system.


halgari (changes requested):
I can log out via the code on `main` but with this PR it doesn't update the UI when logging out. I think it does log out (says so in the log) but the UI doesn't update and my avatar remains in the top right

erri120:
> I can log out via the code on `main` but with this PR it doesn't update the UI when logging out. I think it does log out (says so in the log) but the UI doesn't update and my avatar remains in the top right

Can't repro this. I also didn't touch the code to log out, that has remained the same. Maybe you have an old datastore?

Al12rs:
I tested this and it seems to be working fine, not issues logging in, logging out, then logging back in. I tested both without restarts of the app and with restarts.

Al12rs (approved):
Tested only

halgari (approved):
LGTM

PR #2176, opened 2024-10-16
Updated changelog for 0.6.3 and updated formatting styles

Pickysaurus:
**DO NOT MERGE UNTIL THE NEXT RELEASE**
This PR makes structural changes to the live changelog document which is parsed by the app. We should merge this before the first release containing this PR https://github.com/Nexus-Mods/NexusMods.App/pull/2177

---

- Removed the "Changelog" H1 heading
- All version numbers are now H1 rather than H2
- All subsection headers are now H2 rather than H3
- Bullet points now use `*` instead of `-`
- Horizontal lines should use `---`
- Bold is now used for emphasis rather than headings
- Italics are used for notes rather than headings or bold
- For priority callouts we will now use "Important:" as the prefix with bold styling.
- Other callouts will not include a prefix, but will be italic.
- When including links, use the document title or summary of the linked information rather than "More Info" or "Click here"
- When including links, try to link a full sentence where it makes sense - exceptions include referencing specific external item (e.g. Heroic Launcher) where context may be important.
- Links cannot be bold or italic. The link can be surrounded by text with this formatting but the tags must not wrap around the link itself.


MattSturgeon:
> * talics are used for notes rather than headings or bold
> 
> * For priority callouts we will now use "Important:" as the prefix with bold styling.
> 
> * Other callouts will not include a prefix, but will be italic.

> [!TIP]
> Have you considered using admonitions?

> [!Caution]
> Different markdown flavours use different syntax, so that may be a blocker for you. 

> [!note]
> In GFM, the syntax is similar to blockquotes ([docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)):
>
> ```md
> > [!TIP]
> > This is a "tip" admonition using GFM syntax
> ```

Pickysaurus:
> > * talics are used for notes rather than headings or bold
> > * For priority callouts we will now use "Important:" as the prefix with bold styling.
> > * Other callouts will not include a prefix, but will be italic.
> 
> Tip
> 
> Have you considered using admonitions?
> 
> Note
> 
> In GFM, the syntax is similar to blockquotes ([docs](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)):
> 
> ```
> > [!TIP]
> > This is a "tip" admonition using GFM syntax
> ```
> 
> Caution
> 
> Different markdown flavours use different syntax, so that may be a blocker for you.

This is supported by GitHub but not by the rendered inside the app, we are hamstrung by this limitation when it comes to formatting.

MattSturgeon:
> This is supported by GitHub but not by the rendered inside the app

Sure, makes sense. IDK which markdown engine you're using in-app; but many do have extensions for admonitions.

Maybe it would make sense to pre-compile HTML and render _that_ in-app though?

Not a major issue, of course. Just trying to prompt ideas you may or may not have yet considered :slightly_smiling_face:

insomnious:
We are currently using https://github.com/whistyun/Markdown.Avalonia as our rendering library and we can extend it to add what we need. We do need to have a discussion about what we need to create and when, given the fact that GFM is important for being our code platform and Nexus-flavoured Markdown is quickly going to come along as collections evolve.

I've no idea on the overlap or what has been added to NFM, so that will be fun :)

PR #2178, opened 2024-10-16
Set `target_commitish` in release CI

### LINKED ISSUE #2087: CI Release Github Action is tagging wrong branch
# Bug Report

## Summary
Release CI workflow is tagging main even when releasing from a different branch/commit.
The build happens on the desired branch correctly, but a tag is then pushed on main and the release linked to that tag.

## Details
It seems `softprops/action-gh-release@v1` should be updated to v2 and that we should provide it with `target_commitish` if we wish to push a tag on a different branch.

Fix would be to:
- set `target_commitish` to the branch the workflow is being run on if it is not set
- add optional `target_commitish` input to workflow to allow tagging a specific commit
    - Update the rest of the workflow to use the passed target commit if set, to create the build based on that instead of latest commit.

MattSturgeon:
- **Bump gh-releases action in release CI**
- **Set `target_commitish` in release CI**
- **Simplify `rawVersion` logic in release CI**

I've played around a little in powershell to ensure my syntax is valid, but I haven't tested the workflow as the only way I know to do that is to setup my fork with the required env/secrets/etc and run that. Which I'm far too lazy to do I'm afraid!

If you'd rather work on this internally, I'm happy for you to reference, copy, modify (etc) my PR and/or close this and create your own. No attribution needed.

Fixes #2087

### DIFF
--- a/.github/workflows/build-linux-pupnet.yaml
+++ b/.github/workflows/build-linux-pupnet.yaml
@@ -8,6 +8,10 @@ on:
         required: false
         description: "The version of the application to build"
         default: "1.0.0"
+      Target:
+        description: "The git ref to base the release on (eg: '1957a8e7', default: current head of the branch the workflow is run on)"
+        required: false
+        type: string
       PupNetRepo:
         type: string
         required: false
@@ -62,6 +66,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          ref: ${{ inputs.Target || github.sha }}
           submodules: "recursive"
 
       - name: Transform inputs
--- a/.github/workflows/build-windows-pupnet.yaml
+++ b/.github/workflows/build-windows-pupnet.yaml
@@ -8,6 +8,10 @@ on:
         required: false
         description: "The version of the application to build"
         default: "1.0.0"
+      Target:
+        description: "The git ref to base the release on (eg: '1957a8e7', default: current head of the branch the workflow is run on)"
+        required: false
+        type: string
       PupNetRepo:
         type: string
         required: false
@@ -80,6 +84,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          ref: ${{ inputs.Target || github.sha }}
           submodules: "recursive"
 
       - name: Transform inputs
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -126,7 +126,7 @@ jobs:
 
       - name: Create Release
         id: create-gh-release
-        uses: softprops/action-gh-release@v1
+        uses: softprops/action-gh-release@v2
         with:
           name: ${{ needs.transformInputs.outputs.RawVersion }}
           tag_name: ${{ needs.transformInputs.outputs.TagVersion }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -53,7 +53,7 @@ jobs:
           InputVersion: ${{ inputs.version }}
         run: |
           $tagVersion = $env:InputVersion.StartsWith('v') ? $env:InputVersion : "v" + $env:InputVersion
-          $rawVersion = $env:InputVersion.StartsWith('v') ? $env:InputVersion.Substring(1) : $env:InputVersion
+          $rawVersion = $tagVersion.Substring(1)
           $validation = [System.Version]::Parse($rawVersion)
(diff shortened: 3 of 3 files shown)


Al12rs (changes requested):
Hey, thanks for the PR.
I looked at the implementation and from what I understand this change makes it so that the tag is created on a release_vx.x.x branch, for that release number. 

I guess this would work if we assume we always generate the branch first and then build the release from it. 
What we actually would want though would be to release and tag any commit hash, on any branch, rather than having to create a dedicated branch.

MattSturgeon:
Thanks for taking the time to look over the changes!

> This change makes it so that the tag is created on a `release_vx.x.x` branch, for that release number.
> I guess this would work if we assume we always generate the branch first and then build the release from it.

Up until now, that appears to be the workflow you've been using.

Specifically, #2087 describes how the release tag should always reference a commit on the release branch.

> What we actually would want though would be to release and tag any commit hash, on any branch, rather than having to create a dedicated branch for the release.

This could be achieved by using the workflow's [`${{ github.ref }}`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) context attribute.
- If running a workflow via the web GUI, that'd be the "Use workflow from" `<branch dropdown>` UI.
- If running a workflow via the `gh workflow run` CLI command, that'd be the `--ref` flag.

<details><summary>Web UI screenshot</summary>
<p>

image

</p>
</details> 

Perhaps someone more familiar with how you intend to run the workflow should take the lead here? I'm happy for any changes in this PR to be used in another PR that takes a different approach.

I'm also happy to continue working on this if you have a clear idea of what you want :slightly_smiling_face:

Al12rs:
> Up until now, that appears to be the workflow you've been using.
> 
> Specifically, #2087 describes how the release tag should always reference a commit on the release branch.

Yeah so the main issue we have is that we test a build on main, for release, but while that is getting tested other thing will get merged in the meantime. If the testing goes ok, we want to release that specific commit on main, not the latest commit on main. The way to workaround it at the moment is to create a new branch for the release and then run the workflow on that new branch, since the latest commit on that branch will be the one we want.

We would want to simplify that by adding an optional commit hash as input to use in the workflow, that should be used as the  base for the release build and that should get the tag pushed to it.

We intend to run the workflow from GH web UI. If the commit hash is left empty, the workflow should use the latest commit on the selected branch.

MattSturgeon:
> The way to workaround it at the moment is to create a new branch for the release and then run the workflow on that new branch, since the latest commit on that branch will be the one we want.

Gotcha, so this is already possible even without any additional inputs, using the `${{ github.sha }}` context attribute will allow us to create the new tag on the _commit the workflow was run on_, even if subsequent commits have been made to the _branch_ the workflow was run on.

Additionally, github's "run workflow from" `ref` should also be settable to any commit, not just branch/tag refs. Although if you find it a better UX you could add an optional `input` to specify one.

Al12rs:
This is what we currently have in the Web UI:
image
We would want to add an optional input for the target commit hash there

MattSturgeon:
I've implemented that for the release workflow, however it'll also need to pass the "target" on to the build workflows to ensure they build the correct commit.

This will mean adding a similar "target" input to those workflows.

Specifically, I've added an optional "target" input that defaults to `${{ github.sha }}`. `github.sha` will be the head of the _"Use workflow from"_ branch (`github.ref`) at the time the workflow is run.

Al12rs:
Nice, thank you!
>  however it'll also need to pass the "target" on to the build workflows to ensure they build the correct commit.

Would you like to also take a look at updating the build workflows? I would wait to merge this until we get those updated as well, so that we don't accidentally release the wrong commit.

MattSturgeon:
> Would you like to also take a look at updating the build workflows? I would wait to merge this until we get those updated as well, so that we don't accidentally release the wrong commit.

Yeah, it should ideally be done in this PR. I was short on time earlier so just pushed up what I had so far.

MattSturgeon:
This should now be in a finished state. However as I outlined before, I really don't want to go through the pain of setting up my fork to be able to run the CI and then testing to ensure everything works correctly.

Hopefully the changes are simple enough that most review work can be done from the diff, but you may wish to also try out a "dummy" release too.

I'm also open to wording nits and other minor feedback, so don't hold back :slightly_smiling_face:

Al12rs (approved):
Seems fine from the diff, I guess we will see after merging if there are any issues.
Thank you for the contribution!

PR #2180, opened 2024-10-16
Added back basic Bannerlord support

Aragas:


Pickysaurus:
Thanks for setting this up @Aragas. We have a few bits to finish off before we properly dive into Bannerlord support so don't worry if this takes a while to get reviewed :)

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/MountAndBlade2BannerlordSettings.cs
```diff
@@ -0,0 +1,29 @@
...
+        return settingsBuilder.AddToUI<MountAndBlade2BannerlordSettings>(builder => builder
+            .AddPropertyToUI(x => x.DoFullGameBackup, propertyBuilder => propertyBuilder
+                .AddToSection(Sections.Experimental)
+                .WithDisplayName($"Full game backup: {MountAndBlade2Bannerlord.DisplayName}") 
+                .WithDescription("Backup all game folders, this will greatly increase disk space usage. Should only be changed before managing the game.")
+                .UseBooleanContainer()
+            )
+            .AddPropertyToUI(x => x.BetaSorting, propertyBuilder => propertyBuilder
+                .AddToSection(Sections.Advanced)
+                .WithDisplayName($"Beta Sorting: {MountAndBlade2Bannerlord.DisplayName}")
```
Sewer56, line 22: This setting needs a description that even players new to Bannerlord would be able to understand.

Even as a developer, without doing a deep dive, it's unclear whether 'Beta' here for example refers to 'Beta' in `Bannerlord.LauncherManager`, or whether it refers to a `Beta` version/feature of the game.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/MountAndBlade2BannerlordLoadoutSynchronizer.cs
```diff
@@ -0,0 +1,88 @@
...
+
+                modDirectoryNameToModel[modDirectoryName] = mod;
+            }
+            
+            newFile.LoadoutItem.ParentId = mod.Id;
+        }
+        return ValueTask.CompletedTask;
+    }
+
+    private static bool TryGetSMAPIMod(RelativePath modDirectoryName, Loadout.ReadOnly loadout, IDb db, out ModLoadoutItem.ReadOnly mod)
```
Sewer56, line 59: This function needs renamed; since it was copied from Stardew.

-------

On a misc (unrelated) note, I had a quick walk through the codebase, because I was curious if there's any potential issues with file casings. There shouldn't be in this case, but if you change the casing of an existing game folder on Linux, I imagine stuff would break; because the App would use the casing on original file path as source of truth.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Utils/GamePathProvier.cs
```diff
@@ -0,0 +1,30 @@
...
+using GameStore = NexusMods.Abstractions.GameLocators.GameStore;
+
+namespace NexusMods.Games.MountAndBlade2Bannerlord.Utils;
+
+public static class GamePathProvier
+{
+    private static string GetConfiguration(GameStore store) =>
+        LauncherManagerHandler.GetConfigurationByPlatform(LauncherManagerHandler.FromStore(Converter.ToGameStoreTW(store)));
+
+    public static GamePath PrimaryLauncherFile(GameStore store) =>
```
Sewer56, line 13: The game entry in `MountAndBlade2Bannerlord.cs` calls into this.

Although Xbox store isn't yet fully supported, I imagine you'd want to switch over `store` variable here and change the file name for Xbox store. Because the current logic  would use `TaleWorlds.MountAndBlade.Launcher.exe` when `Launcher.Native.exe` is likely intended when store is Xbox.

Don't have the title on Xbox available to me right now, but I believe this to likely be the case.

(Note: Xbox store basically works but App fails due to missing read privileges on binaries. I know how to fix that, I do it in one of my own projects; so I'll eventually get to pick it up some day 🤞)

Sewer56:
I went through around half of this, will look at the other half this morning.

CC. @Aragas I also started a draft of the info documentation page for the game (to familiarise myself with Bannerlord). I did some digging around for a few hours. I think it's around ~80% of the way there; just need some extra minor details mentioned such as where the load order and enabled mods are saved, etc. added.

Look through the text below and make any improvements/changes to it as needed.
[To get the markdown, just quote this post and copy from there]

-------------------

## General Info

- Name: Mount & Blade II: Bannerlord
- Release Date: 2020
- Engine: Custom - C++ Foundation, C# Scripting

### Stores and Ids

- [Steam](https://store.steampowered.com/app/261550/Mount__Blade_II_Bannerlord/): `261550`
- [GOG](https://www.gog.com/game/mount_blade_ii_bannerlord): `1802539526`, `1564781494`
- [Epic Game Store](https://store.epicgames.com/en-US/p/mount-and-blade-2): `Chickadee`
- [Xbox Game Pass](https://www.xbox.com/en-US/games/store/mount-blade-ii-bannerlord/9pdhwz7x3p03): `TaleWorldsEntertainment.MountBladeIIBannerlord`

### Useful Links

- [Official Mod Docs](https://moddocs.bannerlord.com)
- [Community Mod Docs](https://docs.bannerlordmodding.com)

### Engine and Mod Support

Bannerlord uses a native engine with self-hosted .NET 6 for Steam/GOG/Epic.
Modding is supported out of the box.

Bannerlord has a modding extension [BLSE](https://www.nexusmods.com/mountandblade2bannerlord/mods/1)
that expands the modding capabilities.

It's required to run mods on Xbox and is optional for Steam/GOG/Epic.

## Installing Mods (Outside of NMA)

### Automatic

- [Steam Workshop](https://steamcommunity.com/app/261550/workshop/)

Typically installs immediately via Steam Client

### Manual

- [Nexus](https://www.nexusmods.com/mountandblade2bannerlord)
- [ModDB](https://www.moddb.com/games/mount-blade-ii-bannerlord/mods)

## Mod Types

- [BLSE](https://www.nexusmods.com/mountandblade2bannerlord/mods/1)
- [Module (Standard) Mods](#typical-mod-structure-module).

Since it's not a standard mod.

### BLSE

BLSE needs to be placed beside the main game binary:

- `bin/Gaming.Desktop.x64_Shipping_Client` for Xbox Store
- `bin/Win64_Shipping_Client` for other targets

Outside of that, we will want to launch BLSE rather than the regular game binary if BLSE is installed.

### Typical Mod Structure (Module)

[See: Bannerlord Documentation](https://docs.bannerlordmodding.com/_intro/folder-structure.html#folder-descriptions-file-examples)

```
MyModule
├── AssetPackages
│   └── assetpackage.tpac
├── Atmospheres
│   ├── Interpolated
│   │   └── interpolatedatmosphere.xml
│   └── atmosphere.xml
├── bin
│   └── Win64_Shipping_Client
│       └── MyModule.dll
├── GUI
│   ├── Brushes
│   └── Prefabs
├── ModuleData
├── SceneObj
└── SubModule.xml
```

The only needed file is `SubModule.xml`, this is the metadata file.
All else is optional.

Note that the `bin` folder contains DLLs loaded by the game, don't mistake this tree with
the folder structure of the game itself.

#### Example `SubModule.xml`

Taken from [Player Settlements](https://www.nexusmods.com/mountandblade2bannerlord/mods/7298?tab=files&file_id=44128)

```xml
<Module>
  <Name value="Player Settlement"/>
  <Id value="PlayerSettlement"/>
  <Version value="v6.0.2"/>
  <Url value="https://www.nexusmods.com/mountandblade2bannerlord/mods/7298" />
  <SingleplayerModule value="true"/>
  <MultiplayerModule value="false"/>
  <Official value="false"/>
  <DefaultModule value="false" />
  <ModuleCategory value="Singleplayer" />
  <ModuleType value="Community" />
  <UpdateInfo value="NexusMods:7298" />
  <!-- Used by PlayerSettlement to dynamically add more settlement variants -->
  <PlayerSettlementsTemplates path="ModuleData/Player_Settlement_Templates" />
  <!-- Used by PlayerSettlement to dynamically blacklist settlement variants -->
  <PlayerSettlementsTemplatesBlacklist path="ModuleData/template_blacklist.txt" />
  <DependedModules>
    <DependedModule Id="Bannerlord.Harmony" />
    <DependedModule Id="Bannerlord.ButterLib" />
    <DependedModule Id="Bannerlord.UIExtenderEx" />
    <DependedModule Id="Bannerlord.MBOptionScreen" />
    <DependedModule Id="Native"/>
    <DependedModule Id="SandBoxCore"/>
    <DependedModule Id="Sandbox"/>
    <DependedModule Id="StoryMode" />
  </DependedModules>
  <!-- Community Metadata -->
  <DependedModuleMetadatas>
    <DependedModuleMetadata id="Bannerlord.Harmony" order="LoadBeforeThis" version="v2.2.2" />
    <DependedModuleMetadata id="Bannerlord.ButterLib" order="LoadBeforeThis" version="v2.8.15" />
    <DependedModuleMetadata id="Bannerlord.UIExtenderEx" order="LoadBeforeThis" version="v2.12.0" />
    <DependedModuleMetadata id="Bannerlord.MBOptionScreen" order="LoadBeforeThis" version="v5.10.1" />
    <DependedModuleMetadata id="Native" order="LoadBeforeThis" version="1.0.0.*" />
    <DependedModuleMetadata id="SandBoxCore" order="LoadBeforeThis" version="1.0.0.*" />
    <DependedModuleMetadata id="Sandbox" order="LoadBeforeThis" version="1.0.0.*" />
    <DependedModuleMetadata id="StoryMode" order="LoadBeforeThis" version="1.0.0.*" />
    <DependedModuleMetadata id="CustomBattle" order="LoadBeforeThis" version="1.0.0.*" optional="true" />
  </DependedModuleMetadatas>
  <SubModules>
    <SubModule>
      <Name value="PlayerSettlement"/>
      <DLLName value="PlayerSettlement.dll"/>
      <SubModuleClassType value="BannerlordPlayerSettlement.Main"/>
      <Tags />
    </SubModule>
  </SubModules>
  <Xmls/>
</Module>
```

#### Per-Store Mods

It's possible to have per-store specific code for Bannerlord, namely:

```
📁 Gaming.Desktop.x64_Shipping_Client
    📄 0Harmony.dll (2.2 MB)
    📄 MCMv5.dll (493.1 kB)
    📄 PlayerSettlement.dll (250.9 kB)
    📄 PlayerSettlement.pdb (97.1 kB)
    📄 System.Numerics.Vectors.dll (115.9 kB)
📁 Win64_Shipping_Client
    📄 0Harmony.dll (2.2 MB)
    📄 MCMv5.dll (493.1 kB)
    📄 PlayerSettlement.dll (250.9 kB)
    📄 PlayerSettlement.pdb (97.1 kB)
    📄 System.Numerics.Vectors.dll (115.9 kB)
```

- `Gaming.Desktop.x64_Shipping_Client` applies to Xbox Store
- `Win64_Shipping_Client` applies to other targets

#### Mod Load Order

As seen in `order="LoadBeforeThis`, some mods may be rearranged on boot in terms of load order.
This needs to be expressed in App UI.

#### Mod Names

We should ideally extract the mod name from the `SubModule.xml` if possible,
i.e. from `<Name value="Player Settlement"/>`.

If there are multiple mods in 1 download, we may need to combine it with the mod page name.

## Outside Changes

The game has a built-in mod manager.

If the user chooses to rearrange mods in the 1st party launcher, we will need to ingest 
the changes. Ask team for clarification given last week's discussion.

### Ingesting RuntimeDataCache Folders

Mod modules can generate a `RuntimeDataCache` folder their module at runtime; which contains
assets ready for consumption generated by the game engine (from the module sources).

Computing this folder's contents is a costly operation, for large total conversion mods, this can
even take half an hour, or longer on some machines.

It's imperative we ingest the files generated in this folder right into the App, so the user can
quickly switch between loadouts.

## Diagnostics

[Diagnostics NMA Wiki Link](https://nexus-mods.github.io/NexusMods.App/developers/development-guidelines/Diagnostics/#choosing-the-severity)

### Warning: Gauntlet (non-C#) Modules Must load Before Official Module

[Modding Wiki Link](https://docs.bannerlordmodding.com/_tutorials/modding-gauntlet-without-csharp.html#important)

Severity: `Warning`
Summary: `Codeless Gauntlet UI Mods must be loaded before Official Modules.`
Details: Mods which override Gauntlet UIs from Official Modules MUST be loaded before said module.

To detect this, look at mods which override the `GUI/Prefabs` folder. There may be more folders.

### Suggestion: Mods Should load after Official Modules

[Modding Wiki Link](https://docs.bannerlordmodding.com/_tutorials/modding-gauntlet-without-csharp.html#important)

Severity: `Suggestion`
Summary: `Mods should load after official modules.`
Details: User created mods should load after the following official modules. `Native`, `SandBox`, `Sandbox Core`, `CustomBattle`, `StoryMode`.

Note: [Excludes Gauntlet non-C# Modules](#warning-gauntlet-non-c-modules-must-load-before-official-module).

### Critical: Missing Dependency Submodules

[Relevant Documentation](https://docs.bannerlordmodding.com/_xmldocs/submodule.html#element-descriptions)

Severity: `Critical`
Summary: `Missing Required Module {ModuleA} required by {ModuleB}`
Details: `{ModuleB}` requires `{ModuleA}` but is not present. You must download `{ModuleA}` first.

We check against the `DependedModules` array of `SubModule.xml`. 
Question: Is there a database of Module IDs to download pages?

### Warning: Duplicated Module ID Detected

[Relevant Documentation](https://docs.bannerlordmodding.com/_xmldocs/submodule.html#element-descriptions)

Severity: `Warning`
Summary: `Mod conflict: {ModA} and {ModB} can't run together (using same ID).`
Details: 

```
We found two mods that can't run together:
- {ModA} (version {VersionA})
- {ModB} (version {VersionB})

They're using the same identifier ({ModId}), so you'll need to pick one and disable the other.
```

***if the mod versions match, the text should change a little bit:***

Summary: `{ModA} might be installed twice (found two copies with version {VersionA}).`
Details:

```
We found what looks like a duplicate mod:
- {ModA} (version {VersionA})
- {ModB} (version {VersionA})

Since they're using the same identifier ({ModId}) and have identical versions, you might have
accidentally installed the same mod twice. Check your loadout and remove one copy.
```

To determine this, check ID and version of `SubModule.xml`. 

## Diagnostics (Potential/Future)

### Missing Bin Files for Current Store

[Note: Doublecheck with Community if Copying Stuff across Folders is Ok]

Presumably a mod which has `Win64_Shipping_Client` folder only will not work on Xbox (Game Pass)
but will work on other platforms. Likewise a mod with only `Gaming.Desktop.x64_Shipping_Client` will
only work on Xbox.

Whether we copy `Win64_Shipping_Client` to `Gaming.Desktop.x64_Shipping_Client` and vice versa
will require discussion with rest of team and community.

## Questions (Aragas & Co)

### What's Missing in Existing Mod Managers

Some mod pages say 'Installing with some mod managers may cause a crash'.
I believe this info is out of date, but need to doublecheck.

### Auto Copy Binary Files Across Store Releases

Discuss whether we should [copy binary files across store releases](#missing-bin-files-for-current-store).

### Display Native Game Modules

Should we show the native game modules, `Native`, `SandBox`, `Sandbox Core`, `CustomBattle`, `StoryMode`. 
Doing this may require some new tech in the App.

### Diagnostics for Save Breaking Features

Are there diagnostics for things that break saves?

That's not well defined in docs outside of https://docs.bannerlordmodding.com/_intro/general-recommendations.html (not linked in sidebar).

## Questions (Future Features)

- https://docs.bannerlordmodding.com/_xmldocs/submodule.html#element-descriptions Description is unclear.
  - "XMLs with the same id from two separate mods (or the same mod) will have their assets combined and NOT overwritten." refers to `<XmlName id=`, so multiple entries from those will be combined. 
  - Does 'overwritten' work with regards to when the 'path' field is the same?
  - Should we inform the user of this conflict in load order?
  - Highlight mods with same item IDs as conflicts https://docs.bannerlordmodding.com/_xmldocs/items/item/
  
- Some mods are packed with `.tpac`. (a.k.a. 'Taleworlds Package' named by some people)
    - Does this affect load ordering?
    - Can a `.tpac` file contain any content that 'maps' onto the module folder?
    - Should we emit a diagnostic if a mod can be packed as `.tpac` for performance reasons?

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/MountAndBlade2BannerlordLoadoutSynchronizer.cs
```diff
@@ -0,0 +1,88 @@
...
+        return true;
+    }
+
+    protected override ValueTask MoveNewFilesToMods(Loadout.ReadOnly loadout, IEnumerable<AddedEntry> newFiles, ITransaction tx)
+    {
+        var modDirectoryNameToModel = new Dictionary<RelativePath, ModLoadoutItem.ReadOnly>();
+
+        foreach (var newFile in newFiles)
+        {
+            GamePath gamePath;
```
Sewer56, line 37: Local var is unused.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/MountAndBlade2BannerlordLoadoutSynchronizer.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            mod = default(ModLoadoutItem.ReadOnly);
+            return false;
+        }
+
+        mod = ModLoadoutItem.Load(db, file.AsLoadoutItem().Parent);
+        return true;
+    }
+
+    private static bool IsModFile(GamePath gamePath, out RelativePath modDirectoryName)
```
Sewer56, line 74: `modDirectoryName` could be renamed to `submoduleDirectoryName`, since we're specifically talking about Bannerlord as the 'domain' here so to speak, just make sure it's clear in the docs that the variable extracts `ModuleDirName` from `Modules/{ModuleDirName}/...`

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/FileSystemProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+internal sealed class FileSystemProvider : IFileSystemProvider
+{
+    private readonly ILogger _logger;
+
+    public FileSystemProvider(ILogger<FileSystemProvider> logger)
+    {
+        _logger = logger;
+    }
+
+    public byte[]? ReadFileContent(string filePath, int offset, int length)
```
Sewer56, line 15: I don't mind personally either way since the paths are sourced from the `LauncherManager` module, but I know the other devs would complain if they reviewed this, so I'll bring it up.

Please wrap the string in `NexusMods.Paths` with `AbsolutePath`.
Namely, inject an `IFileSystem` via the constructor, and create a path with `_fs.FromUnsanitizedFullPath(filePath)`. Then perform the read from that.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/FileSystemProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        try
+        {
+            if (offset == 0 && length == -1)
+            {
+                return File.ReadAllBytes(filePath);
+            }
+            else if (offset >= 0 && length > 0)
+            {
+                var data = new byte[length];
```
Sewer56, line 27: It would be preferable to use `GC.AllocateUninitializedArray<byte>(length)` here; since the array does not need to be zeroed given that we're going to immediately overwrite it anyway.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/LauncherManagerNexusMods.cs
```diff
@@ -0,0 +1,61 @@
+using Bannerlord.LauncherManager;
+using Bannerlord.LauncherManager.External;
+using Bannerlord.LauncherManager.Models;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using NexusMods.Abstractions.Settings;
+
+namespace NexusMods.Games.MountAndBlade2Bannerlord.LauncherManager;
+
+public sealed partial class LauncherManagerNexusMods : LauncherManagerHandler,
```
Sewer56, line 10: There's a lot of layers of logic here. Namely, this inherits from `LauncherManagerHandler` which itself consists of many partial classes, segregated by trait-like features.

There needs to be XML Documentation (and notes) of some sort, either here or in the `Bannerlord.LauncherManager` dependency, as for most (almost all?) readers the purpose of this code will be unclear. Especially if they haven't dug into Bannerlord previously.

Just note that this is the `Adapter` class, so to speak; which translates/marshals internal library logic and exposes them via NMA primitives. Some minimal info should be provided on the methods.

- `LauncherState` (`GetState`) seems to toggle between SP/MP, which we may need to inject some sort of button in the UI for.
- `LoadOrder` (`LoadLoadOrder/SaveLoadOrder`) serializes/deserializes load order as a dictionary. Reading in from Bannerlord, and saving from App back to Bannerlord. Every entry here has the index as part of its class layout, so changing one requires changing all other elements.
- `IModuleViewModel` (`GetAllModuleViewModels/GetModuleViewModels/SetModuleViewModels`) is for reading/writing loadouts (no load order).

Some minimal info on how these systems interact with NMA would be needed.
i.e. something along the lines of `Imports the module setup from the game folder to update NMA's internal state`

-----------------

In the long term, I imagine we'd want to use `LauncherManagerNexusMods` to 'ingest' (integrate, import) changes user made in the game's native launcher. Although we've talked about this topic as a  team last week, the result wasn't fully conclusive.

I do think though that using `Bannerlord.LauncherManager` and current disk state as the source of truth would be preferable, so if user rearranges mods in native launcher, we'd apply that; but the mechanisms around that aren't fully set in stone.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Utils/ModuleContext.cs
```diff
@@ -0,0 +1,52 @@
+﻿using Bannerlord.LauncherManager;
+using Bannerlord.ModuleManager;
+using NexusMods.Games.MountAndBlade2Bannerlord.Models;
+
+namespace NexusMods.Games.MountAndBlade2Bannerlord.Utils;
+
+internal class ModuleContext
+{
+    private readonly IDictionary<string, LoadoutModuleViewModel> _lookup;
+    public ModuleContext(IEnumerable<LoadoutModuleViewModel> moduleVMs)
```
Sewer56, line 10: I believe this code already exists in `Bannerlord.LauncherManager.Utils`, as `private class ModuleContext<TModuleViewModel> where TModuleViewModel : class, IModuleViewModel`.

Would it be better to instead make the above public instead of duplicating the code? I'm just curious here.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/NotificationProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace NexusMods.Games.MountAndBlade2Bannerlord.LauncherManager;
+
+internal class NotificationProvider : INotificationProvider
+{
+    private readonly ConcurrentDictionary<string, object?> _notificationIds = new();
+    
+    private Window? _window; // TODO: How to inject the window?
+
+    public void SendNotification(string id, NotificationType type, string message, uint displayMs)
```
Sewer56, line 15: We don't have a built-in notifications provider or abstractions for creating windows but if you need to show something that 'pops up' in front of the user, then the `Overlay` system.

(This is the system used for confirmation dialogs, such as login, deleting mods, etc.)

The caveat though is dismissing it requires user interaction.

To use that, you would use `IOverlayController` from `NexusMods.App.UI.Overlays`. There are some generic ones available like `MessageBoxOkView`. A good example of this is `ShowGameAlreadyRunningError` method. Would just need to lift `IOverlayController` into an abstractions package, I can do that if needed.

If there is a requirement to show info that doesn't require user interaction, we'd need to talk with UX (@captainsandypants). That's not currently in the App's designs.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/LauncherManagerFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+using NexusMods.Games.MountAndBlade2Bannerlord.Utils;
+
+namespace NexusMods.Games.MountAndBlade2Bannerlord.LauncherManager;
+
+public sealed class LauncherManagerFactory
+{
+    private readonly IServiceProvider _serviceProvider;
+    private readonly ConcurrentDictionary<string, LauncherManagerNexusMods> _instances = new();
+
+    public LauncherManagerFactory(IServiceProvider serviceProvider)
```
Sewer56, line 15: Not sure if this will be needed in the long run, given you could invoke the constructor directly (and/or make the methods below static constructors).

I'm going to guess you have some state that takes a while to initialize, and are using this as a cache, in which case it's fine. Just make sure to note that in a comment if that's the case.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/NotificationProvider.cs
Sewer56: CC. @Nexus-Mods/nexusmods-app-developers 

We may want to have some sort of abstraction system for displaying UI elements from game modules, and external code. 

I think this was brought up around start of this year, back when we talked about emitting messages, as well as mediatR pattern and all.

Although we've managed to avoid it so far, there will be a time when game/external code will need to have some specific logic that need to render UI elements. I suppose that time is now.

----------

I guess some examples of messages emitted here (by @Aragas) would be helpful too, as this sort of thing may require design work too.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/MountAndBlade2BannerlordSettings.cs
Aragas: Good point!
I guess we can actually remove the setting, since I don't believe the Beta sort is used a lot. But yea, it's a LauncherManager feature

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Utils/GamePathProvier.cs
Aragas: In our case it's simpler. We don't use the original executable on Xbox.
Xbox is only supported with BLSE, on other sores we support both the Vanilla executable and BLSE

Sewer56: Does Xbox not have a native module manager?

Would we need to eventually emit a diagnostic on Xbox one day to tell the user to download+install BLSE?

Aragas:
@Sewer56 
The game uses .NET Framework on Steam/GOG/Epic for now. The game seems to work when launched via .NET 6 (and BLSE give has a parameter for that), but it's not by default.
Xbox only uses .NET 6

Aragas:
| Whether we copy Win64_Shipping_Client to Gaming.Desktop.x64_Shipping_Client and vice versa
will require discussion with rest of team and community.

It should be generally safe to use `Win64_Shipping_Client` on `Gaming.Desktop.x64_Shipping_Client`. I've added this feature recently in the Vortex extension, generally, works. The only exception was a Visual Basic mod

Aragas:
> Diagnostics for Save Breaking Features
Save files contain the mods and the versions that were used. We produce a warning in both BLSE and Vortex when the version mismatches (should be an error if we downgraded, but I haven't done that yet)
And there should be an error if a mod is missing

Sewer56:
> @Sewer56 The game uses .NET Framework on Steam/GOG/Epic for now. The game seems to work when launched via .NET 6 (and BLSE give has a parameter for that), but it's not by default. Xbox only uses .NET 6

Oh that's whack, so they have an entire untrimmed .NET 6.0.12 runtime present in the game files but that goes entirely unused.
I didn't expect that.

Sewer56:
@Aragas as for the big blob of text. Just make any edits you deem necessary and throw it in the markdown file (alongside other games in `docs/developers/games` folder). :ok_hand:

Sewer56:
> Save files contain the mods and the versions that were used. We produce a warning in both BLSE and Vortex when the version mismatches (should be an error if we downgraded, but I haven't done that yet)
> And there should be an error if a mod is missing

The way we emit warnings in the App is through the 'Health Check' (Diagnostics) view.

image

image

And if it's required by another mod:

image

image

Although there isn't any UX around this currently, the plans are to stop the user if there's any diagnostic which emits a `critical`. In this case, for Bannerlord we could emit a critical if a module is missing, letting the user know before they boot a game.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/FileSystemProvider.cs
Aragas: My only concern is that we'll have to switch to Streams. It's a shame that we won't be able to use `RandomAccess.Read(handle, data, offset);`

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Utils/ModuleContext.cs
Aragas: Good point, made it public and renamed to `DefaultModuleContext`

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/LauncherManagerFactory.cs
Aragas: Correct, the LauncherManager holds a state per game instance.
Yea, there was some PR I had a year or something ago that added a DI scope that was per game instance. In that case the LauncherManager could be a scoped service and this factory would not be needed

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/FileSystemProvider.cs
Sewer56: I'm fine with this code personally, and we could leave it this way, since the paths are provided by the library in any case; so the whole magic of abstraction is lost.
I just more-so worry about what the others would think.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/NotificationProvider.cs
Aragas: Here's some examples
* `SendNotification("module-parsing-error-exception", NotificationType.Error, $"Failed to read SubModule.xml at path '{subModulePath}'!\n{e}", 3000);`
This one is triggered by the LauncherManager becuase it's reading the `/Modules` folder to get all mods. This is a basic operation, so if this fails we give a hint to the user

* `SendNotification("module-parsing-error-invalid-module-id", NotificationType.Error, $"SubModule.xml has an invalid Id at path '{subModulePath}'!", 3000);`
Same

* `SendNotification("sort-finished", NotificationType.Info, new BUTRTextObject("{=J7dh36Dy}Finished sorting!").ToString(), 3000);`
We expect the user to click on a "Auto Sort" button that triggers an asynchronous flow. This will give a confirmation that the operation finished

Aragas: They are present in the Vortex extension

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/LauncherManagerFactory.cs
Sewer56: I think it's fine either way; just note it down in a comment in there, and you'll be good.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/NotificationProvider.cs
Sewer56: Hmm, I see.

We do have a missing piece of UX for reporting general errors that happen at runtime; right now it's all only ever emitted to log. I would like to see emitting messages of this kind to also be possible in the design we'll eventually come up with. I'll make sure to bring this up if others don't see this, sooner rather than later.

Since there's a whole process of asking UX/UI to do it, then discussing it/getting feedback, and then adding the whole abstraction layer.

Aragas:
> > @Sewer56 The game uses .NET Framework on Steam/GOG/Epic for now. The game seems to work when launched via .NET 6 (and BLSE give has a parameter for that), but it's not by default. Xbox only uses .NET 6
> 
> Oh that's whack, so they have an entire untrimmed .NET 6.0.12 runtime present in the game files but that goes entirely unused. I didn't expect that.

It's like this for a long time, we also had some jokes regarding this. The game developers were in the process of migrating .NET FX 472 libraries into .NET Standard2.0, but I haven;t checked for a whle if they finished or not

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/MountAndBlade2BannerlordSettings.cs
Sewer56: By the way feel free to resolve/collapse the segments where the code was changed.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Utils/GamePathProvier.cs
Aragas: Xbox uses the same module management, but the way the game ships kinda broke out main mods.
I don't remember the exact details, I believe it was because we need a controllable entrypoint to fix a critical Harmony issue (the game ships an older Mono.Cecil library that was eagerly loaded by the game's .NET hosting bootstrap). The issue breaks Harmony 2.3.x because we expect a higher version of that library.

The custom executable fixes the issue, but because of that we lose "proper" Xbox support like cloud saves. The game basically falls back to Steam/GOG/Epic behavior and uses save files from there.

For now yea, we need a diagnostic for that. Vortex just ask the user directly to download and install BLSE automatically (with a button confirmation of the action)

Sewer56: Hmm, I see; that makes quite a bit of sense.
In this case, steering those users towards installing does make sense, yup.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/LauncherManager/NotificationProvider.cs
Aragas: We also have a dialog that most likely won't be triggered
```
BUTRLoader detected 0Harmony.dll inside the game's root bin folder!
This could lead to issues, remove it?
```

BUTRLoader was the name for the BLSE backend, not used much anymore. It triggered a simple dialog box with Yes/No buttons

There are some other examples, but they are from an "optional" feature.  
BLSE has the ability to import/export load orders in all formats custom launchers have used (and of course introducing it's [own format](https://xkcd.com/927/)). The feature uses Yes/No dialogs and FileOpen/FileSave dialogs

Sewer56 (approved):

PR #2186, opened 2024-10-21
Alert component

### LINKED ISSUE #2135: "Alerts" UI component
## Currently
We have a "Info banner" component that was created as a temporary way to highlight less discoverable features to the user.

## We want
To replace this component with the more flesh out "Alerts" component. That provides more variants for different contexts and levels of severity.

## Design

[Figma link](https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=2793-15186&t=oFGwtbdf7ybiFBGj-1)

Image

insomnious:
Closes #2135 

A new component that takes Florian's original InfoBanner (only used was as the top of the New Tab panel) and cleans up the usage. It takes the functionality (like being able to track if dismissed by adding an ID to a Settings dictionary) and uses the new properties to show\hide\style aspects of it to match Figma. 

image

- Moved the Debug Banner at the bottom to now use this component
- This component also uses the new `StandardButton` component
- Need to document how to start naming these IDs for tracking dismissals.

Main question I'd like to go through is do we need to use a seperate `AlertSettingsWrapper` (formally `BannerSettingsWrapper`) that is initialised alongside the usage of `<alerts:Alert />`? We could tie that into the new Alert component and offer a property like `ID` that if it's supplied, then we can dismiss on this ID. Usage would end up something like:

```xml
<alerts:Alert x:Name="InfoBanner" ID="new-tab-howto-add-panels" Title="Title goes here" Body="Body text here" />
```

Which feels clean and self-contained but has the potential to limit if something else (other than this Alert) wants to use the Settings object. I would then argue that it's still ok for different components to track in different ways, and they probably should deserve their own settings object.


insomnious:
Need to remove the new Pill component, doesn't really belong in this PR.

src/NexusMods.App.UI/Controls/Alerts/Alert.axaml.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        // hide the body if there is no body text
+        if (ShowBody && string.IsNullOrEmpty(Body))
+        {
+            ShowBody = false;
+        }
+
+        // turn off elements based on properties
+        _dismissButton.IsVisible = ShowDismiss;
+        _bodyTextBorder.IsVisible = ShowBody;
```
Al12rs, line 242:
```suggestion
      
        // turn off elements based on properties
        _dismissButton.IsVisible = ShowDismiss;
        _bodyTextBorder.IsVisible = Content != null && ShowBody && !string.IsNullOrEmpty(Body);
```

src/NexusMods.App.UI/Controls/Alerts/Alert.axaml.cs
```diff
@@ -0,0 +1,259 @@
...
+        _actionsRowBorder.IsVisible = ShowActions;
```
Al12rs, line 243:
```suggestion

        // turn off elements based on properties
        _dismissButton.IsVisible = ShowDismiss;
        _bodyTextBorder.IsVisible = ShowBody && string.IsNullOrEmpty(Body);
        _actionsRowBorder.IsVisible = Content != null && ShowActions;
```

src/NexusMods.App.UI/Controls/Alerts/Alert.axaml.cs
```diff
@@ -0,0 +1,259 @@
...
+
+    /// <summary>
+    /// Defines the Title property of the <see cref="Alert"/>. Defaults to "Default Title".
+    /// </summary>
+    public static readonly StyledProperty<string?> TitleProperty = AvaloniaProperty.Register<Alert, string?>(nameof(Title), defaultValue: "Default Title");
+
+    /// <summary>
+    /// Defines the Body property of the <see cref="Alert"/>. Defaults to "Default Body".
+    /// </summary>
+    public static readonly StyledProperty<string?> BodyProperty = AvaloniaProperty.Register<Alert, string?>(nameof(Body), defaultValue: "Default Body");
```
Al12rs, line 33: I guess you are using these for preview? I would usually have empty values as the default otherwise.

Al12rs (commented):
Talked with erri about the wrapper, and yeah having the id property on the control is good. 
We might still want to define a custom wrapper, in case we want to listen to dismiss events for some reason, so no need to remove the existing code.

captainsandypants (approved):
All good design wise

insomnious:
Due to needing to inject the ServiceProvider and SettingsManager into the component, we aren't really reducing the amount of code by attempting to move the ID to the component at this time. Going to leave it for now and see how we get on when using it.

Al12rs (approved):

Issue #2044, opened 2024-09-16
Add tooltips to toolbar buttons


Pickysaurus:
# Improvement

## Currently 
The toolbar buttons do not have tooltips

## We want
To add tooltips

### Top bar
- Back arrow -> Back
- Forward arrow -> Forward
- Account Icon (Logged In) -> Logged in to Nexus Mods as {USERNAME}
- Log In -> Connect your Nexus Mods account

### My Mods
- Switch View -> Show or hide mod page groups*
- View Mod Files -> Browse the files in the selected mod
- Remove -> Remove the selected mod(s) from this loadout

### Library
- Switch View -> Show or hide mod page groups*
- Delete -> Permanently remove this mod from your Library
- Add -> Add the selected mod(s) to this loadout
- Advanced Add -> Select specific files from a mod to add to your loadout
- From Drive -> Import a mod from your computer
- From Nexus Mods -> Open Nexus Mods

\* Temporary button, so not a priority. 

## Supporting Information
_(Include any supporting screenshots etc to support the above)_

## Design
_(Any designs to required for this feature)_

Pickysaurus:
@captainsandypants for you to review.

captainsandypants:
@Pickysaurus I think these are ok but I wouldn't schedule this work in anytime soon. We're likely to change buttons and labels as we go. We'd also want to add the keyboard shortcuts in for some of these, which haven't been implemented yet. We should keep this ticket open and updated though.

LukeNexusMods:
Simon is working on this. close ticket when complete please.

PR #2188, opened 2024-10-22
Add ToolTip property to StandardButton component

insomnious:
Added a ToolTip property to the component so we can add easy text-based Tooltips.

Usage is `<controls:StandardButton ToolTip="This is a tooltip" />`

We have to make sure that ToolTip is set to `null` for it not to show. Setting it to an empty string will still show the tooltip background but without any content. There is a small check in the code-behind to make sure on this.

This will now allow #2044 to be fixed by migrating the toolbar buttons to use `StandardButton`


erri120:
Do we really need this?

```xaml
<controls:StandardButton ToolTip="This is a tooltip" />
```
vs

```xaml
<controls:StandardButton ToolTip.Tip="This is a tooltip"/>
```

insomnious:
No, we don't necessarily need it. I'd originally tried last night to add some more custom content to a tooltip but then stripped it out this morning for simplicities sake, but didn't strip out enough. I guess we can just document to use `ToolTip.Tip` to add one just as easy, I will adapt the checking for empty string as I think that is still useful to have.

insomnious:
You know what, I'm just closing this as it's not worth the constant check overhead when we aren't doing any other manipulation for tooltips. As much as I like the idea of forcing the ToolTip property to be just a string for consistency sake, we can add that if it becomes an issue when reusing the control in many places.

PR #2195, opened 2024-10-23
Direct download in collections

halgari:
Adds support for Collection download sources that are defined as "direct". In this case we download a file, hash it with MD5 and add it to the library. In the future we can figure out how we want to present these in the UI, but this allows these collections to pass for now.


src/NexusMods.Library/AddDirectDownloadJob.cs
erri120, line 1: Is this a leftover?

halgari: yep

erri120 (approved):

PR #2206, opened 2024-10-29
BG3 Script Extender Health Checks

Al12rs:
- part of #1262 

Adds
- [x] Mod x requires Script Extender, but it is missing
- [x] WINEDLLOVERRIDE suggestion when installing BG3SE on Wine (steam only for now)


Sewer56:
Will look at this tomorrow morning if nobody else does by then.

halgari (approved):

Issue #1467, opened 2024-05-28
Only flag a loadout as needing to apply if changes are necessary. 


Pickysaurus:
# Improvement

## Currently 
The app will show the "Apply" button even if no changes are required to the game folder. 

## We want
Only Apply (or ask the user to apply with the button) when files or other data need to be changed. 

## Supporting Information
Example:
- Install a mod
- Apply changes
- Disable the mod
- Apply changes
- Remove the disabled mod from the loadout
- The app will show the "Apply" button again, but nothing needs to change. 

## Design
_(Any designs to required for this feature)_

LukeNexusMods:
@Pickysaurus Is this fixed now?

Sewer56:
No, it's not.

There are still times where hitting `Apply` does a no-op; we should avoid that as on Windows that may prevent users from running games multiple times at the same time (e.g. for multiplayer testing).

Al12rs:
@Sewer56 I don't think what you describe is the same as the problem reported in this ticket.
The ticket refers to Apply button being enabled and shown to users even if there is nothing to apply. I think this might be fixed.
If not, do you know steps to reproduce?

What you seem to be referring to is more "Don't perform a synchronize operation if there is nothing to apply or ingest"?

Sewer56:
> What you seem to be referring to is more "Don't perform a synchronize operation if there is nothing to apply or ingest"?

@Al12rs yeah that's what I'm referring to.
For example, if you create multiple loadouts, the others will still show both 'Apply' button and perform a 'Synchronize' on game launch. Even though the files to be deployed are exactly identical, thus the Apply action is a no-op.

Al12rs:
That makes sense though, because Syncrhonizing a Loadout sets it as the Last Applied and thus the "Active" loadout.
Even if there is nothing to do between the switch you want to show the Apply button to allow users to consciously switch from one loadout to another.

Sewer56:
I think tnis ticket is good for closing if the original repro is no longer relevant.

I was just trying to point out an edge case there, but after some more work on the code there, the UI isn't equipped to handle that edge case in any case. So it's fine for now.

PR #2209, opened 2024-10-29
Improved: Edge Cases around the Apply & Launch Button

### LINKED ISSUE #1922: Prevent applying/switching loadouts while the Game or a Tool is running
Applying while game is running will cause `UnauthorizedAccessException` if it happens to try to touch files that are in use by the game or Tool. 
This is very likely to happen in particular when switching loadouts, since we are currently reverting to vanilla state and then applying the new loadout, which means we delete all the added files, some of which likely to be in use.

## Impact:
Currently this causes an exception in the log and Apply will fail, it isn't clear if the game folder is left in an inconsistent state. 
But I was able to resume normal operation after closing the game, and applying again. Everything appeared to be as expected in the game folder at that point.

Sewer56:
fixes #1922

This covers the changes from last week and yesterday; with some extra fixes.

Changes
---------

####  Switching loadouts while a game is running in the App now shows the `Running...` bar

image

Previously if you switched to another loadout, you could still click `Apply`, or even launch another copy of the game, which would have been likely to fail if changes were needed.

This is no longer the case, navigating to another loadout now shows games as `running`.

As a result views (Apply Button, Launch Button) now partially synchronize with global state (`GameRunningTracker`) rather than VM local state. 

#### [on Windows] Trying to Apply while the game is already running shows an error dialog.

If the user has launched the game externally, e.g. through Steam; the App will now show an error screen if it detects the PrimaryFile (EXE) is in use.

image

[This dialog was approved by Design]

Previously such a case could have led to the `Apply` operation silently failing. User would click `Apply`, and the button would come back without reporting anything despite an error silently happening under the hood.

Note: Launching games runs `Apply` operation. Therefore running multiple copies of the same game is blocked by
- https://github.com/Nexus-Mods/NexusMods.App/issues/1467 
Should we decide that the user should be able to run multiple copies of the game.

####  Other Changes

- Running Tools (e.g. Running Game) has been migrated to the Jobs system in a minimalistic way.

Future
--------

Currently the code is restricted such that you can only run 1 game at a given time. Behaviour is locked globally, rather than per `GameInstallation`. In the future this can be relaxed; for now I've kept this simple as to not excessively go beyond ticket scope.


src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
```diff
@@ -203,4 +205,46 @@ private IObservable<LoadoutSynchronizerState> CreateStatusObservable(LoadoutId l
...
+        // Edge Cases: - User may have multiple copies of a given game running.
+        //             - Only on Windows.
+        // Solution: Check if EXE (primaryfile) is in use.
+        // Note: This doesn't account for CLI calls. I think that's fine; an external CLI user/caller
+        var game = loadout.InstallationInstance.GetGame() as AGame;
+        var primaryFile = game!.GetPrimaryFile(loadout.InstallationInstance.Store)
+            .Combine(loadout.InstallationInstance.LocationsRegister[LocationId.Game]);
+        if (IsFileInUse(primaryFile))
+            throw new ExecutableInUseException("Game's main executable file is in use.\n" +
+                                               "This is an indicator the game may have been started outside of the App; and therefore files may be in use.");
```
Al12rs, line 222: Clarify what the consequence is for the game files being in use: 
Synchronize operation can't be executed safely while files are in use.

Al12rs (approved):

Issue #2207, opened 2024-10-29
Times and Dates are displaying future times


LukeNexusMods:
![Image](https://github.com/user-attachments/assets/5e65efa9-d012-45d0-bd2d-d21a3f16f71b)

Added times and dates appear far into the future on all pages that show times and dates.

erri120:
I believe this is a result of some recent DB changes. Loading an old data store with a newer app version could cause this issue because we're storing and loading the dates differently. Not really a bug, you just have to reset your data store. It would be a bug if this issue also comes up with new data. Can you add a mod and check the dates?

LukeNexusMods:
This from from a fresh install with no mods installed. I downloaded a collection and those were the dates and times it gave me.

PR #2216, opened 2024-10-30
Fix two bugs found in testing

### LINKED ISSUE #2211: App Mod processing performance drop
When downloading mods there is normally a period you wait where the app is processing them before they appear in the app. Currently this seems a lot longer than normal and its highlighting the issue of not knowing what the app is doing after something has downloaded. 

The time for SMAPI appearing in the app after downloading is normally instant but currently its taking a good 5+ seconds for it to appear. You can see this problem more clearly when downloading collections, I downloaded stardew very expanded and there was a long wait for all of the mods to appear after they had been downloaded even though they are a relatively small size.

halgari:
This PR updates to the latest MnemonicDB release which fixes 2 major issues in the app: 

1) General performance - the `GlobalCompare` method in MnemonicDB was performing 2 allocations on every compare. This was having a massive impact on everything from reads to writes. This issue has been resolved
2) Timestamp regression - we were storing transaction timestamps incorrectly, resulting in the `400 years from now` bug in the UI

Resolves #2211, #2207

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -14,8 +14,8 @@
     <PackageVersion Include="NexusMods.Hashing.xxHash3" Version="3.0.2" />
     <PackageVersion Include="NexusMods.Hashing.xxHash3.Paths" Version="3.0.2" />
     <PackageVersion Include="NexusMods.Paths" Version="0.10.0" />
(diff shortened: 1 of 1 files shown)


halgari:
Going to merge this without a review as all that's here is a deps upgrade

PR #2217, opened 2024-10-30
Fix SDV Manifest parse error

halgari:
After installing a SDV collection we were getting a lot of "Manfifest parse errors". This was due to our ChunkedStreamReader improperly terminating streams it reads in the sync code. This fixes that bug (copies the logic from the async version) which removes the parse errors.

### DIFF
--- a/src/NexusMods.DataModel/ChunkedStreams/ChunkedStream.cs
+++ b/src/NexusMods.DataModel/ChunkedStreams/ChunkedStream.cs
@@ -1,4 +1,5 @@
 ﻿using System.Buffers;
+using System.Diagnostics;
 using Reloaded.Memory.Extensions;
 
 namespace NexusMods.DataModel.ChunkedStreams;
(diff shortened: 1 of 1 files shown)


Sewer56 (approved):

Sewer56 (changes requested):
one moment, don't merge

Sewer56:
This PR had a bug in it, but I should have fixed it just now with an extra commit.

When reading, the `offset` and `count` components of the function signatures were not being considered in the previous commit. Depending on how the caller used the `Read` method, that may have led to incorrect behaviour.

With the use of `Span<T>` in the previous commit, and the blanket implementation that borrows from the `MemoryPool`, this would have led to a buffer overrun, and as a result, an exception. The new commit should correctly-ish consider them.

Tests should probably be more thorough, please check with the original issue/ticket.

Sewer56 (approved):

Sewer56:
(I feel kinda bad about having this merged by non-submitter, since I committed into this branch :sweat_smile: , since there was no response since.)

PR #2223, opened 2024-11-04
List collection downloads

erri120:
No nesting.


src/Themes/NexusMods.Themes.NexusFluentDark/NexusFluentDarkTheme.axaml
```diff
@@ -5,11 +5,11 @@
         <ResourceDictionary>
             <ResourceDictionary.MergedDictionaries>
                 <ResourceInclude Source="/Resources/ResourceIndex.axaml"/>
+                <ResourceInclude Source="avares://NexusMods.App.UI/Pages/LibraryPage/TreeDataGridLibraryResources.axaml"/>
```
Al12rs, line 8: I don't think this should be here:
- It is referencing a resource that isn't in the theme project
- It logically does not belong in the Theme project
- The Library Resources are available everywhere implicitly, there is no line connecting the `TreeDataGrid` using the resources back to the resource file, which makes it hard to find it.

We can instead name the LibraryResources using a Key, and then import them directly in the `TreeDataGrid`s that actually use them, so that there is a clear and explicit path from the usage point to the templates, and this reference isn't here in the Themes project.

Al12rs (changes requested):

src/Themes/NexusMods.Themes.NexusFluentDark/NexusFluentDarkTheme.axaml
erri120: Where should the resource include statement go? It has to be somewhere.

Al12rs: It can't be imported directly in the View file using it then?

erri120: No, we have two views that need it.

Al12rs: It can't be imported in two files?

src/NexusMods.App.UI/Pages/LibraryPage/LibraryView.axaml
```diff
@@ -95,133 +95,13 @@
...
-                                    </DataTemplate>
-
-                                    <!-- fallback -->
-                                    <DataTemplate DataType="{x:Type local:ILibraryItemModel}"/>
-                                </ContentControl.DataTemplates>
-                            </ContentControl>
-                        </DataTemplate>
-                    </TreeDataGrid.Resources>
-                </TreeDataGrid>
+                              HorizontalAlignment="Stretch"/>
```
Al12rs, line 102: Something along these lines should work I think:

```suggestion
                              HorizontalAlignment="Stretch"/>
                  <TreeDataGrid.Resources>
                      <ResourceDictionary>
                          <ResourceDictionary.MergedDictionaries>
                              <ResourceInclude Source="/TreeDataGridLibraryResources.axaml"/>
                          </ResourceDictionary.MergedDictionaries>
                      </ResourceDictionary>
                  </TreeDataGrid.Resources>
                              
```

src/Themes/NexusMods.Themes.NexusFluentDark/NexusFluentDarkTheme.axaml
erri120: It can, but I think it's a bad practice. We've had similar issues when we first created the theme project as well.

Al12rs: Ok, I think we should put it in `App.axaml` under `<Application.Resources>` then.
It's where Avalonia docs says to put it.

PR #2225, opened 2024-11-05
WIP - Improvements to My Games

insomnious:


src/NexusMods.App.UI/Controls/MiniGameWidget/MiniGameWidgetViewModel.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    private readonly ILogger<MiniGameWidgetViewModel> _logger;
+
+    public MiniGameWidgetViewModel(ILogger<MiniGameWidgetViewModel> logger, ISettingsManager settingsManager)
+    {
+        _logger = logger;
+
+        _image = this
+            .WhenAnyValue(vm => vm.Game)
+            // ReSharper disable once ConditionIsAlwaysTrueOrFalseAccordingToNullableAPIContract
```
erri120, line 32: This seems odd.

insomnious: That was copied from the original GameWidget, if it's not needed then I'll strip it out of both

src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
```diff
@@ -22,98 +23,148 @@
...
+                var miniGameWidgetViewModels = supportedGamesAsIGame
+                    .Select(game =>
+                        {
+                            var vm = provider.GetRequiredService<IMiniGameWidgetViewModel>();
+                            vm.Game = game;
+                            vm.Name = game.Name;
+                            // is this supported game installed?
+                            vm.IsFound = _installedGames.Any(installation => installation.Installation.GetGame().Equals(game));
+                            vm.GameInstallations = _installedGames
+                                .Where(installation => installation.Installation.GetGame().Equals(game))
```
erri120, line 156: I don't like this `Equals` check here, that is going to be a `object.Equals` check.

erri120 (dismissed):
See comments.

src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
insomnious: I think a similar comparison was done using an ID. Let me see.

halgari (approved):

PR #2227, opened 2024-11-05
Schema failsafes

halgari:
Provides a framework and set of failsafes for working with schema migrations. 

* Provides a "Fingerprint" hashing function that turns a schema into a single reproducible hash value
* Using Verify, dumps the schema and fingerprint to a `.md` file so we can track changes to the schema. Schema additions will cause this test to fail and we'll then manually validate the data
* Provides a folder for adding compressed `.rocksdb` databases of various states. A test loads this data, migrates to the latest schema version then runs some tests (using verify). These test are basic for now, but can be improved with time.


src/NexusMods.DataModel.SchemaVersions/SchemaFingerprint.cs
```diff
@@ -0,0 +1,32 @@
...
+/// Tools for generating a hash of all the attributes of a schema so that we can detect changes.
+/// </summary>
+public class SchemaFingerprint
+{
+    public static Hash GenerateFingerprint(IDb db)
+    {
+        StringBuilder sb = new();
+        var cache = db.AttributeCache;
+        
+        foreach (var id in cache.AllAttributeIds.OrderBy(id => id.Id))
```
erri120, line 18:
```suggestion
        foreach (var id in cache.AllAttributeIds.OrderBy(id => id.Id, StringComparer.OrdinalIgnoreCase))
```

src/NexusMods.DataModel.SchemaVersions/SchemaFingerprint.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            var aid = cache.GetAttributeId(id);
+            sb.AppendLine(id.ToString());
+            sb.AppendLine(cache.GetValueTag(aid).ToString());
+            sb.AppendLine(cache.IsIndexed(aid).ToString());
+            sb.AppendLine(cache.IsReference(aid).ToString());
+            sb.AppendLine(cache.IsCardinalityMany(aid).ToString());
+            sb.AppendLine(cache.IsNoHistory(aid).ToString());
+            sb.AppendLine("--");
```
erri120, line 27: Should we also include whether an attribute is optional or not here?

src/NexusMods.DataModel.SchemaVersions/NexusMods.DataModel.SchemaVersions.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
```
erri120, line 6:
```suggestion
```

src/NexusMods.DataModel.SchemaVersions/NexusMods.DataModel.SchemaVersions.csproj
```diff
@@ -0,0 +1,21 @@
...
+        <RootNamespace>NexusMods.DataModel.Migrations</RootNamespace>
```
erri120, line 7: What's up with this?

src/NexusMods.DataModel.SchemaVersions/Migrations/UpsertFingerprint.cs
```diff
@@ -0,0 +1,40 @@
...
+        if (!db.AttributeCache.Has(SchemaVersion.Fingerprint.Id))
+            return true;
+        
+        var fingerprints = db.Datoms(SchemaVersion.Fingerprint);
+        // No fingerprint, we need to create it
+        if (fingerprints.Count == 0)
+            return true;
+        
+        var currentFingerprint = SchemaFingerprint.GenerateFingerprint(db);
+        var dbFingerprint = Hash.From(ValueTag.UInt64.Read<ulong>(fingerprints.First().ValueSpan));
```
erri120, line 28:
```suggestion
        var dbFingerprint = Hash.From(UInt64Serializer.Read(fingerprints[0].ValueSpan));
```

src/NexusMods.DataModel.SchemaVersions/Migrations/UpsertFingerprint.cs
```diff
@@ -0,0 +1,40 @@
+using NexusMods.Hashing.xxHash3;
+using NexusMods.MnemonicDB.Abstractions;
+using NexusMods.MnemonicDB.Abstractions.ElementComparers;
```
erri120, line 3:
```suggestion
using NexusMods.MnemonicDB.Abstractions.ElementComparers;
using NexusMods.MnemonicDB.Abstractions.ValueSerializers;
```

tests/NexusMods.DataModel.SchemaVersions.Tests/NexusMods.DataModel.SchemaVersions.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+
+        <IsPackable>false</IsPackable>
+        <IsTestProject>true</IsTestProject>
```
erri120, line 9:
```suggestion
```

tests/NexusMods.DataModel.SchemaVersions.Tests/NexusMods.DataModel.SchemaVersions.Tests.csproj
```diff
@@ -0,0 +1,24 @@
...
+        <RootNamespace>NexusMods.DataModel.Migrations.Tests</RootNamespace>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="coverlet.collector" Version="6.0.0"/>
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0"/>
+        <PackageReference Include="Verify.Xunit" />
+    </ItemGroup>
```
erri120, line 17:
```suggestion
```

tests/NexusMods.DataModel.SchemaVersions.Tests/SchemaFailsafes.cs
```diff
@@ -0,0 +1,86 @@
...
+            .Select(id =>
+                {
+                    var aid = db.AttributeCache.GetAttributeId(id);
+                    return new string[]
+                    {
+                        id.ToString(),
+                        db.AttributeCache.GetValueTag(aid).ToString(),
+                        db.AttributeCache.IsIndexed(aid).ToString(),
+                        db.AttributeCache.IsCardinalityMany(aid).ToString(),
+                        db.AttributeCache.IsNoHistory(aid).ToString(),
```
erri120, line 26: Same as before, I think this should include whether the attribute is optional or not.

tests/NexusMods.DataModel.SchemaVersions.Tests/Resources/Databases/Descriptions.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Adding new test data
+To add new data, set up the app in the state you want to capture. Next log out of
+the app. This is critical as it will clear out your personal data from the database (via excise).
+
+Next zip up the `MnemonicDB.rocksdb` folder from the app folder, and place it in the `Resources/Databases` folder with a name
+that includes the date of the snapshot. Note, the tests expect there to be a `MnemonicDB.rocksdb` folder in the zip file at the root.
+So don't zip up the contents of the folder, zip up the folder itself.
+
+Finally, update the `Descriptions.md` file with the new data.
```
erri120, line 13: We'd probably want to have a CLI command do all this work.

erri120 (changes requested):

src/NexusMods.DataModel.SchemaVersions/SchemaFingerprint.cs
halgari: Optional doesn't make a difference to the database itself, so it's not important here, we can add it later if we need to.

erri120: Changing an attribute from optional to required is something we want to know about. That's the only operation (add, remove, modify) on optional attributes that can result in issues between schema versions.

While the DB probably doesn't care, it's the usage I'm worried about. Going from optional to required means our code now assumes that entities will always have the attribute. If we don't catch this change in our verify test and don't update our code or write a migration to account for this change, we'll end up getting runtime exceptions.

If the schema changes, an attribute from optional to required, we need to write a migration to fill in the blanks with some form of data.

src/NexusMods.DataModel.SchemaVersions/NexusMods.DataModel.SchemaVersions.csproj
halgari: I think it's something with a template that created this file, deleting it

tests/NexusMods.DataModel.SchemaVersions.Tests/SchemaFailsafes.cs
halgari: I can add this in, in the future. Right now this is pulling data from the MnemonicDB internal schema (in the datomstore itself), I'm fine with adding optional as another field on there, but I need to do another release of MnemonicDB first.

src/NexusMods.DataModel.SchemaVersions/SchemaFingerprint.cs
halgari: See below

tests/NexusMods.DataModel.SchemaVersions.Tests/SchemaFailsafes.cs
erri120: That's fine for this PR, but it's something we'd want sooner rather than later.

erri120 (approved):

PR #2239, opened 2024-11-11
Assert LFS download

erri120:
Makes it easier to debug why tests using LFS files are failing.


halgari:
fixed by adding more tests to #2242

PR #2242, opened 2024-11-11
Add a sanity check to the legacy database sanity checks

halgari:
Based on checking if the file is a ZIP file instead of a known string (that maybe could run afoul of newline/encoding issues?) same sort of thing as #2239


erri120 (approved):

erri120:
I'll close #2239 if the tests like this more.

PR #2246, opened 2024-11-13
Use `DateTimeOffset` and `TimeProvider`

erri120:


Al12rs:
Regarding the test failures, the `GeneralModManagementTests` failing because they contain transaction Ids that can randomly change based on the order of things happening in the DB, not relevant to the actual outcome.

I changed those tests to exclude EntityIds and TransactionIds from the verify snapshots in my Load Order PR #2248

Al12rs:
I don't know about the other failures. 
I reviewed the code and it looks good otherwise

Sewer56 (approved):
Mostly skimmed through this given it was already reviewed (but with forgotten approval), looks good to me.

PR #2248, opened 2024-11-13
Load Order v1 (REDmods, simple list ui)

### LINKED ISSUE #2098: Load Order - REDmod backend
The first phase of load order will involve making sure the backend can identify items that require ordering. 

This will include:
RedMod ordering for Cyberpunk 

No UI work will be involved in this initial phase. 

Definition of done:
This item will rely heavily on backend work and could prove difficult to QA before the UI work is completed in phase 2.

### LINKED ISSUE #2099: Load Order UI v1: flat button list, no styling
Phase two of load order will utilise the backend work completed in phase 1 to display a flat list of items that need to be ordered. 

No interaction with this list will be introduced at this stage. It will be a flat list. 

https://www.figma.com/design/1sbIAJGr8URkY3wmisyaJg/%F0%9F%93%B1-Load-order?node-id=264-14154&node-type=section&t=IoDRh4h53c2O0jTt-0

In addition to displaying the items that need to be ordered the backend needs to understand how to load them in an order, ready for the next phase. 

Definition of done
A list can be seen of mods that need to be loaded in order. 
Knowledge of which mods need ordering will help here. 
Installing mods that need ordering and then seeing if they appear in the new list will signify that UI piece of work is complete.

Al12rs:
- closes #2098 : 
- closes #2099 :
    - REDmod sorting prototype UI
    
    
This PR implements a first implementation of load order backend and barebones ui for REDmod sorting.

## System overview:
Games will expose a list of `ISortableItemProviderFactory`, one for each Load Order that the game has (e.g. RedMods and .archives for Cyberpunk).

`ISortableItemProviderFactory` exposes `GetLoadoutSortableItemProvider()` method that returns a `ILoadoutSortableItemProvider`, which can be thought of as load order manager for a specific loadout.

`ISortableItemProviderFactory` implementations should be instantiated on game instance creation (e.g. in the game class constructor), as it needs to listen to Loadouts changes to setup the `ILoadoutSortableItemProvider`s.

For now the loadout providers only expose a collection of sortable items, to be used in the ui, and a `SetRelativePosition()` method to change the order.

The `ILoadoutSortableItemProvider` implementations will be in charge of updating the sorting as changes are made to the loadout (e.g. additions or removals of RedMods).

### Db Models
The sorting information is persisted in the database in the form of two new models:
- `SortOrder`
- `SortableEntry`

`SortOrder` contains a reference to the Loadout that it applies to and serves as the containing group for the `SortableEntry`s.
This is similar enough to a `LoaodutItemGroup` but for now it isn't one.

`SortableEntry` represent an item being sorted and contains the `SortIndex` attribute. Should be extended with implementations that contain the game specific identifier used by the game (e.g. red mod folder name, bg3 .pak uuids, bethesda plugin names).
 
The `SortingIndex` is kept separate from the actual `LoadoutItem`s that are being sorted, since we can actually have multiple LoadoutItems at the same position in the load order, what is being sorted depends on what ID the game uses. E.g. for bethesda, what the LoadOrder defines is the sorting of plugin names, not `LoadoutItem`s.

Keeping `SortingIndex` separate from the actual items allows the possibility of storing backups of load orders that are not bound to a specific configuration of `LoadoutItems`. This also allows keeping the order in case of mod updates, since the `SortingIndex` is not lost on removal of the `LoadoutItem` but is instead kept on a separate db entity that only has the game specific id (e.g. red mod folder name).

### Determining active state and conflict winners
Current implementation is not dealing with conflicts between mods and only approximatively checks if a redmod is actually getting deployed. Data regarding which mod is providing a specific mod in the load order might not be accurate.

### Drag&Drop support
Current ui prototype only allows changing the order through up and down buttons. We want Drag&Drop support, but that will require a writable list that will then need to be kept synchronized. More complexity.

### Generated load order files
Cyberpunk `.archives` and BG3 pak files load orders require wiring out a file containing the order that the game will then use.
Regenerating these files every loadout change and backing up those files in our file store would just create a lot fo churn and should be avoided. 
Basically supporting Load Order there requires implementing a new file type `GeneratedFiles` in the synchronizer.
We had some early concepts of this in our codebase before, but they were removed since they weren't used at the time.

### Eventual consistency nature of load orders
The providers listen to DB changes to know when a new mod is added or removed from the loadout that might matter for the loadorder, this means that when a new mod is added, the commit that adds the mod to the Loadout will not contain the loadorder information for the files inside that mod. That data will only be available in a future commit and there is no way to determine which one.

This can cause issues for systems that also listen to changes in the db and would like to get load order information for newly added files (e.g. diagnostics). 

Multiple solutions could be considered:
- Updating the sorting as part of the commits changing the loadout, this will require the transaction being passed around from wherever items are added or removed from the loadout. 
- Updating the sorting as a separate transaction but communicating the end of a change in the loadorder with a method that doesn't rely on the state of the DB (hooks, events, messages, etc).
- Versioning the SortOrder to specific revisions of the Loadout, to create a link between `SortOrders` and the data they are being computed over. This might require a monotonically increasing revision attribute on the `Loadouts` to reference, and considering whether we would want to recompute the sorting for every Loadout change, or just the latest available. 

Either of these require non insignificant changes in the code and our general approach when it comes to changing the loadout and reacting to it.


erri120:
I don't know if anyone is already reviewing this, but this PR should've been split into at least 2 separate PRs. One for all the restructuring and namespace changes and one for the actual load order changes.

Al12rs:
> I don't know if anyone is already reviewing this, but this PR should've been split into at least 2 separate PRs. One for all the restructuring and namespace changes and one for the actual load order changes.

I agree, I was already in the middle of load order changes when I had to move a couple of things to work around circular dependencies, then since I had to move a couple of files I also moved the rest that made sense to keep a coherent structure. Some UI stuff was moved around, then moved back after we moved out view model stuff from the games.

The code went through quite a few iterations, from trying to get the initial sortable items code from Halgari working, figuring out the issues and changing the approach, changing the model strucutre, moving the view model stuff out of games, experimenting with Cycharp observable collections etc.

I think this feature needed some more upfront design work, to be able to more cleanly split it up and refine it into chunks of work, I guess the assumption was that the design work was already done on the in the starting code.

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProviderFactory.cs
```diff
@@ -0,0 +1,78 @@
+using System.Diagnostics;
+using DynamicData;
+using NexusMods.Abstractions.Games;
+using NexusMods.Abstractions.Loadouts;
+using NexusMods.MnemonicDB.Abstractions;
+using R3;
+
+namespace NexusMods.Games.RedEngine.Cyberpunk2077.SortOrder;
+
+public class RedModSortableItemProviderFactory : ISortableItemProviderFactory
```
Sewer56, line 10: It's not necessarily blocking for this PR, but I think most of this can be lifted out into an abstract class. `ARedModSortableItemProviderFactory`, if we're going by Halgari conventions of the codebase.

There doesn't seem to be anything specifically REDmod specific here; and it seems likely that this boilerplate will be repeated again. `RedModSortableItemProvider` is just an `ILoadoutSortableItemProvider`; so you could do `ARedModSortableItemProviderFactory<TLoadoutSortableItemProvider> where TLoadoutSortableItemProvider : ILoadoutSortableItemProvider` and have almost all the logic, short of `SortOrderName` and `StaticSortOrderTypeId` be moved to the abstract type.

tests/Games/NexusMods.Games.RedEngine.Tests/RedModDeployToolTests.cs
```diff
@@ -0,0 +1,111 @@
...
+using R3;
+using ReactiveUI;
+using StrawberryShake.Extensions;
+using Xunit.Abstractions;
+
+namespace NexusMods.Games.RedEngine.Tests;
+
+public class RedModDeployToolTests : ACyberpunkIsolatedGameTest<Cyberpunk2077Game>
+{
+    private readonly ITestOutputHelper _testOutputHelper;
```
Sewer56, line 20: This was presumably used to log before, it's no longer needed currently.

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProviderFactory.cs
Al12rs: Yeah, it would just be `ASortableItemProviderFactory` without the redMod part, but that is the idea. 
I want to add the abstractions when I add support for the next load orders (`.archive` LO, BG3 pak LO etc), there will probably be more stuff to abstract away at that point.

The factory in particular is likely going to have to include a bunch of UI related info as well.

tests/Games/NexusMods.Games.RedEngine.Tests/RedModDeployToolTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            {
+                if (!tsc1.Task.IsCompleted)
+                {
+                    tsc1.SetResult(Unit.Default);
+                };
+            } );
+        await tsc1.Task;
+        
+        var order = provider.SortableItems;
+        var specificGroup = order.OfType<RedModSortableItem>().Single(g => g.DisplayName == name);
```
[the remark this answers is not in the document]
Al12rs: Good catch

tests/Games/NexusMods.Games.RedEngine.Tests/RedModDeployToolTests.cs
```diff
@@ -0,0 +1,109 @@
...
+{
+    private readonly RedModDeployTool _tool;
+
+    public RedModDeployToolTests(ITestOutputHelper helper) : base(helper)
+    {
+        _tool = ServiceProvider.GetServices<ITool>().OfType<RedModDeployTool>().Single();
+    }
+    
+    [Fact]
+    public async Task LoadorderFileIsWrittenCorrectly()
```
Sewer56, line 28: There's a small caveat with these tests.

The tests here are also asserting the determinism of the underlying mod installer, i.e.
`RedModInstaller.ExecuteAsync`, since one of the archives, `several_red_mods.7z` contains
multiple mods.

The installer *does* sort the metadata files, in `infosList.OrderBy(x => x.InfoJson.Name)` @ `RedModInstaller.cs`, which is good 👍.
However, a note in the source code still needs to be made somewhere that changing that constitutes a possible breaking change.

Either in this method (preferably here since if test fails someone would look here) or in the relevant installer code.
Just want to avoid someone doing unnecessary debugging if possible.

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProvider.cs
```diff
@@ -0,0 +1,387 @@
...
+
+            // Get the current index of the item relative to the full list
+            var currentIndex = stagingList.IndexOf(sortableItem);
+
+            // Get the new index of the group relative to the full list
+            var newIndex = currentIndex + delta;
+
+            // Ensure the new index is within the bounds of the list
+            if (newIndex < 0 || newIndex >= stagingList.Count)
+                return;
```
Sewer56, line 102: We have an unused return value here; think it might be worthwhile to return error cases as a result?
We could also possibly clamp the delta to min/max. Just some random thoughts that came to mind.

tests/Games/NexusMods.Games.RedEngine.Tests/RedModDeployToolTests.cs
Al12rs: Good point, I went over the code and I think we should be good on that front due to this line here:
https://github.com/Nexus-Mods/NexusMods.App/blob/7046eb9583516b4573dd64ad248e12d46ff91bc0/src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProvider.cs#L72

I sort the RedModGroups by name before I iterate over them during the Synchronization operation, so new entries should get added in consistent order, regardless of installer order.

I'm pretty sure the installer does a single commit, which means that we get all the new red mods in a single batch, which then gets sorted. 

So I guess this can fail if the installer is changed to do multiple commits?

Sewer56: Yup, what we have works, just need to note it down somewhere in a comment.

src/Abstractions/NexusMods.Abstractions.Loadouts/NexusMods.Abstractions.Loadouts.csproj.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/CodeInspection/NamespaceProvider/NamespaceFoldersToSkip/=models/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
```
Sewer56, line 2: This rule allows items in the `Models` folder to have a namespace differing from file location.
Just confirming if it's intentional.

If it isn't, can fixup in a successive PR. Not this one, as it will create havoc for visited files on my end; as I'm checking every file out of the 325 off one by one.

Al12rs: Yeah, it was intentional, the idea is to not have `Models` as part of the namespace

Sewer56: :ok_hand:

src/Abstractions/NexusMods.Abstractions.Games/SortableItemViewModel.cs
```diff
@@ -0,0 +1,45 @@
...
+        
+        MoveTo = ReactiveCommand.CreateFromTask<int, Unit>(async index =>
+            {
+                await sortableItem.SortableItemProvider.SetRelativePosition(SortableItem, index - SortIndex);
+                return Unit.Default;
+            }
+        );
+    }
+
+    public ISortableItem SortableItem { get; }
```
Sewer56, line 38: Small style note. We usually put the properties before the constructors in this codebase.
No special specific reason, but figured I'd mention it.

Al12rs: 👍, I'll probably update that in next PR

src/NexusMods.App.UI/Controls/Trees/FileTreeView.axaml
```diff
@@ -9,9 +9,6 @@
                                 xmlns:system="clr-namespace:System;assembly=System.Runtime"
                                 mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
                                 x:Class="NexusMods.App.UI.Controls.Trees.FileTreeView">
-    <Design.DataContext>
-        <trees:ModFileTreeDesignViewModel />
```
Sewer56, line 13: Not strictly related to PR, but not sure why this was removed in the first place. Maybe during some quick rework? Figured I'd just make a quick note.

A basic copy of this is still in the Examples folder, at least.

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItem.cs
```diff
@@ -0,0 +1,39 @@
...
+    
+    public RelativePath RedModFolderName { get; set; }
+
+    public ILoadoutSortableItemProvider SortableItemProvider { get; }
+    public int SortIndex { get; set; }
+    public string DisplayName { get; }
+    public string ModName { get; set; }
+    public bool IsActive { get; set; }
+
+    public int CompareTo(RedModSortableItem? other)
```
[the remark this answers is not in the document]
Al12rs: Good pointer yeah. I'll make a note to do that for the abstraction work 👍

src/NexusMods.App.UI/Controls/Trees/FileTreeView.axaml
Al12rs: Most Design ViewModels were culled and murdered during the LoadoutItem transition period. The survivors were few and rare

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProvider.cs
```diff
@@ -0,0 +1,388 @@
...
+        RedModSortOrder.ReadOnly sortOrderModel,
+        ISortableItemProviderFactory parentFactory)
+    {
+        _connection = connection;
+        LoadoutId = loadoutId;
+        ParentFactory = parentFactory;
+        _sortOrderId = sortOrderModel.AsSortOrder().SortOrderId;
+
+        // load the previously saved order
+        var order = RetrieveSortOrder();
```
Sewer56, line 58: Might be a good idea to rename this to `RetrieveSortableItems()`.
`RetrieveSortOrder` may be confusing given the existing presence of `sortOrderModel`, which itself is the 'sort order', so to speak.

Al12rs: I changed `PersistOrder()` and `RetrieveSortOrder()` to `PersistSortableEntries()` and `RetrieveSortableEntries()`.
What do you think?

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProvider.cs
```diff
@@ -0,0 +1,388 @@
...
+                        // Temp values, will get updated when we load the RedMods
+                        modName: redModSortableItem.RedModFolderName,
+                        isActive: false
+                    );
+                }
+            )
+            .ToList();
+    }
+
+    private static List<RedModSortableItem> SynchronizeSortingToItems(IList<RedModLoadoutGroup.ReadOnly> redModsGroups, List<RedModSortableItem> orderList, RedModSortableItemProvider provider)
```
Sewer56, line 215: Just make a note, via XML docs or otherwise that this function updates the existing load order to remove
no longer available items and add new items that were previously not present. It's not immediately
obvious until you dig into the actual method body.

May be worth to rename the parameters to `availableRedMods` and `currentLoadOrder` too, for clarity.

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProvider.cs
Sewer56: Yeah, that works well for me 👍

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProvider.cs
Al12rs: Good call, I made some adjustments

Al12rs:
Force pushed my merge to fix some issues

PR #2251, opened 2024-11-14
Net 9 and Update Deps

halgari:
This PR does the following:

* Moves the app over to .NET 9
* Updates the action files to use the latest versions and .NET 9
* Updates MnemonicDB and TransparentValueObjects
* Fixes the code to match the new APIs in the above libraries


.github/workflows/clean_environment_tests.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
 
   build-and-test:
     if: github.event_name == 'push' || github.event.pull_request.draft == false
-    uses: Nexus-Mods/NexusMods.App.Meta/.github/workflows/dotnet-build-and-test-with-osx.yaml@ae64a3be780a74e94b59ee463a413083013c8b0c
+    uses: Nexus-Mods/NexusMods.App.Meta/.github/workflows/dotnet-build-and-test-with-osx.yaml@main
```
Al12rs, line 33: Not using a pinned version of the workflows could cause issues.
Changes to the Meta repo that also require changes to the underlying repos would break CI during the mean time. 
E.g. while the PR with the changes is being reviewed, Ci on all other PRs will break. 

Just pointing it out, in case it others have opinions on this

src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/RelativePathAttribute.cs
Al12rs, line 1: Should Paths related attributes really be in the DB repo? 
I thought since those are domain specific it would make sense to keep them here, unless the DB repo is already depending on the Paths library for other things?

Al12rs (commented):
I added a couple of more db attributes in the LoadOrder v1 pr #2248 
There are going to be conflicts there as I also created new modes that use our local attributes, those will need updating as well

erri120 (changes requested):
Some duplicate attributes still exist and should be removed. The new optional values should also be implemented instead of just defaulting to calling `.Value` where the status of the value is unclear. Use a try-get method or manual if-checks instead.

src/Abstractions/NexusMods.Abstractions.MnemonicDB.Attributes/RelativePathAttribute.cs
erri120: The DB uses the paths library, so I've just gone ahead and moved the attributes for `AbsolutePath`, `RelativePath`, and `Size` into the DB repo.

.github/workflows/clean_environment_tests.yaml
halgari: if it causes issues we can always change it back

Issue #1260, opened 2024-04-24
Game Support - Bannerlord


LukeNexusMods:

erri120:
Related: https://github.com/Nexus-Mods/NexusMods.App/issues/45

Sewer56:
- [ ] Core Diagnostics: 
    - https://github.com/Nexus-Mods/NexusMods.App/pull/2252
- [ ] Auto detect mod name
- [ ] Auto ingest existing mods/modules on manage
- [ ] Set EXE parameters/commandline flags.
- [ ] Alias binary files across game stores (Xbox <-> Others), on Deploy
- [ ] Load Order Integration
    - [ ] Import from external formats e.g. `.bmlist`, `.sav`, `.xml`
    - [ ] Show dummy items for official modules [low priority]
- [ ] Exposing Built-in

PR #2252, opened 2024-11-18
Added: Bannerlord Diagnostics & Misc Improvements

Sewer56:
Changelog (simplified):

- Diagnostics: As demo'd last week. 
    - Includes dummy mods to replicate every possible diagnostic.
- Synchronizer: Skip backup of base game mods/modules.
    - Only backup core binaries and module metadata, for future display purposes.

Related:
- https://github.com/Nexus-Mods/NexusMods.App/issues/1260

--------------

Most of the diagnostics code/improvements were submitted upstream, to the codebase shared by multiple mod managers.
So there isn't much here to review, mainly a lightweight integration of the tech, and the other misc changes.


src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Diagnostics/MountAndBlade2BannerlordDiagnosticEmitter.cs
```diff
@@ -0,0 +1,200 @@
...
+                ModName: missingDependencyId.Module.Name
+            ),
+
+            // Base classes, ignore.
+            ModuleVersionMismatchRangeIssue moduleVersionMismatchRangeIssue => null,
+            ModuleVersionMismatchSpecificIssue moduleVersionMismatchSpecificIssue => null,
+            ModuleVersionMismatchIssue moduleVersionMismatchIssue => null,
+            
+            // A new variant that's unknown to our code.
+            _ => throw new ArgumentOutOfRangeException(nameof(issue), issue, null),
```
Al12rs, line 197: Should maybe be Debug assert?

Sewer56: I was hoping to avoid the default case altogether when I originally wrote this, but the compiler wasn't letting me.

Ideally not handling a possible type should be a compile time error here. Can debug assert this, but then nobody would know if something was added, because nobody would ever report an issue.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Helpers.cs
```diff
@@ -0,0 +1,43 @@
...
+            }
+            catch (Exception e)
+            {
+                logger.LogError(e, "Exception while getting manifest for `{Name}`", bannerlordMod.AsLoadoutItemGroup().AsLoadoutItem().Name);
+                continue;
+            }
+
+            yield return (bannerlordMod, resource.Data);
+        }
+    }
```
Al12rs, line 42: This seems weird, I don't think you need the `ToAsyncEnumerable` on the the modules. Wouldn't a normal for loop over the enumerable make sense?

Sewer56: I tried mirroring what the other games did, ala SMAPI mods and the like. I don't think it's strictly needed either, can most likely patch it up when I get the chance.

Al12rs (approved):
See feedback above, can either be handled in this PR or in follow up one

Issue #2254, opened 2024-11-18
Collections UI


LukeNexusMods:
TODO:

- [x] #2548
- [x] #2601
- [x] #2341
- [x] #2581
- [ ] #2552
- [x] #2553
- [x] medium: update cards with collection status
- [x] medium: custom collection loadout page (#2505)
- [x] #391
- [x] unknown: collection loadout page for local collections
- [x] unknown: read-only TreeDataGrid (remember optional mods though)
- [x] medium: add "half-installed" collections, a collection should only be considered installed when all required mods have been installed (#2485)
- [x] #2464
- [x] #2432
- [x] #2444
- [x] #2434
- [x] long: Change download text to "External Download" for external files that have been checked and can't be downloaded directly.
- [x] Add Explore button that opens File Picker to add the downloaded external file.
- [x] medium: Disable "Download all"-buttons when downloads are already running (#2474)
- [x] Change item type for installed downloads into "installed download"-type that doesn't have a "Download"-button (#2440)
- [x] easy: after installation, hide "Install Collection"-button (#2490)
- [x] easy: after installation, show "View installed Collection"-button (#2490)
- [x] easy: Disable "Install"-button when already installing (#2474)
- [x] easy: styling buttons in toolbar
- [x] easy: after installation, change status text to "Ready to play - All required mods installed" (#2474)
- [x] easy: add spinner when downloading or installing (#2474)
- [x] unknown: styling download buttons on each item

insomnious:
Thresholds and Brushes for styling collection rating:

if rating >= 75
  `SuccessStrongBrush`
if rating >= 50
`WarningStrongBrush`
else
`DangerStrongBrush`

Pickysaurus:
Important note on success rating. 
If a collection has less than 5 votes, don't show a rating yet.

PR #2261, opened 2024-11-19
Add missing collection metadata

erri120:
- Removes deprecated tags
- Adds collection category
- Adds collection revision adult status
- Adds collection total downloads

Part of #2254.


Al12rs (approved):
The DB attribute changes are technically breaking, but I guess it doesn't matter if no one has that data in their DBs

PR #2274, opened 2024-11-20
MD5 hash every locally added file

erri120:
Performance can be improved later, this just puts the data in the DB so we have it.


erri120:
> Presumably this is for matching with Nexus API

This is for both querying metadata from the Nexus API via the MD5 hash, and for external collection downloads, so we know when the user downloaded and added the file to the app. This was the result of what was discussed yesterday in a meeting about how to handle external collection downloads that we can't directly download.

Sewer56:
👌

Issue #2155, opened 2024-10-10
App asks you to log in when you are already logged in


LukeNexusMods:
Open the app 
Log in
Close the app
Open the app
App asks you to log in but it looks like you are already logged in as you can see your user profile picture in the corner.

LukeNexusMods:
It might only do this on the second time you open the app. 

On the third it didnt ask me.

LukeNexusMods:
Potentially fixed.

Pickysaurus:
This issue is still present in 0.6.3

LukeNexusMods:
fixed

PR #2283, opened 2024-11-20
Fix "login after exit" bug

halgari:
Fixes https://github.com/Nexus-Mods/NexusMods.App/issues/2155

We had a bit of a race condition here that only presented itself in official builds (as they are more optimized). This PR should fix the issue by making a hard check on the existance of a OAuth login. If we have an OAuth login we won't try and show the login box.

### DIFF
--- a/src/NexusMods.App.UI/Overlays/Login/StartupMessageBox/LoginMessageBoxViewModel.cs
+++ b/src/NexusMods.App.UI/Overlays/Login/StartupMessageBox/LoginMessageBoxViewModel.cs
@@ -32,6 +32,11 @@ public class LoginMessageBoxViewModel : AOverlayViewModel<ILoginMessageBoxViewMo
     
     public bool MaybeShow()
     {
+        // If we're already logged in via OAuth, don't show the login modal
+        if (_loginManager.IsOAuthLogin) 
+            return false;
+        
+        // If we've already shown the modal, don't show it again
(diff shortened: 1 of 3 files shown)


src/NexusMods.App.UI/Overlays/Login/StartupMessageBox/LoginMessageBoxViewModel.cs
```diff
@@ -32,6 +32,11 @@ public LoginMessageBoxViewModel(ISettingsManager settingsManager, ILoginManager
     
     public bool MaybeShow()
     {
+        // If we're already logged in via OAuth, don't show the login modal
+        if (_loginManager.IsOAuthLogin) 
+            return false;
+        
+        // If we've already shown the modal, don't show it again
         if (_settingsManager.Get<LoginSettings>().HasShownModal) return false;
```
Al12rs, line 40: `HasShownModal` ended up being false in release builds here?

erri120: That seems really odd. The point of this method is to check whether the user has already seen the modal or not. The login status doesn't matter. I feel like this change doesn't solve the underlying issue, whatever that is.

erri120:
Should this be closed in favor of #2290?

Al12rs:
Closing in favour of #2290

Issue #2264, opened 2024-11-19
Support Setting Commandline Args on Heroic


Sewer56:
# Improvement

## Currently 

We have no way to set any additional commandline arguments while booting the game with the Heroic launcher.

## We want

A way to set additional commandline arguments when starting the game.

## Supporting Information

[src/backend/protocol.ts#L104](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/3fa5605031d62bff642d2a2ff580d22d33ff4d3b/src/backend/protocol.ts#L104) seems to indicate that it is not (currently) possible to add additional commandline arguments in Heroic itself. We will need to talk with the Heroic devs to see if it's possible to extend the protocol.

This will be required for some more special cased games such as [Bannerlord](https://github.com/Nexus-Mods/NexusMods.App/issues/1260#issuecomment-2482113863).

Pickysaurus:
@Sewer56 Has this been raised with Heroic's devs?

Al12rs:
Heroic merged needed changes for this, so this should no longer be blocked.

This is needed to run `redmod.exe` to configure load order for Cyberpunk on Heroic launcher on Linux.

PR #2284, opened 2024-11-20
Make Bannerlord 'Just Barely Work'

Sewer56:
Feel free to review this; draft because I haven't wrote this full description.
Be aware this PR is full of hacks.

image

Summary:
- Bannerlord boots straight into games with mods enabled
- Bannerlord mods are automatically sorted (LO is very necessary in this game, thankfully we can AutoSort somewhat reliably)
- Bannerlord isn't launnched via Steam directly, instead we skip to the binary.
- Bannerlord Xbox mods can be installed on other stores, and vice versa.
- No longer shows missing modules for game files 
- Processes can now be started with custom commandline arguments
- Bannerlord class names were simplified, now are prefixed with `Bannerlord` rather than `MountAndBlade2Bannerlord`
- Misc Code Cleanups

Problems:
- Heroic is not currently supported
    - https://github.com/Nexus-Mods/NexusMods.App/issues/2264


Sewer56:
Will resolve any feedback in morning. I've stayed up just a bit longer today to bring this over the line for release.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/BannerlordRunGameTool.cs
```diff
@@ -0,0 +1,83 @@
...
+        // Copied from Bannerlord.LauncherManager
+        var manifestPipeline = Pipelines.GetManifestPipeline(_serviceProvider);
+        var modules = (await Helpers.GetAllManifestsAsync(_logger, loadout, manifestPipeline, cancellationToken).ToArrayAsync(cancellationToken))
+            .Select(x => x.Item2);
+        var sortedModules = AutoSort(Hack.GetDummyBaseGameModules()
+            .Concat(modules)).Select(x => x.Id).ToArray();
+        var loadOrderCli = sortedModules.Length > 0 ? $"_MODULES_*{string.Join("*", sortedModules)}*_MODULES_" : string.Empty;
+
+        // Add the new arguments
+        return commandLineArgs.Concat(["/singleplayer", loadOrderCli]).ToArray();
```
Al12rs, line 69: So for now we are limited to Single player and we don't allow users to edit the load order?

I think it's fine to just get people something for now, but I would want to change this to start the Launcher and write the load order to file instead (by adding the load order file to Ignore entries and manually writing it out after Apply, since we don't have GeneratedFIles)

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Hack.cs
```diff
@@ -0,0 +1,747 @@
+using System.Collections;
+using System.Xml;
+using Bannerlord.ModuleManager;
+namespace NexusMods.Games.MountAndBlade2Bannerlord;
+
+/// <summary>
+/// Temporary code, to make game 'usable' while we wait for approval to make the game
+/// properly usable. Scheduled for DELETION.
+/// </summary>
+public static class Hack
```
Al12rs, line 10: I hate this as much as you probably do, we should really implement something to recognize and handle game/external files.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Installers/MountAndBlade2BannerlordModInstaller.cs
Al12rs, line 1: FileName for this is still `MountAndBlade2Bannerlord` do you want to change it to just `Bannerlord` here as well?

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Hack.cs
Sewer56: Yup, it's the best I can do, given the limitations I've been presented with of not ingesting.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Installers/MountAndBlade2BannerlordModInstaller.cs
Sewer56: Yeah, let me get that sorted

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/BannerlordRunGameTool.cs
Sewer56: We could theoretically do that once we have external changes available in the App.
For now, I've went with this simple approach as it's the most 'usable' we can make the game for the end users.

Booting directly into game with singleplayer is actually what the other managers do, so it's consistent with user expectation. That said, ideally I'd love to separate singleplayer and multiplayer out in the UI, and let the user boot the game in either mode directly from our end.

src/Games/NexusMods.Games.MountAndBlade2Bannerlord/Installers/BannerlordModInstaller.cs
```diff
@@ -0,0 +1,216 @@
...
+            var storeFilesToCopy = installResult.Instructions.OfType<CopyStoreInstallInstruction>().ToArray();
+            var hasXboxFiles = storeFilesToCopy.Any(x => x.Store == GameStore.Xbox);
+            var hasNonXboxFiles = storeFilesToCopy.Any(x => x.Store != GameStore.Xbox);
+            var usedDestinations = new HashSet<RelativePath>();
+            
+            foreach (var instruction in storeFilesToCopy)
+            {
+                // Note(sewer) Alias Xbox store with Steam store files, in case mod author
+                //      included files for only one version of the game.
+                //      For more info, see `0004-MountAndBlade2Bannerlord.md`.
```
Al12rs, line 88: Is there no chance that if a mod only contains files for one store, those files are simply not going to work for other stores?
So this aliasing would be useless to negative?

Sewer56: Can you rephrase the question?

In practice: 
- Almost (basically) all mods release with files for non-Xbox. 
- Xbox runs on .NET 6, while other platforms run on .NET Framework. 
- Running non-Xbox mods on Xbox is generally safe
- Running Xbox mods on non-Xbox works depending on `TargetFramework` a mod was compiled with

Al12rs: And there are not API changes from the game side itself that could make one store version not work with the other version?

Sewer56: Edit Note: 

In the 'unsafe' case here, if the Xbox mod doesn't work on non-Xbox, it's a Crash-To-Desktop.
If we don't copy Xbox files to non-Xbox, it's a Crash-To-Desktop.
So by default, chance of success is 0%.

Aliasing Xbox to non-Xbox however means there's a non-zero chance of a CTD, since some mods may work, depending on how they were compiled.

Sewer56: You basically get an error dialog telling you the DLL is missing.
Then game usually goes poof.

Al12rs: Did you discuss this feature with Aragas already? To know whether he things it would be a good idea?

Sewer56: Yeah, I did.
Got the thumbs up from there.

Al12rs: Perfect 👍

Al12rs (approved):
I didn't test since I don't have bannerlord installed, but code looks 👍

PR #2286, opened 2024-11-21
Collections download styling 2

insomnious:
Next bunch of fixes for Collections UI

- Moved to native panels where necessary, using FlexPanel for parent layout controls.
- Color changing on ratings
- Hit [this bug](https://github.com/AvaloniaUI/Avalonia/issues/13395) on description when wrapping and needing ellipses. 
- Added default styles for TabControl as well as it's SubTab class (used later in health check).
- CollectionDownloadView width and height calculations aren't 100% correct, still hitting wrapping bug above, but need to get this in as it 90% fine.

Part of https://github.com/Nexus-Mods/NexusMods.App/issues/2254


Al12rs (approved):

Al12rs:
Nice work with adding the TabControls as well, will be useful for LoadOrder UI

PR #2304, opened 2024-11-26
Collection download status text and fixes

erri120:
- The status text now shows how many required downloads are missing
- Fixed an issue where downloading a file from Nexus Mods didn't do anything. Turns out the code kept on adding new file metadata entities instead of using the associated one...

Part of #2304.


src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadViewModel.cs
```diff
@@ -89,6 +87,24 @@ public CollectionDownloadViewModel(
...
+            Observable
+                .Return(_revision)
+                .OffUi()
+                .SelectMany(revision => _collectionDownloader.RequiredDownloadedCountObservable(revision))
+                .OnUI()
+                .Subscribe(count =>
+                {
+                    if (count == RequiredDownloadsCount)
+                    {
+                        CollectionStatusText = "Ready to install - All required mods downloaded";
```
Sewer56, line 99: This should be a resource, for localization (translation) purposes.

src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadViewModel.cs
```diff
@@ -89,6 +87,24 @@ public CollectionDownloadViewModel(
             TreeDataGridAdapter.Activate();
             Disposable.Create(TreeDataGridAdapter, static adapter => adapter.Deactivate());
 
+            Observable
+                .Return(_revision)
```
Sewer56, line 91: Bit of an unusual way to do it, since this will only ever return one value; I'm guessing it's done this way so `RequiredDownloadedCountObservable` can be queried Off UI in a simple way.

Fine for now. Not an issue, but it does make me curious seeing it.

Sewer56 (approved):

PR #2320, opened 2024-11-27
Added: Support for running Protontricks via Flatpak (and extra)

Sewer56:
This PR adds the following:

- An `AggregateExecutableRuntimeDependency`.
    - This allows us to represent multiple of `ExecutableRuntimeDependency` as a simple runtime dependency.
    - Practical usages include:
        - Multiplexing (e.g. Flatpak & Native) as part of single dependency.
        - Per-OS dependencies (e.g. Run separate command on Linux and OSX).
        
- `AggregateProtontricksDependency`, which executes on either `Flatpak protontricks` or `Native protontricks`, depending on which is available.

-------

Remarks: 

If the user's has custom library directories, they may need to run `flatpak override --user --filesystem=host com.github.Matoking.protontricks` (or similar) to give flatpak's version of `protontricks` access to alternate directories.

### DIFF
--- a/src/Games/NexusMods.Games.Generic/GameToolRunner.cs
+++ b/src/Games/NexusMods.Games.Generic/GameToolRunner.cs
@@ -21,13 +21,13 @@ namespace NexusMods.Games.Generic;
 public class GameToolRunner
 {
     private readonly IProcessFactory _processFactory;
-    private readonly ProtontricksDependency? _protontricks;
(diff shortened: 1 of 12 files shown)


Sewer56:
Since I'm off, if you have any PR feedback/code change requirements, just do them directly on the branch.
It's okay :ok_hand:

erri120 (approved):

PR #2321, opened 2024-11-27
Added: Support for BLSE (Bannerlord Software Extender)

Sewer56:
This PR adds the following:

- Ability to install BLSE
- Ability to start game via BLSE
- Diagnostic for informing the user if BLSE is missing if the 'magic' `BLSE.LoadingInterceptor` or `BLSE.AssemblyResolver` dependencies are required.
- Diagnostic to install `Harmony` (required by BLSE) if missing.

## Starting BLSE On Linux

Starting BLSE on Linux will start it with the more obscure `/forcenetcore` argument of BLSE.

Reasons are multiple:

- Steam wouldn't install .NET Framework into the WINEPREFIX by default.
- The `.NET Framework` installer fails on Proton (for some reason), unlike Wine.
- wine-mono (default Framework implementation) unfortunately doesn't work well with BLSE (and I believe `Harmony`) out the box.

Good chances are the vanilla Bannerlord launcher may already be doing it out of the box, but I haven't yet checked/confirmed it by decompilation.

~~## Draft Notice~~

~~This is a draft because I'm waiting for upstream PR merge/release, this will need a NuGet version bump on the Bannerlord stuff once it's available. I've PR'd new APIs.~~

PR #2322, opened 2024-11-28
Added notes about BLSE and load order for Bannerlord

Pickysaurus:


Sewer56:
BLSE is now supported as of:
- https://github.com/Nexus-Mods/NexusMods.App/pull/2321

So this will require rewording after the merge.

Issue #963, opened 2024-02-16
Setup OSX deployment 


halgari:
# OSX deployment support
We need to make the proper image builders and packaging scripts for a OSX release.

DarthMDev:
I might look into this if no one else has done anything

Sewer56:
There's been a tuny bit of work, but only one of us has access to Mac hardware.

PR #2325, opened 2024-12-01
Add support for OSX deployment

DarthMDev:
Anything else we need before this fixes issue #963 ?


halgari:
Mostly superseeded by #2352

PR #2328, opened 2024-12-02
Rework installing a collection

erri120:
Part of #2254.

If you want to test this, you can't use collections that you've added previously. You must delete old collections and re-add them.

- Reworks how collections are installed.
- Fixes exceptions related to dates
- Filters collections by game
- Delete collections (doesn't remove the library items, it just removes the entities)
- Visit collections on Nexus Mods

### DIFF
--- a/src/NexusMods.Collections/InstallCollectionJob.cs
+++ b/src/NexusMods.Collections/InstallCollectionJob.cs
@@ -18,6 +18,7 @@ using NexusMods.Abstractions.NexusModsLibrary.Models;
 using NexusMods.Games.FOMOD;
 using NexusMods.Hashing.xxHash3;
 using NexusMods.MnemonicDB.Abstractions;
+using NexusMods.MnemonicDB.Abstractions.IndexSegments;
 using NexusMods.Networking.NexusWebApi;
 using NexusMods.Paths;
 using NexusMods.Paths.Extensions;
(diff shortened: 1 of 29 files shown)


src/NexusMods.Collections/InstallCollectionJob.cs
```diff
@@ -16,183 +14,220 @@
...
+        return monitor.Begin<InstallCollectionJob, NexusCollectionLoadoutGroup.ReadOnly>(job);
+    }
 
-        await Parallel.ForEachAsync(root.Mods, context.CancellationToken, async (mod, _) => toInstall.Add(await EnsureDownloaded(mod)));
+    /// <summary>
+    /// Installs the collection.
+    /// </summary>
+    public async ValueTask<NexusCollectionLoadoutGroup.ReadOnly> StartAsync(IJobContext<InstallCollectionJob> context)
+    {
+        var isReady = CollectionDownloader.IsFullyDownloaded(RevisionMetadata, onlyRequired: true, db: Connection.Db);
```
Al12rs, line 110: This will install the same collection multiple times, if it is already installed.
Is that the plan, or just the first iteration?

Is the expected behavior to instead verify whether it is already installed and perhaps install and missing mods in case it was only partially installed?

Al12rs: Plan is to delete and reinstall collection in the future, since they are immutable.

Issue #2277, opened 2024-11-20
Load Order v2: UI & Styling


Al12rs:
## Outline
Implement UI and Styling for Load Order view to match figma designs:
- #1959 
https://www.figma.com/design/1sbIAJGr8URkY3wmisyaJg/%F0%9F%93%B1-Rules?node-id=619-8636&t=ozUYvrxkTATN0dB4-4

## Details
![Image](https://github.com/user-attachments/assets/162addf9-54af-46d5-b90e-fbb8af900c15)

The UI will be implemented inside a separate page (`LoadOrderWIPPage`) during development. The page should contain a single View and a single ViewModel of `SortingSelectionView` that will show the tabs to switch between the types of load order that are present for a specific game.

The contents of a particular tab should be the actual `LoadOrderView`.

Notes:
- Add empty state in case the load order is empty (ask Design for one)
- Groups should be ignored for now
- Trophy bar on the left of the order should point up or down depending on some variable state (e.g. bool IsWinnerTop)
- See Figma docs for further details
- Only REDmods for cyberpunk are currently supported

Links for some REDmods to test:
- https://www.nexusmods.com/cyberpunk2077/mods/10395
- https://www.nexusmods.com/cyberpunk2077/mods/9237
- https://www.nexusmods.com/cyberpunk2077/mods/6181
Can find more by searching REDmod on cyberpunk nexus

PR #2335, opened 2024-12-03
Load order styling - Part 4

insomnious:
Part of https://github.com/Nexus-Mods/NexusMods.App/issues/2277

Includes:

- Text changes for tooltips and empty state
- Extra tooltip for loser arrow
- Displays ordinal number instead of zero-based sort index
- Refactored styles into correct files (still more to be done, especially stripping styles out of views)

### DIFF
--- a/src/NexusMods.App.UI/Pages/Sorting/LoadOrder/ILoadOrderItemModel.cs
+++ b/src/NexusMods.App.UI/Pages/Sorting/LoadOrder/ILoadOrderItemModel.cs
@@ -13,5 +13,5 @@ public interface ILoadOrderItemModel : ITreeDataGridItemModel<ILoadOrderItemMode
     
     public string ModName { get; }
     public bool IsActive { get; }
-    public string SortIndexWithSuffix { get; }
+    public string SortOrdinalNumber { get; }
 }
--- a/src/NexusMods.App.UI/Pages/Sorting/LoadOrder/ILoadOrderItemModel.cs
+++ b/src/NexusMods.App.UI/Pages/Sorting/LoadOrder/ILoadOrderItemModel.cs
@@ -20,7 +20,7 @@ public interface ILoadOrderItemModel : ITreeDataGridItemModel<ILoadOrderItemMode
     /// </summary>
     /// <param name="sortIndex">The zero-based index to convert</param>
     /// <returns>A string representing the ordinal number with the appropriate suffix.</returns>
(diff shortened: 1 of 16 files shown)


src/NexusMods.App.UI/Pages/Sorting/LoadOrder/ILoadOrderItemModel.cs
```diff
@@ -13,4 +13,5 @@ public interface ILoadOrderItemModel : ITreeDataGridItemModel<ILoadOrderItemMode
     
     public string ModName { get; }
     public bool IsActive { get; }
+    public string SortOrdinalNumber { get; }
```
Al12rs, line 16: What do you think of `DisplaySortIndex` for this?

insomnious: I was curious if there was any precedent for a string representation of an index

src/NexusMods.App.UI/Pages/Sorting/LoadOrder/ILoadOrderItemModel.cs
```diff
@@ -13,4 +13,24 @@ public interface ILoadOrderItemModel : ITreeDataGridItemModel<ILoadOrderItemMode
...
     public string ModName { get; }
     public bool IsActive { get; }
+    public string DisplaySortIndex { get; }
+    
+    /// <summary>
+    /// Converts a zero-based index to an ordinal number string with the appropriate suffix.
+    /// </summary>
+    /// <param name="sortIndex">The zero-based index to convert</param>
+    /// <returns>A string representing the ordinal number with the appropriate suffix.</returns>
+    internal static string ConvertZeroIndexToOrdinalNumber(int sortIndex)
```
Al12rs, line 23: If you put it here I think you can make it protected maybe?
But I think having it in the VM implementation makes sense since it is an implementation detail.
I guess it doesn't matter too much.

Al12rs (approved):

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/TreeDataGrid/SortOrderStyles.axaml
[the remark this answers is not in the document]
insomnious: If it's ok with you, save this for the tidy up PR after I've moved all the styles out of the views and into the right seperate files, now we have some terms defined

PR #2351, opened 2024-12-05
Game hashing

halgari:


erri120:
Will take a look when it's ready for review.

erri120:
You'd probably wanna use `JsonAugment` on value types instead of making a custom JSON serializer.

PR #2352, opened 2024-12-06
Mostly working app on MacOS

halgari:
Somehow it all clicked once I got a box that can actually build the app. This does the following:

- Moves standard app logic into `NexusMods.App.Common`
- Makes `NexusMods.App` now purely focused on Windows/Linux support. It still targets base `net-9.0`, and `WinExe`
- Adds NexusMods.App.MacOS that is a `net9.0-macos` app focused on settings and startup utilities needed for mac support
- Adds to the above project the `Info.plist` settings required for URL dispatching and setup of icons
- Adds the proper Iconset files
- Adds a python script for resizing images for the iconset (this is rarely needed)
- Adds changes to the core of the Avalonia startup to handle URL events. This is now cross-platform, but I'm not aware of any other OSes that send files this way. 
- Fixes a bug with URL opening on macOS when the URL contains spaces (like login URLs do)

There's still a few bugs here. The app crashes in release mode, and the `7zz` files aren't getting copied into the app bundle. I'll research that later. But this is miles ahead of anything we've had so far. 

Next week I'll make sure this doesn't break anything in the other OSes, but it's looking good so far.


DarthMDev:
Lets go well done 👏 ill try to look into those issues when i can if you don't beat me to it

PR #2355, opened 2024-12-09
Implement event tracking

### LINKED ISSUE #2266: Matomo event tracking
Design @captainsandypants
- Website ID
- Event IDs

Tech:
- `Telemetry.SendEvent(...)` should be static, single line to add to commands, and sync
- Use a queue or similar
- Background thread collects and sends events
- No UI block

erri120:
Resolves #2266.

### DIFF
--- a/src/NexusMods.App/Services.cs
+++ b/src/NexusMods.App/Services.cs
@@ -70,6 +70,7 @@ public static class Services
 
                 .AddSingleton<ITelemetryProvider, TelemetryProvider>()
                 .AddTelemetry(telemetrySettings ?? new TelemetrySettings())
+                .AddTracking(telemetrySettings ?? new TelemetrySettings())
 
                 .AddSingleton<CommandLineConfigurator>()
                 .AddCLI()
--- a/src/NexusMods.Telemetry/Constants.cs
+++ b/src/NexusMods.Telemetry/Constants.cs
@@ -7,10 +7,15 @@ namespace NexusMods.Telemetry;
 [PublicAPI]
 internal static class Constants
 {
-    // NOTE(erri120): don't change this, this has to be in sync with the backend
+    // NOTE(erri120): don't change this, ask me if you have questions
+    public static readonly Uri MatomoTrackingEndpoint = new("https://pw.nexusmods.com/matomo.php");
+    public const string MatomoSiteId = "7";
+
+    // NOTE(erri120): don't change this, this has to be in sync with the backend, ask me if you have questions
     public static readonly TimeSpan ExportInterval = TimeSpan.FromMinutes(1);
     public static readonly TimeSpan ExporterTimeout = TimeSpan.FromSeconds(10);
 
+    // NOTE(erri120): don't change this, ask me if you have questions
     public static readonly Uri MetricsEndpoint = new("https://collector.nexusmods.com/v1/metrics");
     public static readonly Uri TracesEndpoint = new("https://collector.nexusmods.com/v1/traces");
 
--- a/src/NexusMods.Telemetry/EventMetadata.cs
+++ b/src/NexusMods.Telemetry/EventMetadata.cs
@@ -1,0 +1,44 @@
+using System.Diagnostics;
+using JetBrains.Annotations;
+
+namespace NexusMods.Telemetry;
+
+/// <summary>
+/// Provides metadata to an event.
+/// </summary>
+[PublicAPI]
+public readonly struct EventMetadata
+{
+    /// <summary>
+    /// Current time.
+    /// </summary>
+    public readonly TimeOnly CurrentTime;
+
+    /// <summary>
+    /// Name of the event.
+    /// </summary>
+    public readonly string Name;
+
+    /// <summary>
+    /// Constructor.
+    /// </summary>
+    [Obsolete(error: true, message: "Don't use the default constructor!")]
+    public EventMetadata()
+    {
+        throw new UnreachableException();
+    }
+
+    /// <summary>
+    /// Constructor.
+    /// </summary>
+    public EventMetadata(string name, TimeProvider? timeProvider = null)
+    {
+        Name = name;
+        CurrentTime = TimeOnly.FromDateTime((timeProvider ?? TimeProvider.System).GetLocalNow().LocalDateTime);
+    }
+
+    /// <summary>
+    /// Checks whether the struct wasn't default initialized.
+    /// </summary>
+    public bool IsValid() => Name is not null;
+}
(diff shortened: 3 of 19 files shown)


src/NexusMods.Telemetry/Constants.cs
```diff
@@ -7,10 +7,15 @@ namespace NexusMods.Telemetry;
 [PublicAPI]
 internal static class Constants
 {
-    // NOTE(erri120): don't change this, this has to be in sync with the backend
+    // NOTE(erri120): don't change this, ask me if you have questions
+    public static readonly Uri MatomoTrackingEndpoint = new("https://pw.nexusmods.com/matomo.php");
+    public const string MatomoSiteId = "7";
```
Al12rs, line 12: I have questions, are the details supposed to be private? Otherwise we should add some more context in the comments, so that somebody can wrangle the meaning if you are not available.

src/NexusMods.App/Services.cs
```diff
@@ -70,6 +70,7 @@ public static IServiceCollection AddApp(this IServiceCollection services,
 
                 .AddSingleton<ITelemetryProvider, TelemetryProvider>()
                 .AddTelemetry(telemetrySettings ?? new TelemetrySettings())
+                .AddTracking(telemetrySettings ?? new TelemetrySettings())
```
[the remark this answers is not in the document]
erri120: They'll work with the same values, just not the same instance, which doesn't matter.

src/NexusMods.Telemetry/EventMetadata.cs
```diff
@@ -0,0 +1,44 @@
...
+    public readonly string? Name;
+
+    /// <summary>
+    /// Constructor.
+    /// </summary>
+    [Obsolete(error: true, message: "Don't use the default constructor!")]
+    public EventMetadata()
+    {
+        throw new UnreachableException();
+    }
```
Al12rs, line 29: Why does this exist?

src/NexusMods.Telemetry/Events.cs
```diff
@@ -0,0 +1,51 @@
...
+        public static readonly EventDefinition RemoveGame = new(Category, Action: "Remove Game");
+    }
+
+    public static class Loadout
+    {
+        private const string Category = "Loadout";
+        public static readonly EventDefinition CreateLoadout         = new(Category, Action: "Create Loadout");
+        public static readonly EventDefinition CreateLoadoutCopy     = new(Category, Action: "Create Loadout Copy");
+        public static readonly EventDefinition ApplyLoadoutChanges   = new(Category, Action: "Apply Loadout Changes");
+        public static readonly EventDefinition PreviewLoadoutChanges = new(Category, Action: "Preview Loadout Changes");
```
Al12rs, line 25: There are unused events definitions here, is the plan to add the event submissions in future PRs or in this one?

src/NexusMods.Telemetry/Constants.cs
erri120: There are no details or hidden meaning, this is just the endpoint we're sending data to and the site ID is essentially the namespace for the app, otherwise we'd be mixing app data with website data.

src/NexusMods.Telemetry/Events.cs
erri120: These events are mostly just WIP, @captainsandypants and I are still figuring out a list of events.

src/NexusMods.Telemetry/Constants.cs
Al12rs: We can say that in the comments then. 
`// The site id is essentially the namespace for events coming from the app`

src/NexusMods.Telemetry/Events.cs
Al12rs: Yeah, some of these seem a bit spammy, but I guess they are more an example for now

src/NexusMods.Telemetry/EventSender.cs
```diff
@@ -0,0 +1,253 @@
...
+
+            isFirst = false;
+            SerializeEvent(writer, eventTuple, userId);
+        }
+
+        using (var sb = ZString.CreateUtf8StringBuilder(notNested: true))
+        {
+            sb.Append("] }");
+            sb.CopyTo(writer);
+        }
```
Al12rs, line 117: Why create a different StringBuilder for every addition?
Why not creating a single string builder for everything and do a single `sb.CopyTo(writer)` at the end?

src/NexusMods.Telemetry/EventSender.cs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    /// <summary>
+    /// Serializes the event into the buffer as an UTF8 encoded string.
+    /// </summary>
+    [SuppressMessage("ReSharper", "StringLiteralTypo")]
+    private static void SerializeEvent(IBufferWriter<byte> writer, ValueTuple<EventDefinition, EventMetadata, ulong> tuple, Optional<UserId> userId)
```
Al12rs, line 124: Should probably serialize to passed StringBuilder instead of directly onto the buffer

src/NexusMods.Telemetry/EventSender.cs
```diff
@@ -0,0 +1,253 @@
...
+        // https://lx3rzx16x9.csb.app/#Windows%20NT%2010.0
+        if (platform == OSPlatform.Windows)
+        {
+            return $"Windows NT {osVersion.Version.ToString(fieldCount: 2)}";
+        }
+
+        // https://lx3rzx16x9.csb.app/#Macintosh;%20Intel%20Mac%20OS%20X%2010_1
+        if (platform == OSPlatform.OSX)
+        {
+            return $"Macintosh; Intel Mac OS X {osVersion.Version.Major}_{osVersion.Version.Minor}";
```
Al12rs, line 248: We can't detect the CPU architecture? For distinguishing Apple silicon?

erri120: You can ignore this method, it's not used. Matomo user agent detection is really whack.

src/NexusMods.Telemetry/EventSender.cs
erri120: See above.

src/NexusMods.Telemetry/EventSender.cs
Al12rs: There doesn't seem to be a `FlushTo()`, to at least reuse the same string builder, so I guess this works, even though it doesn't seem the most intuitive.

Al12rs (approved):

PR #2360, opened 2024-12-10
Better state checks for collections

erri120:
- Install button will be disabled if you haven't downloaded all required mods or have already installed the collection
- Download button will be disabled if you have downloaded everything already or if you aren't premium

Part of #2254.


Al12rs (approved):
! This changes the DB schema again, but for collections stuff, and I think it doesn't break stuff for now. 
Just something to keep track of.

PR #2364, opened 2024-12-11
Move toolbar into tabs

erri120:
Part of #2254. UI styling needs updating.

### DIFF
--- a/src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadView.axaml
+++ b/src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadView.axaml
@@ -46,6 +46,22 @@
                 </MenuItem.Header>
             </MenuItem>
         </MenuFlyout>
+
+        <StackPanel x:Key="TabItemContent" Orientation="Vertical" Spacing="0">
+            <Border x:Name="ListHeaderRowBorder" DockPanel.Dock="Top">
+                <panels:FlexPanel x:Name="ListHeaderRow">
+                    <TextBlock x:Name="TextCollectionStatus"/>
+                    <controls:StandardButton x:Name="ButtonDownloadRequiredItems" Text="Download all required mods" />
+                    <controls:StandardButton x:Name="ButtonInstallRequiredItems" Text="Install Collection" />
+
+                    <controls:StandardButton x:Name="ButtonDownloadOptionalItems" Text="Download all optional mods" />
+                    <controls:StandardButton x:Name="ButtonInstallOptionalItems" Text="Install downloaded optional mods" />
+                    <controls:StandardButton Flyout="{StaticResource CollectionMenuFlyout}" Text="..." />
+                </panels:FlexPanel>
+            </Border>
+
+            <TreeDataGrid x:Name="DownloadsTree" />
+        </StackPanel>
     </reactiveUi:ReactiveUserControl.Resources>
 
     <DockPanel x:Name="Body">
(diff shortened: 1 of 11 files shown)


src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadView.axaml
```diff
@@ -46,6 +46,22 @@
                 </MenuItem.Header>
             </MenuItem>
         </MenuFlyout>
+
+        <StackPanel x:Key="TabItemContent" Orientation="Vertical" Spacing="0">
```
Al12rs, line 50: You shouldn't put a `TreeDataGrid` in a vertical `StackPanel`, because the `StackPanel` would have infinite heigh and that messes with the `ScrollViewer` inside the `TreeDataGrid`.

Either use a different panel or add a note for later

src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadViewModel.cs
```diff
@@ -78,13 +78,22 @@ public CollectionDownloadViewModel(
...
 
-        InstallCollectionCommand = _canInstall.ToReactiveCommand<Unit>(
+        CommandDownloadOptionalItems = _canDownloadOptionalItems.ToReactiveCommand<Unit>(
+            executeAsync: (_, cancellationToken) => collectionDownloader.DownloadItems(_revision, itemType: CollectionDownloader.ItemType.Optional, db: connection.Db, cancellationToken: cancellationToken),
+            awaitOperation: AwaitOperation.Drop,
+            configureAwait: false
+        );
+
+        // TODO: implement this button
+        CommandInstallOptionalItems = _canInstallOptionalItems.ToReactiveCommand<Unit>();
```
Al12rs, line 94: This for later PRs?

erri120: Yes

Al12rs (changes requested):
See comment about TreeDataGrid inside a StackPanel

Al12rs (changes requested):
Can't press the down

src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadViewModel.cs
```diff
@@ -130,39 +139,59 @@ public CollectionDownloadViewModel(
...
+                .OnUI()
+                .Subscribe(tuple =>
+                {
+                    var (numDownloadedOptionalItems, isPremium) = tuple;
+                    var hasDownloadedAllOptionalItems = numDownloadedOptionalItems == OptionalDownloadsCount;
+
+                    CountDownloadedOptionalItems = numDownloadedOptionalItems;
+                    _canInstallOptionalItems.OnNext(numDownloadedOptionalItems > 0);
+                    _canDownloadOptionalItems.OnNext(!hasDownloadedAllOptionalItems && isPremium);
+                }).AddTo(disposables);
```
Al12rs, line 193: Seems this is never executed, and the initial values for the canExecute are all false

erri120:
> You can't interact with the buttons right now with this change: image

What collection is that?

Al12rs:
> > You can't interact with the buttons right now with this change: image
> 
> What collection is that?

This one, I can test with others
https://next.nexusmods.com/stardewvalley/collections/cwovfq

PR #2366, opened 2024-12-11
Added empty text when no game installs found

insomnious:
Quick bug fix so we show some text when no games are discovered

image

### DIFF
--- a/src/NexusMods.App.UI/Pages/MyGames/MyGamesView.axaml.cs
+++ b/src/NexusMods.App.UI/Pages/MyGames/MyGamesView.axaml.cs
@@ -12,21 +12,31 @@ public partial class MyGamesView : ReactiveUserControl<IMyGamesViewModel>
         InitializeComponent();
 
         this.WhenActivated(d =>
-        {
-            this.WhenAnyValue(view => view.ViewModel!.InstalledGames)
-                .BindToView(this, view => view.DetectedGamesItemsControl.ItemsSource)
-                .DisposeWith(d);
-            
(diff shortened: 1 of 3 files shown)


src/NexusMods.App.UI/Pages/MyGames/MyGamesView.axaml.cs
```diff
@@ -12,21 +12,31 @@ public MyGamesView()
...
+                
+                this.WhenAnyValue(view  => view.ViewModel!.InstalledGames)
+                    .Select(games => games.Count == 0)
+                    .Subscribe(isEmpty =>
+                        {
+                            NoGamesDetectedText.IsVisible = isEmpty;
+                            DetectedGamesItemsControl.IsVisible = !isEmpty;
+                        }
+                    )
+                    .DisposeWith(d);
```
Al12rs, line 38: This only updates if the InstalledGames property is assigned, so if the entire collection is replaced, because the value of the property is the reference to the ObservableCollection.

Observe changes on the Count property of the observable collection instead. Since the collection is observable, the Count property should emit propertyChanged events.


```suggestion
                this.WhenAnyValue(view  => view.ViewModel!.InstalledGames.Count)
                    .Select(installedCount  => installedCount == 0)
                    .Subscribe(isEmpty =>
                        {
                            NoGamesDetectedText.IsVisible = isEmpty;
                            DetectedGamesItemsControl.IsVisible = !isEmpty;
                        }
                    )
                    .DisposeWith(d);
```

TBF, I don't think we actually detect if a game is getting installed while the app is open anyways, but worth making the code more correct.

Al12rs (approved):

PR #2376, opened 2024-12-12
My Games design change

### LINKED ISSUE #2375: Add pictograms to the app
Design have created branding elements that will need to be used in the app. The first of these are a SVG pictograms at various sizes. The app will need to be able to display these alongside content on pages.

They do not all need to be loaded into the app but we need to establish how they will be when needed on an individual bases.

[Figma link](https://www.figma.com/design/TtEARAAYixcdcJMSllMIO4/Brand-Assets?m=auto&node-id=4-3662&t=uTZO7joP4TSDd2QB-1)

Image

insomnious:
My Games design change required adding new brand pictograms into our existing icon system. Needed a Skia setting change to fix opacity changes. (See https://github.com/AvaloniaUI/Avalonia/pull/9964 for details).

Fixes #2375 and #2378

### DIFF
--- a/src/NexusMods.App.UI/Pages/LibraryPage/LibraryView.axaml
+++ b/src/NexusMods.App.UI/Pages/LibraryPage/LibraryView.axaml
@@ -81,8 +81,7 @@
 
         <!-- Empty State -->
         <controls:EmptyState x:Name="EmptyState" Grid.Row="1" 
-                             Header="{x:Static resources:Language.EmptyLibraryTitleText}" 
-                             IsActive="True">
+                             Header="{x:Static resources:Language.EmptyLibraryTitleText}">
             <controls:EmptyState.Subtitle>
                 <StackPanel Orientation="Vertical" Spacing="16">
                     <TextBlock x:Name="EmptyLibraryTextBlock" Text="Subtitle"/>
--- a/src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutView.axaml
+++ b/src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutView.axaml
@@ -47,8 +47,7 @@
 
 
         <controls:EmptyState x:Name="EmptyState" Grid.Row="1" 
-                             Header="Header"
-                             IsActive="True">
+                             Header="Header">
             
             <controls:EmptyState.Subtitle>
                 <StackPanel  Orientation="Vertical" Spacing="16">


src/NexusMods.App/Startup.cs
```diff
@@ -100,7 +100,8 @@ public static AppBuilder BuildAvaloniaApp(IServiceProvider serviceProvider)
             .UsePlatformDetect()
             .LogToTrace()
             .UseR3()
-            .UseReactiveUI();
+            .UseReactiveUI()
+            .With(new SkiaOptions { UseOpacitySaveLayer = true });
```
erri120, line 104: Please annotate why this is needed.

src/NexusMods.App.UI/Pages/MyGames/MyGamesView.axaml
```diff
@@ -25,25 +25,38 @@
...
                         x:Name="DetectedGamesItemsControl">
                         <ItemsControl.ItemsPanel>
                             <ItemsPanelTemplate>
-                                <panels:FlexPanel 
-                                    RowSpacing="24" 
+                                <panels:FlexPanel
+                                    RowSpacing="24"
                                     ColumnSpacing="24"
                                     Wrap="Wrap" />
                             </ItemsPanelTemplate>
```
erri120, line 62: This should be using `EmptyState`.

erri120 (changes requested):
See comments.

src/NexusMods.App/Startup.cs
insomnious: Opacity doesn't work on SVGs without this (see https://github.com/AvaloniaUI/Avalonia/pull/9964). It's needed for when we want to disable\fade SVG icons.

insomnious: Want me to add as a comment in the code ?

src/NexusMods.App.UI/Pages/MyGames/MyGamesView.axaml
insomnious: Good point

src/NexusMods.App/Startup.cs
erri120: Please add a comment in the code, linking to the relevant information.

src/NexusMods.App.UI/Pages/MyGames/MyGamesView.axaml
insomnious: It's now using the `EmptyState` control and I've changed the code so that if Header or Icons aren't set, then those controls `IsVisibility =False` so that they aren't using up empty UI real estate. This matches other controls. Also removed some extra styles and needed to tweak some just for it's usage here.

insomnious:
More refactoring than originally planned but EmptyState is now a bit more generalized with how it's getting used now with trying not to repeat too much styling and keeping the default style appropriate for majority of usage. It still needs a bit more work long term.

erri120 (approved):

PR #2381, opened 2024-12-12
Open collection JSON file in text editor

erri120:
Adds a menu item to open the collection JSON file in a text editor (read-only). Allows us to better debug collection issues.

Part of #2254.

Also documents and adds some more JSON properties.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Collections/Json/ModSourceType.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Collections/Json/ModSourceType.cs
@@ -1,4 +1,7 @@
 // ReSharper disable InconsistentNaming
+
+using System.Text.Json.Serialization;
+
 namespace NexusMods.Abstractions.Collections.Json;
 
 /// <summary>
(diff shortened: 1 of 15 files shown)


src/Abstractions/NexusMods.Abstractions.Collections/Json/ModSourceType.cs
```diff
@@ -1,13 +1,35 @@
...
+    /// <summary>
+    /// Bundled with the collection archive.
+    /// </summary>
+    [JsonStringEnumMemberName("bundle")]
+    Bundle,
+
+    /// <summary>
+    /// Downloaded externally via an URL.
+    /// </summary>
+    [JsonStringEnumMemberName("Browse")]
```
Sewer56, line 27: Can you add a quick note that 'browse' usually indicates there's an expectation of user interaction with the URL? And that sometimes it's miscategorised by collection devs.

Might be worth noting, not everyone may always know/remember.

src/NexusMods.App.UI/Pages/TextEdit/TextEditorPageViewModel.cs
```diff
@@ -138,7 +144,7 @@ public TextEditorPageViewModel(
                     };
 
                     Document = document;
-                    IsReadOnly = false;
+                    IsReadOnly = context.FileId.IsT1;
```
Sewer56, line 147: This feels like a bit of a hack; but I'm not concerned about this for now, we can leave it as-is.

Eventually down the road, mod authors will want to be able to edit their own files; and they may try doing it in the App too, so there likely will be a case where you would want non-RO LibraryFiles. But such use case is very far from now, so I'm not concerned.

Sewer56:
Only misc notes. Mergeable as-is as far as I'm concerned.

PR #2386, opened 2024-12-15
Set StartupWMClass in freedesktop Desktop Entry

MattSturgeon:
This is used to associate the Desktop Entry with the running
application, e.g. in taskbars, panels, and docks.Set `StartupWMClass` so that the Desktop Entry gets properly associated with the running app window.

image

When the app is associated with its Desktop Entry, the DE knows which icon to use for the app's window.

With this PR, the window name and icon show correctly in the dock as: 
image

---

Inspecting with GNOME's Looking Glass (`Alt+F2`, `lg`), shows the WM Class for the app's window is unsurprisingly `NexusMods.App`.
Screenshot From 2024-12-15 20-42-20

After this PR, it shows association with the Desktop Entry: 
image

---

Additionally, set `SingleMainWindow` as a hint to DEs for whether to offer a "New Window" button.

image

This makes sense for NexusMods.App, because it only supports having one "main" window open.


erri120 (approved):
Awesome!

PR #2387, opened 2024-12-15
Set `SingleMainWindow` in Desktop Entry

MattSturgeon:
Follow up to #2386, this tells the DE that we don't need a "New Window" button because we typically only have one single "main" window.

`SingleMainWindow` is a hint to DEs for whether to offer a "New Window" button.

image

This makes sense for NexusMods.App, because it only supports having one "main" window open.

---

The CI is unhappy about `SingleMainWindow`:

> error: file contains key "SingleMainWindow" in group "Desktop Entry", but keys extending the format should start with "X-"

https://github.com/Nexus-Mods/NexusMods.App/actions/runs/12342433935/job/34442250123?pr=2386#step:8:1375

The CI is validating against [Desktop Entry Spec v1.0](https://specifications.freedesktop.org/desktop-entry-spec/1.0/recognized-keys.html), however `SingleMainWindow` was added in [v1.5](https://specifications.freedesktop.org/desktop-entry-spec/1.5/recognized-keys.html).

I'm not sure if we can have this as an `X-` prefixed option and it still work, need to test that out. Otherwise, is it possible for the linter to support a newer Desktop Entry Spec version?


erri120 (approved):
Change is fine, I'll try and find out if we can upgrade to a newer spec version, I think I ran into this issue before.

Sewer56 (approved):
This is good for the time being.

Long term we are supposed to expose multi window (in fact, we previously did); but opted to disable it for the time being due to some technical caveats.

erri120:
@Sewer56 https://github.com/Sewer56/PupNet-Deploy/tree/main/PupNet/Assets it looks like we have appimagetool builds in the deploy project that are outdated. I think we can just upgrade that.

Sewer56:
@erri120 Sounds good to me.
I don't have Linux set up here are the office; I can install it, but might take a bit.
Wanna PR it instead? Then we can PR that to upstream; if upstream's willing to accept it (if you remember, upstream stopped maintaining for time being).

MattSturgeon:
> The CI is validating against [Desktop Entry Spec v1.0](https://specifications.freedesktop.org/desktop-entry-spec/1.0/recognized-keys.html), however `SingleMainWindow` was added in [v1.5](https://specifications.freedesktop.org/desktop-entry-spec/1.5/recognized-keys.html).

Has this CI been removed? My latest rebase seemingly only triggered [one CI job](https://github.com/Nexus-Mods/NexusMods.App/actions/runs/14452089438/job/40526926308?pr=2387). :thinking:

erri120:
> > The CI is validating against [Desktop Entry Spec v1.0](https://specifications.freedesktop.org/desktop-entry-spec/1.0/recognized-keys.html), however `SingleMainWindow` was added in [v1.5](https://specifications.freedesktop.org/desktop-entry-spec/1.5/recognized-keys.html).
> 
> Has this CI been removed? My latest rebase seemingly only triggered [one CI job](https://github.com/Nexus-Mods/NexusMods.App/actions/runs/14452089438/job/40526926308?pr=2387). 🤔
> 
> EDIT: or maybe the full CI test suite doesn't run on PRs labeled as `stale`?

The CI only runs if code changes were made, changes to the `.desktop` file doesn't trigger the CI.

MattSturgeon:
Ah. So has the previous desktop entry validation CI been removed?

erri120:
Closing this for now, we can't merge this until PubNet updates and includes the latest `.desktop` schema version, otherwise CI will always fail as it validates on build.

MattSturgeon:
> Closing this for now, we can't merge this until PubNet updates and includes the latest `.desktop` schema version, otherwise CI will always fail as it validates on build.

Thanks for the response. Is there a PubNet PR/issue open we can track?

---

Slightly meta, but I'm not sure closing a blocked PR is necessary the right way to deal with "we can't merge until xyz other thing happens".

If the PR is complex enough to get stale, then sure; close it once it grows stale. E.g., once it has merge conflicts and isn't being maintained.

Otherwise I believe it's better to do one of the following:
- Submit a "request changes" review to block the PR; could reference what the pr is blocked by
- Add a note to the PR description
- Mark the PR as a draft; i.e., shouldn't be merged

PR #2391, opened 2024-12-16
Remove `IImageCache`

erri120:


src/Games/NexusMods.Games.FOMOD.UI/Step/GuidedInstallerStepViewModel.cs
```diff
@@ -54,21 +58,41 @@ public GuidedInstallerStepViewModel(IImageCache imageCache)
             .WhereNotNull()
             .Select(optionVM => optionVM.Option.Image)
             .WhereNotNull()
-            .Select(optionImage =>
+            .OffUi()
+            .SelectMany(async optionImage =>
+            {
+                try
+                {
+                    if (optionImage.TryPickT0(out var uri, out var imageStoredFile))
```
Sewer56, line 66: This is technically not incorrect, but it 'feels' a bit weird. Technically you should prefer loading a local file, so you really should be asking 'do I have a local file' instead of 'do I have a URL'.

While in this case there isn't a functional difference, because it's a OneOf; I thought I'd point it out. Someone looking at this code may misinterpret it; since the `Try` methods in a branch like this are usually used with dictionaries.

Sewer56 (approved):
Rest is ok, just the comment above.

src/Games/NexusMods.Games.FOMOD.UI/Step/GuidedInstallerStepViewModel.cs
erri120: I've simplified the code using `Match`.

PR #2393, opened 2024-12-17
Fix URL encoding for events

erri120:
Forgot to URL encode event names and values.


halgari:
This needs a description of what its doing and why

Al12rs (approved):

PR #2430, opened 2025-01-06
Fix nested mod files being treated as as archives and extracted 

### LINKED ISSUE #2367: Files recognized as archives inside mods are extracted and not properly added to the mod
# Bug Report

## Outline:
If a mod added to the library contains a file that the app recognizes as an archive, the app will attempt to extract the contents of that archive, and the file entry is recorded as 0 sized file. It's unclear where the nested files end up.

This breaks installation of any mod that contains a file in archive format that should not be extracted.

## Details
This was noticed in #2193, where a `Credits.xlsx` file (archive format containing Excel sheet) was missing from the App installation of a SDV collection.
Investigation revealed that the file was being recognized as an archive, and a 0 sized ghost entry was present rather than the full archive as expected.

## Steps to reproduce
- Install https://www.nexusmods.com/stardewvalley/mods/20414?tab=files
- Preview Apply changes, notice Credits.xlsx file with 0 size.
- Apply and Credits file should be missing

Not sure yet. 
The reason this happens in the first place is to support the installation of certain mods that come with nested archives that need to be extracted during installation. We want to have all the files available upfront, before installation.

Tentatively, my assumption would be that we would want to potentially index all the child files, but also index the parent archive. 

This could have the very bad downside of the app backing up data twice, both the archive contents and the archive itself. 
If we encounter an entire game where such archives are common (e.g. if this where to happen for something like BSA archives), that would be pretty disastrous space efficiency wise and have big negative impact on Download finalization time (overhead of extracting the inner data).

Al12rs:
- fixes #2367 

Only recursively extract known archive formats in case of files nested inside mods.
To avoid extracting files that are meant to be left intact.

This might be more of a temporary fix, in case we encounter games/mods that need to keep and deploy `.zip` or other known archive files inside the mods.


erri120:
Can you add some tests for this?

Al12rs:
Added some tests for this

erri120 (approved):
Probably fine for now, but it doesn't feel like the right solution. We might need to come back to this later on.

Al12rs:
Yeah, I'm not satisfied with it either, but should prevent issues in the short term

PR #2440, opened 2025-01-07
Installing optional downloads

### LINKED ISSUE #2432: Change downloaded optional mods into installable items for collections
Allows users to selectively install downloaded optional mods from installed collections.

erri120:
Resolves #2432, part of #2254.

Allows users to install individual optional downloads. Also makes it so that installed required downloads now show up as installed.

### DIFF
--- a/src/NexusMods.App.UI/Pages/NexusModsDataProvider.cs
+++ b/src/NexusMods.App.UI/Pages/NexusModsDataProvider.cs
@@ -127,8 +127,8 @@ public class NexusModsDataProvider : ILibraryDataProvider, ILoadoutDataProvider
                     .TransformImmutable(datom => LibraryLinkedLoadoutItem.Load(_connection.Db, datom.E))
                     .TransformImmutable(item =>
                     {
-                        var parent = item.AsLoadoutItemGroup().AsLoadoutItem().Parent;
-                        return parent.ToCollectionGroup().ToNexusCollectionLoadoutGroup();
+                        if (!LoadoutItem.Parent.TryGetValue(item, out var parentId)) return default(NexusCollectionLoadoutGroup.ReadOnly);
+                        return NexusCollectionLoadoutGroup.Load(item.Db, parentId);
                     })
                     .FilterImmutable(static parent => parent.IsValid())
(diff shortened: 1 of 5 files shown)


src/NexusMods.App.UI/Pages/NexusModsDataProvider.cs
```diff
@@ -56,58 +58,114 @@ public IObservable<IChangeSet<ILibraryItemModel, EntityId>> ObserveCollectionIte
...
+
+        // NOTE(erri120): different behavior depending on whether the download is optional or not.
+        // Optional downloads can be individually installed by the user, Required downloads should only
+        // show that they are installed, but not offer an install button.
+        var observable = nexusModsDownload.AsCollectionDownload().IsOptional ? isInLibraryObservable : isInLibraryObservable.SelectMany(IObservable<bool> (isInLibrary) =>
+        {
+            if (!isInLibrary) return System.Reactive.Linq.Observable.Return(false);
+            if (!CollectionDownloader.TryGetDownloadedItem(nexusModsDownload, _connection.Db, out var libraryItem)) throw new NotImplementedException();
+            return GetIsInstalledObservable(libraryItem.AsLibraryItem(), revisionMetadata);
+        }).DistinctUntilChanged();
```
Al12rs, line 93: If you remove an optional download, it correctly returns ready for install.
If you remove some or all required mods you can't install them again at the moment, as the Install all button doesn't get enabled again.

What should be the expected behavior for required mods?

erri120: Collections are supposed to be read-only, so you shouldn't be able to remove required mods.

Al12rs: What is going to be the expected way to remove an installed collection from a loadout (without deleting it from the library as well)?

erri120: Not sure, I haven't seen designs for that yet.

Al12rs: Worth asking Design then, in any case not a problem for this PR

Al12rs (approved):

PR #2446, opened 2025-01-08
Added: Mod & Pager Thumbnails to Library Page, Collections Page, Loadout Page

### LINKED ISSUE #1974: Add mod page thumbnails to mod page loadout items
## We want:

The mod page image to be displayed on the mod page loadout item rows.

To show the amount of children (mod files) as a count label on the thumbnail.

## Designs:

[Figma link](https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=2260-15881&t=ZTeXIYzZDrF8NLnF-1)

Image

Sewer56:
Styling from @insomnious.
Design signed off by @captainsandypants.
Code from @Sewer56.

Actual PR review should be on the simple side, most of the time was spent figuring out how stuff works for the most part; and making sure design, visual and implementation align, especially with regards to behaviour as there is some very specific behaviour, e.g. don't show mod thumbnails in tree view but show in flat view, etc.


erri120:
I'll review the PR after the merge conflicts have been resolved.

Sewer56:
> Opening the Library page locks up the app for me.

@erri120 Save your DataStore.
I wonder if it's a migration thing, this worked fine on at least 2 machines, but these were mostly clean datastores.

A small amount of the code I had was redone during merging with main, as main changed the code I originally modified; hopefully it's not that; since I tested afterward, but I can also recheck.

erri120:
Opening the library doesn't appear to lock up the app anymore. Not sure what changed.

Al12rs:
> Opening the library doesn't appear to lock up the app anymore. Not sure what changed.

Cached data from previous run? Or di you try from a clean setup?

Sewer56:
I made a callback async, and didn't wait on result of a threadpool call while doing misc cleanup in yesterday's standup, could be related.

erri120:
Are you going to implement the rest of the designs as well?

image
image
image

Sewer56:
Not in this PR, there is still some work around Library in general.

This PR was implemented to the scope I was asked to implement it to, and verified (UI wise) by @captainsandypants

src/NexusMods.App.UI/Pages/LibraryPage/NexusModsModPageLibraryItemModel.cs
```diff
@@ -32,10 +36,29 @@ public NexusModsModPageLibraryItemModel(IObservable<IChangeSet<NexusModsLibraryI
...
+            // SAFETY: This can't have race condition, code is executed on UI, so can't be executed in parallel.
+            var first = LibraryItems.FirstOrDefault();
+            if (!first.IsValid()) return;
+            try
+            {
+                var thumbNail = await modPageThumbnailPipeline.LoadResourceAsync(first.ModPageMetadataId, CancellationToken.None);
+                Thumbnail.Value = thumbNail.Data.Value;
+            }
+            catch (Exception) { /* suppress unhandled error in case lookup fails silently */ }
+        });
```
erri120, line 61: I hate this, we're doing async/await inside a void lambda. You can do something like this instead:

```csharp
OnUI().Select(changeSet => {
  LibraryItems.ApplyChanges(changeSet);
  return LibraryItems.FirstOrDefault();
}).Where(item => item.IsValid())
.SelectMany(async item => /* load */).SubscribeWithErrorLogging();
```

erri120:
image

this also seems weird

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutItemModel.cs
```diff
@@ -47,8 +55,23 @@ public LoadoutItemModel(LoadoutItemId loadoutItemId)
             return ids.Select(id => (Id: id, ShouldEnable: shouldEnable)).ToArray();
         });
 
-        _modelActivationDisposable = WhenModelActivated(this, static (model, disposables) =>
+        _modelActivationDisposable = WhenModelActivated(this, (model, disposables) =>
         {
```
erri120, line 59: Keep the `static` you have access to the class instance and can pass anything else as state into the lambda.

Sewer56:
We're aware, I was just told to leave it for later; the push for the alpha is going quite hard.

src/NexusMods.App.UI/Pages/NexusModsDataProvider.cs
```diff
@@ -372,6 +374,31 @@ public IObservable<IChangeSet<LoadoutItemModel, EntityId>> ObserveNestedLoadoutI
...
+                            ImagePipelines.CreateObservable(nexusLibraryItem.ModPageMetadataId, modPageThumbnailPipeline)
+                                .Take(1)
+                                .ObserveOnUIThreadDispatcher()
+                                .Subscribe(this, (bitmap, _) => model.Thumbnail.Value = bitmap);
+                        }
+                        
+                        return;
+                    }
+                });
+
```
erri120, line 401: This is a complete mess. Just go to `FakeParentLoadoutItemModel` and subscribe to changes there.

erri120 (changes requested):

src/NexusMods.App.UI/ImagePipelines.cs
```diff
@@ -182,6 +210,43 @@ private static IResourceLoader<EntityId, Bitmap> CreateCollectionBackgroundImage
...
+                partitionId: PartitionId.User(ImagePartitionId)
+            )
+            .Decode(decoderType: DecoderType.Qoi)
+            .ToAvaloniaBitmap()
+            .UseFallbackValue(ModPageThumbnailFallback)
+            .UseScopedCache(
+                keyGenerator: static tuple => tuple.Item1,
+                keyComparer: EqualityComparer<EntityId>.Default,
+                capacityPartition: new FavorWarmPartition(totalCapacity: 333, warmRatio: FavorWarmPartition.DefaultWarmRatio)
+            )
```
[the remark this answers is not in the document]
Sewer56: I don't think it's a big difference.

I did it because the example code at the time had it set up this way.

I just stuck with it because I figured it might reduce lag when using `Switch View` (all images are reloaded because the tree & non-tree rows use a different VM); loading thumbs could cause lag for non-SSD users; though I suppose there's quite a few of these nowadays.

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutItemModel.cs
Sewer56: The `state` parameter is constrained by `where T : IReactiveR3Object`.

In order to pass the variables via state, I'd need to persist the remaining constructor fields to the class; however those fields are only used during initialisation. 

It can be solved by wrapping `this` in a struct, and delegating the whole interface implementation to what's under the hood, but then I'd be getting complaints about unnecessary boilerplate in the code. Or do a temporary field but that would be a waste of memory. Can't really win here.

erri120: You can just use `WhenActivated` instead of `WhenModelActivated`.

Sewer56: Alright, done, I actually didn't spot the other overload earlier, despite it being right below the other method.
Works for me, and is preferable.

erri120 (changes requested):
The changes are backwards. Nexus Mods mod pages are the only items that actually have images yet the code works backwards and assigns parents images based on the children, even though the children don't have the images.

erri120 (approved):

PR #2447, opened 2025-01-08
Fixed: Token Expiry Date Not using UTC Time

### LINKED ISSUE #2442: Users in time zones 6+ hours ahead cannot use the app
# Bug Report

## Summary
After a lengthy investigation with "saffy" (aka saffy on Nexus Mods) on Discord, we have discovered they were unable to use the collections feature as their OAuth token was always showing as expired and would constantly try to refresh.

We troubleshooted the issue for several hours and eventually, @Sewer56 discovered that with the user being in a UTC +11 timezone, the expiry time of the OAuth tokens is being calculated incorrectly. This issue will affect any user who is 6 or more hours ahead of the UK due to the token expiry time.

## Steps to reproduce
1. Set your PC timezone to UTC+11.
2. Log into the app.
3. Attempt to download a collection.
4. Check the logs.

The correct time is used when calculating if the user's token has expired.

Sewer56:
fixes #2442

I tested this on both Windows and Linux; should be fine.

Normally something like this you'd unit test with a DateTimeProvider, and mock that; however, time for test build is now, not later.

### DIFF
--- a/src/Networking/NexusMods.Networking.NexusWebApi/Auth/JWTToken.cs
+++ b/src/Networking/NexusMods.Networking.NexusWebApi/Auth/JWTToken.cs
@@ -63,7 +63,7 @@ public partial class JWTToken : IModelDefinition
 
         tx.Add(entityId, AccessToken, reply.AccessToken);
         tx.Add(entityId, RefreshToken, reply.RefreshToken);
-        tx.Add(entityId, ExpiresAt, DateTimeOffset.FromUnixTimeSeconds(reply.CreatedAt).DateTime + TimeSpan.FromSeconds(reply.ExpiresIn));
+        tx.Add(entityId, ExpiresAt, DateTimeOffset.FromUnixTimeSeconds(reply.CreatedAt).UtcDateTime + TimeSpan.FromSeconds(reply.ExpiresIn));
 
         return entityId;
     }
--- a/src/Networking/NexusMods.Networking.NexusWebApi/Auth/JWTToken.cs
+++ b/src/Networking/NexusMods.Networking.NexusWebApi/Auth/JWTToken.cs
@@ -63,7 +63,7 @@ public partial class JWTToken : IModelDefinition
 
         tx.Add(entityId, AccessToken, reply.AccessToken);
         tx.Add(entityId, RefreshToken, reply.RefreshToken);
-        tx.Add(entityId, ExpiresAt, DateTimeOffset.FromUnixTimeSeconds(reply.CreatedAt).UtcDateTime + TimeSpan.FromSeconds(reply.ExpiresIn));
+        tx.Add(entityId, ExpiresAt, DateTimeOffset.FromUnixTimeSeconds(reply.CreatedAt) + TimeSpan.FromSeconds(reply.ExpiresIn));
 
         return entityId;
     }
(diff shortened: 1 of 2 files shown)


src/Networking/NexusMods.Networking.NexusWebApi/Auth/JWTToken.cs
```diff
@@ -63,7 +63,7 @@ public static Optional<EntityId> Create(IDb db, ITransaction tx, JwtTokenReply r
 
         tx.Add(entityId, AccessToken, reply.AccessToken);
         tx.Add(entityId, RefreshToken, reply.RefreshToken);
-        tx.Add(entityId, ExpiresAt, DateTimeOffset.FromUnixTimeSeconds(reply.CreatedAt).DateTime + TimeSpan.FromSeconds(reply.ExpiresIn));
+        tx.Add(entityId, ExpiresAt, DateTimeOffset.FromUnixTimeSeconds(reply.CreatedAt).UtcDateTime + TimeSpan.FromSeconds(reply.ExpiresIn));
```
Al12rs, line 66: What is the reason for converting to a DateTime? Shouldn't we just keep using the DateTimeOffset directly?
```suggestion
        tx.Add(entityId, ExpiresAt, DateTimeOffset.FromUnixTimeSeconds(reply.CreatedAt)+ TimeSpan.FromSeconds(reply.ExpiresIn));
```

erri120, line 66: The `DateTime` conversion was a left-over from when I updated the attributes to use `DateTimeOffset`.

```suggestion
        tx.Add(entityId, ExpiresAt, DateTimeOffset.FromUnixTimeSeconds(reply.CreatedAt) + TimeSpan.FromSeconds(reply.ExpiresIn));
```

erri120 (changes requested):

Sewer56:
Gimme a moment to reboot twice again 😅

Sewer56:
Okay, I retested, should be good.

erri120 (approved):

Sewer56:
I do find it a bit unusual that we store the user's timezone information with each timestamp in the App (DB) though; you wouldn't normally expect that. I'm used to normalizing all timestamps as DateTime with UTC, haha.

erri120:
> I do find it a bit unusual that we store the user's timezone information with each timestamp in the App (DB) though; you wouldn't normally expect that. I'm used to normalizing all timestamps as DateTime with UTC, haha.

We store the `DateTimeOffset` as unix milliseconds in the DB:

https://github.com/Nexus-Mods/NexusMods.MnemonicDB/blob/main/src/NexusMods.MnemonicDB.Abstractions/Attributes/TimestampAttribute.cs

Sewer56:
Ya, I get that. I specifically put 'DB' in brackets because that's the part that enforces this contract.
DB stores the value correctly as `Int64` (minor precision loss aside).

What I moreso meant is the fact we're using the `DateTimeOffset` type in the App as if it was a `UtcDateTime` type.
Rather than storing just a normalized timestamp, we store a normalized timestamp + offset to user's timezone, in each timestamp.

All timestamps are generally made in the same timezone, so that offset is redundant in each case.

PR #2449, opened 2025-01-08
Design tweaks for Health Check

insomnious:
Left aligned, fixed width, and up to date design of the details view.

This is good enough for now to get us up to date for next release. It can and will be tidied\refactored more when we do the next pass on it after Stardew. One of the things to be refactored, based on usage, will be translucent severity colors which we've not used anywhere else yet.


src/NexusMods.App.UI/Pages/Diagnostics/Details/DiagnosticDetailsView.axaml.cs
erri120, line 1: For the explanation text use the descriptions we already have: https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/Abstractions/NexusMods.Abstractions.Games.Diagnostics/DiagnosticSeverity.cs

erri120 (approved):

Issue #2452, opened 2025-01-08
SMAPI is not installing correctly


Pickysaurus:
# Bug Report

## Summary
Built from source on 20ac0a87e009555d5cf75f56ca36c492e2da6288

SMAPI no longer installs correctly

## Steps to reproduce
1. Build from master on 20ac0a87e009555d5cf75f56ca36c492e2da6288
2. Install SMAPI
3. See the fallback installer

## What is the expected behaviour?
SMAPI installs

## Other information
_(Logs, screenshots, possible fixes, expected behaviour, etc.)_

Log files can be found in these locations:

- Windows: `%LOCALAPPDATA%\NexusMods.App\Logs`
- Linux: `~/.local/state/NexusMods.App/Logs`

Please attach both `nexusmods.app.main` and `nexusmods.app.slim` log files with the most recent dates.

PR #2453, opened 2025-01-08
Fix SMAPI install

Al12rs:
Fixes https://github.com/Nexus-Mods/NexusMods.App/issues/2452


halgari (approved):

erri120:
It would be good to know why our existing tests didn't catch SMAPI being broken.

Al12rs:
> It would be good to know why our existing tests didn't catch SMAPI being broken.

I don't think we have a test installing SMAPI actually. If there was one maybe it got lost during one of the purges?

Issue #2433, opened 2025-01-06
Left Menu rework


Al12rs:
Redo left menu with following enhancements:

- [x] ~Consider using `Menu` control with `MenuItems` for the UI implementation~ Decided against, as there were no real benefits. MenuItems are assumed to be placed inside a menu, which is meant to be a something like the right click menu to top bar one, not a navigation bar with custom requirements, interactions, behaviors and styling. Simple buttons are simpler and more flexible.
- Go through designs, list all the types of buttons we need (selection, badges, dropdowns etc):
    - [x] Selection enabled menu items
    - [x] Collection toggles
    - [x] Installed mod count in left menu
    - [ ] Health Check specific badges
    - [ ] Library notification badges
    - [ ] Collapsible sections

- [x] Update Left menu structure
    - [ ] Update icons

insomnious:
### New Icons

My Loadouts `IconValues.Package`
My Games `IconValues.GamepadOutline` 
Library `IconValues.LibraryOutline` 
Health Check `IconValues.Cardiology`

PR #2455, opened 2025-01-08
WIP Left menu button styles

insomnious:
Styles\layout for the various Left Hand Nav button variations. Just text, text with Badges, text with Icon, text with Toggle.

I've still more tweaks to do for ToggleSwitch but the structures are there so Al can see and continue with functionality. 

Part of #2433


insomnious:
no longer needed

PR #2456, opened 2025-01-08
Basic exception overlay

halgari:
This PR makes the following changes: 

* Updates the OK Message box to accept an optional markdown renderer that can be used for "supporting documentation"
* Moved around a few of the parameters in the existing uses of this overlay to allow for the markdown VM to be pulled in.
* Adds a new logging target in `NexusMods.App/Program.cs` that logs to a implementation of `NexusMods.Abstractions.Logging.IObservableExceptionSource`
* MainWindowViewModel wires the IObservableExceptionSource up to the Message Box Ok dialog when exceptions come through. 

No exception spamming on my install so far. A few that we likely need to look at, like observable `null reference exceptions` and a json parse exception on loading the window settings. Overall I think this is good for pointing out where we have otherwise silent errors happening in the app.


Al12rs (changes requested):
I tried triggering it (I threw the exception) and the exception details text doesn't wrap or scroll, and can't be read or copied.
image

Al12rs:
Close and reopen to re trigger PR builds

erri120:
I still don't believe this is the right approach to error handling, but if you want to go down this route, we should at least have a toggle in the settings to disable this dialog.

PR #2457, opened 2025-01-09
Cli fixes and escape hatches

halgari:
This PR does the following

* Fixes consoles on Windows
  * When compiling with WinExe the console isn't connected
  * When compiling with Exe we always get a console, even when running in UI mode
  * This is a windows issue. Fixed by compiling with WinExe mode and doing some P/Invoke magic to attach any existing consoles
* Changes up the Proxy Console code a bit so we can create paths of modules for verbs instead of prefixing verbs. For example instead of `nexus-login` we can have `nexus login`, `nexus verify` etc. This is opt-in and we can convert verbs over as time goes on. 
* Modules (verb prefixes) can have doc strings so we can do `nexus --help` to find out what console commands are available in that module
* Verb options can now be optional
* Verb options that are not optional are now *always* required
* Reorganized/Cleaned up the loadout management verbs
* Added some new render helpers, so we can do code like `(IEnumerable<(string, int)>).RenderTable(renderer, "Name", "Count");` making a lot of the render code easier to read
* Added some exception catching to the verb dispatch code
* Added a way to delete files based on a file glob from specific mods via the CLI. So now we have an escape hatch for files that get stuck in `Overrides`.

### DIFF
--- a/src/NexusMods.App/ConsoleHelper.cs
+++ b/src/NexusMods.App/ConsoleHelper.cs
@@ -1,11 +1,13 @@
 using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
 
 namespace NexusMods.App;
 
 /// <summary>
 /// Helpers for consoles on Windows
 /// </summary>
-public class ConsoleHelper
+[SupportedOSPlatform("windows")]
+public static class ConsoleHelper
 {
     // ReSharper disable once InconsistentNaming
     private const int ATTACH_PARENT_PROCESS = -1;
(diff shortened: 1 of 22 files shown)


Al12rs:
Just checked the failing tests, it seems they need to be updated to use the new CLI incantations

src/NexusMods.App/ConsoleHelper.cs
```diff
@@ -0,0 +1,30 @@
+using System.Runtime.InteropServices;
+
+namespace NexusMods.App;
+
+/// <summary>
+/// Helpers for consoles on Windows
+/// </summary>
+public class ConsoleHelper
```
erri120, line 8:
```suggestion
[SupportedOSPlatform("windows")
public static class ConsoleHelper
```

Issue #1676, opened 2024-06-24
Consider using XDG Desktop Portal for Linux OS Interop


erri120:
- [Docs](https://flatpak.github.io/xdg-desktop-portal/) and [API docs](https://flatpak.github.io/xdg-desktop-portal/docs/)
- [Wiki](https://wiki.archlinux.org/title/XDG_Desktop_Portal)
- https://github.com/erri120/LinuxDesktopUtils

Benefits:
- Flatpak friendly (we don't have a Flatpak yet, but if we want to have one, we need portals).
- Uniform access to desktop features independent of desktops and toolkits:
  - [OpenURI](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.OpenURI.html) for opening links, files, other in the program of choice instead of relying on `xdg-open`.
  - [File Chooser](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html) for picking a file (this is already implemented by [Avalonia](https://github.com/AvaloniaUI/Avalonia/tree/master/src/Avalonia.FreeDesktop)).
  - [Notifications](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Notification.html).
  - [Secrets](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Secret.html) for JWT tokens and other secrets. (although we'd probably wanna use [libsecret](https://gnome.pages.gitlab.gnome.org/libsecret/)).
  - [Trash](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Trash.html) for deleting files by putting them into a "trash" folder.

Downsides:
- D-Bus.
- APIs that spawn dialogs kinda require a [window handle](https://flatpak.github.io/xdg-desktop-portal/docs/window-identifiers.html). It's possible to pass an empty string instead, but I don't know how great of an experience that is.

LukeNexusMods:
Will help with installing app on steamdeck. Linux is flatpak, steamdeck users use them too. 

Flatpaks help with compatibility.

erri120:
I've been cooking up a library that implements XDG related functionality: https://github.com/erri120/LinuxDesktopUtils

MattSturgeon:
[...]

Issue #2436, opened 2025-01-06
Flatpak release


erri120:
Docs: https://docs.flatpak.org/en/latest/building.html

Needs investigating:
- Use of desktop portals (#1676)
- Flatpak permissions
- Pupnet support for publishing Flatpaks
- Validation of [`com.nexusmods.app.metainfo.xml`](https://github.com/Nexus-Mods/NexusMods.App/blob/main/src/NexusMods.App/com.nexusmods.app.metainfo.xml)
- Ownership of `com.nexusmods.app` ID on Flathub

erri120:
Example `com.nexusmods.app.yaml` build file:

```yaml
id: com.nexusmods.app
runtime: org.freedesktop.Platform
runtime-version: '24.08'
sdk: org.freedesktop.Sdk
sdk-extensions:
  - org.freedesktop.Sdk.Extension.dotnet9
build-options:
  prepend-path: "/usr/lib/sdk/dotnet9/bin"
  append-ld-library-path: "/usr/lib/sdk/dotnet9/lib"
  prepend-pkg-config-path: "/usr/lib/sdk/dotnet9/lib/pkgconfig"

command: "NexusMods.App"

finish-args:
  - --device=dri          # rendering
  - --device=input        # inputs, specifically controllers
  - --socket=x11          # windowing
  - --share=ipc           # for X11 shared memory extension
  - --share=network       # network
  - --socket=session-bus  # D-Bus
  - --env=DOTNET_ROOT=/app/lib/dotnet

modules:
  - name: dotnet
    buildsystem: simple
    build-commands:
    - /usr/lib/sdk/dotnet9/bin/install.sh

  - name: app
    buildsystem: simple
    build-options:
      env:
        DOTNET_CLI_TELEMETRY_OPTOUT: 'true'
        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 'true'
    sources:
      - type: git
        url: "https://github.com/Nexus-Mods/NexusMods.App.git"
        tag: "main"
      - ./nuget-sources.json
    build-commands:
      - dotnet publish -r linux-x64 src/NexusMods.App/NexusMods.App.csproj --no-self-contained --source ./nuget-sources
      - mkdir -p ${FLATPAK_DEST}/bin
      - cp -r src/NexusMods.App/bin/Release/net9.0/linux-x64/publish/* ${FLATPAK_DEST}/bin
```

[...]

PR #2461, opened 2025-01-09
Use XDG Desktop Portal to open URIs and files

erri120:
Resolves https://github.com/Nexus-Mods/NexusMods.App/issues/1676 and related to https://github.com/Nexus-Mods/NexusMods.App/issues/2436.


Sewer56 (approved):
Seems fine to me. Only reviewed App code, not your NuGet library. (ps. BATMAN)

When I rebooted and tested on Windows, `/select` was broken, but it seems you already caught on to that.

As for #1676, I think it's probably ok to close for now. For Flatpak we'll need host filesystem access most likely to scan for games and write to game folders; but outside of that, we've got it mostly covered.

Al12rs (approved):
Didn't test on Linux, but looks good

PR #2465, opened 2025-01-13
Move collection cards into library

### LINKED ISSUE #2464: Move collection cards into the Library
Part of #2254.

erri120:
Resolves #2464.

Apparently this is the first expander in the app, which is why it doesn't have any styles yet.

image


Al12rs (approved):
View is a bit jank without the extra styling.
Might want to start with the expander closed or an empty state if there are no collections.
But can be handled with a styling PR I think

PR #2467, opened 2025-01-13
Composite Item Models

erri120:
Keeping the good stuff from #2143 and reworking the annoying parts.

With the old system, it was annoying and difficult to add new shapes or change existing shapes of data. You'd have to create or update column definitions, update all the different item types and implement a lot of convoluted code.

The new system massively improves on this aspect. Instead of creating a custom item model type for different data types, there is now a single type: `CompositeItemModel`. Using composition over inheritance, you can simply create this item model and add components to it.

Each component holds a piece of data and is identified by a key:

```csharp
var itemModel = new CompositeItemModel<EntityId>(x.Id);

itemModel.Add(SharedColumns.Name.StringComponentKey, new StringComponent(value: x.Name));

itemModel.Add(LoadoutColumns.IsEnabled.ComponentKey, new LoadoutComponents.IsEnabled(
    initialValue: x.InitialIsEnabled,
    valueObservable: x.IsEnabled
));
```

Columns aren't interfaces anymore but empty classes that only serve as definitions:

```csharp
public sealed class Name : ICompositeColumnDefinition<Name>
{
    public static int Compare<TKey>(CompositeItemModel<TKey> a, CompositeItemModel<TKey> b) where TKey : notnull
    {
        var aValue = a.GetOptional<StringComponent>(StringComponentKey);
        var bValue = b.GetOptional<StringComponent>(StringComponentKey);
        return aValue.Compare(bValue);
    }

    public const string ColumnTemplateResourceKey = Prefix + "Name";
    public static readonly ComponentKey StringComponentKey = ComponentKey.From(Prefix + nameof(Name) + nameof(StringComponent));

    public static string GetColumnHeader() => "Name";
    public static string GetColumnTemplateResourceKey() => ColumnTemplateResourceKey;
}
```

The thing that stayed pretty similar is the `DataTemplate`, but with some custom behavior to work with components. Each column now chooses which components it uses to display things. With this approach a column can have different representation for different shapes of data:

```xaml
<DataTemplate x:Key="{x:Static controls:SharedColumns+Name.ColumnTemplateResourceKey}">
    <DataTemplate.DataType>
        <x:Type TypeName="controls:CompositeItemModel" x:TypeArguments="abstractions:EntityId"/>
    </DataTemplate.DataType>

    <controls:ColumnContentControl x:TypeArguments="abstractions:EntityId" Content="{CompiledBinding}">
        <controls:ColumnContentControl.AvailableTemplates>
            <controls:ComponentTemplate x:TypeArguments="controls:StringComponent" ComponentKey="{x:Static controls:SharedColumns+Name.StringComponentKey}">
                <controls:ComponentTemplate.DataTemplate>
                    <DataTemplate DataType="{x:Type controls:StringComponent}">
                        <TextBlock Text="{CompiledBinding Value.Value}" />
                    </DataTemplate>
                </controls:ComponentTemplate.DataTemplate>
            </controls:ComponentTemplate>
        </controls:ColumnContentControl.AvailableTemplates>

        <controls:ColumnContentControl.Fallback>
            <TextBlock Text="-"/>
        </controls:ColumnContentControl.Fallback>
    </controls:ColumnContentControl>
</DataTemplate>
```

XAML limitations are holding this back a bit, I might be looking at that in a different PR.


src/NexusMods.App.UI/Tickers.cs
```diff
@@ -0,0 +1,18 @@
+using R3;
+
+namespace NexusMods.App.UI;
+
+public static class Tickers
+{
+    private static readonly TimeSpan Period = TimeSpan.FromSeconds(30);
+    public static Observable<DateTimeOffset> Primary { get; }
```
Al12rs, line 8: I would name this based on the period amount, or something vaguely descriptive, like `SlowPoll`

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutView.axaml.cs
```diff
@@ -14,6 +14,7 @@ public LoadoutView()
     {
         InitializeComponent();
 
+        // TreeDataGridViewHelper.SetupTreeDataGridAdapter<LoadoutView, ILoadoutViewModel, CompositeItemModel<EntityId>, EntityId>(this, TreeDataGrid, vm => vm.Adapter);
```
Al12rs, line 17: Do you still need this?

src/NexusMods.App.UI/Controls/TreeDataGrid/CompositeItemModel.cs
```diff
@@ -0,0 +1,314 @@
...
+                var (_, disposables) = tuple;
+                if (change.Value is not IReactiveR3Object reactiveR3Object) return;
+                reactiveR3Object.Activate().AddTo(disposables);
+            }).AddTo(disposables);
+
+            self._components.ObserveDictionaryRemove().ObserveOnUIThreadDispatcher().Subscribe((self, disposables), static (change, tuple) =>
+            {
+                var (_, disposables) = tuple;
+                if (change.Value is not IReactiveR3Object reactiveR3Object) return;
+                var didRemove = disposables.Remove(reactiveR3Object);
```
Al12rs, line 61: Who is supposed to handle the disposal of the removed component? Whoever issued the removal?

erri120: This `Remove` disposes the object. I can put a comment on there as it's implementation detail.

Al12rs (approved):
I went through all the code, and it makes sense to me.
We already discussed the Key thing and I think what is proposed here works for now.
The Rider bug is unfortunate and annoying, but for now the workaround will do.

There is some code in here that is just for placeholder for previewing, so that will need to be cleaned up once this starts getting integrated into views.

Only thing I would consider doing for now is adding more comments about what is meant to stay and what is temporary, so that if we only come back to this in a few months we can more easily pick it up, but it is fine either way.

PR #2475, opened 2025-01-14
Text change to Install Mods title

insomnious:
Small change to text


Al12rs:
I'll fix this on my end with Translations

PR #2477, opened 2025-01-14
Add semaphore to prevent apply/ingest of multiple loadouts at once.

### LINKED ISSUE #1196: Apply/Ingest - Prevent multiple Apply/Ingest from all loadouts of same game install while already Applying/Ingesting
## Summary:
While an Apply or Ingest operation is underway, the game install should be locked to prevent any other Apply/Ingest operation from being executed. 
From both the current loadout, and all other loadouts that point to the same Game Install.

Starting multiple Apply/Ingest operations on the same game installation would be quite disastrous and should be avoided.

`ApplyService` should serve as the locking manager

axmanalad:
Fixes #1196

This PR blocks CLI command `loadout synchronize -l [loadout_id]` from processing when there is a loadout that is currently being applied. I believe me and @Michael-Kowata could implement more prevention to applying/ingesting if we understand how `ApplyService` works. More clarification about "ApplyService" as the locking manager is appreciated.


src/NexusMods.DataModel/CommandLine/Verbs/LoadoutManagementVerbs.cs
```diff
@@ -43,6 +43,8 @@ private static async Task<int> Synchronize([Injected] IRenderer renderer,
         [Option("l", "loadout", "Loadout to apply")] Loadout.ReadOnly loadout,
         [Injected] ISynchronizerService syncService)
     {
+        if (await syncService.IsAnyLoadoutSynchronizing())
+            return await renderer.InputError("Another loadout is already being synchronized.");
         await syncService.Synchronize(loadout);
```
Al12rs, line 48: Hi there, thanks for contributing!

With the approach here, there is still a potential concurrency issue, if Synchronize operation is started in between `IsAnyLoadoutSynchronizing()` returning and the `Synchronize()` call.

A solution could be to have a `bool TrySynchronyze()` that tests and runs atomically or returns false if busy.
Or, the async Synchronize operation is queued if another is running, either by awaiting on a `SlimSemaphore` or by implementing an actual queue.

Being able to queue Synchronize operations is actually something we would like to have on the future, but might be out of scope for this pr.

Al12rs (changes requested):

Al12rs:
It's already late for me, but I can give a rundown of the `SynchronizeService` (previously called `ApplyService`) on discord tomorrow morning

src/NexusMods.DataModel/CommandLine/Verbs/LoadoutManagementVerbs.cs
axmanalad: > With the approach here, there is still a potential concurrency issue, if Synchronize operation is started in between `IsAnyLoadoutSynchronizing()` returning and the `Synchronize()` call.
> 
> A solution could be to have a `bool TrySynchronyze()` that tests and runs atomically or returns false if busy. Or, the async Synchronize operation is queued if another is running, either by awaiting on a `SlimSemaphore` or by implementing an actual queue.

Thank you for responding with useful feedback about `SynchronizerService` and `bool TrySynchronize()` in Discord; it definitely clarified a lot of confusion about the issue stated.

> Being able to queue Synchronize operations is actually something we would like to have on the future, but might be out of scope for this pr.

Both of us saw how `SemaphoreSlim` operated like a locking mechanism and felt that it was reasonable enough to apply an addition of queues when calling the synchronize command through CLI. The reason I implemented queues was an intention to place messages in CLI regarding how many `Synchronize()` operations are left before it started to apply from awaiting. In that regard of the CLI messages, I would suggest to place a text-based queue indication into the front-end in the future.

axmanalad:
> Being able to queue Synchronize operations is actually something we would like to have on the future, but might be out of scope for this pr.

Both of us saw how `SemaphoreSlim` operated like a locking mechanism and felt that it was reasonable enough to apply an addition of queues when calling the synchronize command through CLI. The reason I implemented queues was an intention to place messages in CLI regarding how many `Synchronize()` operations are left before it started to apply from awaiting. In that regard of the CLI messages, I would suggest to place a text-based queue indication into the front-end in the future.

src/NexusMods.DataModel/CommandLine/Verbs/LoadoutManagementVerbs.cs
```diff
@@ -43,8 +43,24 @@ private static async Task<int> Synchronize([Injected] IRenderer renderer,
...
+        {
+            var queueCount = syncService.GetQueueCount();
+            var position = syncService.GetQueuePosition(loadout.LoadoutId);
+            var remainingQueueCount = queueCount - position;
+            var currentQueuePosition = position - queueCount;
+            if (remainingQueueCount == -1)
+                return;
+            var positionText = currentQueuePosition == 0 ? "Syncing..." : "Waiting in queue...";
+            await renderer.Text("Queued Synchronize operations left: {0}\t\t{1}\n", remainingQueueCount, positionText);
+        }
```
Al12rs, line 63: We really don't want to have this kind of event based callback system, it makes for uglier code to read/maintain and has other downsides like management of event subscriptions. 

The queue system implemented here doesn't really do what we would like it to do later on, which is to deduplicate multiple consecutive Synchronize requests for the same game installation.

Ideally the queue would be completely transparent to the caller, which would only have to await a task.
The requirements for that system haven't been formalized and I think it would result in a complicated system that is outside the scope of the original ticket, of preventing multiple concurrent synchronize operations.

The code for locking with the semaphore seems Ok here, so I would remove the Queue stuff from this pr and just do the mutual exclusion part.

If you are interested on working on a queue system we should formalize the requirements for that a bit more before proceeding, maybe we can chat on discord about it and put it in a ticket.

axmanalad:
Hello AL, sorry for one week of delay. Our schools had recently resumed which had shifted our time for a bit.

> The code for locking with the semaphore seems Ok here, so I would remove the Queue stuff from this pr and just do the mutual exclusion part.

We removed the queue implementation as you had requested.

> If you are interested on working on a queue system we should formalize the requirements for that a bit more before proceeding, maybe we can chat on discord about it and put it in a ticket.

Thank you for elaborating on the queue system. We are interested to making a queue system as seen from the previous commit and understand that there is not really a design implemented yet to plan out how the queue system goes. For now, we can wait for the requirements of this system to be formalized then.

Al12rs (commented):
Nice, thanks! 
I think our code isn't ready yet for the queue stuff yet, something we will look at later on.

In the meantime, one improvement we could do would be to only limit concurrent Synchronize operations, if they are performed on the same game installation path.

This is a complication of the code, since rather than having a simple semaphore at the beginning of the Synchronize method, we would now need some kind of logic to determine whether there is a need to wait or not depending on what other operations are in effect. 
A pitfall in these scenarios is to check whether the way is clear and then acquire the lock as two separate operations, rather than doing so atomically, which opens up the possibility of another thread acquiring the lock in the meantime and placing the system in  an inconsistent state.
If you guys are interested we can chat more on discord or I can make a ticket where the details are set out.

Al12rs (approved):

PR #2479, opened 2025-01-15
Added: Fuzzy Search/File Map for Detecting File Updates

Sewer56:
This is a very small PR; just the stuff I added yesterday morning & afternoon.

This implements a 'fuzzy search' strategy for mapping that aren't the `file_updates` field in the V1 API; and updates all of the related documentation around updates, 

This should be sufficient to make it work on 40-70% of mod pages (random guess), in the absence of proper links.

There are still some improvements that can be made, but for now I've decided to keep it relatively simple; i.e. covering most possible cases in minimum amount of work. Just enough to make it work on hopefully over half the mods; for a working prototype.


halgari:
I thought we were just going with whatever file was newer, until we got the rest of the UI done for this.

Al12rs (approved):
A lot of this code will have to be thrown out when we get to actually writing an accurate update check, and tests will need to be updated accordingly. 
That is unfortunate, but I don't think there is a point in delaying the PR to removing it now or something as opposed to later on.

PR #2484, opened 2025-01-15
Fixes lack of Library scrolling

insomnious:
Added ScrollViewer for whole page as per @captainsandypants request. Added bottom border to expanders to visually separate when both are collapsed.

### DIFF
--- a/src/NexusMods.App.UI/Pages/LibraryPage/LibraryView.axaml
+++ b/src/NexusMods.App.UI/Pages/LibraryPage/LibraryView.axaml
@@ -14,141 +14,146 @@
     mc:Ignorable="d" d:DesignWidth="1000" d:DesignHeight="450"
     x:Class="NexusMods.App.UI.Pages.LibraryPage.LibraryView">
 
-    <Grid RowDefinitions="Auto, Auto" ColumnDefinitions="*">
-        <Expander x:Name="ExpanderCollections" Grid.Row="0" IsExpanded="True" ExpandDirection="Down" HorizontalAlignment="Stretch">
-            <Expander.Header>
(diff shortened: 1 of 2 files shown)


Al12rs:
This makes it so the entire page contents are scrolled rather than just the contents of the TreeDataGrid:
image


While in the loadout the headers of the treedatagrid stay in view:
image

Sewer56:
I noted that internally yesterday, there's a discussion thread about it.
@insomnious 's going to discuss more with @captainsandypants later

Al12rs:
To make it work with just the Tree scrolling and keeping the header in view (you have to collapse the collections section to make room to actually see of course), you can remove the `ScrollViewer` at the top, replace the top `StackPanel` with a `<Grid RowDefinitions="Auto, *">`, and that should just work

Sewer56:
Oh, that actually worked for you?

It was the literal first thing I tried yesterday, and I didn't have as good luck; maybe I missed something in there given I've only poked around for 15 minutes.

insomnious:
I also tried that first and it doesn't work in all eventualities. We don't need 3 people on this so don't worry about it for now as I'm going through with Laurence some different options to get us through to the milestone. As was discussed on Slack.

Sewer56:
@insomnious 

Please delete the above comment and remove the name of the chat application before reposting. If you don't, @ erri120 will, as they usually do.

insomnious:
Removed page-wide vertical scroll
Moved EmptyState for mods so it stays within the expander but doesn't show the toolbar.
Fixed height on Collections expander content
Signed off by @captainsandypants

Al12rs (approved):
TreeDataGrid seems to be virtualizing the items correctly

PR #2502, opened 2025-01-20
Re-order delete operations

### LINKED ISSUE #2496: Deleting a collection from the download page should redirect back to the library
# Improvement

## Currently 
If a Collection is deleted from the Library from the "Collection Download" page, the user remains on the page for the deleted content and can still interact with it. 
## We want
Redirect the user back to the Library (can we do anything about the page history for the back/forward buttons). 

## Supporting Information

Image

## Design

erri120:
Fixes #2496.

Collection will now be deleted after we switch the page.


Pickysaurus:
> I guess this will result in reports that the collection doesn't get deleted after deletion if it still takes a while to delete after

If this is the case, we might want to consider a "Deleting" spinner or something for the Library page.

PR #2505, opened 2025-01-20
Add custom collection loadout page

erri120:
Part of #2254.

### DIFF
--- a/src/NexusMods.App.UI/Pages/LoadoutPage/ICollectionLoadoutViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/LoadoutPage/ICollectionLoadoutViewModel.cs
@@ -9,6 +9,10 @@ public interface ICollectionLoadoutViewModel : IPageViewModelInterface
 {
     LoadoutTreeDataGridAdapter Adapter { get; }
 
+    bool IsLocalCollection { get; }
+
+    bool IsReadOnly { get; }
+
     bool IsCollectionEnabled { get; }
 
     string Name { get; }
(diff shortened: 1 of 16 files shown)


src/NexusMods.App.UI/Pages/LoadoutPage/ICollectionLoadoutViewModel.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace NexusMods.App.UI.Pages.LoadoutPage;
+
+public interface ICollectionLoadoutViewModel : IPageViewModelInterface
+{
+    LoadoutTreeDataGridAdapter Adapter { get; }
+
+    bool IsLocalCollection { get; }
+
+    bool IsReadOnly { get; }
```
Al12rs, line 14: Could you add some comments on these? To explain what the definition is meant to be?

src/NexusMods.App.UI/Pages/LoadoutPage/ICollectionLoadoutViewModel.cs
```diff
@@ -0,0 +1,40 @@
...
+using R3;
+
+namespace NexusMods.App.UI.Pages.LoadoutPage;
+
+public interface ICollectionLoadoutViewModel : IPageViewModelInterface
+{
+    LoadoutTreeDataGridAdapter Adapter { get; }
+
+    /// <summary>
+    /// Gets whether the collection is local or remote.
```
Al12rs, line 13: So for now there is only the Default collection that is local?
Would a writable copy of a remote collection be local?

Is there any case where `IsReadOnly` and `IsLocalCollection` wouldn't match?

erri120: I don't know yet, we're not there yet.

Al12rs: Fair answer

Al12rs (approved):

PR #2506, opened 2025-01-20
Show mod count in left menu

Al12rs:
- part of #2433 

## Details:
- Allow passing a string observable for the text label of LeftMenuItems
- Show num mods next to Loadout entry in left menu

### DIFF
--- a/src/NexusMods.App.UI/LeftMenu/Items/ILeftMenuItemViewModel.cs
+++ b/src/NexusMods.App.UI/LeftMenu/Items/ILeftMenuItemViewModel.cs
@@ -1,5 +1,6 @@
 using System.Reactive;
 using NexusMods.Abstractions.UI;
+using NexusMods.App.UI.Controls;
 using NexusMods.App.UI.Controls.Navigation;
 using NexusMods.Icons;
 using ReactiveUI;
@@ -8,7 +9,7 @@ namespace NexusMods.App.UI.LeftMenu.Items;
 
 public interface ILeftMenuItemViewModel : IViewModelInterface 
 {
-    public string Text { get; set; }
+    public StringComponent Text { get; set; }
     
     public IconValue Icon { get; set; }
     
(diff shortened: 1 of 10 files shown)


src/NexusMods.App.UI/LeftMenu/Items/LeftMenuItemViewModel.cs
```diff
@@ -14,7 +15,7 @@ namespace NexusMods.App.UI.LeftMenu.Items;
 
 public class LeftMenuItemViewModel : AViewModel<ILeftMenuItemViewModel>, ILeftMenuItemViewModel
 {
-    [Reactive] public string Text { get; set; } = "";
+    [Reactive] public StringComponent Text { get; set; } = new("");
```
erri120, line 18: This shouldn't be reactive but instead passed into the constructor.

src/NexusMods.App.UI/LeftMenu/Items/ILeftMenuItemViewModel.cs
```diff
@@ -8,7 +9,7 @@ namespace NexusMods.App.UI.LeftMenu.Items;
 
 public interface ILeftMenuItemViewModel : IViewModelInterface 
 {
-    public string Text { get; set; }
+    public StringComponent Text { get; set; }
```
erri120, line 12: Shouldn't have a setter.

erri120 (changes requested):

src/NexusMods.App.UI/LeftMenu/Items/LeftMenuItemViewModel.cs
Al12rs: Good point, didn't change that after switching from normal text.

To avoid having to change all the code I made it Init with a default though, is that ok?

erri120 (approved):

PR #2514, opened 2025-01-21
Use StringComponent for Left Menu Item ToolTips

### LINKED ISSUE #2433: Left Menu rework

Al12rs:
Should be final PR for Left Menu rework:
- Closes #2433 

Reorganized the ToolTips to use an optional StringComponent and fallback on Text if not set.

### DIFF
--- a/src/NexusMods.App.UI/LeftMenu/Items/LeftMenuItemViewModel.cs
+++ b/src/NexusMods.App.UI/LeftMenu/Items/LeftMenuItemViewModel.cs
@@ -1,6 +1,7 @@
 using System.Reactive;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
+using Avalonia.Controls;
 using DynamicData;
 using DynamicData.Binding;
 using NexusMods.Abstractions.UI;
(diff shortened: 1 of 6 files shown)


src/NexusMods.App.UI/LeftMenu/Items/LeftMenuItemViewModel.cs
```diff
@@ -17,10 +18,13 @@ public class LeftMenuItemViewModel : AViewModel<ILeftMenuItemViewModel>, ILeftMe
 {
     public StringComponent Text { get; init; } = new("");
     [Reactive] public IconValue Icon { get; set; } = new();
+    [Reactive] public string ToolTipText { get; private set; }
     [Reactive] public bool IsActive { get; private set; }
     [Reactive] public bool IsSelected { get; private set; }
     public ReactiveCommand<NavigationInformation, Unit> NavigateCommand { get; }
 
+    public StringComponent? ToolTip { get; init; }
```
erri120, line 26: This doesn't have to be nullable. You can just provide a default value:

```csharp
ToolTip { get; init; } = new(value: string.Empty);
```

This way you don't need an extra `ToolTipText` property and can just use this.

erri120 (changes requested):

src/NexusMods.App.UI/LeftMenu/Items/LeftMenuItemViewModel.cs
Al12rs: I made it nullable to be able to know whether it was set or not, if not set it is meant to fallback to showing the Text as the tooltip. 
I think it would be more messy to try to make the fallback work using a default value.

The ToolTipText property also makes things easier since I need to call `ToolTip.SetTip()` in the code behind, rather than doing a binding.

erri120 (approved):

PR #2517, opened 2025-01-21
Fixed "DWrite.dll" getting installed into "Bin" instead of "bin", BG3SE

Eddieddg1:
Just as the title describes.


Al12rs (approved):
Thanks for the fix!

Eddieddg1:
It was annoying me enough to do fix it.

Issue #391, opened 2023-07-05
As a user, I want to update a collection


Greg-Nexus:
Design: https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=1940-32020&t=Jv5UvMd7hKsGosAM-1

Collection download page checks for new revisions and shows an "Install Update" button if there are new revisions. The button adds the new revision to the app and navigates to the collection download page for the new revision.

Collection download page checks if old revisions are installed and offers to replace the old revision that deletes it from the app.

PR #2520, opened 2025-01-22
Preperations for collection updates

erri120:
- can delete individual revisions
- can fetch all or newer revision numbers
- library displays revisions instead of first revision of collections

Part of #391.

### DIFF
--- a/src/Networking/NexusMods.Networking.NexusWebApi/NexusModsLibrary.Collections.cs
+++ b/src/Networking/NexusMods.Networking.NexusWebApi/NexusModsLibrary.Collections.cs
@@ -61,11 +61,28 @@ public partial class NexusModsLibrary
     }
 
     /// <summary>
+    /// Deletes a revision and all downloaded entities.
+    /// </summary>
+    public async ValueTask DeleteRevision(CollectionRevisionMetadataId revisionId)
+    {
+        var db = _connection.Db;
+        using var tx = _connection.BeginTransaction();
+
+        var downloadIds = db.Datoms(CollectionDownload.CollectionRevision, revisionId);
+        foreach (var downloadId in downloadIds)
+        {
+            tx.Delete(downloadId.E, recursive: false);
+        }
+
+        tx.Delete(revisionId, recursive: false);
+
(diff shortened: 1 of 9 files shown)


src/Networking/NexusMods.Networking.NexusWebApi/NexusModsLibrary.Collections.cs
```diff
@@ -61,11 +61,73 @@ public partial class NexusModsLibrary
...
+            .Select(static number => RevisionNumber.From((ulong)number))
+            .ToArray();
+
+        return revisionNumbers;
+    }
+
+    /// <summary>
+    /// Deletes a revision and all downloaded entities.
+    /// </summary>
+    public async ValueTask DeleteRevision(CollectionRevisionMetadataId revisionId)
```
Al12rs, line 111: I'm a bit confused as to why `DeleteRevision` and `DeleteCollection` methods are here in `NexusMods.Networking.NexusWebApi`, wouldn't it make more sense for this stuff to live somewhere in the DataModel or something?

Al12rs (changes requested):
See comment above

src/Networking/NexusMods.Networking.NexusWebApi/NexusModsLibrary.Collections.cs
erri120: `NexusModsLibrary` has somewhat become the place for all these operations.

Al12rs: I don't want to necessarily fix that in this PR, but would you agree that it shouldn't be that way?

erri120: They could go into `CollectionDownloader`

Al12rs: Want to move in separate PR?

Al12rs (approved):

PR #2521, opened 2025-01-22
Updated Health Check UI empty state\diagnostic details.

insomnious:
- Updated Health Check to match designs and include a better empty state
- Made a UI component for the Page Header as they are being reused
- Fixed button styles not being selected correctly

[Figma link](https://www.figma.com/design/hsUiqro210T3tsK1F9l1fQ/%F0%9F%93%B1%E2%9D%A4%EF%B8%8F%E2%80%8D%F0%9F%A9%B9-Health-Check?node-id=0-1&m=dev)


src/NexusMods.App.UI/Pages/Diagnostics/List/DiagnosticListView.axaml
```diff
@@ -47,30 +48,71 @@
...
+                </controls:EmptyState.Styles>
+
+                <controls:EmptyState.Subtitle>
+                    <Border
+                        VerticalAlignment="Top"
+                        Margin="0,16,0,0"
+                        CornerRadius="4"
+                        BorderThickness="2"
+                        Padding="24"
+                        Background="#0D86EFAC"
```
insomnious, line 92: Fixed color values until Laurence says they will be used elsewhere

src/NexusMods.App.UI/Controls/PageHeader/PageHeader.axaml.cs
```diff
@@ -0,0 +1,119 @@
+using Avalonia;
+using Avalonia.Controls;
+using Avalonia.Controls.Primitives;
+using NexusMods.Icons;
+
+namespace NexusMods.App.UI.Controls.PageHeader;
+
+public class PageHeader : ContentControl
```
Al12rs, line 8: We should be careful when collapsing common features into components this way.
The resulting code isn't necessarily more concise, and they come at the cost of flexibility. 

E.g. if Designs needs us to add a button or a link in one of the Descriptions of the page headers, then we either have to revert to a manual implementation of the `PageHeader` UI, resulting in fragmentation of the code when it comes to impleming a page header, or we need to spend time to add the new capabilities to `PageHeader`, making that in turn more articulated and complex.

Just a consideration for the future when componetizing UI, not really something specifically for this PR.

insomnious: I do agree, it's a bit murky around what is or isn't worth the effort in order to minimize repetition

Al12rs (approved):

PR #2527, opened 2025-01-22
Game Hash DB Builder

halgari:
This PR closes the loop from our Steam/GOG file hashing, through the creation of a MnemonicDB with the hashes, to the downloading of that DB to a very simple usage of that DB. 

The files hashed by the CLI commands emit data into the format found in http://github.com/Nexus-Mods/game-hashes. This results in about 58MB of JSON data. Not only is this data too large, it's not indexed in any useful way. So the CLI command `game-hashes-db update` will process this JSON, insert it into a MnemonicDB, zip the DB up and prepare it for a release. The release file `game_hashes_db.zip` should then be attached to a github release, on the game-hashes repo,  as an asset. Example here: https://github.com/Nexus-Mods/game-hashes/releases/tag/v28521F70FFC7C0FA

As of now, Stardew Valley is around 7MB. This is primarilly because SDV has a large number of releases, and a large number of files in the release, around 3500 files. For contast most games shipe with around 100 files, and update once a year if ever. We'll monitor this archive to see if the size becomes too large over time, and discuss alternatives at that point. 

The new `IFileHashService` provides an interface for getting this database. It's a read-only database so no connection is exposed, only a single IDb. When a new release arrives the `IFileHashService` will automatically update to the new version. Currently it checks for updates once per application launch. I expect new hash databases to be rare (once per month), so it's unlikely that this will cause a problem. 

For now this database is used in only one location: to reduce the full file hashes required when indexing a game the first time. This is such a low-profile change that it's seemed like a good place to start integrating the database. 

Other related changes
* Moves JSON parser DI calls around so that they get included in more tests by default
* Adds more CLI utilities, such as parsing games by name. So you can pass in `stardewvalley` or `Stardew Valley` to a CLI command expecting a `ILocatableGame`
* The GOG/Steam indexers now expect a game not an app ID, they infer the AppId from the game
* Fixed a few error handling issues in the CLI system

### DIFF
--- a/src/NexusMods.Games.FileHashes/FileHashesService.cs
+++ b/src/NexusMods.Games.FileHashes/FileHashesService.cs
@@ -52,16 +52,7 @@ public class FileHashesService : IFileHashesService, IDisposable
             path.DeleteDirectory(true);
         }
 
-        OpenLatest();
-        
-        // Trigger an update
-        Task.Run(() => Task.FromResult(CheckForUpdate()));
-    }
 
-    private void OpenLatest()
-    {
-        if (_currentDb is not null)
-            return;
         
         // Open the latest database
         var latest = ExistingDBs().FirstOrDefault();
@@ -69,8 +60,11 @@ public class FileHashesService : IFileHashesService, IDisposable
         {
             _currentDb = OpenDb(latest.PublishTime, latest.Path);
         }
+        
+        // Trigger an update
+        Task.Run(() => Task.FromResult(CheckForUpdate()));
     }
-    
+
     private ConnectedDb OpenDb(DateTimeOffset timestamp, AbsolutePath path)
     {
         lock (_databases)
@@ -143,7 +137,6 @@ public class FileHashesService : IFileHashesService, IDisposable
             if (gameHashesReleseFileName.FileExists && gameHashesReleseFileName.FileInfo.LastWriteTimeUtc + _settings.HashDatabaseUpdateInterval > DateTime.UtcNow)
             {
                 _logger.LogTrace("Skipping update check due a check limit of {CheckIterval}", _settings.HashDatabaseUpdateInterval);
-                OpenLatest();
                 return;
             }
             
@@ -155,10 +148,7 @@ public class FileHashesService : IFileHashesService, IDisposable
             var existingReleases = ExistingDBs().ToList();
 
             if (existingReleases.Any(r => r.PublishTime == release.CreatedAt))
-            {
-                OpenLatest();
                 return;
-            }
 
             var tempZipPath = _settings.HashDatabaseLocation.ToPath(_fileSystem) / $"{release.CreatedAt.ToUnixTimeSeconds()}.tmp.zip";
 
--- a/tests/Games/NexusMods.Games.TestFramework/DependencyInjectionHelper.cs
+++ b/tests/Games/NexusMods.Games.TestFramework/DependencyInjectionHelper.cs
@@ -79,6 +79,10 @@ public static class DependencyInjectionHelper
                     new ConfigurablePath(baseKnownPath, $"{baseDirectory}/Archives"),
                 ],
             })
+            .OverrideSettingsForTests<FileHashesServiceSettings>(settings => settings with
+            {
+                HashDatabaseLocation = new ConfigurablePath(baseKnownPath, $"{baseDirectory}/FileHashes"),
+            })
             .AddSettingsManager()
             .AddFileHashes()
             .AddFileExtractors();


src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -967,26 +983,51 @@ public async Task IndexNewState(GameInstallation installation, ITransaction tx)
...
+            {
+                // Match on the relative path as well as the minimal hash
+                foreach (var path in matchingHash.Paths)
+                {
+                    if (path.Path == gamePath.Path)
+                        return (gamePath, matchingHash.XxHash3);
+                }
+                
+            }
+        }
```
Al12rs, line 1024: It seems this code is making two DB lookups in case there are size matches.
Wouldn't it be more efficient to only lookup the hash matches in the already filtered relations that match by size?

I'm not sure about the performance numbers, but my instincts would be to:
- lookup by size
- filter by matching paths
- perform minimal hash only if a match is found

I guess this might get better with better query syntax anyways in the not far future

src/Abstractions/NexusMods.Abstractions.Games.FileHashes/IFileHashesService.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Force an update of the file hashes database
+    /// </summary>
+    public Task ForceUpdate();
+    
+    /// <summary>
+    /// Get the file hashes database, downloading it if necessary
+    /// </summary>
+    public ValueTask<IDb> GetFileHashesDb();
```
Al12rs, line 18: So the idea is to have a single DB that contains all the hashes from all the versions of all the supported games?

Would it make sense in the future to potentially split up this DB based on games and only fetch the relevant DBs when a user starts managing a game?
And likewise, only query for updates to the DBs for which the user has games being managed?

halgari:
@Nexus-Mods/nexusmods-app-developers give this a look over even if it doesn't build yet. It passes locally but we're hitting Github API limits. I'm going have to find a solution for that.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -904,10 +912,15 @@ await Parallel.ForEachAsync(location.Value.EnumerateFiles(), async (file, token)
...
-
-                            // If the files don't match, update the entry
-                            if (fileInfo.LastWriteTimeUtc > entry.LastModified || fileInfo.Size != entry.Size)
+                            
+                            // Dates in computers are so utterly borked it's not even funny. Turns out here there's an issue, a DateTimeOffset
+                            // when run through DateTime and/or an ulong conversion will lose a very small amount of precision, about 150μs worth. 
+                            // But this precision loss is negated if everything just uses a long. So convert both sides to a long and compare.
+                            // If you ever want to try and fix this, simply subtract one from the other, and you'll see a timespan that does not
+                            // equal zero. This is the precision loss, I (halgari) am not sure how else to fix it.
+                            if (fileInfo.LastWriteTimeUtc.Date.ToFileTimeUtc() > entry.LastModified.Date.ToFileTimeUtc() || fileInfo.Size != entry.Size)
```
Al12rs, line 921: I'm not sure I follow why `.Date` is used here, was it to fix the accuracy issue? 
From reading the docs, `.Date` returns a `DateTime` of just the date information and time set at midnight, so this just checks whether there is a difference of more than a day?

src/Networking/NexusMods.Networking.GOG/Client.cs
```diff
@@ -53,13 +55,15 @@ internal class Client : IClient
     /// <summary>
     /// TTL time for the secure URL cache.
     /// </summary>
-    private static readonly TimeSpan CacheTime = TimeSpan.FromMinutes(5);
+    private static readonly TimeSpan CacheTime = TimeSpan.FromSeconds(5);
```
Al12rs, line 58: Just making sure, this change isn't a leftover of debug/testing, right?

Al12rs (commented):
Some questions and comments above, plus the failing tests to solve

src/NexusMods.Games.FileHashes/VerbImpls/Build.cs
```diff
@@ -0,0 +1,330 @@
...
+        _connection = new Connection(provider.GetRequiredService<ILogger<Connection>>(), _datomStore, provider, []);
+    }
+    
+    public IServiceProvider Provider { get; set; }
+
+    public async Task BuildFrom(AbsolutePath path, AbsolutePath output)
+    {
+        await AddHashes(path);
+        await AddGogData(path);
+        await AddSteamData(path);
```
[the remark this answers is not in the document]
halgari: No, the hashes and data is stored in the /game-hashes github repo. That is incremental. This packages up all the data and uploads the database and only needs to run whenever the files in that repo change.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
halgari: I commented it a bit, the chances for collisions here are rare, and we can improve this in the future if we need to. For now I think it'll run pretty fast.

src/Abstractions/NexusMods.Abstractions.Games.FileHashes/IFileHashesService.cs
halgari: Yes for now, a single DB, it's just easier to maintain for now. It's easy enough in the future to split it up if it gets too large.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
halgari: I tried date, and it's still off as well. I think there's strange culture/timezone stuff in all these cases, so the method that we know always works is just to use the underlying values

src/Networking/NexusMods.Networking.GOG/Client.cs
halgari: Yep, it was

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
Al12rs: The code is doing the hash lookup on the entire DB, should it not only check the Size matches? Or does it not matter due to const lookup times?

halgari: It doesn't matter here so much because all we're doing is finding a relation between a `size, path, minimal hash` and a `xxhash3`. We will be much more restrictive when we start building loadouts. In otherwords, a minimal hash in a given place, for any game is unlikely to collide with any other minimal hash in the same place from a different game.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
Al12rs: Current code uses `.Date`, which cuts off all the time data to midnight. 
So I think the current code is wrong in the sense that it checks if the created Date is greater, with 1day resolution, rather than comparing TimeStamps.

halgari: The issue is deeper than this and will require a DB migration, so I'm pulling that change out, and will get a new PR up with this change soon

src/NexusMods.Games.FileHashes/FileHashesService.cs
```diff
@@ -0,0 +1,216 @@
...
+        
+        // Open the latest database
+        var latest = ExistingDBs().FirstOrDefault();
+        if (latest.Path != default(AbsolutePath))
+        {
+            _currentDb = OpenDb(latest.PublishTime, latest.Path);
+        }
+        
+        // Trigger an update
+        Task.Run(() => Task.FromResult(CheckForUpdate()));
```
Al12rs, line 65: We have a `FireAndForget()` helper to catch exceptions in non awaited tasks like this.
Do you think it would be worth using here?

Al12rs (approved):

src/NexusMods.Games.FileHashes/FileHashesService.cs
halgari: I'll swap that over sometime soon

PR #2534, opened 2025-01-23
[Updates 1/?] Added: Initial Support for Displaying Updates in Library

Sewer56:
This adds the basic functionality for running an update check, and showing updates in UI.
Quick notes:

- Update All button is hidden until we make it work, but is present in code
- We check for updates on entering library; or when pressed explicitly by user
- Behaviour of `update` button matches that of `install` button
    - With regards to how numbers are displayed on the UI
    - Confirmed with @captainsandypants as new intended behaviour.
    - FIGMA will be updated, if it hasn't been already.
- Updates work in both Tree and Flat View

Took a bit longer than expected, due to fixes I had to make last second; and being sick middle of week.

Checking for updates may be slow with a lot of mods, due to borked caching. (In order of many seconds)
Fix for that is in next PR; which in fact is already ready.

Known Issues:

- Switching from Tree/Flat View doesn't reuse previous update check results (needs to run update manually).

Plenty of more to come, small and iterative.

------------

~~CC. @erri120~~

~~I broke lazy initialisation here for children of mod pages, meaning the child viewmodels get created immediately when the grid is first created, rather than on activation of the parent model. After looking around at the problem for a bit; I figured I'd do the minimum to get things to 'work' here; and that's not lazy loading at all; because I'm under the impression this will all immediately get ported to `CompositeItemModel`s directly after merge; so the previous code will be recycled.~~

~~There is no further UI work to my awareness that would be blocking the migration; I've mostly got code behind left.~~

~~The problem is, when the user runs the update check (on entering page, or manually), the update procedure needs to write to ViewModels which have not been loaded in yet. For instance suppose you enter to library, with 83 mods:~~

image

~~If a mod at the bottom of the list, gets an update, say `SMAPI`; there is no way to update the row without actually loading the row itself. We could use some sort of global state that's read at initialisation time to keep the lazy properties, but that's also extra complexity. Or we could just load the expensive parts of the row (e.g. thumbnails) lazily, while keeping all the models pre-loaded.~~

~~I don't know the nature of the new system you wrote yet (was not involved in review, etc.); but since we're canning the current system, I figured I'd leave that up to what you feel is best.~~


src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridItemModel.cs
```diff
@@ -88,59 +89,61 @@ public bool IsExpanded
...
+        IObservable<IChangeSet<TModel, TKey>>? childrenObservable = null)
     {
+        HasChildrenObservable = hasChildrenObservable ?? Observable.Return(false);
+        ChildrenObservable = childrenObservable ?? Observable.Empty<IChangeSet<TModel, TKey>>();
         _childrenView = _children.ToNotifyCollectionChanged();
 
-        _modelActivationDisposable = WhenModelActivated(this, static (model, disposables) =>
+        // NOTE(sewer): If you're reading this during PR review; this code is marked
+        // for obsoletion as it will be replaced by `CompositeItemModel` system.
+        // So am not building anything more elaborate here.
```
erri120, line 105: `TreeDataGridItemModel` is not marked as obsolete. It's still the base class for all things TreeDataGrid related as it facilitates logic for expanding and showing child item models.

The changes in this file should be reverted.

src/NexusMods.App.UI/Pages/LibraryPage/NexusModsModPageLibraryItemModel.cs
```diff
@@ -103,14 +110,42 @@ public NexusModsModPageLibraryItemModel(IObservable<IChangeSet<NexusModsLibraryI
...
+                    .Max(x => NuGetVersion.TryParse(x.Version.Value, out var version) 
+                        ? version : new NuGetVersion(0, 0, 0));
+
+                if (maxVersion != null)
+                {
+                    // SAFETY: Not null because >= 1 item.
+                    _preUpdateVersion = maxVersion!.ToString();
+                    Version.Value = _preUpdateVersion;
+                } 
+            });
```
erri120, line 142: I think going to change `ChildrenObservable` to a private getter to prevent this. It's never supposed to be accessed by anything other than `TreeDataGridItemModel` itself.

src/NexusMods.App.UI/Pages/LibraryPage/LibraryViewModel.cs
```diff
@@ -219,8 +230,158 @@ public LibraryViewModel(
...
+    }
+    
+    private void HandleUpdateMessage(ActionMessage message, CancellationToken cancellationToken)
+    {
+        void StartLibraryItemUpdate(LibraryItemId id)
+        {
+            // By definition, only works on Nexus library items.
+            var libraryItem = LibraryItem.Load(_connection.Db, id);
+            if (!libraryItem.IsValid() || !libraryItem.TryGetAsNexusModsLibraryItem(out var nexusLibraryItem)) 
+                return;
```
erri120, line 267: Can just do `NexusModsLibraryItem.Load` directly and check if that's valid.

src/NexusMods.App.UI/Pages/LibraryPage/LibraryViewModel.cs
```diff
@@ -219,8 +230,158 @@ public LibraryViewModel(
...
+            // Reuse known newest version in local storage, obtained via
+            // call to make starting this update possible in first place.
+            var newerItems = RunUpdateCheck.GetNewerFilesForExistingFile(nexusLibraryItem.FileMetadata);
+            var mostRecentVersion = newerItems.FirstOrDefault();
+            if (!mostRecentVersion.IsValid()) // Catch case of no newer items.
+                return;
+
+            var modFileUrlFormat = "https://www.nexusmods.com/Core/Libs/Common/Widgets/DownloadPopUp?id={0}&game_id={1}";
+            var modFileUrl = string.Format(modFileUrlFormat, mostRecentVersion.Uid.FileId, mostRecentVersion.Uid.GameId);
+            var osInterop = _serviceProvider.GetRequiredService<IOSInterop>();
```
erri120, line 278: URL formatting should go into `NexusModsURLBuilder`.

src/NexusMods.App.UI/Pages/LibraryPage/LibraryViewModel.cs
```diff
@@ -219,8 +230,158 @@ public LibraryViewModel(
...
+        }
+        else
+        {
+            StartLibraryItemUpdate(singleId);
+        }
+    }
+
+    // Note(sewer): ValueTask because of R3 constraints with ReactiveCommand API
+    private async ValueTask RefreshUpdates(CancellationToken token) 
+    {
```
erri120, line 295: I really hate this entire method. The entire approach feels wrong, and not just because it's accessing the child item models directly and switching on deprecated item model types. The architecture is completely backwards. This method didn't have to exist. All you needed to show that a library item has an update available is some service, probably `RunUpdateCheck` that gives you an observable like this:

```csharp
IObservable<Optional<NexusModsFileMetadata.ReadOnly>> GetNewestVersionObservable(NexusModsFileMetadata.ReadOnly current); 
```

Or something similar depending on what input data you have. Maybe it accepts the mod page instead or something. Regardless, the `NexusModsDataProvider` would just use that observable to change the UI. That's how everything else works in the data providers, they just get an observable for some piece of data and use that to update the UI.

This entire method should just be 10 lines of calling the API to update the mod pages and fetch the newest files. All of the UI updating would be done in the data providers using the observable described above.

erri120 (changes requested):

src/NexusMods.App.UI/Pages/LibraryPage/NexusModsModPageLibraryItemModel.cs
```diff
@@ -103,14 +112,44 @@ public NexusModsModPageLibraryItemModel(IObservable<IChangeSet<NexusModsLibraryI
...
```
erri120, line 146: This should be removed. As with the other stuff, direct access of the children shouldn't be done. This data should be supplied by the data provider.

erri120 (changes requested):

src/NexusMods.App.UI/Pages/LibraryPage/NexusModsModPageLibraryItemModel.cs
Sewer56: I knos what you mean, gimme a bit, just gotta get my stuff sorted before I scuffle to my PC

Sewer56: sorted

erri120 (approved):

PR #2535, opened 2025-01-23
[Updates 2/?] Fixed/Optimized: Slow Update Checks on Large Collections due to Small Oversights in Caching Strategy

Sewer56:
This PR Follows:
- https://github.com/Nexus-Mods/NexusMods.App/pull/2534

Merge this after the other one.
***Do not review until the other one is merged.***

---------------------

This PR improves the caching mechanism used to ensure we don't unnecessarily update mod page info.
Existing logic was mostly correct, but was missing a few things.

- Added a timestamp which signifies last time a mod page was updated in the DataStore
    - This avoids unnecessarily pulling info for every mod which was updated over 28 days ago each time.
- Checking for updates of non-cached pages is now parallelised, 16x.
- Mod pages which no longer exist are now treated as Up-To-Date (if less than 28 days old) or Out of Date otherwise.
    - If 'out of date', we suffer a small speed penalty; due to how code is structured.
    - But on Nexus we don't remove fully pages, except for DMCA.
        - And even then I don't know if the page is actually fully removed from API, so this may be a nothingburger.

More exact details can be found by examining code, but basically, we only now update mod page info if our local copy is older than 28 days. Otherwise we use a cached copy in all cases, successfully. Previously we used timestamps of the mod pages themselves, now we use timestamps of when we last fetched the data.

After this PR, update checks should be roughly 200-400ms in runtime depending on your connection to Nexus servers when you have around 100 mods installed. A (very tiny) bit longer for cold uncached reads, but now we're also sending these in parallel.


src/Networking/NexusMods.Networking.NexusWebApi/RunUpdateCheck.cs
```diff
@@ -49,28 +49,56 @@ public static async Task<PerFeedCacheUpdaterResult<PageMetadataMixin>> CheckForM
     public static async Task UpdateModFilesForOutdatedPages(IDb db, ITransaction tx, ILogger logger, INexusGraphQLClient gqlClient, PerFeedCacheUpdaterResult<PageMetadataMixin> result, CancellationToken cancellationToken)
     {
         // Note(sewer): Undetermined items may be removed items from the site; or
-        // caused by programmer error, so wr should log these whenever possible,
+        // caused by programmer error, so we should log these whenever possible,
         // but they should not cause a critical error; in case it's simply the result
         // of mod removal such as DMCA takedown.
-        foreach (var mixin in result.UndeterminedItems)
+        using var semaphore = new SemaphoreSlim(16);
```
Al12rs, line 55: Why 16? Should we use something like `Environment.ProcessorCount` instead? They are going to execute on the same thread, so it doesn't quite track, but 16 seems arbitrary.

Al12rs (changes requested):

src/Networking/NexusMods.Networking.NexusWebApi/RunUpdateCheck.cs
```diff
@@ -49,28 +49,56 @@ public static async Task<PerFeedCacheUpdaterResult<PageMetadataMixin>> CheckForM
...
-                await UpdateModPage(db, tx, gqlClient, cancellationToken, mixin);
+                await semaphore.WaitAsync(cancellationToken);
+                try
+                {
+                    await UpdateModPage(db, tx, gqlClient, cancellationToken, mixin);
+                }
+                finally
+                {
+                    semaphore.Release();
+                }
```
Al12rs, line 71: `semaphore.WaitAsync(cancellationToken);` can actually return without acquiring the resource if the token is already cancelled on first enter, which can lead to a release of something that was never acquired in the `finally` block.

`WaitAsync(cancellationToken);` returns false if the resource is never acquired, so you can use that do determine whether you need to release or not in the finally.

src/Networking/NexusMods.Networking.NexusWebApi/RunUpdateCheck.cs
Sewer56: [This is for others reading]
This is an I/O bound workload, not a CPU bound one.

What we're actually doing is throttling the amount of concurrent requests to the Nexus API, more strictly saying, we only want 16 requests to the API at any given time.

What actually happens is when we await, it yields, and one of the other tasks gets picked up; we're actually limiting the requests to 16 at a time in transit.

The number is arbitrary (as we DM'd to each other too); I'll make a note of this in the code.

src/Networking/NexusMods.Networking.NexusWebApi/RunUpdateCheck.cs
Al12rs: Yeah, right I forgot about the specific overload thing

src/Networking/NexusMods.Networking.NexusWebApi/RunUpdateCheck.cs
```diff
@@ -49,28 +49,70 @@ public static async Task<PerFeedCacheUpdaterResult<PageMetadataMixin>> CheckForM
...
+        using var semaphore = new SemaphoreSlim(16);
+        var tasks = new List<Task>();
+
+        // Helper function to process a single mixin with error handling
+        async Task ProcessMixin(SemaphoreSlim sema, PageMetadataMixin mixin, bool isUndetermined)
         {
             try
             {
-                await UpdateModPage(db, tx, gqlClient, cancellationToken, mixin);
+                var isTaken = await sema.WaitAsync(-1, cancellationToken);
```
Al12rs, line 76: Use `Timeout.Infinite` instead of -1

Sewer56: Sure, it's same thing either way.

Al12rs: Yeah, just for code clarity

src/Networking/NexusMods.Networking.ModUpdates/Mixins/PageMetadataMixin.cs
```diff
@@ -22,8 +22,17 @@ public struct PageMetadataMixin : IModFeedItem
...
+    public DateTime GetLastUpdatedDateUtc()
+    {
+        // Local update time in database. Not on remote server.
+        if (NexusModsModPageMetadata.DataUpdatedAt.TryGetValue(_metadata, out var result))
+            return result.DateTime;
+        
+        // If not in DB for whatever reason, default to min, will be refreshed on next
+        // update check.
+        return DateTime.MinValue;
+    }
```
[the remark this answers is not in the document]
Sewer56: Let me change that up quickly, this API (`IModFeedItem`) predates the migration work there, I'll need to change it in another struct or two too.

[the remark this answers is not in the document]
Sewer56: Thankfully I got tests for all of this, so it's very easy to verify the logic.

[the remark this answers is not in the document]
Sewer56: Should be good, adjusted the API, ran the tests, seems good to me.

src/Networking/NexusMods.Networking.NexusWebApi/RunUpdateCheck.cs
```diff
@@ -49,28 +49,70 @@ public static async Task<PerFeedCacheUpdaterResult<PageMetadataMixin>> CheckForM
...
+                if (isUndetermined)
+                {
+                    var id = mixin.GetModPageId();
+                    logger.LogError(e, "Failed to update metadata for Mod (GameID: {Page}, ModId: {ModId})", id.GameId, id.ModId);
+                }
+                else
+                {
+                    // Rethrow for non-undetermined items as these should be truly exceptional
+                    throw;
+                }
```
Al12rs, line 98: I don't think you need two try catch here. I would declare `isTaken = false;` outside then do a single try catch and finally with the  `if (isTaken) sema.Release();`

Sewer56: All done 👏

Al12rs (approved):

PR #2538, opened 2025-01-27
Fix `ApplyChanges` exceptions in the library/loadout

### LINKED ISSUE #2529: Fix `NullReferenceException`: `R3Extensions.ApplyChanges[TKey,TValue]`
```
00:05:11.054 [ERROR] (NexusMods.App.Startup) Exception from Observable in onNext|System.NullReferenceException: Object reference not set to an instance of an object.
   at NexusMods.Abstractions.UI.Extensions.R3Extensions.ApplyChanges[TKey,TValue](IDictionary`2 dict, IChangeSet`2 changes) in C:\Dev\Repos\NexusMods.App\src\Abstractions\NexusMods.Abstractions.UI\Extensions\R3Extensions.cs:line 133
   at NexusMods.App.UI.Pages.LibraryPage.IHasLinkedLoadoutItems.<>c__DisplayClass4_0`1.<SetupLinkedLoadoutItems>b__1(IChangeSet`2 changes) in C:\Dev\Repos\NexusMods.App\src\NexusMods.App.UI\Pages\LibraryPage\ILibraryItemModel.cs:line 47
   at NexusMods.App.UI.ReactiveUiExtensions.ExceptionObserverWithOnNext`1.OnNextCore(T value) in C:\Dev\Repos\NexusMods.App\src\NexusMods.App.UI\ReactiveUIExtensions.cs:line 185
```
## Steps:
Install a new collection, or remove a collection. In general it seems to trigger when large amounts of changes get pushed quickly.

Al12rs:
- Fixes #2529 

This `DisposeMany` seems to wreak havoc on items that are still in view or similar. I don't know if it is necessary to replace this with a different mechanism for disposing the items, but I think it should happen higher level to avoid items getting disposed before they are fully removed from view and other collections.

### DIFF
--- a/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
+++ b/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
@@ -7,6 +7,7 @@ using NexusMods.Abstractions.UI.Extensions;
 using NexusMods.App.UI.Extensions;
 using ObservableCollections;
 using R3;
+using System.Reactive.Linq;
 
 namespace NexusMods.App.UI.Controls;
 
@@ -99,8 +100,9 @@ public abstract class TreeDataGridAdapter<TModel, TKey> : ReactiveR3Object
                     return self
                         .GetRootsObservable(viewHierarchical)
                         .OnUI()
+                        .Do(changeSet => self.Roots.ApplyChanges(changeSet))
+                        .DisposeMany()
                         .ToObservable()
-                        .Do(self, static (changeSet, self) => self.Roots.ApplyChanges(changeSet))
                         .Select(viewHierarchical, static (_, viewHierarchical) => viewHierarchical);
                 })
                 .Switch()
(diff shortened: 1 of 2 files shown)


erri120 (changes requested):
Since the exception mentioned in #2529 is for `IHasLinkedLoadoutItems` which will cease to exist soon, I don't believe disabling disposing will achieve anything. In fact, it will make the app worse as the new `CompositeItemModel` type is more reliant on proper disposing than our old models.

If you really want to a temp fix, you should look at `IHasLinkedLoadoutItems` where the exception is actually being thrown.

Al12rs:
> Since the exception mentioned in #2529 is for `IHasLinkedLoadoutItems` which will cease to exist soon, I don't believe disabling disposing will achieve anything. In fact, it will make the app worse as the new `CompositeItemModel` type is more reliant on proper disposing than our old models.
> 
> If you really want to a temp fix, you should look at `IHasLinkedLoadoutItems` where the exception is actually being thrown.

I think the exception in `IHasLinkedLoadoutItems` is the symptom, not the issue. The issue is that items are getting disposed before they are fully removed from relevant collections. 
Dispose should happen after the items have been fully removed from our models structure, otherwise we get into inconsistent states, like what we currently see in `IHasLinkedLoadoutItems`.
I don't think `DisposeMany` there is a good solution for disposing the items as it is too soon

erri120 (approved):

PR #2545, opened 2025-01-28
Switch timestamps to ticks

halgari:
This PR is the outcome of some testing and bug fixing in my game hashes PR. There's a problem currently in our timestamps, they are off by roughly 1microsecond. What seems to cause this is that we are storing timestamps in MnemonicDB as unix timestamps (ms since Unix Epoch) in a 64 bit integer. However `DateTimeOffset` stores these dates in Ticks. A tick is 100 nanoseconds and the epoch is Jan 1st, year 1 of the Gregorian calendar. This conversion causes some loss in the datetime conversion. 

This is a rather major issue in the synchronizer as it means we always end up synching all the files in the game folder during apply. Or at least that vast majority of them. 

This PR does the following: 

* Pulls in the new version of MnemonicDB which stores times as ticks in the TimestampAttribute
* Reworks the DB migrations to be closer to what we actually want
* Sets up a naming convention and migration system for the database. 
* Stores logs about migrations in the MnemonicDB database
* Automatically performs migrations as needed
* Updates the schema version tests to follow the new format
* Implements a schema migration for converting from unix times to ticks for all TimestampAttributes in the db. 

Nice thing here is that the values are so radically different that we can detect attributes that are already converted, as the year for these dates will be clearly visible once run through the conversion routines. No, a date for the 41st millenium is not valid, contrary to the view of several people in the Nexus Mods office.


Sewer56:
As a fun piece of trivia before I nap.

The 100ns per tick is actually inherited from the Windows API, or specifically from the NT kernel a.k.a. 'NT Time Epoch'. 

In .NET they just offset the starting date to extend the range of possible dates, so two way conversion is down to a pure integer addition/subtraction.

For Unix based systems, conversion is a bit more involved since format is different.

src/NexusMods.DataModel.SchemaVersions/Migrations/_0001_ConvertTimestamps.cs
```diff
@@ -0,0 +1,61 @@
...
+    
+    /// <summary>
+    /// Perform the actual data conversion
+    /// </summary>
+    private static long ConvertTimestamps(long oldTimestamp)
+    {
+        try
+        {
+            var dt = DateTimeOffset.FromUnixTimeMilliseconds(oldTimestamp);
+            return dt.Ticks;
```
erri120, line 53: Should be `UtcTicks`.

erri120 (changes requested):

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -495,12 +495,13 @@ await _fileStore.ExtractFiles(toExtract.Select(item =>
                     throw new InvalidOperationException("File found in tree processing is not a loadout file, this should not happen (until generated files are implemented)");
                 }
 
+                var writeTimeUtc = register.GetResolvedPath(entry.Path).FileInfo.LastWriteTimeUtc;
```
erri120, line 498:
```suggestion
                var writeTimeUtc = new DateTimeOffset(register.GetResolvedPath(entry.Path).FileInfo.LastWriteTimeUtc);
```

Functionality the same as before except it's now explicit that we're using `DateTimeOffset` instead of using the implicit conversion from `DateTime` to `DateTimeOffset`.

We've had funky exceptions going from `DateTime` to `DateTimeOffset` before, especially when dealing with default or epoch values, this just makes sure any exceptions related to the conversion happen in this line.

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -906,7 +906,7 @@ await Parallel.ForEachAsync(location.Value.EnumerateFiles(), async (file, token)
                             var fileInfo = file.FileInfo;
 
                             // If the files don't match, update the entry
-                            if (fileInfo.LastWriteTimeUtc > entry.LastModified || fileInfo.Size != entry.Size)
+                            if (fileInfo.LastWriteTimeUtc != entry.LastModified || fileInfo.Size != entry.Size)
```
erri120, line 909: Same as above, would suggest doing the conversion explicit:

```csharp
var fileInfo = file.FileInfo;
var writeTimeUtc = new DateTimeOffset(fileInfo.LastWriteTimeUtc);

// If the files don't match, update the entry
if (writeTimeUtc != entry.LastModified || fileInfo.Size != entry.Size)
{
    var newHash = await file.XxHash3Async();
    tx.Add(entry.Id, DiskStateEntry.Size, fileInfo.Size);
    tx.Add(entry.Id, DiskStateEntry.Hash, newHash);
    tx.Add(entry.Id, DiskStateEntry.LastModified, writeTimeUtc);
    changes = true;
}
```

src/NexusMods.DataModel.SchemaVersions/Migrations/_0001_ConvertTimestamps.cs
halgari: They're already in UTC format, so I'm not sure it matters, but I'll swap it over anyways just incase

erri120 (approved):

PR #2551, opened 2025-01-28
Collections: implement Install Optional mods button

### LINKED ISSUE #2550: Optional Mods page has no download all if all required mods installed
Optional Mods page has no download all if all required mods installed. 

@erri120  confirms this and will be fixing it in the next release.

Al12rs:
- closes #2550 
I think ticket above wasn't quite correct, what was missing was the Install button, not the downloads button.

### DIFF
--- a/src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadViewModel.cs
@@ -106,8 +106,16 @@ public sealed class CollectionDownloadViewModel : APageViewModel<ICollectionDown
             configureAwait: false
         );
 
-        // TODO: implement this button
-        CommandInstallOptionalItems = IsInstalling.CombineLatest(_canInstallOptionalItems, static (isInstalling, canInstall) => !isInstalling && canInstall).ToReactiveCommand<Unit>();
+        CommandInstallOptionalItems = IsInstalling.CombineLatest(_canInstallOptionalItems, static (isInstalling, canInstall) => !isInstalling && canInstall).ToReactiveCommand<Unit>(executeAsync: async (_, _) => { await InstallCollectionJob.Create(
+                serviceProvider,
+                targetLoadout,
+                source: libraryFile,
+                revisionMetadata,
+                items: CollectionDownloader.GetItems(revisionMetadata, CollectionDownloader.ItemType.Optional)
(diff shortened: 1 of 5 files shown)


src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadViewModel.cs
```diff
@@ -295,13 +306,16 @@ public CollectionDownloadViewModel(
...
-                .Subscribe(numDownloadedOptionalItems =>
+                .Subscribe(tuple =>
                 {
+                    var (numDownloadedOptionalItems, hasInstalledAllOptionals) = tuple;
                     var hasDownloadedAllOptionalItems = numDownloadedOptionalItems == OptionalDownloadsCount;
 
                     CountDownloadedOptionalItems = numDownloadedOptionalItems;
-                    _canInstallOptionalItems.OnNext(numDownloadedOptionalItems > 0);
+                    HasInstalledAllOptionalItems.Value = hasInstalledAllOptionals;
+                    _canInstallOptionalItems.OnNext(numDownloadedOptionalItems > 0 && !hasInstalledAllOptionals);
```
erri120, line 318: You should only be able to install all optional items if you have downloaded all optional items, not just one.

Al12rs: Hm, I'm not sure actually. Since perhaps I would want to just download most of them but not all of them (or one of the external optional ones fails or something) and then I would still want to have the convenience of the button at the top.
It should dynamically update based on whether there are downloaded optional items that need to be installed or not.

Al12rs: I think it already works as you said it should:
image

erri120: It's harder to do it that way, something we can do in a later PR. This PR should just have a button for install all optional items.

Al12rs: Changed it. It was already behaving that way due to the IsVisible logic, so it didn't really matter, but I guess it's better to have consistent code.

erri120 (approved):

Al12rs:
Closing and re opening for fixing tests

PR #2556, opened 2025-01-28
Fix migration check for existing data models

Al12rs:
The tested path is a directory, not a file, which makes the check always return false, even if the DB actually exists.
Let me know if it can be a file as well, in which case the code needs to be updated to check for either.


Al12rs:
Duplicate of #2557

PR #2559, opened 2025-01-28
`CompositeItemModel` for the Library

erri120:
Moves the library models to `CompositeItemModel`.


Al12rs:
I think this is breaking deleting mods from the library

erri120:
> I think this is breaking deleting mods from the library

Works fine on my end.

src/NexusMods.App.UI/Pages/LibraryPage/LibraryViewModel.cs
```diff
@@ -311,15 +261,14 @@ await Parallel.ForAsync(
...
-
-        var ids2 = Adapter.SelectedModels
-            .OfType<IIsChildLibraryItemModel>()
-            .Select(static model => model.LibraryItemId);
+        var ids = Adapter.SelectedModels
+            .Select(static model => model.GetOptional<LibraryComponents.InstallAction>(LibraryColumns.Actions.InstallComponentKey))
+            .Where(static optional => optional.HasValue)
+            .SelectMany(static optional => optional.Value.ItemIds)
+            .Distinct()
+            .ToArray();
```
Al12rs, line 269: This doesn't select installed items for Removal command

erri120: What mods do you use to repro this?

Al12rs: Mods installed with a collection

erri120: Resolved in a call, issue was a subscription not being active thus IDs collection was not populated.

src/Networking/NexusMods.Networking.NexusWebApi/ModUpdateService.cs
```diff
@@ -66,75 +65,56 @@ await RunUpdateCheck.UpdateModFilesForOutdatedPages(
...
-                newestDate = mostRecentItem.UploadedAt;
-                newestItem = mostRecentItem;
-                newestItemOldVersion = modFile.Version;
-                isAnyOnModPageNewer = true;
-            }
+        var modPageToNewerFiles = existingFileToNewerFiles
+            .SelectMany(static kv => kv.Value)
+            .DistinctBy(static fileMetadata => fileMetadata.Id)
+            .GroupBy(static fileMetadata => fileMetadata.ModPageId)
+            .ToDictionary(static group => group.Key, static group => group.ToArray());
```
Sewer56, line 97: Just a quick repeat since I've mentioned after standup in call.

Design wants to show details of the mod that updated most recently on the mod page row in the library when there's an update. That may not possibly work right when there are multiple mods on a mod page.

Although I haven't 100% verified this, I believe the code works something like this:

- `existingFileToNewerFiles` contains a Key->Value where Value contains newer files in sorted order
- we merge files from multiple mods via `SelectMany`, concatenating all items, preserving order
- then we group by the mod page.
- when the grouping is done, the order from the `SelectMany` is preserved

The problem is, if you have 2 mods on a page which have an update, and the second one has a file update newer than the first one, the first one will be shown anyway. If I'm understanding the code right.

I think you just want to sort the group array. Design didn't define what 'newest update on page' means, so feel free to just use newest uploaded file.

src/NexusMods.App.UI/Pages/LibraryPage/ILibraryItemModel.cs
```diff
@@ -14,75 +14,3 @@ namespace NexusMods.App.UI.Pages.LibraryPage;
```
Sewer56, line 14: The using directives in this file are now unused.
Just clean up in interest of reducing warnings, maybe one day we'll come closer to 0 again.

Sewer56:
First time looking at the CompositeItemModel system; pretty neat. Seems fun to use; composes fairly cleanly, aside a bit of boilerplate in the XAML.

I think I have a very, very rough grasp of how it works from reading this code: 
- TreeDataGrid rows bind to CompositeItemModel
- The XAML columns use `ComponentControl` which pull the inner `IItemModelComponent` via the `ComponentKey`; and bind to that.
- TreeDataGrid creates CustomTemplateColumn from `ICompositeColumnDefinition` (via ColumnCreator)

The XAML binds to the actual `CompositeItemModel`, but the inner control `ComponentTemplate` extracts the component with a matching key, seems to check out.

There's a lot of back and fourth between the various classes, but reading this code was a good exercise to see how this system works at a basic level. I think I know enough-ish to do some simple manipulation on existing trees, but in the long run this is probably better documented when we get some time after the SDV Alpha. I'm not (yet) currently sure as to how to cleanly integrate update button with this.

Smallest thing I've noticed there is that `AssertComponent` is a single function that throws without using a ThrowHelper; which prevents it from being inlined. GIven that the method is just one check with a branch, it'd be good to wrap the `InvalidOperationException` in a `ThrowHelper`, so the check itself gets inlined while the throw is a cold call.

Based on my limited knowledge of what I'm working with so far, the code seems good; I'll leave it to @Al12rs to do the final honours. Timing here is a bit unfortunate, as we've ended up dropping the 'update' button from this code; which is a bit blocking on my end, but that can be done as a separate PR. I haven't worked enough with this system yet to know how to do that cleanly however; without chucking both install and update into a single column and somehow sharing the data.

Al12rs:
> First time looking at the CompositeItemModel system; pretty neat. Seems fun to use; composes fairly cleanly, aside a bit of boilerplate in the XAML.
> 
> I think I have a very, very rough grasp of how it works from reading this code:
> 
> * TreeDataGrid rows bind to CompositeItemModel
> * The XAML columns use `ComponentControl` which pull the inner `IItemModelComponent` via the `ComponentKey`; and bind to that.
> * TreeDataGrid creates CustomTemplateColumn from `ICompositeColumnDefinition` (via ColumnCreator)
> 
> The XAML binds to the actual `CompositeItemModel`, but the inner control `ComponentTemplate` extracts the component with a matching key, seems to check out.

Yeah, that is a pretty accurate description of the system. The advantage is that the resulting `CompositeItemModels` are mutable, you can add and remove Components on an existing model and have the contents of the columns update accordingly.

The previous system used interfaces to define what a particular ItemModel would display. This made it so that you would have to remove the entire model and replace it with another one that had a different combination of interfaces just to change the contents of one column. The combination of possible concrete implementations was getting out of hand, making the system not scalable.

erri120:
@Sewer56 I'm going to re-add the update button. For the action column we have the `MultiComponentControl` which we currently don't use, but the idea is that the control will switch it's content based on what components the item model has in a mutually exclusive selection process that goes from top to bottom in XAML.

Sewer56:
Works for me, I did see that when reading through thr code.

I would guess both AL and I probably reas through yesterday's code by now. If you want to merge and do in successive PR, that's fine, can easily check this off and get it merged. All up to you.

Al12rs (approved):

PR #2564, opened 2025-01-29
Fix Active loadout state bug

### LINKED ISSUE #2563: Last Applied Loadout status is bugged when switching loadouts
Image

## Summary
After switching between two empty loadouts, they both appear to be the Active loadouts.

## Steps to reproduce
Switch between two empty loadouts.

The old loadout should not longer show up as the last applied one, only the new one should

## Other information:
Bug appears to be in `SynchronizerService.StatusForLoadout()`. 
Early guess is that a long Diff operation is triggered, the status is changed, then the diff operation finishes, and the result overrides the previously set result.

Al12rs:
- Fixes #2563

Fix incorrect LastApplied states being returned due to long diff operation returning out of order.
Using R3 we can ensure that the async operations are executed sequentially, and longer operations don't override later results.

### DIFF
--- a/src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
+++ b/src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
@@ -135,7 +135,7 @@ public class SynchronizerService : ISynchronizerService
             .Select(e => e.Value ? GameSynchronizerState.Busy : GameSynchronizerState.Idle);
     } 
     
-    private readonly Dictionary<LoadoutId, IObservable<LoadoutSynchronizerState>> _statusObservables = new();
+    private readonly Dictionary<LoadoutId, Observable<LoadoutSynchronizerState>> _statusObservables = new();
     private readonly SemaphoreSlim _statusSemaphore = new(1, 1);
 
     /// <inheritdoc />
@@ -145,11 +145,11 @@ public class SynchronizerService : ISynchronizerService
         try
         {
             // This observable may perform heavy diffing operation, so it needs to be shared between all subscribers
-            if (_statusObservables.TryGetValue(loadoutId, out var observable)) return observable;
+            if (_statusObservables.TryGetValue(loadoutId, out var observable)) return observable.AsSystemObservable();
 
             observable = CreateStatusObservable(loadoutId);
             _statusObservables[loadoutId] = observable;
-            return observable;
+            return observable.AsSystemObservable();
         }
         finally
         {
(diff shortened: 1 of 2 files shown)


src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
```diff
@@ -170,49 +170,51 @@ private IObservable<LoadoutSynchronizerState> CreateStatusObservable(LoadoutId l
...
-        var statusObservable = Observable.CombineLatest(isBusy,
-                lastApplied,
+        var statusObservable = isBusy.CombineLatest(lastApplied,
                 revisions,
                 (busy, last, rev) => (busy, last, rev.loadout, rev.revDbTx)
             )
             .DistinctUntilChanged()
-            .SelectMany(
-                async tuple =>
+            .ToObservable()
```
erri120, line 180: Should try and keep as much of the observable stream in R3. I'd suggest moving this further up.

src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
```diff
@@ -170,49 +170,51 @@ private IObservable<LoadoutSynchronizerState> CreateStatusObservable(LoadoutId l
...
                     return diffFound ? LoadoutSynchronizerState.NeedsSync : LoadoutSynchronizerState.Current;
-                }
+                },
+                awaitOperation: AwaitOperation.Sequential
             )
             .Replay(1)
             .RefCount();
 
-        return statusObservable;
+        return statusObservable.AsSystemObservable();
```
erri120, line 217: At this point we should return an R3 observable.

src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
Al12rs: I changed most of the code in the function and called to be R3. I didn't change the interface signature as that would have required changing all the users and I didn't want to go too out of scope

erri120 (approved):

src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
```diff
@@ -145,69 +145,71 @@ public async Task<IObservable<LoadoutSynchronizerState>> StatusForLoadout(Loadou
...
+                            var diffTree = GetApplyDiffTree(loadoutId);
+                            var diffFound = diffTree.GetAllDescendentFiles().Any(f => f.Item.Value.ChangeType != FileChangeType.None);
+                            _logger.LogTrace("Changes found in loadout {LoadoutId}: {DiffFound}", loadoutId, diffFound);
+                            return diffFound;
+                        }, cancellationToken);
 
                     return diffFound ? LoadoutSynchronizerState.NeedsSync : LoadoutSynchronizerState.Current;
-                }
+                },
+                awaitOperation: AwaitOperation.Sequential
```
erri120, line 212: Should be `Switch`.

erri120 (changes requested):

erri120 (approved):

PR #2573, opened 2025-01-29
Add a copy button on the error dialog

Al12rs:
image


Sewer56 (approved):
Approving this, effective immediately; so it can be pushed down onto the imminent release.
Very simple change; whether we want an abstraction that doesn't require TopLevel will be a topic for another day.

PR #2574, opened 2025-01-29
Memory usage reduction

halgari:
This fixes a few issues with memory usage in the app:

* It uses R3's `ThrottleFirstLast` in the Synchronizer service so we only ever check synchronization of one loadout revision at a time, and drop any itermediate revisions that are created while a loadout is being checked for updates. Without this fix we run a unlimited number of synchronizer calls at one time causing a unlimited amount of memory usage. 
* It uses Parallel.ForEachAsync in the Diagnostic manager, so that we don't have to wait for a sequential running of all diagnostics. 

After introducing these two changes I saw memory usage go from 8GB during a SDV collection install, to staying roughly at 1GB.

### DIFF
--- a/src/NexusMods.DataModel/Diagnostics/DiagnosticManager.cs
+++ b/src/NexusMods.DataModel/Diagnostics/DiagnosticManager.cs
@@ -80,7 +80,7 @@ internal sealed class DiagnosticManager : IDiagnosticManager
 
         try
         {
-            ConcurrentBag<Diagnostic> diagnostics = new();
+            List<Diagnostic> diagnostics = [];
             
             await Parallel.ForEachAsync(diagnosticEmitters.OfType<ILoadoutDiagnosticEmitter>(), cancellationToken, async (emitter, token) =>
             {
@@ -89,7 +89,12 @@ internal sealed class DiagnosticManager : IDiagnosticManager
                 try
                 {
                     await foreach (var diagnostic in emitter.Diagnose(loadout, token))
-                        diagnostics.Add(diagnostic);
+                    {
+                        lock(diagnostics)
+                        {
+                            diagnostics.Add(diagnostic);
+                        }
+                    }
                 }
                 catch (TaskCanceledException)
                 {


src/NexusMods.DataModel/Diagnostics/DiagnosticManager.cs
```diff
@@ -78,43 +80,41 @@ private async Task<Diagnostic[]> GetLoadoutDiagnostics(Loadout.ReadOnly loadout,
...
-                    var start = DateTimeOffset.UtcNow;
-
-                    try
-                    {
-                        var res = await emitter
-                            .Diagnose(loadout, cancellationToken)
-                            .ToListAsync(cancellationToken);
+            List<Diagnostic> diagnostics = [];
+            
+            await Parallel.ForEachAsync(diagnosticEmitters.OfType<ILoadoutDiagnosticEmitter>(), cancellationToken, async (emitter, token) =>
```
Al12rs, line 85: The caller of `GetLoadoutDiagnostics` (line 60) is using `SelectMany` which doesn't guarantee order, and isn't passing in cancellation tokens. 

We should switch it to `R3.Observable` with `ToObservable()` and then use `SelectAwait` with `AwaitOperation.Switch`, cancellation should hopefully work better here, but we can also use `AwaitOperation.ThrottleFirstLast` if that is not the case.

Sewer56: I had a look through this code as well.
This seems to check out.

To be specific, `SelectMany` on an observable returns items in the order that they complete in. If you're running on multiple threads, and a context switch happens (or `await GetLoadoutDiagnostics` stalls for longer for whatever reason), it is possible for the items to be emitted out of order; meaning we may receive diagnostics for an earlier loadout state.

In practice the chance of that happening though, is actually basically 0, especially given we've a Throttle right before that. However, it's better to be technically correct rather than correct simply due to external circumstance.

Al12rs (approved):
Current code should be good, I'll try to add the `SelectMany` -> `SelectAwait` in diagnostics 
 in a separate pr later today if I get the time.
 
 I'll merge this in the meantime and test if I can repro the deadlock issue with this in

PR #2580, opened 2025-01-30
Use custom fallback installer for collection installs

### LINKED ISSUE #2553: Install unrecognized Stardew Valley mods into `Mods` folder as a fallback for collections
Instead of using the advanced installer as a fallback, when we install a collection we want to install mods for Stardew Valley into the `Mods` folder as a fallback. This makes the app behave more similar to Vortex and prevents the installation process from being interrupted by the advanced installer as requested by @captainsandypants.

erri120:
Resolves #2553.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Games/AGame.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Games/AGame.cs
@@ -1,3 +1,4 @@
+using DynamicData.Kernel;
 using JetBrains.Annotations;
 using Microsoft.Extensions.DependencyInjection;
 using NexusMods.Abstractions.Diagnostics.Emitters;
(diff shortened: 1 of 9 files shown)


Al12rs:
Not sure about this, my idea was to add a SDV installer that only worked for collection installs or something like that.
Not something affecting every game. Having the advanced installer pop up was also a good indication of mod formats we were missing support for. 
Now they will just likely be installed incorrectly by default instead. 

Basically I don't like that this makes all collections for every game work as if everything went well, even if we literally have no installers for that game.

erri120:
I can make this opt-in for now.

Al12rs:
> I can make this opt-in for now.

Sounds good 👍

Al12rs (approved):

Issue #2341, opened 2024-12-05
Show the pre-install instructions for a Collection.


Pickysaurus:
Show install instructions the curators have written before starting the Collection install process. 

These instructions should also be accessible again during the installation. 

Data is here: https://github.com/Nexus-Mods/NexusMods.App/blob/3c79088073574a19e181ec699939af511a762062/src/Abstractions/NexusMods.Abstractions.Collections/Json/CollectionInfo.cs#L21-L22

Just use a Markdown renderer and a new tab and be done.

## Designs

https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=1643-21262&t=trI7PJedHiwe9jMT-1

![Image](https://github.com/user-attachments/assets/c5c46c3b-19de-4c6c-8c07-8938bde3196f)

LukeNexusMods:
These are instructions that are currently shown in Vortex when the collection is installing.

LukeNexusMods:
@captainsandypants This will need some thought as to how we display this.

LukeNexusMods:
This was discussed and the current information that is in the install instructions specifically relates to Vortex. If we implement showing these instructions at this stage it will confuse users that are trying the app. Sending users to the website for instructions could be a possibility.

LukeNexusMods:
After receiving feedback from collection curators I think we will need to implement a way for users to see their instructions within the app. The way that instructions are used vary greatly between curators. They also impact the success of the collection as they may contain instructions needed to get the collection working.  It is true that currently these instructions are based on vortex but this will change over time as the app supports more and more games. 
Our implementation of this should be extremely light touch to begin with so we can gather feedback about it. Users should be presented with the collection information/mod information as they are installing the collection and have a way to view it again if they need to.

PR #2600, opened 2025-02-04
Collection instruction styling

insomnious:
Styling of Additional Instructions tab on the Collection Download page.

[Figma link
](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=284-1336&m=dev)

Please note that this won't show the last minute tweaks from @captainsandypants

Part of https://github.com/Nexus-Mods/NexusMods.App/issues/2341.

### DIFF
--- a/src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadView.axaml
+++ b/src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadView.axaml
@@ -11,7 +11,7 @@
     xmlns:controls="clr-namespace:NexusMods.App.UI.Controls"
     xmlns:spinner="clr-namespace:NexusMods.App.UI.Controls.Spinner;assembly=NexusMods.App.UI"
     xmlns:navigation="clr-namespace:NexusMods.App.UI.Controls.Navigation"
-    mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="600"
+    mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="800"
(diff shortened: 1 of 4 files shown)


src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadView.axaml
```diff
@@ -199,34 +202,57 @@
...
-                </StackPanel>
+                <ScrollViewer>
+                    <StackPanel Orientation="Vertical">
+
+                        <Expander Header="Collection Instructions" x:Name="CollectionInstructionsExpander"
+                                  IsVisible="False" IsExpanded="True">
+                            <Grid>
+                                <Grid.ColumnDefinitions>
+                                    <ColumnDefinition Width="*" MaxWidth="1024" />
+                                </Grid.ColumnDefinitions>
```
erri120, line 216: This seems odd, what happened here?

insomnious: This is the easiest way for a MaxWidth to work like you'd expect it to. i.e. it shrinks when it doesn't need to be 1024 but never grows wider. If we you set MaxWidth (or any other Width or Height numerical value) on other controls, it then  completely ignores the HorizontalAlignment = Stretch

insomnious: I've used the same technique in a few different places (`DiagnosticListView.axaml` being one of them) and was a solution from online sources

erri120 (approved):

Issue #2598, opened 2025-02-04
Collections - Read only collections


LukeNexusMods:
As a creator
I want my collection to be read only
So that it simplifies support and keeps the collection together

This is to help simplify updates but also make sure that collections are kept together and are not editable. The library will need to know and show which mods are in collections and not show updates for those.  Collections are updated through downloading new revisions only. Mods within collections cannot be updated singularly.

Users can deactivate mods in a collection but not delete or update.

Eventually we will add the functionality for users to create a copy of a collection and be able to edit it when we start to implement collection creation.

UI will need to be created for what this looks like in the library. Design are aware and are working on it.

jonatino:
Really wish this could be overridden. Plenty of mod packs (especially) have mods that can be individually removed without breaking major things. Especially in cyberpunk since its very forgiving for mods. Not being to override this would suck.

Al12rs:
There will be a simple option to create an editable copy of the collection. The idea is to make aware users that they are responsible for the collection breaking if they change things

PR #2604, opened 2025-02-05
Add missing attributes and models for collections

erri120:
- Added missing library link for mods installed with predefined FOMOD choices
- Adds new model for replicated items
- Updates model for bundled items
- Adds new model for all items containing a reference to the original download and whether the item is required or optional

Required for #2598.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Collections/NexusCollectionItemLoadoutGroup.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Collections/NexusCollectionItemLoadoutGroup.cs
@@ -1,0 +1,30 @@
+using NexusMods.Abstractions.Loadouts;
+using NexusMods.Abstractions.NexusModsLibrary.Models;
+using NexusMods.MnemonicDB.Abstractions.Attributes;
+using NexusMods.MnemonicDB.Abstractions.Models;
+
+namespace NexusMods.Abstractions.Collections;
+
+[Include<LoadoutItemGroup>]
+public partial class NexusCollectionItemLoadoutGroup : IModelDefinition
+{
+    private const string Namespace = "NexusMods.Collections.NexusCollectionItemLoadoutGroup";
+
+    /// <summary>
+    /// Whether the item is required or optional.
+    /// </summary>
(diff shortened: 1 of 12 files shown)


src/Abstractions/NexusMods.Abstractions.Collections/NexusCollectionItemLoadoutGroup.cs
```diff
@@ -0,0 +1,30 @@
+using NexusMods.Abstractions.Loadouts;
+using NexusMods.Abstractions.NexusModsLibrary.Models;
+using NexusMods.MnemonicDB.Abstractions.Attributes;
+using NexusMods.MnemonicDB.Abstractions.Models;
+
+namespace NexusMods.Abstractions.Collections;
+
+[Include<LoadoutItemGroup>]
+public partial class NexusCollectionItemLoadoutGroup : IModelDefinition
```
Al12rs, line 9: Should this not be a `LibraryLinkedLoadoutItem`?

erri120: Bundled mods kinda screw with that since the files are inside the collection archive. While you could make a link to the collection archive, since it's just a library item, it doesn't exactly feel right. If we need to, we can change this later, as adding a reference to the archive is easily done via a migration.

Al12rs: I'm worried that a bunch of our queries in the Library are based on `LibraryLinkedLoadoutItems`, so they might skip over the collection items. 
It might be fine for some queries, like for updates, but not ok for Removal I think. 
Like checking if there are any linked items before being able to say it can be safely removed.

erri120: Every other collection item except bundled mods are linked loadout items, so that all works fine. It's just the bundled mods that are weird.

Al12rs: Ah, ok they are already linked. Cool, I thought we weren't setting that.

Al12rs (approved):

PR #2605, opened 2025-02-05
Provider user-agent

erri120:
Adds OS information to events.

### DIFF
--- a/src/NexusMods.Telemetry/EventSender.cs
+++ b/src/NexusMods.Telemetry/EventSender.cs
@@ -1,14 +1,17 @@
 using System.Buffers;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Net;
 using System.Net.Http.Headers;
 using System.Net.Mime;
 using System.Runtime.InteropServices;
+using System.Text;
 using Cysharp.Text;
 using DynamicData.Kernel;
 using JetBrains.Annotations;
 using NexusMods.Abstractions.NexusWebApi;
 using NexusMods.Abstractions.NexusWebApi.Types;
+using NexusMods.Paths;
 
 namespace NexusMods.Telemetry;
 
@@ -28,6 +31,8 @@ internal class EventSender : IEventSender
     private readonly ILoginManager _loginManager;
     private readonly HttpClient _httpClient;
     private readonly ArrayBufferWriter<byte> _writer;
+
+    private static readonly byte[] EncodedUserAgent = CreateUserAgent();
 
     public EventSender(ILoginManager loginManager, HttpClient httpClient)
     {
@@ -134,11 +139,8 @@ internal class EventSender : IEventSender
         sb.Append("&rec=1"); // Required for tracking, must be set to one
         sb.Append("&apiv=1"); // API version to use, currently always 1
 
-        // NOTE(erri120): The device detector they have really doesn't like us
-        // https://github.com/matomo-org/device-detector
-        // test: https://lx3rzx16x9.csb.app
-        // sb.Append("&ua="); // User agent
-        // sb.Append(ApplicationConstants.UserAgent);
+        sb.Append("&ua="); // User agent
+        sb.Append(EncodedUserAgent);
 
         sb.Append("&send_image=0"); // Matomo will respond with an HTTP 204 response code instead of a GIF image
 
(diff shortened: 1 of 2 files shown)


src/NexusMods.Telemetry/EventSender.cs
```diff
@@ -240,27 +238,45 @@ private static ReadOnlySpan<ValueTuple<EventDefinition, EventMetadata, ulong>> P
...
+            return "Mozilla/5.0 (X11; Linux x86_64)";
         }
 
-        // https://lx3rzx16x9.csb.app/#Windows%20NT%2010.0
+        // https://devicedetector.lw1.at/Mozilla%2F5.0%20(Windows%20NT%2010.0;%20Win64;%20x64)
         if (platform == OSPlatform.Windows)
         {
-            return $"Windows NT {osVersion.Version.ToString(fieldCount: 2)}";
+            Debug.Assert(osVersion.Version.Major == 10);
+            Debug.Assert(architecture == Architecture.X64);
```
Sewer56, line 272: Is there any specific reason you'd be asserting a *specific* NT version number and arch?

For NT version, you probably want to check if Major is at least 10, since that's what the .NET Runtime (and by extension we) support. However, although Win10 and Win11 do both use 10.0 version number, that could still technically one day change.  (Note: Version number should really be 6.4, but they went with 10.0 for marketing reasons, bleh.) You don't want to unnecessarily trip someone building the app up in the future, be it a person building an old App version for whatever reason, or being on a new upcoming OS.

As for the Arch, although we don't test it, it's expected that there will be users on aarch64 (ARM64); we may not be testing for it right now, but the users do and will exist. And they'll be running games via the built in translation layer. Our stuff should generally 'just work' on ARM, outside anything missing in built-in OS emulation; same way we have a mostly working macOS build on ARM. More specifically, on the Windows end, you have the recent Snapdragon X Elite chips on aarch64 (ARM64).

I don't want to risk the App terminating in the case one of these users receives a Debug build for Testing/Debugging purposes. Instead, we should instead put up a info modal/box/banner/etc. if the user is using a configuration that we don't get to test.

Sewer56 (approved):
Rest is fine, just the note above, needs a small change.
It's a trivial change, so not worth blocking merging for.

Please do avoid asserting here; if the concern is about checking for 'supported' platforms, better do more at the App level. This runs in Telemetry code, which is optional, so you'll have uses on 'unsupported' targets skip the assert by not enabling telemetry, for instance.

Ideally, just throw that info in the logs and/or the UI, as mentioned above.
Go patch up however you see fit and merge.

Sewer56:
(Do read the note above, if you haven't)

src/NexusMods.Telemetry/EventSender.cs
erri120: We only support Windows 10 and newer on x64. If we ever support something else, we can update this code.

Sewer56: This doesn't address the fact that this logic only runs in *optional* telemetry logic, or that the message the user would receive would just be a nasty stack trace.

I really, really don't want to be coming back to this unnecessarily in the future; where someone reports an issue and we also have to change this for test builds just cause.

In addition, there is no arch restriction on other targets here, despite us not 'supporting' aarch64 Linux or Intel macOS.

Sewer56: I care very little for what the end users are actually running. Even if it's not a hard limit, adding arbitrary restrictions on unsupported hardware for arbitrary reasons is quite literally against the spirit of modding itself, which is an open and free discipline we are fundamentally promoting.

You can also just report any misc platform we don't exactly know as unsupported, or as the exact user agent it should be, depends on what Matomo supports.

For all we know, those Snapdragon chips may be selling well, or their successors will. We'll probably want to know that, and adjust our strategy accordingly as needed.

erri120: I think you miss the point of the code, the strings we're creating are fake. I manually created it to satisfy the arbitrary parsing rules of matomo. The only information in those strings that are of importance is the OS name: `Linux`, `Windows`, and `Mac`.

Sewer56: Exactly, the asserts here are completely arbitrary and not related in any way at all to what we're trying to track here. Per your own comment in the source code.

Regardless of whether the strings are fake, a user running on 'unsupported' hardware will get a nasty stack trace with 3 buttons, without any sort of other context whatsoever, should they boot a debug build.

Suppose that user happens to be one that reports an issue, and we send them a locally compiled test build to see if it resolves their issue. Then they'll get a very nasty error dialog, or have the App outright die. What does that gain us? It doesn't gain us anything, really. I objectively don't care about the nature of the configuration, for all I know it could be an issue completely unrelated to that.

Even if you want to have such a check, this sort of check does not belong in the telemetry code, but rather more at init.

PR #2606, opened 2025-02-05
Mod updates UI Phase One: Various Fixes

Sewer56:
This is a small number of fixes I've done yesterday related to the Mod Updates UI.
There isn't much code, but it's very much all over the place, since it's all refinements.

- Update button in the UI is now localizable.
    - Spent way longer here than I would have expected, I didn't think the (English) string would be put in the code raw, especially since the code being rewritten was already localized.
    - So I had a funny moment of being really confused why changing resources didn't change translation, haha.

- Changed the 'Update' button text to reflect the text for Phase One.
    - The text change will be reverted for Phase Two to its intended final form.

- Ignore files in ReadOnly collections from being checked for updates.
    - I would much rather work on a whitelist system rather than a blacklist one, but ~~writing it inverse with blacklist was easier in this case.~~
    - So basically we filter out any mod that's from a read-only collection from receiving update notifications; since those mods should be updated by updating the collection revision itself.
    - This is functionally the inverse of 'whitelist every mod that's not installed, or installed in a read-write collection' , this can be observed if you draw a venn diagram.

- Entering the library now does an automatic update check (confirmed OK with Laurence).
- Moved the Library toolbar contents to the left of the bar, to match the design.
    - This is the only toolbar in the App which is left aligned.
    - The alignment of our toolbars is standardized via styles. Because this is the odd one out, I put the override locally in the view itself; leaving a note about it in the xaml.

- Added the 'Premium Download Experience', that is, automatically downloading files when Premium when starting an update.
- Fixed Download Link for Update, so when Free user is sent to website, it now downloads using nxm handler.
- Fixed Mod Page row not using the most recent update info.
    - If a mod page has multiple files with updates, the mod page row is supposed to show the details of the latest file.
    - This may not always have been true up till now; depending on which file was the latest one.
    - I put the link to the details of that in the code.

This is all mostly refinement work, very little code change but all over the place as mentioned above.


erri120 (approved):

src/NexusMods.App.UI/Pages/LibraryPage/LibraryViewModel.cs
```diff
@@ -224,19 +228,35 @@ public LibraryViewModel(
                 .Bind(out _collections)
                 .Subscribe()
                 .AddTo(disposables);
+
+            // Auto check updates on entering library.
+            RefreshUpdatesCommand.Execute(Unit.Default);
```
Al12rs, line 233: What is the cache on this? I don't want us to send network requests every time the library is open

Sewer56: There isn't at the moment.

I was considering adding a 1 minute one but given that I get razed for being overkill typically I didn't add one to the update service.

I don't consider the user rapidly switching between library and loadout on left menu to be a concern.

I do consider them mashing refresh to be a concern, but due to the above I've 'kept it simple'.

src/Networking/NexusMods.Networking.NexusWebApi/ModUpdateService.cs
```diff
@@ -111,10 +176,36 @@ public void NotifyForUpdates()
...
+{
+    /// <summary>
+    /// Returns the newest file on the mod page.
+    /// </summary>
+    /// <remarks>
+    /// Note(sewer): We by definition don't create this struct with empty arrays,
+    /// so the first element is always present. 
+    /// </remarks>
+    public NexusModsFileMetadata.ReadOnly NewestFile() => Files[0];
+}
```
Al12rs, line 211: I would add Debug assert in the constructor to ensure Files is not empty

Sewer56: I don't think it's needed here actually. If the assumption is false, that will trigger an unhandled exception which will propagate to the top level error handler, that being the error dialog.

So regardless of whether there is an assert or not, the user would be notified.

src/NexusMods.App.UI/Pages/LibraryPage/LibraryViewModel.cs
Al12rs: > I don't consider the user rapidly switching between library and loadout on left menu to be a concern. No sane user would be doing that

All users will be doing that. We either add a 5 minute timeout or no automatic update.

Also this throws if you are logged out. Exception is captured at some point I think, but we shouldn't be trying.

Sewer56: I will add a rate limit on the service in the upcoming PR.

Sewer56: I've still plenty of bugs to fix around the functionality in general, I'll put this on my shortlist and handle it in follow up.

src/Networking/NexusMods.Networking.NexusWebApi/ModUpdateService.cs
```diff
@@ -65,18 +67,73 @@ await RunUpdateCheck.UpdateModFilesForOutdatedPages(
...
+        // not those which are [read-only].
+        //
+        // Although it is easier to code it, so we look for all mods, and take
+        // away the [read-only] ones, whitelisting [read-write] ones instead
+        // will help us avoid risks.
+        //
+        // It's better to have a false negative (can't update that should be updatable)
+        // than a false positive (can update that shouldn't be updatable); because
+        // the former has no potential negative side effects, while the latter
+        // can potentially cause some level of damage.
```
Al12rs, line 94: Talked with Design, show updates on all items.
We don't want to miss updates for relevant mods like smapi because it is used in one collection.
We don't have replace that would break stuff, so checking all is fine.

Al12rs (changes requested):

src/Networking/NexusMods.Networking.NexusWebApi/ModUpdateService.cs
Sewer56: Reverted for now

Al12rs (approved):

PR #2607, opened 2025-02-05
Add ITransactionalMigration.cs

halgari:
For simpler migrations we can use this vs the scanning migrations


halgari:
@erri120 feel free to merge/copy this code, but here's how I was planning on handling these simpler migrations

PR #2610, opened 2025-02-06
Fix adding duplicate mod pages

### LINKED ISSUE #2608: Cannot complete a collection download as a specific mod is not recognised.
# Bug Report

## Summary
As a free user, I cannot complete the install of [Aesthetic Valley | Witchcore](https://next.nexusmods.com/stardewvalley/collections/g14kxi) for Stardew Valley as the app won't recognise [Gwen's Paths - AT Version](https://www.nexusmods.com/stardewvalley/mods/8683?tab=files&file_id=78396&nmm=1), even when I have added it to my library. 

This issue does not occur when downloading as a Premium user. 

## Steps to reproduce
1. Start with a clean app installation.
2. Log in as a free user. (NXMMember is our test account)
3. Manage Stardew Valley
4. Download [Aesthetic Valley | Witchcore](https://next.nexusmods.com/stardewvalley/collections/g14kxi)
5. From the Collection install page, attempt to install [Gwen's Paths - AT Version](https://www.nexusmods.com/stardewvalley/mods/8683?tab=files&file_id=78396&nmm=1)
6. See that it is not recognised.
7. Install any other mod in the collection.
8. See that it works.

The mod is properly recognised as part of the collection. 

## Other information

[nexusmods.app.main.current.log](https://github.com/user-attachments/files/18686323/nexusmods.app.main.current.log)
[nexusmods.app.slim.current.log](https://github.com/user-attachments/files/18686322/nexusmods.app.slim.current.log)

erri120:
Fixes #2608 but not retroactively.

We use `UidForMod` as a key to figure out if a mod page has already been added or not. However, the struct never had an equality method implemented, and was thus useless as a key.

The bug #2608 surfaced multiple bugs in our current code:

1) Free users download via NXM links, we never check if the file has already been downloaded. Premium users don't run into this issue because we download files for them and use checks to disable the download button.
2) When adding a collection, we add any missing mod page and file metadata entities. However, if a collection contains two files from the same mod page, and that mod pages wasn't in the DB already, we'd create two mod page entities, one for each file metadata entity.

Adds a migration to removes all duplicate file and mod page metadata entities.

### DIFF
--- a/tests/NexusMods.DataModel.SchemaVersions.Tests/LegacyDatabaseSupportTests.TestDatabase_name=Issue-2608.rocksdb.zip.verified.txt
+++ b/tests/NexusMods.DataModel.SchemaVersions.Tests/LegacyDatabaseSupportTests.TestDatabase_name=Issue-2608.rocksdb.zip.verified.txt
@@ -1,7 +1,7 @@
 ﻿{
   Name: Issue-2608.rocksdb.zip,
   OldId: 1,
-  NewId: 1,
+  NewId: 3,
   Loadouts: 1,
   LoadoutItemGroups: 2,
   Files: 3788,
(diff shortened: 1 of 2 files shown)


Sewer56 (approved):
Pretty simple, LGTM

Al12rs (approved):

PR #2617, opened 2025-02-10
Add locked component for loadout items

erri120:
Part of #2598.

Shows a lock icon for required mods of collections in the loadout views for now.

### DIFF
--- a/src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutComponents.cs
+++ b/src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutComponents.cs
@@ -1,4 +1,5 @@
 using DynamicData;
+using JetBrains.Annotations;
 using NexusMods.Abstractions.Loadouts;
 using NexusMods.Abstractions.UI;
 using NexusMods.Abstractions.UI.Extensions;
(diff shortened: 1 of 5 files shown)


src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutComponents.cs
```diff
@@ -85,21 +86,28 @@ protected override void Dispose(bool disposing)
             base.Dispose(disposing);
         }
     }
+
+    public sealed class Locked : ReactiveR3Object, IItemModelComponent<Locked>, IComparable<Locked>
```
Al12rs, line 90: I would rename to `LockedComponent` for clarity. 
Same thing with `IsEnabled` -> `IsEnabledComponent` tbh

erri120: These are child classes of `LoadoutComponents`. If you want to access them, you always have to do `LoadoutComponents.IsEnabled`. Adding the `Component` suffix is somewhat redundant as you'd have `LoadoutComponents.IsEnabledComponent`.

Al12rs: It would still be clearer, it isn't blowing the name out

Al12rs:
There seems to be the same scrolling issue where the Toggle is first shown and then replaced with the lock icon when scrolling up and down and revealing new items, similar to the Actions buttons issue we saw for the collection download page

Al12rs:
> Thanks to observables the value to show the locked icon will only be set after activation.

I guess the solution would be to not have the toggle component if it is a locked item, so it doesn't show it while observable is receiving the value.
But you are using IsEnabled component for sorting so eh

PR #2618, opened 2025-02-10
Add observable tracker page

erri120:
For devs only. The page taps into the R3 observable tracker system and allows us to visualize all active observables.


Al12rs (approved):
Nice one, I think we might not be disposing some ReactiveCommand CanActivate subscriptions

Al12rs:
Try to rebase PR on latest main to fix tests

halgari:
I'd love to have more of these sorts of pages over time. Is this behind a "dev mode" setting?

erri120:
> I'd love to have more of these sorts of pages over time. Is this behind a "dev mode" setting?

Only available in debug builds atm.

PR #2621, opened 2025-02-11
Updated states for loadout items

erri120:
Part of #2598.

- Adds the "Collections"-column for loadout items
- Adds the `IsLocked` state to `LoadoutComponents.IsEnabled`
- Adds a new component for the disabled state of the parent collection

I'll do the more difficult state stuff in another PR.


Al12rs (changes requested):
The behaviour of mod page entries isn't correct in this moment (shows locked icon if one of the children in locked, even if other children are not). 
But we can fix that in future PRs

Al12rs (approved):
Misclicked

PR #2622, opened 2025-02-11
TreeDataGrid styles update

insomnious:
- Column headers capitalized
- Mod update styles
- Updated outline Expander icons in
- TreeDataGrid styles refactored into separate files (this will be ongoing but a good enough start for now)
- Moved various Collection button text to Language resource
- Hidden Switch View button


erri120 (changes requested):
The uppercasing of column headers should be done in styles with a custom value converter:

https://github.com/Nexus-Mods/NexusMods.App/blob/c66e55222fc648288eab8390dd29551821bca133/src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/TreeDataGrid/TreeDataGridStyles.axaml#L254

erri120 (approved):

PR #2629, opened 2025-02-12
Collection mod rules

### LINKED ISSUE #2368: Collections: Apply rules for mod asset conflict priority (mod level and file level)
## Outline
Collections contain rules between mods and between files that determine the winners in case of file conflicts.
The app needs to interpret and apply these rules to obtain a final result that is consistent with the original collection.

## Example:
[Stardew Very Expanded](https://next.nexusmods.com/stardewvalley/collections/tckf0m?tab=mods) contains the [Stardew Very Configured mod](https://www.nexusmods.com/stardewvalley/mods/20870) containing overriders for configuration files from other mods in the collection. 
SDV Very Configured needs to win conflicts with the other mods, to replace the configuration files, otherwise the output of the installed collection will not be the desired one.

### LINKED ISSUE #2627: Stardew Valley - SMAPI has two different names when installed via a collection.
# Bug Report

## Summary
SMAPI can have inconsistent naming when installed via a collection vs from the Library. 

## Steps to reproduce
1, Install https://next.nexusmods.com/stardewvalley/collections/g14kxi
2. Go to the Library, find SMAPI, and select the row.
3. Click "Add" on the toolbar to add SMAPI to "My Mods".
4. Go to "All" and find the SMAPI row, expand it to see both SMAPIs from the collection and My Mods.
5. See the two different names.

Image

SMAPI files have a consistent name.

### LINKED ISSUE #2630: Bundled Mods are sometimes installed in the wrong location
# Bug Report

## Summary
I installed [Aesthetic Valley | Witchcore](https://next.nexusmods.com/stardewvalley/collections/g14kxi/mods). One of the bundled mods is `[CP] Medieval Aurora Refurbished` and the contents of that mod are being installed into the game root `Stardew Valley/` instead of `Mods/`. This happens to be a bundled mod, so perhaps that is why no other mods have this same issue. 

## Steps to reproduce
Install the above collection, look at `preview changes`, sit and ponder why one mod is different from the rest.

The mod should be in `Mods/`

erri120:
- Fixes #2368
- Fixes #2627 
- Fixes #2630

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Collections/Json/ModRule.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Collections/Json/ModRule.cs
@@ -31,7 +31,7 @@ public class ModRule
     /// Other item.
     /// </summary>
     [JsonPropertyName("reference")]
-    public required VortexModReference Reference { get; init; }
+    public required VortexModReference Other { get; init; }
 }
 
 /// <summary>
--- a/src/Abstractions/NexusMods.Abstractions.Collections/Json/ModRule.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Collections/Json/ModRule.cs
@@ -1,7 +1,7 @@
 using System.Diagnostics;
 using System.Text.Json.Serialization;
 using JetBrains.Annotations;
-using NexusMods.Abstractions.Hashes;
+using NexusMods.Abstractions.MnemonicDB.Attributes;
 using NexusMods.Paths;
 
 namespace NexusMods.Abstractions.Collections.Json;
@@ -45,7 +45,7 @@ public class VortexModReference
 #region Vortex IReference
 
     [JsonPropertyName("fileMD5")]
-    public Md5 FileMD5 { get; init; }
+    public Md5HashValue FileMD5 { get; init; }
 
     [JsonPropertyName("fileSize")]
     public Size FileSize { get; init; }
@@ -57,11 +57,17 @@ public class VortexModReference
     /// See this shit: https://github.com/Nexus-Mods/Vortex/blob/1bc2a0bca27353df617f5a0b0f331cf9d23eea9c/src/extensions/mod_management/util/testModReference.ts#L102-L125
     /// </remarks>
     [JsonPropertyName("versionMatch")]
-    public string? VersionMatch { get; init; }
+    public string VersionMatch { get; init; } = string.Empty;
 
     [JsonPropertyName("logicalFileName")]
     public string? LogicalFileName { get; init; }
 
+    /// <summary>
+    /// Either an exact match against a mod name or a glob match against the archive name without the file extension.
+    /// </summary>
+    /// <remarks>
+    /// https://github.com/Nexus-Mods/Vortex/blob/1bc2a0bca27353df617f5a0b0f331cf9d23eea9c/src/extensions/mod_management/util/testModReference.ts#L226-L240
+    /// </remarks>
     [JsonPropertyName("fileExpression")]
     public string? FileExpression { get; init; }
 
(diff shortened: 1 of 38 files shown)


src/Abstractions/NexusMods.Abstractions.Collections/Json/ModRule.cs
```diff
@@ -0,0 +1,138 @@
...
+    /// </summary>
+    [JsonPropertyName("idHint")]
+    public string? IdHint { get; init; }
+
+    /// <summary>
+    /// I have no fucking idea why this and <see cref="FileMD5"/> exists on the same object.
+    /// Vortex also does some really questionable stuff: https://github.com/Nexus-Mods/Vortex/blob/1bc2a0bca27353df617f5a0b0f331cf9d23eea9c/src/extensions/mod_management/util/dependencies.ts#L37-L47
+    /// This is likely never included in any JSON files.
+    /// </summary>
+    public Md5HashValue MD5Hint { get; init; }
```
Al12rs, line 90: Should probably improve the wording of these comments a bit 😅

erri120: I think the adequately describe the experience of working this DTO. I also only added them for completion, we'll never end up using them.

src/NexusMods.DataModel.SchemaVersions/Migrations/_0005_MD5Hashes.cs
```diff
@@ -0,0 +1,72 @@
...
+            .Where(entity =>
+            {
+                foreach (var datom in entity.IndexSegment)
+                {
+                    if (_attributesToRemove.Contains(datom.A)) return true;
+                }
+
+                return true;
+            })
+            .ToArray();
```
Al12rs, line 49: Should we just make the mapping here, like a dictionary of LibraryFile ids -> (md5, localFileId to retract)?

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -895,13 +897,76 @@ private bool ShouldWinWrapper(IDb db, in GamePath path, in SyncNodePart existing
         // Return the newer one
         return ShouldWin(path, existingLoadoutItem, replacementLoadoutItem);
     }
-    
+
+    /// <summary>
+    /// Returns true if <paramref name="a"/> should win over <paramref name="b"/> based
+    /// on rules defined in a collection.
+    /// </summary>
+    protected Optional<bool> ShouldOtherWinWithCollectionRules(LoadoutItem.ReadOnly a, LoadoutItem.ReadOnly b)
```
Al12rs, line 905: Comment and function name don't seem match. `ShouldWin` returns true if the second parameter should win, while this function does the opposite while using same naming

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -895,13 +897,76 @@ private bool ShouldWinWrapper(IDb db, in GamePath path, in SyncNodePart existing
...
+            (CollectionDownloadRules.Other, downloadB)
+        );
+
+        foreach (var id in ids)
+        {
+            var rule = CollectionDownloadRules.Load(db, id);
+            if (!rule.IsValid()) continue;
+
+            // `source` goes before `other`, meaning `a` doesn't win over `b`
+            if (rule.RuleType == CollectionDownloadRuleType.Before) return false;
```
Al12rs, line 940: Should return true, since B (other) wins

Al12rs:
```suggestion
            // `source` goes before `other`, meaning `b` wins over `a`
            if (rule.RuleType == CollectionDownloadRuleType.Before) return true;
```

Al12rs (approved):

PR #2636, opened 2025-02-13
Fixes steam game indexing, better performance, etc

halgari:
I went to index steam files for Cyberpunk and ended up finding a bunch of small issues. So this does the following:

* Fixes the game hashes db builder so it no longer silently failswhen the version definitions are invalid (like referencing a manifest we've never seen. Now it ignores those versions and logs info about them.
* When critical errors happen we now return -1 so that gihub actions won't continue after a failed run
* Added support for `SteamCache` endpoints and defaults to those
* When we get a failure from a Steam server we mark the server as failed and try the next
* Added support for prefetching in the chunked stream reader. Some steam servers have very bursty responses, this evens it out a bit. 
* Added some parallelism to the steam indexing code
* Added support for logging into Steam via credentials stored in environment variables, useful for github actions

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.IO/ChunkedStreams/ChunkedStream.cs
+++ b/src/Abstractions/NexusMods.Abstractions.IO/ChunkedStreams/ChunkedStream.cs
@@ -1,4 +1,5 @@
 ﻿using System.Buffers;
+using System.Collections.Concurrent;
 using System.Diagnostics;
 
 namespace NexusMods.Abstractions.IO.ChunkedStreams;
@@ -13,16 +14,19 @@ public class ChunkedStream<T> : Stream where T : IChunkedStreamSource
     private readonly T _source;
     private LightweightLRUCache<ulong, IMemoryOwner<byte>> _cache;
     private readonly MemoryPool<byte> _pool;
+    private Dictionary<ulong, Task<IMemoryOwner<byte>>> _preloadingTasks = new();
+    private readonly int _preLoad;
 
     /// <summary>
     /// Main constructor, creates a new Chunked stream from the given source, and with a LRU cache of the given size
     /// </summary>
-    public ChunkedStream(T source, int capacity = 16)
+    public ChunkedStream(T source, int capacity = 16, int preLoad = 0)
     {
         _position = 0;
         _source = source;
         _pool = MemoryPool<byte>.Shared;
         _cache = new LightweightLRUCache<ulong, IMemoryOwner<byte>>(capacity);
+        _preLoad = preLoad;
     }
 
     /// <inheritdoc />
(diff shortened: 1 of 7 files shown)


src/Abstractions/NexusMods.Abstractions.IO/ChunkedStreams/ChunkedStream.cs
```diff
@@ -92,16 +96,44 @@ public override async ValueTask<int> ReadAsync(Memory<byte> buffer,
...
+            var chunk = await task;
+            _cache.Add(index, chunk);
+            _preloadingTasks.Remove(index);
+            return chunk.Memory;
+        }
+
+        var memoryOwner = await AllocateAndGetChunk(index, token);
+        _cache.Add(index, memoryOwner);
+        return memoryOwner.Memory;
+    }
```
erri120, line 129: The idea behind `ChunkedStream` is to fetch data in chunks, prefetching chunks is like fetching chunks in chunks. Saying I want to load data in 4KB chunks, but prefetch 4 chunks, is the same as saying I want to load data in (1 + 4) * 4KB chunks. At least that is my understanding of the idea.

The implementation of this prefetching also has issues, this is never prefetching anything. If we set `_preLoad = 2` and we're requesting chunk `index = 0` then this should load 3 chunks with indices `0`, `1`, and `2`. Instead, if we request chunk `index = 0`, we create tasks for getting chunks `1` and `2` but never run them until we request those chunks. In which case we're not prefetchin, we're just complicating the normal loading of data.

src/NexusMods.App/Startup.cs
```diff
@@ -102,7 +102,7 @@ public static AppBuilder BuildAvaloniaApp(IServiceProvider serviceProvider)
             .LogToTrace()
             .UseR3()
             .UseReactiveUI()
-            .With(new SkiaOptions { UseOpacitySaveLayer = true });
+            .With(new SkiaOptions { UseOpacitySaveLayer = false });
```
erri120, line 105: What happened here? As mentioned in the comment below by @insomnious, we need this setting to disabled/fade SVG icons.

erri120 (changes requested):

src/Abstractions/NexusMods.Abstractions.IO/ChunkedStreams/ChunkedStream.cs
halgari: This "... this is never prefetching anything..." this is incorrect. If you run this code in a console you will see 0-9 printed on the screen then the text "waiting for tasks" then the ending lines.

```
List<Task> tasks = new();

for (int i = 0; i < 10; i++)
{
    tasks.Add(MyFunc(i));
}

Console.WriteLine("Waiting for tasks to complete");

await Task.WhenAll(tasks);

return;


async Task MyFunc(int i)
{
    Console.WriteLine($"Start {i}");
    await Task.Delay(1000);
    Console.WriteLine($"End {i}");
}
```


 Meaning that calling an async function without awaiting starts the async process immediately. This is also why we get the warning in the IDE all the time of "there is no await in this method so it will run on the context of the calling thread". Because everything up to the first call to await is run on the caller thread at the time the method is called.

halgari:
```
The idea behind ChunkedStream is to fetch data in chunks, prefetching chunks is like fetching chunks in chunks. Saying I want to load data in 4KB chunks, but prefetch 4 chunks, is the same as saying I want to load data in (1 + 4) * 4KB chunks. At least that is my understanding of the idea.
```

If the prefetch count is ever anything other than 0, then that means when ever we load a given chunk we will ensure that that chunk and the next `prefetch` chunks have started loading. For a disk stream, or sources with large chunks this doesn't make a lot of sense but steam uses 1MB chunks, so large files are quite slow when we have to load them sequentially. 

This code is single threaded, so if we're reading forward through a file we never load the same chunk multiple times, instead we look at any previously existing tasks before prefetching more.

Sewer56: My email is popping off here.

In any case, the highlighted code makes sense to me. No explanation needed in my case, the purpose was immediately clear.

The whole point here is to schedule an asynchronous I/O task in the background while letting the main thread do its thing. 

You basically want the data to already be 'ready' by next call, otherwise you'll be stuck waiting for I/O. When you want to do CPU bound computation, like hashing, you don't want to be stuck waiting until the next chunk loads. This is specifically referred to as 'prefetching', in the caching world.

----

I think the confusion might have come from understanding of Async in C# here. While it is true that in some languages asynchronous routines are lazy (e.g. in Rust), in C#, they are not. Tasks start as soon as you fire the method.

halgari: Ah, that would make sense if Rust is lazy. I've only ever used async in Clojure, C# and JS, and those are all eager (I think JS is, I've never checked)

erri120: Right, I missed the task getting started. What worries me now is starting 32 tasks at once to fetch 1MB chunks. That seems quite aggressive, but we can tune that later.

erri120 (approved):

PR #2637, opened 2025-02-13
Fix apply preview and exceptions when switching loadouts

Al12rs:
Preview view was showing wrong data, now it should show all the changes that will happen to the last analyzed disk state if the loadout is applied. 

This means that if something is on disk but not in the loadout it will appear as a normal entry in the preview, since it will not change. The preview will only show how the game folder will be affected by Apply, not how the loadout will be affected by apply.

I added a different method to determine whether Synchronize should be needed, since the previous one was ignoring a number of cases that still required a synchronize.

Also fixed exceptions that would happen due to missing LastAppliedLoadout data after switching lodouts.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -535,8 +535,10 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
         return loadout.Rebase(result.Db);
     }
 
-    /// <inheritdoc />
-    public async Task RunActions(Dictionary<GamePath, SyncNode> syncTree, GameInstallation gameInstallation)
+    /// <summary>
+    /// Alternative to <see cref="RunActions"/> that ignores changes and optionally clears the last sync loadout metadata
+    /// </summary>
+    public async Task RunActions(Dictionary<GamePath, SyncNode> syncTree, GameInstallation gameInstallation, bool clearLastSyncedLoadout = true)
     {
         using var _ = await _lock.LockAsync();
         using var tx = Connection.BeginTransaction();
@@ -591,7 +593,7 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
             }
         }
 
-        if (gameMetadata.Contains(GameInstallMetadata.LastSyncedLoadout))
+        if (clearLastSyncedLoadout && gameMetadata.Contains(GameInstallMetadata.LastSyncedLoadout))
         {
             tx.Retract(gameMetadataId, GameInstallMetadata.LastSyncedLoadout, (EntityId)gameMetadata.LastSyncedLoadout);
             tx.Retract(gameMetadataId, GameInstallMetadata.LastSyncedLoadoutTransaction, (EntityId)gameMetadata.LastSyncedLoadoutTransaction);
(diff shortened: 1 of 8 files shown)


src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1471,7 +1502,7 @@ public async Task ResetToOriginalGameState(GameInstallation installation, string
         ProcessSyncTree(desiredState);
 
         // Run the groupings
-        await RunActions(desiredState, installation);
+        await RunActions(desiredState, installation, clearLastSyncedLoadout);
```
halgari, line 1505: I'd like some documentation and/or tests on why this change is required. It adds another dimetion to this code that I think may make it harder to debug in the long run. Why can't always clear or not clear the state? Also if we've reset the game state, wouldn't the last applied state no longer be valid?

halgari (changes requested):

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
Al12rs: The problem is that this code is currently being used to switch loadouts, in which case we don't want to reset the state.

The best solution would probably be to have a dedicated optimized code path to switch loadouts, rather than resetting the game folder to default and then applying the new loadout, but I think that would be outside the scope of this.

In the case of a loadout switch we don't want to delete the `LastSyncedLoadoutTransaction` and `LastSyncedLoadout`, as they will still be needed to determine the loadout Synchronized state (should apply).

I think the current loadout switching tests don't catch this since the errors happen in the `ShouldApply` check, that is only performed if there is at least a subscription active on `StatusForGame()` observable. 

I'll try adding that to the tests.

halgari:
Why don't we want to reset the states? If we're swapping loadouts (a
relatively rare occurrence) we're doing exactly what this code says:
deactivate the old account and reactivate the new one. What I'm trying to
avoid here is introducing a difference between applying a loadout for the
first time and swapping a loadout. The fewer unique states we have the more
consistent testing should be.

Al12rs:
> Why don't we want to reset the states? If we're swapping loadouts (a relatively rare occurrence) we're doing exactly what this code says: deactivate the old account and reactivate the new one. What I'm trying to avoid here is introducing a difference between applying a loadout for the first time and swapping a loadout. The fewer unique states we have the more consistent testing should be.

The code that checks the loadout status (`ShouldSynchronize()` and `LoadoutToDiskDiff()`) needs to know what the previous `DiskState` was. Without that information the results become inaccurate.

After a game is managed, that code assumes we always have a loadout applied, up until we unmanage the game. 
The only case where resetting to default game state should happen is either as an intermediate step of switching loadouts, or when unmanaging the game.

There should be no need to clear the last `LastSyncedLoadout` and the `LastSyncedLoadoutTransaction` when switching loadouts. That is still information we need for the purposes of computing the synchronize status of loadouts and diffs.


I updated the loadout switching test to more accurately follow what the app does when switching loadouts and I added the subscription to the state of the loadouts to listen to any errors that happen there.

That test now catches the exceptions that this PR fixes.
I also found that the  `StubbedFileHasherService` was regenerating the game indexing data every call, which was causing issues.

Al12rs:
I changed the loadout sync state functions to fallback to the last indexed state if the `LastSynched` state is missing, which is going to be incorrect, but the incorrect data should be replaced by more accurate data once the loadout switch fully completes.

This way deactivating a loadout always does the same thing, and no exceptions are thrown during sync state checks

PR #2640, opened 2025-02-13
Added: Suppress that Annoying DBus Error on Some Wayland Compositors [for now]

Sewer56:
Just something I added on my own local machine last night, might aswell chuck it in here.
This suppresses the annoying DBus error that some Wayland compositors get while displaying tooltips in the UI.

Since we're releasing to a wider audience we can't really have constant error dialogs be thrown at end users; it's rough enough when testing, and probably much worse in actual use.

We need to report this upstream, if it's not reported already.
I left a note with the suppression in the source code.

### DIFF
--- a/src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
+++ b/src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
@@ -19,6 +19,7 @@ using NexusMods.App.UI.Overlays.Updater;
 using NexusMods.App.UI.WorkspaceSystem;
 using ReactiveUI;
 using ReactiveUI.Fody.Helpers;
+using Tmds.DBus.Protocol;
 
 namespace NexusMods.App.UI.Windows;
 
@@ -118,6 +119,9 @@ public class MainWindowViewModel : AViewModel<IMainWindowViewModel>, IMainWindow
         return source.Exceptions
             .Subscribe(msg =>
                 {
+                    if (!ShouldShowError(msg))
+                        return;
+                    
                     var title = "Unhandled Exception";
                     var description = msg.Message;
                     string? details = null;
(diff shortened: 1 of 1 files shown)


src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
```diff
@@ -19,6 +19,7 @@
 using NexusMods.App.UI.WorkspaceSystem;
 using ReactiveUI;
 using ReactiveUI.Fody.Helpers;
+using Tmds.DBus.Protocol;
 
 namespace NexusMods.App.UI.Windows;
```
halgari, line 24: This shouldn't be in .Windows. Instead we want it in .Linux, and if we have to pull in a `DBus` dependency for this as well, we probably want it in a separate project (like we discussed with the Linux/Windows/OSX split

erri120: This is `NexusMods.App.UI.Windows`, the `Windows` is not referring to the OS but to the concept of windows, this is the `MainWindowViewModel.cs` file.

src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
```diff
@@ -135,6 +139,31 @@ private IDisposable ConnectErrors(IServiceProvider provider)
...
+        //
+        // This needs to be fixed on Avalonia's end, unfortunately, which may take a while
+        // and wouldn't quite cut it for Stardew Valley Beta.
+        //
+        // For now, we opt to ignore this exception for the time being, while either us
+        // or the Avalonia folks come up with a solution. Whichever is first.
+        // ReSharper disable once MergeIntoPattern
+        if (msg.Exception is AggregateException aggregateException 
+            && aggregateException.InnerException is DBusException)
+            return false;
```
erri120, line 162: I don't think we should be excluding all D-Bus exceptions. D-Bus is used for selecting and opening files and links, among other things.

erri120 (changes requested):

src/NexusMods.App.UI/Windows/MainWindowViewModel.cs
Sewer56: Do you have a better suggestion? We can't necessarily filter out by using the exception message; that may be localized; or contain localized parts.

The HRESULT is 0x80131500, which, as far as I can tell would likely be shared by other DBus Services.

The Stack Trace does not have any Avalonia code etc. we don't even know the caller of the issue. Nothing we could reference there either.

We have no hope of getting Avalonia to merge our changes, even if we were to fix it ourselves within the next 4-8 working days.

Sewer56:
Completed via #2802

PR #2642, opened 2025-02-13
Added: Remainder of Phase One of Updates.

### LINKED ISSUE #2051: Mod Updates UI: Phase 1
## Summary
(To be delivered before Stardew Valley release)

Show updates in the Library UI

* Add rows in the library for new files on nexus mod pages for which we have updates
    * Add button on row to either:
        * Download file if Premium
            * Show spinner while downloading
        * Open website if non Premium
* Add toolbar button to start the update check
    * Spinner in this as well
* Skip ignore update button for this phase
* Column to display which collection/s the mod is installed into

## Designs:

[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=1711-32336&t=Cjrt5ovAMCoknmqP-1)

Image

Sewer56:
fixes #2051

This adds the remainder of 'Phase One' functionality related to updates; and some miscellaneous related changes.

## Real Time Update Status Updates

[liveupdatestuff.webm](https://github.com/user-attachments/assets/c815a40c-01c7-4375-aaa4-2476625fbe2b)

Update status of individual mods is now updated in real time as items are added/removed from the library.

To support this, I also added an optimized minimal update notification routine for when Library Items are Installed/Removed. Rather than recomputing everything from scratch the optimized routine recomputes *only* the mods and mod pages that may have been affected by the library additions/deletions.

This means that if a user has 3000+ mods, and user adds 1 mod, we only refresh update info for 1 mod page (all mods on that page), etc. Same with removals. 

This avoids loading unnecessary entities from the DB, saving many milliseconds of time each time you install a mod with an already large library.

I haven't quite profiled it extensively enough to know the exact savings on large datasets, but I would imagine it could save around a second or two if you install 300 mods from a collection with 1000 already installed.

My main concern however was moreso in not hitching the UI or stalling the DB. Stuff should feel responsive, and optimized implementation this should help a bit.

Notes included in the commit.

## Local Rate Limit of Mod Update Checks

I added a 30 second rate limit for mod update checks to guard against users spamming the Nexus API. You can skip it from API/code, but not from the UI.

This makes a worst case scenario of `120*numberOfManagedGames` per hour in terms of Nexus API calls, should the user put an autoclicker on the refresh button. The average API calls per game each update check is 1.

When the local rate limit is hit, we note that in the logger.

## Lift out non-Generic Parts of TreeDataGridItemModel

Lifted out parts of the `TreeDataGridItemModel<TModel, TKey>` to its base variant, `TreeDataGridItemModel`, while exposing more of its functionality via the interface. 

I originally did that for the whole 'expander' functionality (expand row if free user and clicking 'update all'), which we couldn't get working in the end without a nasty hack that wouldn't pass PR, due to a TreeDataGrid bug we encountered and need to report upstream. However, I figured the tiny refactor was good to keep.

Previously access to such functionality wasn't (really) possible without a nasty cast into a generic; I would rather avoid that, so we're going for a non-generic interface for functionality such as 'expander' here.

While going at that, I noticed that more behaviour could be lifted out to the base class, so I did, in case there are future needs similar to this. I didn't notably lift out model activation, I could; but I'm deciding to leave that be for now, since that involves a tiny bit of generic logic I want to avoid touching; since we'd kind of be breaking the nice chain of readable notes.

## Small Rework to Mod Page Update Info Observables (UpdateService)

Mod Page Update info observable now provide info for mods on the given page at an individual level.

Previously we returned all files that are newer on a page as a flat list, assuming the first file as the most recent.
However, this meant that it was difficult to distinguish which files belonged to which mod.
If there were multiple mods on a mod page, that was problematic; we couldn't easily tell them apart.

Which was problematic if user clicks 'update all' on a mod page row, etc.

In addition, we are also now able to determine if we may already have a newer version of an existing file much easier as a result of the rework.

## Added RaiseAndSetIfChanged for ReactiveR3Object class

Added a `RaiseAndSetIfChanged` method for raising `PropertyChanged` and `PropertyChanging` was added to ReactiveR3Object. The previous method was renamed to `RaiseAndSetIfChangedReactive`.

Although I'm not exactly sure of the exact purpose of the old method, as it is not used anywhere, I've decided to keep it around per @erri120 comment in voice call.

## Expand Mod Page row if Free User clicks 'Update' and there are Multiple Mods

Added the behaviour from the design, however this code is commented out and replaced with alternate temporary behaviour (after discussion with design) due to the TreeDataGrid bug.

We will restore the code once the bug is fixed.

## Misc Changes

- Added a note on why we subscribe on UI thread for Update Button
- Added a note on where/why we exclude update checks to items on disk.


src/Abstractions/NexusMods.Abstractions.UI/ReactiveR3Object.cs
```diff
@@ -61,22 +66,57 @@ protected virtual void Dispose(bool disposing)
...
+    ///     This is auto set via <see cref="CallerMemberNameAttribute"/> if invoking the property directly.
+    /// </param>
     protected void RaiseAndSetIfChanged<T>(ref T backingField, T newValue, [CallerMemberName] string? propertyName = null)
     {
+        // If they are equal, no action should be taken.
+        if (EqualityComparer<T>.Default.Equals(backingField, newValue))
+            return;
+
+        if (propertyName is null)
+            ThrowHelpers.ArgumentNullException(nameof(propertyName));
```
erri120, line 97: There already is a throw helper: `ArgumentNullException.ThrowIfNull`

src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridItemModel.cs
erri120, line 1: Excluding moving some properties to the interface, what's with the other changes? `ChildrenCollectionInitialization` and `HasChildrenInternal`. I don't see this used anywhere.

Sewer56: If you're fine with me lifting out

```
model._childrenObservableSerialDisposable.Disposable = model.ChildrenObservable
    .OnUI()
    .DisposeMany()
    .SubscribeWithErrorLogging(changeSet => model._children.ApplyChanges(changeSet));
```

To a method in the generic parent, while moving the rest to the non-generic base class, I'd be happy to do that.
I didn't do that because I felt like the whole comment chain is more valuable as one.

erri120: I still don't understand why the `TreeDataGridItemModel.cs` file was changed at all besides using a non-ReactiveUI `RaiseAndSetIfChanged` method. There should be no need to do anything else in there.

Sewer56: I changed it to expose the expander via the interface originally, so methods can be written that interact with that work on any TreeDataGrid row.

Outside of that though, I tried moving everything (as much as possible) to non-generic code when investigating the TreeDataGrid bug in hopes it could help, since the properties were now no longer in generic types. It made no difference, sadly.

Sewer56: Just like my life, redundant, bleh.

I do have a question though.
Why is there a base class at all then? 

What's in there is not anywhere close to sufficient to represent a single row. Now, it very much is, aside from lifting the constructor logic out.

If you ever want to make another derivative then what is here is better, as it's just 99 seconds to move the constructor, lifting out the innermost scope (last subscription) with generic logic into an abstract method.

If this is to be the only implementation, the base class should be removed, as it is not necessary.

erri120: >  Why is there a base class at all then?

Before `CompositeItemModel<>` we used to have dozen of sub-classes of `TreeDataGridItemModel` and sub-interfaces of `ITreeDataGridItemModel`. I just haven't cleaned up the `TreeDataGridItemModel.cs` file yet, mainly because we still have a model that hasn't been moved over to `CompositeItemModel<>` yet (load order models).

That's why any changes to the abstraction in the file are mostly a waste of time, there is no need to abstract if there is only one concrete type or any use-case in the abstraction.

For this PR, the only change we'd want here is the `RaiseAndSetIfChanged` fix and that's it.

Sewer56: Okie, should be done.

erri120 (approved):

PR #2643, opened 2025-02-17
Add Delete button to mod contents File Tree view

### LINKED ISSUE #2324: Deleting files outside the App results in impossible to deploy files
## Outline:
Deleting files outside the app results in irreversible ban of the file from deployment.
Only solution is to abandon loadout and restart from scratch.

## Problem:
If you delete a file outside the app, the deletion gets picked up by the app and placed in the hidden Overrides mod.

That deletion entry will from then on prevent any file with the same destination from being deployed.
The deletion entry itself can't be removed, cleared, or in any way be interacted with. It is also in no way visible to the users.

The only way a user has to allow the app to deploy files to that destination again is by starting from scratch, abandoning their loadout with all their mods.

This experience is horrible for users, that don't know why some files are not getting deployed. It is also horrible for giving support as users will rarely be able to tell whether they have deleted some file at some point or not.

## Steps:
To reproduce:
- Install a BG3 mod with a pak file
- Delete the BG3 pak file from BG3 `%localappdata%\Larian Studios\Baldur's Gate 3\Mods`.
- Attempt to install the same pak file from the app again
- pak file will not be deployed
- pak file is now cursed

### LINKED ISSUE #2603: Outside Changes/Overrides/External files
As a user 
I want to see a list of files that are not game files that were there before I installed the NMA
So that I know NMA has picked them up and gives me options

One of the most reported problems for SDV are that they claim Vortex eats their files because it does something with them and they don't know what. This exposes any files/mods that were already there before NMA started managing the game and gives them visibility and options of what to do with them. 

More options will be added in time but to begin with they are able to delete the file Or turn off or on all external changes. 

#### Edit 11/02/25:
After discussing with design and considering implementation complications and edge cases, we decided to:
- [ ] Show "External Changes" entry under "Utilities" in Left menu
    - [ ] Open View Mod Contents page for Overrides when pressed
    - [ ] Add button to delete entries in view

Showing Overrides in My Mods collection poses a number of complications:
- A migration is required to move it under "My Mods"
- Synchronizer interactions with disabled Override are unknown and potentially problematic
- Toggling "My Mods" off would also toggle Overrides mod off, with the same potential Synchronizer problems
- Overrides isn't a mod users have installed, so its presence in My Mods could be confusing.

Al12rs:
- Closes #2324
- Closes #2603 

Since the tree is static it doesn't react to changes, the only way to refresh the view is to recompute the tree from scratch, which resets the expanded states.

Functionality is there, experience isn't great. We need to move the file tree to new TreeDataGrid tech to make it dynamic and fix the performance/memory problems.

### DIFF
--- a/src/NexusMods.App.UI/Pages/ItemContentsFileTree/ItemContentsFileTreeViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/ItemContentsFileTree/ItemContentsFileTreeViewModel.cs
@@ -1,6 +1,4 @@
-using System.Reactive;
 using System.Reactive.Disposables;
-using System.Reactive.Linq;
 using Avalonia.Controls.Selection;
 using JetBrains.Annotations;
 using Microsoft.Extensions.Logging;
@@ -17,6 +15,7 @@ using NexusMods.MnemonicDB.Abstractions;
 using NexusMods.MnemonicDB.Abstractions.TxFunctions;
 using ReactiveUI;
 using ReactiveUI.Fody.Helpers;
+using R3;
 
 namespace NexusMods.App.UI.Pages.ItemContentsFileTree;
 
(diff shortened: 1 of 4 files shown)


src/NexusMods.App.UI/Pages/ItemContentsFileTree/ItemContentsFileTreeViewModel.cs
```diff
@@ -35,64 +38,106 @@ public ItemContentsFileTreeViewModel(
...
                 .Select(group => new LoadoutItemGroupFileTreeViewModel(group))
-                .BindToVM(this, vm => vm.FileTreeViewModel)
+                .Do(viewModel => FileTreeViewModel = viewModel)
+                .Subscribe()
                 .DisposeWith(disposables);
-
+            
+            // Observe selected items
             this.WhenAnyValue(vm => vm.FileTreeViewModel!.TreeSource.Selection)
+                .ToObservable()
```
Al12rs, line 127: @erri120 Do you know if there is a better way to do this with R3 that doesn't require using system observables? 

`ObservePropertyChanged` only supports immediate properties, no chains and no nullable chains.

erri120 (approved):

PR #2646, opened 2025-02-17
Check parent id for installation status

### LINKED ISSUE #2626: A Collection can be considered "Complete" without mods installed as collection components.
# Bug Report

## Summary
It's possible to get a collection into a state where the app thinks it's fully installed but the mods in the loadout aren't actually from a collection. 

## Steps to reproduce
1, Install https://next.nexusmods.com/stardewvalley/collections/g14kxi
2. Go to the Library, find SMAPI, and select the row.
3. Click "Add" on the toolbar to add SMAPI to "My Mods".
4. Go to "All" and find the SMAPI row, expand it to see both SMAPIs from the collection and My Mods.
5. Remove the collection SMAPI.
6. The collection is still in the left menu as the app thinks it is fully installed. 
7. Remove the non-collection SMAPI. 
8. See that the left menu item disappears. 

In terms of collection integrity, we need to verify that the installed mod is _from_ the collection, as there may be customisations such as "replicate", "binary patch" or "FOMOD options" that would differ from the intended experience when installed in another collection.

erri120:
Fixes #2626.

Before, the app would only check the loadout id, but never made sure the parent is the collection.

### DIFF
--- a/src/NexusMods.Collections/CollectionDownloader.cs
+++ b/src/NexusMods.Collections/CollectionDownloader.cs
@@ -401,8 +401,16 @@ public class CollectionDownloader
             .FilterOnObservable(item =>
             {
                 return groupObservable
-                    .Select(optional => optional.Convert(static group => group.AsLoadoutItemGroup().AsLoadoutItem().LoadoutId))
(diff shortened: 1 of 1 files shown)


Al12rs:
Doesn't seem to be completely fixing #2626 
If you follow the steps in the ticket and then delete the collection mod, the collection still shows up as installed, even though the collection mod has been removed. So it seems the collection isn't checking things correctly either

PR #2649, opened 2025-02-17
Fixes for `DirectDownloadLibraryFile`

### LINKED ISSUE #2638: Directly downloaded files of a collection don't show up in the library
Test with [Mega Expanded Stardew 600+ Revision 63](https://next.nexusmods.com/stardewvalley/collections/g3i395/revisions/63/mods) and download the external file `Anything Anywhere`. It doesn't appear in the library because the library only shows local files and files from Nexus Mods.

erri120:
Fixes #2638.

- `DirectDownloadLibraryFile` entities now show up in the library
- `DirectDownloadLibraryFile` entities now have proper names


erri120:
> What happens if you already have a local file that matches the MD5 hash for the external file? Do we convert it to a `DirectDownloadLibraryFile`? Do we leave it as a `LocalFile`? 

If you already have the file downloaded then we don't download it again, so you'll only ever have one of these.

Al12rs (approved):

PR #2653, opened 2025-02-17
Upgrade GameFinder to 4.5.0

### LINKED ISSUE #2401: Stardew Valley not detected on Linux Heroic
# Bug Report

## Summary
Stardew Valley (Linux version) is not detected through Heroic games store. 

## Steps to reproduce
1. Install the native SDV build via Heroic.
2. See it is not detected in the app. 

I can manage SDV

## Other information

[installed.json](https://github.com/user-attachments/files/18167992/installed.json)

[nexusmods.app.main.current.log](https://github.com/user-attachments/files/18168027/nexusmods.app.main.current.log)
[nexusmods.app.main.01.log](https://github.com/user-attachments/files/18168032/nexusmods.app.main.01.log)
[nexusmods.app.main.02.log](https://github.com/user-attachments/files/18168033/nexusmods.app.main.02.log)

erri120:
Fixes #2401

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -136,7 +136,7 @@
     <PackageVersion Include="BitFaster.Caching" Version="2.5.2" />
     <PackageVersion Include="CliWrap" Version="3.6.7" />
     <PackageVersion Include="DynamicData" Version="9.0.4" />
-    <PackageVersion Include="GameFinder" Version="4.4.0" />
+    <PackageVersion Include="GameFinder" Version="4.5.0" />
     <PackageVersion Include="Humanizer" Version="2.14.1" />
     <PackageVersion Include="ini-parser-netstandard" Version="2.5.2" />
     <PackageVersion Include="Mutagen.Bethesda.Skyrim" Version="0.44.0" />
--- a/src/NexusMods.Games.FileHashes/FileHashesService.cs
+++ b/src/NexusMods.Games.FileHashes/FileHashesService.cs
@@ -300,7 +300,10 @@ public class FileHashesService : IFileHashesService, IDisposable
             foreach (var gogId in locatorMetadata)
             {
                 if (!ulong.TryParse(gogId, out var parsedId))
-                    throw new InvalidOperationException("GOG locator metadata is not a valid ulong");
+                {
+                    _logger.LogWarning("Unable to parse `{Raw}` as ulong", gogId);
+                    return false;
+                }
 
                 var gogBuild = GogBuild.FindByBuildId(Current, BuildId.From(parsedId))
                     .FirstOrDefault();
--- a/src/NexusMods.StandardGameLocators/HeroicGogLocator.cs
+++ b/src/NexusMods.StandardGameLocators/HeroicGogLocator.cs
@@ -1,3 +1,4 @@
+using System.Runtime.InteropServices;
 using GameFinder.Launcher.Heroic;
 using GameFinder.StoreHandlers.GOG;
 using Microsoft.Extensions.DependencyInjection;
@@ -45,6 +46,7 @@ public class HeroicGogLocator : IGameLocator
         {
             if (!_cachedGames.TryGetValue(GOGGameId.From(id), out var found)) continue;
             var fs = found.Path.FileSystem;
+            var gamePath = found.Path;
 
             if (found is HeroicGOGGame heroicGOGGame)
             {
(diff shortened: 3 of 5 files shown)


Sewer56 (commented):
I had a long-ish comment here.
Unfortunately, my browser crashed, and I lost all the text.
Here's a ***short*** summary from what I remember. [compared to original text anyway]

```
System.InvalidOperationException: GOG locator metadata is not a valid ulong
   at NexusMods.Games.FileHashes.FileHashesService.TryGetGameVersionDefinition(GameInstallation installation, IEnumerable`1 locatorMetadata, ReadOnly& versionDefinition) in /home/sewer/Project/NexusMods.App/src/NexusMods.Games.FileHashes/FileHashesService.cs:line 303
   at NexusMods.Games.FileHashes.FileHashesService.TryGetGameVersion(GameInstallation installation, IEnumerable`1 locatorMetadata, String& version) in /home/sewer/Project/NexusMods.App/src/NexusMods.Games.FileHashes/FileHashesService.cs:line 284
   at NexusMods.App.UI.Controls.GameWidget.GameWidgetViewModel.<>c__DisplayClass1_0.<<-ctor>b__11>d.MoveNext() in /home/sewer/Project/NexusMods.App/src/NexusMods.App.UI/Controls/GameWidget/GameWidgetViewModel.cs:line 49
``` 

image

(On every App boot)

It seems we're missing a BuildID here.

I've not worked with GameFinder code here, but a quick debug session later, it seems, that we're pulling BuildID from Heroic's own metadata, and they don't store the BuildID. Presumably because it comes from installer, and not GOG's Content-System. You can only get the BuildID from game files post install.

[Note: BuildID is `58211149695218771` per `goggame-1453375253.info` for latest Linux GOG release. This is a proper unique BuildID, not some copied ID from a non-Linux release.]

Opinions would differ here I imagine, but I would consider fetching this data from the store launcher (in this case Heroic) to be the 'correct' thing here. Ideally, long term (in my opinion), we should be editing the launchers' store metadata when we do an in-App game version switch.

This is to prevent the store or the user from accidentally restoring the files for a different game version; for example when user clicks 'Verify Game Files' in their launcher; or risking the launcher auto-replacing files, e.g. if they launch game from launcher.

---------

Per @halgari comment though, it doesn't seem like we're going to be indexing Native GOG Linux SDV for now, as that's not served by GOG's Content-System, and we're not quite equipped to handle that.

Without a game-hashes DB entry you can't manage the game, so I would recommend against closing 
- https://github.com/Nexus-Mods/NexusMods.App/issues/2401

Sewer56:
The code is fine. I've checked the extra commit yesterdag after regaining power.
I need to check if the game can be managed and played however.

If I recall, GameFinder yesterday was returning folder to where the installer put its stuff in, but the actual game files are in a subfolder. (They made the wrapper script).

This might cause us to deploy stuff to wrong folder, I'll need to see and test that. I haven't been emotionally there this morning, so it may take a short while.

erri120:
@Sewer56 I've updated the code to use the `game` sub-path. It's a hack but we don't really have anything better right now.

Sewer56:
Ya, that works for me. I just ran into another bug where I have a GOG Stardew install in Wine. 

image

The item that shows up in the Manage Games page is the Heroic Install.
The one passed to ManageGame/CreateLoadout is the Heroic install.

You go to game page, it picks the Wine install location.
Making a ticket for this, after I'll clean registry and test with the hack (which I also wrote locally, haha).

Sewer56:
I had to make a commit here to get this to work, since we should be overwriting the result of the GameFinder query (`_cachedGames`), last commit only replaced the `GameLocatorResult`, but later code would pull data from `_cachedGames`, which didn't match up.

I also had to wipe datastore, it seems that somewhere, we cache the install location in MMDB. If you manage game without the patch, apply the patch, then manage the game again, then the `Installation` instance used for diagnostics etc. will still see the path from the first manage. Not sure if worth reporting as bug, since this is induced by `

Outside of that, things mostly work.

image

[CC. @Al12rs , the Diagnostic Emitters think this is the Windows version of SDV 1.6.15, probably because it was closest match]

But you can install mods and run the game:

image

image

Sewer56:
@erri120 I can't push to branch since it's on your fork, so here's the git patch file for the changes I needed to make

[227ce1.patch.zip](https://github.com/user-attachments/files/18882321/227ce1.patch.zip)

Sewer56:
CC @Al12rs in case the reason I tagged you isn't clear.
From my experience here, I think there may be issues around managing games that aren't indexed. 
There's a risk an unmanage after a game update we don't index etc. may nuke a game, like it did here. (and the other diagnostic)

Sewer56 (approved):
The PR itself is good.
Still got some issues, but they're unrelated to PR. See discussion.

Sewer56:
Files will be deleted either way if unrecognized (overrides), and solving it properly for unknown versions will take some time, so we can leave it as is for the release, given there's not time for it for the initial beta.

We will need to solve this sooner rather than later though. We don't want users' optional files (DLCs) to be wiped on unmanage.

Al12rs:
> We will need to solve this sooner rather than later though long term.
I agree, made a ticket: https://github.com/Nexus-Mods/NexusMods.App/issues/2692

PR #2655, opened 2025-02-17
Update to the latest version of MnemonicDB

halgari:
Well this was a much bigger pain than I thought. This updates to the latest version of MnemonicDB, which does the following: 

* Fixes the deadlock issues with nested observables
* Implements observables in a race condition tolerant way. Essentially we no longer care about race conditions during subscribers and instead drop or recreate out-of-order database updates. Updates that skip DB versions are fast-fowarded (we create deltas-on-the-fly for the missing updates), those that are older are dropped. 
* There was a serious performance regression due to the way we unsubscribed from updates, implemented a bulk item removal in the IntervalTree code and switched to a custom build of that code.


tests/NexusMods.Collections.Tests/CollectionInstallTests.CanInstallCollections_slug=jjctqn_revisionNumber=1.verified.txt
erri120, line 1: Why did these verify files changed?

erri120: These names might've changed due to #2649 and the tests that produce these files aren't run in CI.

Issue #2493, opened 2025-01-20
Command Line Window appearing when starting the app


Pickysaurus:
# Bug Report

## Summary
Building a release from main (https://github.com/Nexus-Mods/NexusMods.App/commit/ca81f4d9e13aeac77ee60cceba09cc40dfc7eeeb) results in the command prompt window opening again on Windows. This is undesirable behaviour and is a regression of https://github.com/Nexus-Mods/NexusMods.App/issues/1264

## Steps to reproduce
1. Build from main
2. Open the app on Windows
3. See command prompt window.

## What is the expected behaviour?
The command prompt window does not open.

PR #2656, opened 2025-02-18
Don't show the console on windows

halgari:
fixes: [#2493](https://github.com/Nexus-Mods/NexusMods.App/issues/2493)

This will continue to be an issue in the future due to the utter jank way Windows handles CLI commands and compilation modes. But this work for now. 

Known issue: When you invoke a CLI command from the app, you instantly get a new prompt then the info from the app pops up afterwards. This is confusing and annoying but not required for our beta release. At least now the console doesn't pop up when you run the UI, and yet you can still use the CLI.

### DIFF
--- a/src/NexusMods.App/ConsoleHelper.cs
+++ b/src/NexusMods.App/ConsoleHelper.cs
@@ -24,7 +24,7 @@ public static class ConsoleHelper
     /// <param name="forceNewConsoleIfNoParent">If there is no parent console, should one be created?</param>
     public static void EnsureConsole(bool forceNewConsoleIfNoParent = false)
     {
-        if (!AttachConsole(ATTACH_PARENT_PROCESS) && !forceNewConsoleIfNoParent)
+        if (!AttachConsole(ATTACH_PARENT_PROCESS) && forceNewConsoleIfNoParent)
         {
             AllocConsole();
         }
--- a/src/NexusMods.App/NexusMods.App.csproj
+++ b/src/NexusMods.App/NexusMods.App.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <OutputType>Exe</OutputType>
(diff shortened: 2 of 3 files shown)


Sewer56:
I'll double check this stuff as a sanity test after rebooting in a bit but looks fine at first glance.
I have some experience around these APIs, due to mod loader development.

There are some caveats normally, if you launch with WinExe and manually attach console. Been a while and not fresh in my head but IIRC stdout may not be wired either depending on runtime used. (Specifically for native binaries, so possibly no console from native DLLs, .NET wires up stuff fine under the hood). There were also some caveats with stdin, if you pause to read a line.

In our App, we don't run into these edge cases though, one command to the CLI to init, and stdout from .NET, where it works just fine. I think it's a-ok to go.

Sewer56:
More specifically, if you launch the process without a console window, and then later attach to a console, or alloc a console, then stdin/stdout/stderr will not be wired up to that console in `msvcrt`, i.e. native msvc built binaries, i.e. all our native dependencies.

Normally you need to manually wire these up

```c
freopen("CONIN$", "r", stdin);
freopen("CONOUT$", "w", stdout);
freopen("CONOUT$", "w", stderr);
```

Such that printing to stdout will actually redirect to the console.
I don't remember if the .NET `Console.SetOut` etc. functionality properly redirect these streams, I remember playing around all that for a very long time though, it was a hassle.

I think I might have brought this up all the way back in 2023, and possibly even did something before the SingleProcess stuff. Memory is vague by now, but I remember thinking about it in App context.

Thankfully this is very minor, we have few native dependencies but it is a thing worth knowing/noting.

Sewer56 (approved):
Can confirm that when launching process using an external console, the stdin/stdout/stderr handles are correctly set up (after inheriting from parent); thus this is good to merge. I even tested with a real MSVC based native DLL, just in case.

Have not tested that with spawning new console (`forceNewConsoleIfNoParent`), since we're not intending to use that for now.

PR #2679, opened 2025-02-19
Count loadout items properly

### LINKED ISSUE #2676: Mod count implementations don't include bundled mods
Add `IObservable<int> CountLoadoutItems()` to `ILoadoutDataProvider` and use that for counting mods instead of relying on previous implementations that don't include bundled mods or any other new type of items we'd want to add in the future.

- [x] `LoadoutUserFilter` count (used in loadout cards mod counts)
- [x] `LoadoutLeftMenuViewModel`  `All (#)` mod count
- [x] `TelemetryProvider` mods per loadout count

erri120:
Resolves #2676.


src/NexusMods.App/TelemetryProvider.cs
```diff
@@ -35,6 +37,30 @@ public TelemetryProvider(IServiceProvider serviceProvider)
...
+
+                    var gameName = loadout.Installation.Name;
+                    _loadoutCounts.TryGetValue(gameName, out var counts);
+
+                    _loadoutCounts[gameName] = counts + kv.Value;
+                }
+
+                return 0;
+            })
+            .SubscribeWithErrorLogging()
```
Al12rs, line 62: This is grouping by game install not by loadout. This is counting all the mods for a specific game, not all the mods for a loadout. 
Should we return the average number of mods per loadout per game? The max?

erri120: Updated to use original behavior.

Al12rs (approved):

PR #2693, opened 2025-02-20
Event Bus and reacting to downloads

### LINKED ISSUE #1537: Add Event Bus
An event bus fixes the issue of having unrelated components try to talk with each other.

We should discuss what kind of API we want.

### LINKED ISSUE #2548: Open collections download page after user added a collection
After a user added a collection from the website to the app, the app should navigate to the collection download page of the added revision. If possible, the app should also be brought into the foreground to grab the users attention.

### LINKED ISSUE #2689: Bring app window to front when it opens a download from browser
# Improvement

## Currently 
App remains in background when a mod or collection is downloaded from a browser that is set to open the download in the app.

## We want
To bring the app to front when this happens.
This can later be a setting so the user has a choice, but the default should be it always opens in front when called.

erri120:
- Implements a simple event bus (Resolves #1537)
- Opens the collection download page when a collection gets added and brings the app window to the front (resolves #2548)
- Brings the app window to the front when a download is added (resolves #2689)
- Adds a setting for users to toggle the behavior of bringing the app window to the front

### DIFF
--- a/src/NexusMods.App.UI/Settings/BehaviorSettings.cs
+++ b/src/NexusMods.App.UI/Settings/BehaviorSettings.cs
@@ -1,0 +1,20 @@
+using NexusMods.Abstractions.Settings;
+
+namespace NexusMods.App.UI.Settings;
+
+public record BehaviorSettings : ISettings
+{
+    public bool BringWindowToFront { get; set; } = true;
+
+    public static ISettingsBuilder Configure(ISettingsBuilder settingsBuilder)
+    {
+        return settingsBuilder.AddToUI<BehaviorSettings>(builder => builder
+            .AddPropertyToUI(x => x.BringWindowToFront, propertyBuilder => propertyBuilder
+                .AddToSection(Sections.General)
+                .WithDisplayName("Bring app window to front")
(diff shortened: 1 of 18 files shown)


src/NexusMods.App.UI/Settings/BehaviorSettings.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public bool BringWindowToFront { get; set; } = true;
+
+    public static ISettingsBuilder Configure(ISettingsBuilder settingsBuilder)
+    {
+        return settingsBuilder.AddToUI<BehaviorSettings>(builder => builder
+            .AddPropertyToUI(x => x.BringWindowToFront, propertyBuilder => propertyBuilder
+                .AddToSection(Sections.General)
+                .WithDisplayName("Bring app window to front")
+                .WithDescription("When enabled, operations like adding a collection will bring the app window to the foreground")
```
Al12rs, line 15:
```suggestion
                .WithDescription("When enabled, operations like starting a mod download will bring the app window to the foreground")
```

Al12rs (approved):

PR #2696, opened 2025-02-20
Open Library when managing game

### LINKED ISSUE #2554: Creating a Loadout should open the Library
Create a new loadout
Click on that loadout
First page you see is 'All'

The first page you should see should be the Library so you are presented with options of which collection or mods to install to that loadout.

Al12rs:
- Closes #2554


Sewer56 (approved):
Has some whitespace to trim in diff, but otherwise is good.

PR #2697, opened 2025-02-20
Add tests for Stardew Valley diagnostic emitters

erri120:
This doesn't have tests for the emitters that work on game versions due to how annoying it is to set up game versions in tests.


tests/Games/NexusMods.Games.StardewValley.Tests/DependencyDiagnosticEmitterTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        // Content Patcher 2.5.3 (https://www.nexusmods.com/stardewvalley/mods/1915?tab=files)
+        await InstallModFromNexusMods(loadout, ModId.From(1915), FileId.From(124659));
+
+        await ShouldHaveNoDiagnostics(loadout, because: "The required dependency Content Patcher has been installed");
+
+        await VerifyDiagnostic(diagnostic);
+    }
+
+    [Fact]
+    public async Task Test_DisabledRequiredDependencyMessageData()
```
Al12rs, line 50: We are missing the case where parent collection of the dependency is disabled.
We should add a test for that once we have fixed any issues we have there.

Al12rs (approved):

Issue #1721, opened 2024-07-04
Allow users to share their SMAPI log directly from the app. 


Pickysaurus:
# User story

**As a** user

**I want to** get a sharable link to my SMAPI log file

**So that** I can give it to others who are helping me. 

## Requirements
We should ensure we append the log with our application name and version so that it's clear they have been using our app to manage the game. 
Vortex adds this `[10:28:12 INFO Vortex] Log exported by Vortex 1.11.7.`

### How to send the log
SMAPI's log website allows users to submit log files as part of a POST request. 

1. POST `https://smapi.io/log` with the raw text of the latest SMAPI log passed as `x-www-form-urlencoded` under the key `input`.
2. A HTML page will be returned. 
3. Look for the following section of the HTML document
```html
<div class="banner success" v-pre>
                <strong>Share this link to let someone else see the log:</strong>
                <code>https://smapi.io/log/8a59751b85d9496ea30483489ac5cea0</code><br />
        (Or <a href="/log">upload a new log</a>.)
            </div>
```
4. Extract the link from the HTML and allow users to copy/share it. (Or just open the webpage in their default browser) 

## Design
_(Any designs required for this story)_

## DOD
_(Describe the expected behaviour from the users point of view. Create multiple DoD's if required to fully describe what should happen. Number each DoD)_

## Scenario 1
**Given** _______

**When** _______

**Then** _______

Pathoschild:
If it's easier, SMAPI also supports Pastebin IDs. You can [upload a file to Pastebin](https://pastebin.com/doc_api) and then pass its ID in the URL (like [pastebin.com/vsCY4xtX](https://pastebin.com/vsCY4xtX) → [smapi.io/log/vsCY4xtX](https://smapi.io/log/vsCY4xtX)).

PR #2699, opened 2025-02-21
Added SMAPI crash diagnostic

Pickysaurus:
Added a new diagnostic for Stardew Valley which will show a Critical Health Check warning if the last log generated by SMAPI is a crash log.

It works by:
1. Checking the `%appdata%\StardewValley\ErrorLogs` (or Linux equivalent (needs testing!)) 
2. Compares the last edit time of all log files.
3. A diagnostic is raised if the latest log file is `SMAPI-crash.txt` and it was last edited less than 12 hours ago.
4. When the latest log file is `SMAPI-log.txt` the diagnostic will not show.

Possible improvements:
- Detect the last time we launched the game (if possible?)  and use that to work out how relevant the log is
- Localise the date/time based on the users app locale
- Add tests


erri120:
I'm not sure what prompted this PR, whether this is just a pet project or something else. I think the only open issue related to SMAPI logs is #1721 which is about sharing logs. We had talks about a log viewer and parser before, but that got never scheduled or even written down.

Regardless of the ticket state, log parsing and diagnostics should be two separate systems. The idea behind diagnostics is to provide static analysis of the loadout and notify the user about every issue we can identify statically _before_ they launch the game. Logs are only produced _after_ launching the game. In the ideal world, we'd catch every issue with diagnostics before they start the game, not notify them afterwards using log parsing.

Log parsing is still helpful, especially for mod creators, and for debugging issues that can't be statically analyzed like runtime conflicts, which is why a separate system that is more fleshed out and specialized for logs would be needed. Probably a new left menu entry that opens a log viewer with a parser, a share button and many other gadgets to help work with logs. There are just too many features that we'd want in regards to logs that the diagnostic system doesn't provide because it was built for something else entirely.

I don't know what you want to do with this PR, probably fine to keep it as a draft PR for now until we figure out what we want to do for logs.

erri120:
Closing this for, we can take another look when dealing with #1721 or similar.

PR #2707, opened 2025-02-24
Manual download journey for collections

erri120:
Implements the new designs for the manual download journey.


Al12rs:
What is a good collection to test this with?

erri120:
> What is a good collection to test this with?

https://next.nexusmods.com/stardewvalley/collections/g3i395

Al12rs:
Not sure why some items don't have any buttons:
image

Al12rs (approved):
I tested and noted a couple of weird things above, but otherwise it seems to work correctly

insomnious:
> I tested and noted a couple of weird things above, but otherwise it seems to work correctly

Laurence and I are aware of this. Too big a job right now to start changing properties and swapping controls based on the current state. Will come back around to it.

Issue #2652, opened 2025-02-17
UI tweaks for pre-SDV release


insomnious:
These are in priority order

- Move give feedback button from top bar to bottom banner
- Update top bar menu items 
- Update alpha modal 
- left hand menu separators seem too bright
- on the lefthand menu item, active state should only change foreground to strong
- left align buttons in all toolbars
- tab titles are currently semi bold, they need to be just normal

PR #2708, opened 2025-02-24
Various UI tweaks 3

insomnious:
Sorry for the fiddly PR - it hit a lot of files that haven't been touched in a while to get the toolbars looking consistent.

Main changes are:

- SplitButton can now ussed!
- Right aligned bottom banner
- Added new Fill entry for StandardButton to not use translucent brushes
- Refactored Toolbar Border class
- Surface coloured fallback collection background image

Closes https://github.com/Nexus-Mods/NexusMods.App/issues/2652


insomnious:
image

src/NexusMods.App.UI/Controls/StandardButton.cs
```diff
@@ -107,9 +107,14 @@ public enum Fills
...
         /// <summary>
-        /// Weak fill.
+        /// Weak fill (translucent).
         /// </summary>
         Weak,
+        
+        /// <summary>
+        /// Weak solid fill.
+        /// </summary>
+        WeakAlt,
```
Al12rs, line 117: Does this follow figma nomenclature? It doesn't seem the most intuitive name

insomnious: I'll check further

Al12rs (approved):
Just the notes above, looks good otherwise

src/NexusMods.App.UI/Controls/StandardButton.cs
insomnious: To report back from @captainsandypants, it's to make sure that the abstracted name isn't exactly the visual representation of it in case we change it. It's why I added the comment to at least help devs using the properties and changing the comment is less disruptive. `WeakAlt` at the moment is using solid fills but in the future it may be a diff color or use diff opactities to `Weak`

Al12rs (approved):

PR #2714, opened 2025-02-24
Refactor `IFileHashesService` and implementations

erri120:
Cleanup PR. Functionality shouldn't be affected.

- Adds two new value types: `VanityVersion` and `LocatorId`
- Removes and updates many methods from `IFileHashesService`
- Refactors the implementations of `IFileHashesService`


halgari (approved):
Approved, if for no other reason that I love `VanityVersion`

PR #2715, opened 2025-02-25
Update to latest MnemonicDB that fixes deadlocks

halgari:
This updates to the latest version of MnemonicDB which resolves the deadlock issue by doing the following:

1. Added a queue to `IConnection` this queue handles all DB updates and changes to datom subscriptions. 
2. `ObserveDatoms` now essentially enqueues `(from, to, observer)` onto this queue whenever a new observer wishes to connect
3. DB updates go through this same queue

This means that there is now no need for locks because the same thread adds the observable and pushes updates. This also reduces a lot of complexity about delta calculations, we can send the initial datoms when we get a subscription event, and on the next DB update send all the deltas, no race conditions or locks. 

There is a *slight* change in semantics here that boils down to the question: how quickly after `.ObserveDatoms(...).Subscribe(...)` is called should the initial state be passed to the subscribe observer? In the old code it was instant, perhaps even before the `.Subscribe` call exited. This behavior is untenable as our application gets more complex. The semantics with this update are now: "as soon as possible, often very quickly". The connect is handed to a queue, and so the execution time can vary. 

However, in normal Rx code the behavior here is undefined, or at least not commonly relied on, so it's unlikely that this has a imact in our code. I ran the app through its paces and most of it seems to be working fine, perhaps even a bit faster than in the past. 

(this PR also contains the previously discussed changes in test files due to changes made in other PRs)


halgari:
I'm aware that the tests are failing here, not sure why, I'll try it out on Linux/MacOS tomorrow and try and track it down. in the mean time feel free to test this and I'll try and get the tests passing asap

halgari:
Disabled a test that was non-determinstic

Al12rs (commented):
From the load order test results I think the load order code made some assumptions as for the ordering of events received from the DB matching the order of individual Commit calls, which I think might not be the case anymore (I think multiple commits might get compacted into a single changeset depending on subscription timing?).

This probably makes the loadorder code non deterministic. I'm not sure how much of a problem that is. I'm thinking that we may have to rely less on DB sourced events and more on application sourced events for load order specifically.

halgari:
We'll need to loop back around and look at that code, because it shouldn't be dependant on the results form the DB or even be dependant on the observables at all. Thankfully we don't need that code for the SDV release

Al12rs (approved):
I tested te code a lot and was not able to reproduce any deadlocks.

I did see a couple of instances where values load it now, for example the left menu collection toggles, but I haven't spotted anything that might be a problem yet.

I have a couple of questions regarding the latest DB code.
Some of the code and comments suggest the possibility of race conditions that should no longer be possible now that everything is executing on the same `_eventThread`.

I also read up on RX Schedulers and dispatchers yesterday evening, as well as R3 equivalents, I have some takeaways from that regarding assumptions for subscription/OnNext synchronicity. 

I'd like to talk briefly after standup about the above.

Issue #2615, opened 2025-02-09
Unhandled Exception - Exception sending events


maaneeack:
# Bug Report

## Summary
With Telemetry enabled, if it's unable to reach the server an exception pop up happens every 3-10 seconds

## Steps to reproduce
Enable Telemetry
Block pw.nexusmods.com

## What is the expected behaviour?
A quiet handling of the error, or notification of some sort

## Other information

![Image](https://github.com/user-attachments/assets/3c4ccb4d-cbad-4721-9820-686e57a14afe)

I asked about it on the discord, Nothus, when I removed the block from pw.nexusmods.com on my pihole the error stopped occurring.

data./pw./sso.nexusmods.com are being blocked by the hosts file list here: https://github.com/KodoPengin/GameIndustry-hosts-Template

PR #2725, opened 2025-02-25
Silence the event sender errors, but log them with a debug message

halgari:
Fixes: #2615

Fixes for other network related errors will come in future PRs. This logs a debug message on failed sends for telemetry events, for testing purposes, being a "debug" level it won't show up in most users logs.

### DIFF
--- a/src/NexusMods.Telemetry/EventSender.cs
+++ b/src/NexusMods.Telemetry/EventSender.cs
@@ -32,10 +32,11 @@ internal class EventSender : IEventSender
     private readonly ILoginManager _loginManager;
     private readonly HttpClient _httpClient;
     private readonly ArrayBufferWriter<byte> _writer;
+    private readonly ILogger<EventSender>? _logger;
 
     private static readonly byte[] EncodedUserAgent = CreateUserAgent();
 
-    public EventSender(ILogger<EventSender> logger, ILoginManager loginManager, HttpClient httpClient)
+    public EventSender(ILogger<EventSender>? logger, ILoginManager loginManager, HttpClient httpClient)
     {
         _logger = logger;
         _loginManager = loginManager;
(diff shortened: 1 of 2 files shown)


src/NexusMods.Telemetry/EventSender.cs
```diff
@@ -200,14 +203,22 @@ void AppendBytes(byte[] data)
...
+        try
+        {
+            var response = await httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
+            response.EnsureSuccessStatusCode();
+        }
+        catch (Exception ex)
+        {
+            // NOTE (halgari): Yes, this will add more bloat to the log, but only in testing where DEBUG logging is enabled.
+            _logger.LogDebug(ex, "Error occured while sending events");
+        }
```
Al12rs, line 221: We were talking about adding a circuit breaker with Polly to avoid the spam attempts. 
I guess this works fine for now. 

I would consider using warning, but if we don't do any circuit breaking this could get spammy, so I guess debug is fine if we go with this.

Al12rs (approved):

Al12rs (changes requested):
Tests are failing, missing `ILogger<EventSender> logger` parameter when calling `EventSender` constructor in tests

src/NexusMods.Telemetry/EventSender.cs
halgari: Took a look at polly and I feel like it's a tad overkill in this case. The circuit breaker is great, but it has settings like "half-open", "minimum throughput" and "failure ratio" that don't really make sense on something that's already rate limited like telemetry. We don't really care about retries either as we'll just try again on the next polling event.

Al12rs (approved):

PR #2726, opened 2025-02-25
Respect health check settings in the summaries and side menu

### LINKED ISSUE #2720: Updating the Health Check Sensitivity setting does not fully hide lower severity notifications
# Bug Report

## Summary
Health check diagnostics not fully hidden when sensitivity setting is updated.

## Pre-requisites

- Add Stardew valley and no mods
- At the time of recording there is a suggestion that is always displayed for "Game files are not backed up", if this is no longer the case then you will need to trigger another suggestion diagnostic
- Install a mod e.g. "NPC Map Locations" (do not yet install "SMAPI")
- There should be a warning diagnostic to inform that "SMAPI is not installed"

## Steps to reproduce

- Go to settings and update the Health Check sensitivity to "Warning"
- Open the health check page
- Click on the suggestions filter

There should be no indication of diagnostics with a severity lower than the selected level in the total, also the tabs for the lower severity level should be hidden

## Other information

Image

Image

Image

Log files:

[nexusmods.app.slim.current.log](https://github.com/user-attachments/files/18965357/nexusmods.app.slim.current.log)
[nexusmods.app.main.current.log](https://github.com/user-attachments/files/18965358/nexusmods.app.main.current.log)

halgari:
Fixes #2720, also only displays the icons in the left menu when the diagnostics are not filtered by the severity settings.

### DIFF
--- a/src/NexusMods.App.UI/LeftMenu/Items/HealthCheckLeftMenuItemViewModel.cs
+++ b/src/NexusMods.App.UI/LeftMenu/Items/HealthCheckLeftMenuItemViewModel.cs
@@ -2,6 +2,7 @@ using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using NexusMods.Abstractions.Diagnostics;
 using NexusMods.Abstractions.Loadouts;
+using NexusMods.Abstractions.Settings;
 using NexusMods.App.UI.WorkspaceSystem;
 using ReactiveUI;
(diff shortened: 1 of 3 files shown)


Al12rs:
I just noticed that the Library left menu notifications don't filter by game:
image
Cyberpunk is empty, notifications are for SDV:
image

Don't know if you want to sneak that in as well

Al12rs (approved):

PR #2730, opened 2025-02-26
Hide unsupported games from My Games

### LINKED ISSUE #2313: Feature switch currently in progress games
On the initial release the only game that should be viewable in the app is Stardew Valley. 

Other games in progress should be locked behind a switch in settings to enable experimental games.

erri120:
Resolves #2313.

Before:

After:

image

### DIFF
--- a/src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
@@ -26,6 +26,8 @@ using ReactiveUI;
 using System.Reactive;
 using System.Reactive.Linq;
 using DynamicData.Aggregation;
+using NexusMods.Abstractions.Settings;
+using NexusMods.App.UI.Extensions;
 using NexusMods.App.UI.Overlays;
 using NexusMods.App.UI.Overlays.AlphaWarning;
 using NexusMods.App.UI.Pages.LibraryPage;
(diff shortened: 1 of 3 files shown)


Sewer56 (approved):
Works for me.

src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
```diff
@@ -89,6 +94,12 @@ public MyGamesViewModel(
         this.WhenActivated(d =>
             {
                 gameRegistry.InstalledGames
+                    .Where(game =>
+                    {
+                        if (experimentalSettings.EnableAllGames) return true;
+                        return experimentalSettings.SupportedGames.Contains(game.Game.GameId);
+                    })
+                    .ToReadOnlyObservableCollection()
```
Al12rs, line 102: We could check if game is already managed, and only hide if it isn't managed and it isn't supported

Sewer56: Too overkill in my opinion.
My games view no longer adds loadouts; like it used to back in the day.

The loadouts view is still unrestricted.
So functionally, there wouldn't be much difference here.

Sewer56: Wait... do we not have a way to remove individual loadouts anymore?
I don't see that anywhere in the UI.

Sewer56: Oh, I see. You can only delete loadouts that aren't the first loadout. So you need to still be able to see managed games in `My Games`, to unmanage them. Seems AL's right here; we do need to show these.

Al12rs:
Making this into an observable filter would make this more complicated that it needs to

PR #2733, opened 2025-02-26
Update collection card status

erri120:


insomnious (approved):
I have some questions around where the various states should be linking, and if it's installed it felt like it should be going to the CollectionLoadoutView and not the CollectionDownloadView. This PR works and is a step in the right direction

PR #2736, opened 2025-02-26
Use fallback page instead of breaking the workspace

### LINKED ISSUE #2735: Workspace system breaks if exception is thrown in page constructor or factory
If opening a page throws an exception in the page constructor or factory, the workspace will be broken:

Image

erri120:
Fixes #2735 opening the default page. This should guarantee that we don't end up in a situation where the workspace is completely broken.

### DIFF
--- a/src/NexusMods.App.UI/WorkspaceSystem/Page/PageFactoryController.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/Page/PageFactoryController.cs
@@ -2,6 +2,7 @@ using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using DynamicData.Kernel;
+using NexusMods.App.UI.Pages;
 using NexusMods.App.UI.Windows;
 
 namespace NexusMods.App.UI.WorkspaceSystem;
@@ -46,6 +47,7 @@ public class PageFactoryController
     {
         var factory = GetFactory(pageData);
         var page = factory.Create(pageData.Context);
+
         page.ViewModel.WindowId = windowId;
         page.ViewModel.WorkspaceId = workspaceId;
         page.ViewModel.PanelId = panelId;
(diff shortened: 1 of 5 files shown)


src/NexusMods.App.UI/WorkspaceSystem/Page/PageFactoryController.cs
```diff
@@ -44,8 +45,29 @@ public IPageFactory GetFactory(PageData pageData)
...
+                Context = new ErrorPageContext
+                {
+                    Exception = e,
+                    FailedPage = pageData,
+                },
+            };
+
+            var factory = GetFactory(errorPageData);
+            page = factory.Create(errorPageData.Context);
+        }
```
Al12rs, line 69: Wouldn't it have been better to handle the exception in the callers and just log an error so that an exception dialog is surfaced?

src/NexusMods.App.UI/WorkspaceSystem/Workspace/WorkspaceViewModel.cs
```diff
@@ -408,7 +411,22 @@ private void OpenPageReplaceTab(PageData pageData, OpenPageBehavior.ReplaceTab r
...
+        try
+        {
+            newTabPage = _factoryController.Create(pageData, WindowId, Id, panel.Id, tab.Id);
+        }
+        catch (Exception e)
+        {
+            _logger.LogError(e, "Exception creating page for factory `{FactoryId}` with context `{Context}`", pageData.FactoryId, pageData.Context);
+
+            var defaultPageData = GetDefaultPageData();
+            OpenPageReplaceTab(defaultPageData, replaceTab, selectTab, checkOtherPanels: false);
```
Al12rs, line 426: I think it would be better not to recall the function inside the catch, and just use the data newTabPage data instead.
Unless there is a reason you didn't do that?

erri120: We can do that.

Al12rs (approved):

PR #2737, opened 2025-02-26
Improve observable tracker page

erri120:
Adds selection and detailed information about the selected item with both a slim and a full stack trace.

Also added a pie chart for the top 10 most used types.

image

### DIFF
--- a/src/NexusMods.App.UI/Pages/ObservableInfo/ObservableInfoPageViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/ObservableInfo/ObservableInfoPageViewModel.cs
@@ -42,27 +42,24 @@ public class ObservableInfoPageViewModel : APageViewModel<IObservableInfoPageVie
         {
             Observable
                 .Timer(dueTime: TimeSpan.Zero, period: TimeSpan.FromSeconds(1), timeProvider: TimeProvider.System)
-                .Do(this, static (_, self) =>
+                .Select(this, static (_, self) =>
                 {
                     self._trackingStatesBuffer.Clear();
+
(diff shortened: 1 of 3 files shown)


src/NexusMods.App.UI/Pages/ObservableInfo/ObservableInfoPageViewModel.cs
```diff
@@ -36,13 +52,52 @@ public ObservableInfoPageViewModel(IWindowManager windowManager) : base(windowMa
...
                     });
 
-                    self._trackingStatesBuffer.Sort();
+                    var counts = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
+
+                    foreach (var state in self._trackingStatesBuffer)
+                    {
+                        var count = counts.GetValueOrDefault(state.FormattedType, 0);
+                        counts[state.FormattedType] = count + 1;
+                    }
```
Al12rs, line 61: Do this in the loop above to avoid iterating twice
```csharp
                    var counts = new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase);
                    
                    ObservableTracker.CheckAndResetDirty();
                    ObservableTracker.ForEachActiveTask(trackingState =>
                    {
                        if (self._trackingStatesBuffer.Add(trackingState))
                        {
                            var count = counts.GetValueOrDefault(trackingState.FormattedType, 0);
                            counts[trackingState.FormattedType] = count + 1;
                        }
                        
                    });
```

Al12rs: Not sure if you actually want the `if`, if you duplicates you might want to count those, but not sure

Issue #2390, opened 2024-12-16
TreeDataGrid column sorting doesn't work


LukeNexusMods:
The alphabetical order on the library page doesn't work correctly and keeps giving you a different order that is not alphabetical. 

You can see this most when you have installed a large collection.

erri120:
The new column tech we have doesn't sort properly and I don't know why. Might be a weird interaction with `TreeDataGrid` specifically, but the comparison operators we have should be fine.

erri120:
This hasn't been fixed.

LukeNexusMods:
File a ticket with Avalonia to find out what is happening.

LukeNexusMods:
Are we able to manually give the models values in the code?

halgari:
Did some investigation on this. As of now (in main) Size is broken, so I started there. After a bit of digging it's because we're setting `subscribeWhenCreated=false` on the `SizeComponent`. Setting that to `true` and inverting the sort order gives the expected behavior. 

At least for this column it's all down to subscriptions not enabling when we create the rows, so that we can't sort columns because the interneal sort values are `0` when the row is created, and because the rows aren't in view they never get the official value so that the sort can work correctly. 

Another test for this is to download ~4 mods, then notice that the sorting works fine because all 4 rows are in view. 

I'll let @erri120 comment/dig into this one a bit and give a recommendation.

erri120:
It's pretty clear that the sorting doesn't work properly because the item models aren't activated if they are out of view which means every component isn't activated as well and some components aren't added because the subscriptions for adding components aren't active.

Sorting (computer science) works by comparing two values, but in our case item models and components that haven't been activated don't have a value. Regardless of how the Avalonia TreeDataGrid code does sorting, even if we completely replace it, we'll still have the issue of having no values. The only solution for that is to activate every model in the TreeDataGrid regardless of whether they are in view or not. @Al12rs and I have been chatting about this from time to time and performance is the only reason we're currently only activating item models that are in view.

I think we can activate every item model if we improve many of the performance issues we currently have. It would definitely simplify a bunch of code we have, we already have many places where subscriptions are active even if the item model isn't in view because selections have to work on all items (eg: select all items and hit "remove", it should remove all items and not just the items in view).

[...]

PR #2739, opened 2025-02-26
Activate every model

erri120:
Related to #2390. Also implements the action column ordering for the collection download page.

### DIFF
--- a/src/NexusMods.App.UI/Controls/TreeDataGrid/CompositeItemModel.cs
+++ b/src/NexusMods.App.UI/Controls/TreeDataGrid/CompositeItemModel.cs
@@ -22,6 +22,7 @@ public sealed class CompositeItemModel<TKey> : TreeDataGridItemModel<CompositeIt
 
     private readonly Dictionary<ComponentKey, IDisposable> _observableSubscriptions = new();
     private readonly ObservableDictionary<ComponentKey, IItemModelComponent> _components = new();
+    private readonly CompositeDisposable _trackedDisposables = new();
 
     /// <summary>
     /// Gets the dictionary of all components currently in the item model.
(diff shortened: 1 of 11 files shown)


Al12rs (changes requested):
Current code has some issues:

- Child models are never activated
- `BeforeModelActivationHook` and `BeforeModelDeactivationHook` are never called, meaning that some subscriptions are missing

erri120:
> Current code has some issues:
> 
>     * Child models are never activated (or deactivated)
> 
>     * `BeforeModelActivationHook` and `BeforeModelDeactivationHook` are never called, meaning that some subscriptions are missing

Should be fixed. I'll probably re-do some parts of this tomorrow.

halgari:
Aside from the above comments, this seems to be working fine to me.

erri120:
I've updated the code with new activation logic. Root rows now get activated before TreeDataGrid gets to see them, this should give the observables enough time to fill in data before the first render. Child rows are still activated via the previous TreeDataGrid `RowPrepared` functionality.

A big change was the removal of row deactivation. If our models are always activated, the only time when they would be deactivated is when they would also be disposed. If we already dispose, we don't need to deactivate, it would be a redundant step. `DisposeMany` from Dynamic Data will dispose any removed items, I've verified that this works using the observable tracker page.

This also meant that `BeforeModelDeactivationHook` can be removed. The subscriptions that get set up in `BeforeModelActivationHook` are now tracked by the model itself instead of the adapter.

src/NexusMods.App.UI/Controls/TreeDataGrid/CompositeItemModel.cs
```diff
@@ -315,6 +316,22 @@ public IDisposable SubscribeToComponent<TComponent, TState>(
...
     }
 
+    /// <summary>
+    /// Subscribes to a component with the given key in the item model.
+    /// </summary>
+    /// <remarks>
+    /// Differs from <see cref="SubscribeToComponent{TComponent,TState}"/> in that the
+    /// disposable is tracked by the model itself.
+    /// </remarks>
+    public void TrackSubscriptionToComponent<TComponent, TState>(
```
Al12rs, line 326: Suggestion:
`SubscribeAndTrackToComponent`

Al12rs (commented):
I have just run into a deadlock while testing this.

I also think we may be leaking stuff since memory usage kept going up after each time the library or other tree view was opened. I got to 2.7GB before the deadlock.

Sewer56:
> I also think we may be leaking stuff since memory usage kept going up after each time the library or other tree view was opened. 

I've experienced something like this on main branch too to some extent. Not sure if related.

halgari:
Merged in #2776 after I fixed the tests. We'll have to make a pass on RedMod sorting as part of the Cyberpunk sprint coming up.

PR #2742, opened 2025-02-26
Added: Miscellaneous Bug Fixes around Updates

### LINKED ISSUE #2678: Mod update check is inaccurate
# Bug Report

## Summary
Our current mod update check is very approximate and can result in inaccurate update data and unexpected behaviour.

## Steps to reproduce
SMAPI:
- Download SMAPI 3.18.2 
- Press update button
- 4.0.8 will be installed rather than 4.1.10 or grater

SpaceCore (https://www.nexusmods.com/stardewvalley/mods/1348?tab=files):
- Download 1.2- SpaceCore - 1.0.1
- No updates are detected instead of getting the latest one
- Download 1.16.1
- Press download update 
- 1.27.0 is downloaded correctly
- but download update button is still available
- pressing button again downloads 1.23.1?

### LINKED ISSUE #2711: Updating one mod removes update button from other mods
Updating one mod removes update button from other mods. You have to navigate away from the library and then back again to get the update buttons to appear again.

### LINKED ISSUE #2712: Updating a mod in the library drop down keeps downloading different versions of the mod
If you are in the library and you expand a mod view, if that mod has an update and you use the button from the expanded view and it has multiple versions between the one its on and the update, you can keep clicking update and it keeps downloading all the versions one at a time.

### LINKED ISSUE #2747: UI update version discrepancy
# Bug Report

## Summary
_(Summarize the bug encountered concisely ensuring you state the current behaviour)_

Image

Following on from Slack conversation with @Sewer56 

> It's selecting the highest version as the previous version (as intended). But doesn't check if that highest version has an update.

### LINKED ISSUE #2748: Library Download Update button number is inconsistent
Download two older versions of Content patcher with mod manager
Number is 1:
Image

Press update button and then delete the update,
Number is 2:
Image

Not actually sure what the expected number should be, 
- 1 as that is how many files will be downloaded?
- 2 as that is how many library items need to be updated?

### LINKED ISSUE #2759: Bug: Updates for Items from Collections Don't Work
## Summary

Checking for updates for items (mods) downloaded via collections is broken.

## Steps to reproduce

1. Download Collection
2. Download Mods
3. Mods from collection don't get updates.

Mods from collections get updates.

## Other information

This happens because we don't fetch files when fetching the mod page when downloading a collection.
But the 'Last Updated' timestamp is still updated, meaning update check wouldn't update the files.

Normally we fetch files when doing single downloads, but not with collections.

Sewer56:
- fixes #2711
- fixes #2712
- fixes #2678
- fixes #2747
- fixes #2748
- fixes #2759

Basically what I got done today, and a bit of yesterday.
Fixes the reported bugs around updating, and adds a bunch of tests around the functionality; to ensure it never breaks again.


src/Networking/NexusMods.Networking.NexusWebApi/ModUpdateService.cs
```diff
@@ -109,17 +115,17 @@ public async Task<PerFeedCacheUpdaterResult<PageMetadataMixin>> CheckAndUpdateMo
...
+                return PerFeedCacheUpdaterResult<PageMetadataMixin>.WithStatus(CacheUpdaterResultStatus.Throttled);
             }
-            _lastUpdateCheckTime = DateTimeOffset.UtcNow;
+            _lastUpdateCheckTime = _timeProvider.GetUtcNow();
         }
         else
         {
-            _lastUpdateCheckTime = DateTimeOffset.UtcNow;
+            _lastUpdateCheckTime = _timeProvider.GetUtcNow();
         }
```
Al12rs, line 129: You can just set the lastUpdateTime outside the if, since the only case you don't want it has an early return

Al12rs: Also update the text of the log to say throttling rather than rate limiting.

Sewer56: Solved

Al12rs (approved):

PR #2744, opened 2025-02-27
Upgrade to the latest version of MnemonicDB

halgari:
Upgrades to the latest version of MnemonicDB. There's only one small semantic change in this version: the In-memory DB no longer exists in its previous form, instead we use RocksDB but with a in-memory environment. Essentially it's normal RocksDB but with files kept in memory instead of on disk. This means future versions of MnemonicDB can tie closer to RocksDB removing some layers of abstractions. 

There's also a lot of microoptimizations in this version, such as using structs internally and keeping data on the stack while reading it into IndexSegments. None of these changes require our code to change.


src/NexusMods.DataModel/Services.cs
```diff
@@ -44,39 +45,21 @@ public static IServiceCollection AddDataModel(this IServiceCollection coll)
...
-            if (settings.UseInMemoryDataModel)
-            {
-                return sp.GetRequiredService<MnemonicDB.Storage.InMemoryBackend.Backend>();
-            }
-            else
-            {
-                var datomStoreSettings = sp.GetRequiredService<DatomStoreSettings>();
-
-                if (!datomStoreSettings.Path.DirectoryExists()) 
-                    datomStoreSettings.Path.CreateDirectory();
```
Al12rs, line 75: Since this was removed, the App will fail to start from a clean setup, due to:
```
RocksDbSharp.RocksDbException: IO error: No such file or directory: Failed to create a directory: C:/Users/Al12r/AppData/Local/NexusMods.App/DataModel/MnemonicDB.rocksdb: The system cannot find the path specified.
```

Issue #2687, opened 2025-02-19
[WIP] Stardew Valley - Health Check issues


Pickysaurus:
# Bug Report

## Summary
A working list of my findings from testing the SDV Health Check messages

- [ ] RequiredDependencyIsOutdated
	- Appears when the dependency is disabled on the loadout - should only appear when enabled
- [ ] SMAPIRequiredButDisabled
 	- Does not appear when SMAPI is in a disabled collection.
- [ ] ModCompatabilityObsolete
 	- Did not work for https://www.nexusmods.com/stardewvalley/mods/501?tab=files
- [ ] ModCompatabilityAssumeBroken
 	- Did not work for https://www.nexusmods.com/stardewvalley/mods/10755
 	- Worked for https://www.nexusmods.com/stardewvalley/mods/16881

- [ ] GameVersionOlderThanMinimumGameVersionTemplate
     - Did not trigger with game version 1.6.12 and SMAPI 4.1.10 
     ![Image](https://github.com/user-attachments/assets/b44883bc-c44c-46c7-b36b-e755e58bb58f)

Pickysaurus:
I've gone through all of them and the above list is all the issues I could find.

Al12rs:
https://smapi.io/mods

Al12rs:
@Pickysaurus I investigated the last one and there is no entry for in in the [`metadata.json`](https://github.com/Pathoschild/SMAPI/blob/fa7fed4ed3b20e0f974e225ea5ef46afef658ec2/src/SMAPI.Web/wwwroot/SMAPI.metadata.json) file that SMAPI and the app use for compatibility checking.

That mod is marked as broken in the smapi logs after SMAPI actually fails to manually load it:
https://github.com/Pathoschild/SMAPI/blob/fa7fed4ed3b20e0f974e225ea5ef46afef658ec2/src/SMAPI/Framework/SCore.cs#L1996-L2007

The mod is apparently not added to the list of broken mods in the manifest on purpose, since it is preferred to have the stack trace of the failure rather than a generic `AssumedBroken` message:
https://github.com/Pathoschild/SMAPI/blob/fa7fed4ed3b20e0f974e225ea5ef46afef658ec2/src/SMAPI.Web/wwwroot/SMAPI.metadata.json#L44-L45

[...]

PR #2745, opened 2025-02-27
SDV Health Check fixes

Al12rs:
- Fixes parts of #2687

Fixed three bugs, each covered by new tests cases:

- RequiredDependencyIsOutdated
Appears when the dependency is disabled on the loadout - should only appear when enabled 
- SMAPIRequiredButDisabled
Does not appear when SMAPI is in a disabled collection.
- DisabledRequiredDependency
Appears if there is at least a disabled version of the dependency, even though there are enabled versions

I think remaining bugs listed on #2687 might be fixed by updating SMAPI submodule, but I haven't gotten around to looking into that yet.

### DIFF
--- a/tests/Games/NexusMods.Games.StardewValley.Tests/DependencyDiagnosticEmitterTests.cs
+++ b/tests/Games/NexusMods.Games.StardewValley.Tests/DependencyDiagnosticEmitterTests.cs
@@ -72,7 +72,22 @@ public class DependencyDiagnosticEmitterTests : ALoadoutDiagnosticEmitterTest<De
 
         var dependent = disabledRequiredDependencyMessageData.SMAPIMod.ResolveData(ServiceProvider, Connection);
         dependent.AsLoadoutItem().ParentId.Should().Be(farmTypeManager.LoadoutItemGroupId, because: "Farm Type Manager is the dependent");
+        
+        // Add extra enabled copy in another collection
+        var collectionA = (await CreateCollection(loadout.Id, "Collection A")).AsLoadoutItemGroup().LoadoutItemGroupId;
+        // Content Patcher 2.5.3 (https://www.nexusmods.com/stardewvalley/mods/1915?tab=files)
+        var contentPatcherCollectionA = await InstallModFromNexusMods(loadout, ModId.From(1915), FileId.From(124659), parent: collectionA);
+        
+        await ShouldHaveNoDiagnostics(loadout, because: "The dependency ContentPatcher in collectionA  is enabled");
 
+        await DisableItem(collectionA);
+        var parentDisabledDiagnostic = await GetSingleDiagnostic(loadout);
+        diagnostic.Should().BeOfType<Diagnostic<Diagnostics.DisabledRequiredDependencyMessageData>>(because: "Content Patcher parent collection is disabled and required by Farm Type Manager");
+        
+        await EnableItem(contentPatcher);
+        
+        await ShouldHaveNoDiagnostics(loadout, because: "The dependency ContentPatcher outside the collection is enabled");
+        
         await VerifyDiagnostic(diagnostic);
     }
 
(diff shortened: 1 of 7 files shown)


tests/Games/NexusMods.Games.StardewValley.Tests/DependencyDiagnosticEmitterTests.cs
```diff
@@ -72,7 +72,22 @@ public async Task Test_DisabledRequiredDependencyMessageData()
...
+        // Add extra enabled copy in another collection
+        var collectionA = (await CreateCollection(loadout.Id, "Collection A")).AsLoadoutItemGroup().LoadoutItemGroupId;
+        // Content Patcher 2.5.3 (https://www.nexusmods.com/stardewvalley/mods/1915?tab=files)
+        var contentPatcherCollectionA = await InstallModFromNexusMods(loadout, ModId.From(1915), FileId.From(124659), parent: collectionA);
+        
+        await ShouldHaveNoDiagnostics(loadout, because: "The dependency ContentPatcher in collectionA  is enabled");
+
+        await DisableItem(collectionA);
+        var parentDisabledDiagnostic = await GetSingleDiagnostic(loadout);
+        diagnostic.Should().BeOfType<Diagnostic<Diagnostics.DisabledRequiredDependencyMessageData>>(because: "Content Patcher parent collection is disabled and required by Farm Type Manager");
```
erri120, line 85: You're asserting on `diagnostic` instead of `parentDisabledDiagnostic`.

tests/Games/NexusMods.Games.StardewValley.Tests/DependencyDiagnosticEmitterTests.cs
```diff
@@ -102,6 +117,19 @@ public async Task Test_RequiredDependencyIsOutdated()
         var dependent = requiredDependencyIsOutdatedMessageData.Dependent.ResolveData(ServiceProvider, Connection);
         dependent.AsLoadoutItem().ParentId.Should().Be(farmTypeManager.LoadoutItemGroupId, because: "Farm Type Manager is the dependent");
 
+        // Test disabled cases
+        await DisableItem(farmTypeManager);
+        
+        (await GetAllDiagnostics(loadout)).OfType<Diagnostic<Diagnostics.RequiredDependencyIsOutdatedMessageData>>()
+            .Should().BeEmpty(because: "The dependant Farm Type Manager is disabled");
```
erri120, line 124: This should just be `ShouldHaveNoDiagnostics`.

tests/Games/NexusMods.Games.StardewValley.Tests/DependencyDiagnosticEmitterTests.cs
```diff
@@ -102,6 +117,19 @@ public async Task Test_RequiredDependencyIsOutdated()
...
+        
+        await EnableItem(farmTypeManager);
+        await DisableItem(contentPatcher);
+        
+        (await GetAllDiagnostics(loadout)).OfType<Diagnostic<Diagnostics.RequiredDependencyIsOutdatedMessageData>>()
+            .Should().BeEmpty(because: "The outdated dependency Content Patcher is disabled");
```
erri120, line 130: This should just be `ShouldHaveNoDiagnostics`.

erri120 (changes requested):
Just some minor test changes, you should also move the collection specific test cases into their own test methods.

tests/Games/NexusMods.Games.StardewValley.Tests/DependencyDiagnosticEmitterTests.cs
Al12rs: Good catch

tests/Games/NexusMods.Games.StardewValley.Tests/DependencyDiagnosticEmitterTests.cs
Al12rs: That wouldn't work because you actually get the `DisabledRequiredDependencyMessageData` instead

erri120: Right, then you should assert that you got a single diagnostic with `DisabledRequiredDependencyMessageData` and no diagnostics with `RequiredDependencyIsOutdatedMessageData`.

Al12rs: I think we should get `MissingRequiredDependency` instead tbh, basically check that the disabled dependency actually matches min required version.

tests/Games/NexusMods.Games.StardewValley.Tests/DependencyDiagnosticEmitterTests.cs
Al12rs: Same as above, we get a different diagnostic if it is disabled.

erri120 (approved):

PR #2758, opened 2025-02-27
Create workflow for issue maintenance

erri120:
Can be expanded upon depending on our needs. Currently has two jobs:

1) Writes a comment if an issue gets labeled with `needs logs`
2) Writes a comment if an issue gets closed as completed and has the `user raised` label


Pickysaurus (approved):
Fine for now, we can refine it later.

Issue #2372, opened 2024-12-12
Use localized string representation of numbers


Pickysaurus:
# Improvement

Relates to https://github.com/Nexus-Mods/NexusMods.App/issues/1694

## Currently 
When we show larger numbers in the app, we don't comma-separate them. (e.g. `1000`)

## We want
To add comma-separation for consistency with the website. 

We may need to discuss localisations here, as in some countries they use decimal places instead of commas (`1.000` instead of `1,000`)

## Supporting Information
_(Include any supporting screenshots etc to support the above)_

## Design
_(Any designs to required for this feature)_

erri120:
![Image](https://github.com/user-attachments/assets/41b7d841-13d3-4695-b00d-44479778f435)

The default integer representation uses the "general format" (`G`). We probably want `N` for numbers.

MattSturgeon:
> We may need to discuss localisations here, as in some countries they use decimal places instead of commas (`1.000` instead of `1,000`)

I know in Java there's a "localized number format(er)" that can use comma, dot, apostrophe, etc as appropriate for the current l10n. Presumably .NET has something similar?

Pickysaurus:
Further example of this:

![Image](https://github.com/user-attachments/assets/55ac80c0-64d6-465c-9c30-af0f653d352b)

PR #2762, opened 2025-02-27
Displaying Sizes and Numbers using Humanize

insomnious:
It's been a long time coming but here goes...

Size (`NexusMods.Paths.Size`) now has a ReactiveUI `IBindingTypeConverter` which is globally registered. Whenever a Size is bound to a UI control (normally a TextBlock's Text property) it automatically kicks in.

When SizeComponent is used in a TreeDataGrid, Humanize is directly used without the above converter.

Humanize returns either a single decimal point if 1 GB in size or greater, no decimal points if not.

Numbers can be formatted either by Humanize (if need it to say 1.1k instead of 1123) or by using "N0" format for it to have thousand separators. Generally applied in the code-behind in the OneWayBind's

Should we also be using Humanize if we just want thousand separators ?

image

image

Closes https://github.com/Nexus-Mods/NexusMods.App/issues/2372

### DIFF
--- a/src/NexusMods.App.UI/Controls/TreeDataGrid/SizeComponent.cs
+++ b/src/NexusMods.App.UI/Controls/TreeDataGrid/SizeComponent.cs
@@ -14,7 +14,13 @@ public sealed class SizeComponent : AFormattedValueComponent<Size>, IItemModelCo
 {
     /// <inheritdoc/>
     public int CompareTo(SizeComponent? other) => Value.Value.CompareTo(other?.Value.Value ?? Size.Zero);
-    private static string _FormatValue(Size value) => ByteSize.FromBytes(value.Value).Humanize();
+
+    private static string _FormatValue(Size value)
+    {
+        var byteSize = ByteSize.FromBytes(value.Value);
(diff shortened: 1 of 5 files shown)


src/NexusMods.App.UI/Controls/TreeDataGrid/SizeComponent.cs
```diff
@@ -14,22 +14,38 @@ public sealed class SizeComponent : AFormattedValueComponent<Size>, IItemModelCo
 {
     /// <inheritdoc/>
     public int CompareTo(SizeComponent? other) => Value.Value.CompareTo(other?.Value.Value ?? Size.Zero);
-    private static string _FormatValue(Size value) => ByteSize.FromBytes(value.Value).Humanize();
+
+    private static string _FormatValue(Size value)
+    {
+        var byteSize = ByteSize.FromBytes(value.Value);
+        return byteSize.Gigabytes < 1 ? byteSize.Humanize("0") : byteSize.Humanize("0.0");
```
erri120, line 21: What is this supposed to achieve?

erri120 (approved):

PR #2767, opened 2025-03-02
[CLI] Hyperlinks for collection downloads for free users

### LINKED ISSUE #2766: CLI opens tabs for all downloads for free users
# Bug Report

## Summary
When installing a large collection in CLI using `install-collection`, it opens all required mods that are not downloaded in your local browser.

## Steps to reproduce
1. Assume that the user is not premium.
2. Run `install-collection -l <loadout> -s <slug> -r <revision>`
3. A bunch of tabs will open that directs the user to download each mod.

It should limit the user to opening URLs with a limit (ex. 10 or less mods).

## Other information
PR: #2767

axmanalad:
Fixes #2766 

This PR provides hyperlinks to all mods that are missing for collections when a free user tries to install a collection through CLI. Instead of opening all mod URLs at once for free users, this also cancels out future jobs which would have otherwise made an more exceptions due to missing mods.

Original PR: *This PR gets the number of required downloads for a collection and checks if it passes the limit condition (in this case 10 or less) before opening the URLs (only for free users).*

### DIFF
--- a/src/NexusMods.Collections/CollectionDownloader.cs
+++ b/src/NexusMods.Collections/CollectionDownloader.cs
@@ -281,25 +281,25 @@ public class CollectionDownloader
     }
 
     /// <summary>
-    /// Returns the hyperlinks for the missing items in CLI format.
+    /// Returns all missing downloads and Uris.
     /// </summary>
-    public static List<string> GetMissingHyperlinks(CollectionRevisionMetadata.ReadOnly revision, IDb db, ItemType itemType = ItemType.Required)
+    public static IReadOnlyList<(CollectionDownload.ReadOnly Download, Uri Uri)> GetMissingDownloadLinks(CollectionRevisionMetadata.ReadOnly revision, IDb db, ItemType itemType = ItemType.Required)
     {
-        var missingDownloads = revision.Downloads.Where(download =>
-            DownloadMatchesItemType(download, itemType) &&
-            !GetStatus(download, db).IsDownloaded()).ToArray();
-        
-        var hyperlinks = new List<string>();
-        
-        foreach (var item in missingDownloads)
+        var results = new List<(CollectionDownload.ReadOnly Download, Uri Uri)>();
+        var downloads = GetItems(revision, itemType).Where(download => GetStatus(download, db).IsNotDownloaded());
+
+        foreach (var download in downloads)
         {
-            if (!item.TryGetAsCollectionDownloadNexusMods(out var d)) continue;
-            var url = d.FileMetadata.GetUri();
-            var text = $"{d.FileMetadata.Name}";
-            hyperlinks.Add($" - \u001b]8;;{url}\u0007{text}\u001b]8;;\u0007");
+            if (download.TryGetAsCollectionDownloadNexusMods(out var nexusModsDownload))
+            {
+                results.Add((download, nexusModsDownload.FileMetadata.GetUri()));
+            } else if (download.TryGetAsCollectionDownloadExternal(out var externalDownload))
+            {
+                results.Add((download, externalDownload.Uri));
+            }
         }
 
-        return hyperlinks;
+        return results;
     }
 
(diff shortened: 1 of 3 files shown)


src/NexusMods.Collections/Verbs.cs
```diff
@@ -47,8 +47,18 @@ private static async Task<int> InstallCollection([Injected] IRenderer renderer,
...
+        {
+            await collectionDownloader.DownloadItems(revisionMetadata, itemType: CollectionDownloader.ItemType.Required, db: connection.Db,
+                cancellationToken: token
+            );
+        }
+        catch (OperationCanceledException e)
+        {
+            await renderer.Error(e.Message);
+            return 0;
+        }
```
erri120, line 60: The code that you added to the `CollectionDownloader` regarding getting the links should be added here.

You'd wanna check with the `ILoginManager` whether the user is premium or not, if they are you can use `collectionDownloader.DownloadItems`, if they are a free user then you should print all download links.

erri120 (changes requested):

src/NexusMods.Collections/CollectionDownloader.cs
```diff
@@ -280,6 +280,28 @@ public IObservable<bool> IsCollectionInstalledObservable(
...
+        foreach (var item in missingDownloads)
+        {
+            if (!item.TryGetAsCollectionDownloadNexusMods(out var d)) continue;
+            var url = d.FileMetadata.GetUri();
+            var text = $"{d.FileMetadata.Name}";
+            hyperlinks.Add($" - \u001b]8;;{url}\u0007{text}\u001b]8;;\u0007");
+        }
+
+        return hyperlinks;
+    }
```
erri120, line 303:
```suggestion
    /// <summary>
    /// Returns all missing downloads and Uris.
    /// </summary>
    public static IReadOnlyList<(CollectionDownload.ReadOnly Download, Uri Uri)> GetMissingDownloadLinks(CollectionRevisionMetadata.ReadOnly revision, IDb db, ItemType itemType = ItemType.Required)
    {
        var results = new List<(CollectionDownload.ReadOnly Download, Uri Uri)>();
        var downloads = GetItems(revision, itemType).Where(download => GetStatus(download, db).IsNotDownloaded());

        foreach (var download in downloads)
        {
            if (download.TryGetAsCollectionDownloadNexusMods(out var nexusModsDownload))
            {
                results.Add((download, nexusModsDownload.FileMetadata.GetUri()));
            } else if (download.TryGetAsCollectionDownloadExternal(out var externalDownload))
            {
                results.Add((download, externalDownload.Uri));
            }
        }

        return results;
    }
```

src/NexusMods.Collections/Verbs.cs
```diff
@@ -47,8 +50,22 @@ private static async Task<int> InstallCollection([Injected] IRenderer renderer,
...
+            );
+        }
+        else
+        {
+            var links = CollectionDownloader.GetMissingHyperlinks(revisionMetadata, db: connection.Db);
+            if (links.Count > 0)
+            {
+                await renderer.Text("The following mods are missing from the collection:" + Environment.NewLine + string.Join(Environment.NewLine, links));
+                return 0;
+            }
```
erri120, line 66:
```suggestion
            var tuples = CollectionDownloader.GetMissingDownloadLinks(revisionMetadata, db: connection.Db, itemType: CollectionDownloader.ItemType.Required);
            if (tuples.Count > 0)
            {
                await renderer.TextLine($"Missing {tuples.Count} downloads:");
                await renderer.Table(["Name", "Uri"], tuples.Select(t => new object[] { t.Download.Name, t.Uri.ToString() }));
                return 0;
            }
```

erri120 (changes requested):
Just some small cleanup and then we can merge.

src/NexusMods.Collections/Verbs.cs
erri120: We can render the results as a table with two columns `Name` and `Uri`.

erri120 (approved):

Pickysaurus:
Just a note on this: opening lots of NexusMods.com web pages in a short time is the fastest way to get rate limited!

PR #2773, opened 2025-03-04
Fix deadlock for component subscriptions

erri120:
@Al12rs and I found another deadlock that happened because the DB event thread was calling `R3.BehaviorSubject.OnNext` which locked the subject and then triggered an activation on the UI thread which does a subscribe on the subject which is locked and thus we got a deadlock.

The subject in question was on the collection download page https://github.com/Nexus-Mods/NexusMods.App/blob/e18a22d31d51de5618f53facea31168bfb5f20eb/src/NexusMods.App.UI/Pages/CollectionDataProvider.cs#L140 the entire method needs some cleaning up later.

This PR should prevent any future deadlocks from happening in regards to subscriptions.


Al12rs (approved):

Al12rs:
I'm worried that we are only fixing the symptoms rather than the problem

PR #2786, opened 2025-03-05
Hide game hashes diagnostic message for now. 

Pickysaurus:
Don't show this error from the Health Checks as users cannot resolve this issue through the UI:
image


Al12rs:
The diagnostic should remain, but the text should be updated to more helpful to user trying to solve the issue.

In particular it should suggest verifying the game files using the game store (like steam or GOG).

PR #2795, opened 2025-03-05
Add correct icons and loadout resource tweak

insomnious:
Part of #2598

Added correct icons and a quick control change in the TDG Loadout Resources to get started


Al12rs:
This is included in #2805

PR #2802, opened 2025-03-06
Disable DBus menu for X11

### LINKED ISSUE #2799: X11 Tooltip exception investigation
On some Linux systems you get a DBus exception `The name is not activatable` when hovering over an item that will open a tooltip.

Here are my findings:

Using `dbus-monitor` we can see the messages being send over DBus. When a tooltip opens, this happens:

```
method call time=1741258811.468008 sender=:1.142 -> destination=com.canonical.AppMenu.Registrar serial=144 path=/com/canonical/AppMenu/Registrar; interface=com.canonical.AppMenu.Registrar; member=RegisterWindow
   uint32 27263131
   object path "/net/avaloniaui/dbusmenu/db202851e5414c6989c5773735428af5"
error time=1741258811.468016 sender=org.freedesktop.DBus -> destination=:1.142 error_name=org.freedesktop.DBus.Error.ServiceUnknown reply_serial=144
   string "The name is not activatable"
```

The interface that causes the issue is `com.canonical.AppMenu.Registrar`. The method on that interface `RegisterWindow` is called by Avalonia X11 for every new window:

https://github.com/AvaloniaUI/Avalonia/blob/603a2bdb43b7e8b2d1c8919f9d0669af7b33f117/src/Avalonia.FreeDesktop/DBusMenuExporter.cs#L110-L122

https://github.com/AvaloniaUI/Avalonia/blob/603a2bdb43b7e8b2d1c8919f9d0669af7b33f117/src/Avalonia.X11/X11Window.cs#L239-L240

The reason we get this for tooltips is because tooltips are implemented as popups, which are implemented as new X11 windows:

https://github.com/AvaloniaUI/Avalonia/blob/603a2bdb43b7e8b2d1c8919f9d0669af7b33f117/src/Avalonia.X11/X11Window.cs#L1221-L1222

As such, when a tooltip opens up, we create a new X11 Window, which tries to register the window on the DBus.

We can either disable DBus native menus or simply don't have tooltips as new X11 windows.

erri120:
Fixes #2799. See #2799 for details.

### DIFF
--- a/src/NexusMods.App/Startup.cs
+++ b/src/NexusMods.App/Startup.cs
@@ -100,13 +100,21 @@ public class Startup
         var app = AppBuilder
             .Configure(serviceProvider.GetRequiredService<App>)
             .UsePlatformDetect()
+            .With(new X11PlatformOptions
+            {
+                // NOTE(erri120): Prevents DBus exceptions for tooltips.
+                // For details see https://github.com/Nexus-Mods/NexusMods.App/issues/2799
+                UseDBusMenu = false,
+            })
+            .With(new SkiaOptions
+            {
+                // NOTE(insomnious): Opacity doesn't work on SVGs without this SkiaOptions set. It's needed for when we want to disable\fade SVG icons.
+                // For details see https://github.com/AvaloniaUI/Avalonia/pull/9964
+                UseOpacitySaveLayer = true,
+            })
             .LogToTrace()
             .UseR3()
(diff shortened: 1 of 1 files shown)


Sewer56:
Also closes:
- #2640

Sewer56 (approved):
Works as intended.
This obviously doesn't fix the underlying issue, but at least it prevents the error dialog.

Issue #2782, opened 2025-03-05
Missing "off" state for mod pages in a disabled collection


captainsandypants:
# Improvement

## Currently 
Mod page items have a "on" toggle state even when they are disabled by their parent collection. 

![Image](https://github.com/user-attachments/assets/32fcfd5c-75fd-4b90-92fd-e86c24dc32e1)

## We want
To give them the "off" toggle state described in the designs.

## Design
[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=2360-51132&t=JTwrRyO8CHnWrR7Y-1)

![Image](https://github.com/user-attachments/assets/64dc38e5-21bc-451e-9ffe-fa755f7ee191)

PR #2805, opened 2025-03-06
Refactor and separate `IsEnabled` loadout column component into separate states

Al12rs:
- Part of #2782 
- And part of #2598 

The `IsEnabled` component contained the normal toggle, plus the `IsLocked` state plus the `LoadoutItemIds`.
To implement all the possible state combinations for parent and child items it is better to separate these.

This also includes the UI and icons changes by @insomnious from #2795

This shouldn't change the current behaviour much, except for showing the Toggle rather than parent items showing Toggle rather than Locked state if at least one child is Togglable.

### DIFF
--- a/src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutComponents.cs
+++ b/src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutComponents.cs
@@ -96,18 +96,6 @@ public static class LoadoutComponents
             });
         }
 
-        public EnabledStateToggle(
-            ValueComponent<bool?> valueComponent,
-            IObservable<IChangeSet<LoadoutItemId, EntityId>> childrenItemIdsObservable)
-        {
-            _valueComponent = valueComponent;
-
-            _activationDisposable = this.WhenActivated((childrenItemIdsObservable), static (self, state, disposables) =>
-            {
-                self._valueComponent.Activate().AddTo(disposables);
-            });
-        }
-
         private bool _isDisposed;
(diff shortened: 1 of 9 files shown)


src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutComponents.cs
```diff
@@ -48,39 +84,28 @@ public int CompareTo(IsEnabled? other)
...
                 self._valueComponent.Activate().AddTo(disposables);
             });
         }
 
-        public IsEnabled(
+        public EnabledStateToggle(
             ValueComponent<bool?> valueComponent,
-            IObservable<IChangeSet<LoadoutItemId, EntityId>> childrenItemIdsObservable,
-            ValueComponent<bool> isLockedComponent)
+            IObservable<IChangeSet<LoadoutItemId, EntityId>> childrenItemIdsObservable)
```
erri120, line 101: This is unused and can be removed. The entire constructor can be removed.

Al12rs: It seems to be used though

erri120: It's passed as state into `WhenActivated` below, but the state is never used.

Al12rs: Ah, I see what you mean

erri120 (approved):

Issue #2806, opened 2025-03-06
Revamp diagnostic system


erri120:
The first iteration of the diagnostic system is slowly reaching it's limit on what's possible without a revamp. While we've been steadily improving the user-facing messages, we don't have interactivity yet. Many diagnostic emitters needed to be patched up to work with collections, with game files no longer being part of the loadout, or some other core changes we've recently done.

The current implementation of the diagnostic system has become tech debt, this epic encompasses both user-facing and backend changes.

Discussion points for meetings:

1) Installers put marker types on entities which are used by diagnostic emitters. This breaks when files are installed differently (eg: advanced installer, collection default installer). As an example, SMAPI mods are installed via the SMAPI installer which finds the `manifest.json` files and puts a marker on them. The various Stardew Valley diagnostics only look at loadout item groups that have a `manifest.json` file with the required marker. In essence, we only consider a mod to be a "SMAPI Mod" if it was installed via the "SMAPI Mod Installer". A different approach would be a data-driven approach where we'd a that a "SMAPI Mod" is any loadout item group with a SMAPI `manifest.json` file. The installer wouldn't put markers on the entities, it would just put the files in the correct place and the diagnostic emitters would have to find the mod based on some query. Since installers and emitters are linked currently via the markers, any change in one would require changes in other and possibly a migration.
2) Diagnostic emitters get the entire loadout and can do anything with it. We've hit problems where we create diagnostics for files that aren't enabled or that are not going to be written on disk because they are being overwritten by something else. The diagnostic emitters need to know how the files will be deployed. This also links to point number 1, the query could run on all files in the loadout or just on files to be deployed.
3) Interactivity of the rendered markdown. If we reference a loadout item it's currently being rendered as text. We can have interactivity by rendering a custom component that can take the user to the loadout item in the TreeDataGrid.
4) Automatic solutions to problems. If we find that you're missing SMAPI we currently just provide a link to SMAPI. Instead of doing that, for premium users we can offer a button instead that will automatically download and install the correct version of SMAPI for you. Other interactions could be added as well depending on the diagnostic.

Pickysaurus:
From my limited experience with the current system, it might be cool to have something like an `ActionItems` object alongside the `MessageData` that can include buttons with different styles (Primary, Secondary, Open External URL). For simplicity, we could always include the buttons at the bottom of the Diagnostic message or come up with fancy MD syntax to place them inside the message.

Al12rs:
[...]

PR #2808, opened 2025-03-06
Select highest installed version for SDV dependency diagnostics

### LINKED ISSUE #2807: Outdated dependency warning shows in health check when there is a newer version of the dependent mod installed
# Bug Report

## Summary
An outdated dependency warning can be triggered in the health check even when a newer version of the dependency is installed. 

## Steps to reproduce

- Install BirbCore
- Install an old version of Content Patcher e.g. 2.2.0
- Install a newer version of Content Patcher e.g. 2.5.3

I initially installed both of these mods from this collection (https://next.nexusmods.com/stardewvalley/collections/2d0gvh) and then installed the different versions into the My Mods collection afterwards

Image

- If there is a version of the dependency mod that is of the correct version or greater installed then it should be recognised and the error diagnostic should not be displayed

## Other information

Image

[nexusmods.app.slim.current.log](https://github.com/user-attachments/files/19111925/nexusmods.app.slim.current.log)
[nexusmods.app.main.current.log](https://github.com/user-attachments/files/19111924/nexusmods.app.main.current.log)

Al12rs:
- Fixes #2807

### DIFF
--- a/src/Games/NexusMods.Games.StardewValley/Emitters/DependencyDiagnosticEmitter.cs
+++ b/src/Games/NexusMods.Games.StardewValley/Emitters/DependencyDiagnosticEmitter.cs
@@ -77,9 +77,10 @@ public class DependencyDiagnosticEmitter : ILoadoutDiagnosticEmitter
         IGrouping<string, KeyValuePair<SMAPIModLoadoutItemId, SMAPIManifest>> g)
     {
         // TODO: select winner based on synchronizer winner
-        // For now prefer enabled mods over disabled ones
-        var enabledItem = g.FirstOrOptional(kv => LoadoutItem.Load(loadout.Db, kv.Key).IsEnabled());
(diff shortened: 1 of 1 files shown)


Sewer56 (approved):

Sewer56:
I would argue not to use LINQ in Diagnostics as that's a hot path that reacts to our UI actions and database changes but that's beyond the scope for this PR.

Pickysaurus:
@Al12rs should we be doing it this way? Really we should look at which version is winning the conflict and compare that?

Al12rs:
> @Al12rs should we be doing it this way? Really we should look at which version is winning the conflict and compare that?

Yes, but we don't have the facilities to do that at the moment and even if we did, users wouldn't have the facilities to control it either. So this is the most intuitive behaviour in the meantime.

- #2806 contains the stuff we need for the system to account for conflicts

PR #2810, opened 2025-03-10
Handle TaskCanceledException during delay in TrackingService

Al12rs:


erri120 (commented):
Why do we need this?

Al12rs:
The delay throws exception and it should just be handled. This happens on app close so not super impactful for users, but it is just annoying for me while debugging.

erri120 (approved):

PR #2815, opened 2025-03-10
Improved in-memory storage for pipeline results

erri120:
We've been using the resource cache wrong this entire time. The resource cache is for situations where you want a cache of "hot" items, meaning items that are frequently accessed (or other metrics).

The pipelines that used this cache were pipelines that parsed some loadout files (SMAPI manifests, BG3 pak files). The issue is that we didn't use these caches as caches but as in-memory stores. That's why we set arbitrary size limits of `300`. Additionally, we can run into situations where items would never be removed from the cache.

This PR introduces an actual in-memory store for resource pipeline data. It's thread-safe and has a mechanism of deleting items that are no-longer valid. I considered three approaches for the deleting mechanism:

1) using observables -> checking items to be deleted when the observable produces a value
2) timer based -> checking items periodically
3) access based -> checking items after someone accesses the data

I choose the timer approach because it can be used universally and fits better into the access patterns we currently have. The observable approach only works if you actually have an observable, otherwise you'd also be using a timer based approach. The access based mechanism would be too much of a performance hit, as you'd essentially have to check every time someone accesses the data. With our current access pattern, the timer based approach works the best.


Al12rs:
> I choose the timer approach because it can be used universally and fits better into the access patterns we currently have. The observable approach only works if you actually have an observable, otherwise you'd also be using a timer based approach.

Why not use the Observable based approach that uses a timer as the default? Wouldn't this offer the best flexibility?
I don't think I like the timer for the BG3 pak data for example. In some cases there is no reason to evict, except if there are too many items I guess.

Sewer56:
I've identified that as a growing part of our heap last Thursday, so it'd be good for me to measure here, hence self assign.

Timer based sounds good, this is what was actually pitched up to be added to MMDB, for disposing observables in cases where we have lack of DB writes. Right now MMDB only disposes datom observables on writes, so it's easy to rack up a lot in cache by just moving around the pages without changing a loadout, for instance.

This is a similar caveat, so is in theme with what I'm working with this time around.

erri120:
> Why not use the Observable based approach that uses a timer as the default? Wouldn't this offer the best flexibility? I don't think I like the timer for the BG3 pak data for example. In some cases there is no reason to evict, except if there are too many items I guess.

With the timer the in-memory store can act outside the DB observable shenanigans. That means you have batch deletes and you don't get congestion on the lock.

Sewer56:
And importantly, deletes during Idle periods, where they would not affect latency/throughput of operations as much.

Sewer56 (commented):
[This comment is regarding allocations only, nothing else]

There's a bunch of Heap allocations sprinkled in here.

Enough for me to comment on it; as we generate quite a bit of garbage on the GC heap (Gen0), while cleaning garbage elsewhere; and that's not quite ideal. The fact this runs on a timer in the background makes it a bit more acceptable; but I'm somewhat concerned about scalability long term.Some of the problems could be addressed over a longer span of time if needed.

They can be broken down into the following.

## Deletes (Consumer Side)

Some things to mind come with `getKeysToDelete`:

- Using LINQ to filter items creates an allocation.
- Loading the items one by one to determine if an entity exists.
- Allocating a final (empty) array with `ToArray`.

Two of these are avoidable here, the other (loading items in MMDB), not so much currently I believe.

There is for example, `<MMDBModel>.PrimaryAttribute.IsIn(entity);`, but I imagine that would return an outdated value, due to an out of date `IndexSegment` being used. I think an API on MMDB's side would do pretty good there. I would raise that as an issue.

## Deletes (Backend Side)

Calling `getKeysToDelete` in `await getKeysToDelete(_dictionary.Keys.ToArray(), cancellationToken);` performs an allocation to lift out all of the keys to a new array. That's roughly 56 bytes/element (excluding overhead of array object itself).

It would be preferable to do one of the following:

- Pass the dictionary KeyCollection itself. [1 allocation / run]
- Pass an enumerable that lifts out the needed items from KeyCollection. [2 allocations / run]
    - Can use [StructLinq](https://github.com/reegeek/StructLinq) or [LinqGen](https://github.com/cathei/LinqGen) to avoid the allocation, but not strictly needed.

-----

Basically allocations are fine, but per item allocations are expensive.

erri120:
@Sewer56 I can also change the delegate to be a predicate returning `ValueTask<bool>` for no allocations at all.

Sewer56 (commented):
It doesn't need to be lengthy but we also need some sort of guideline in the code to tell when to use the current cache method instead of the previous cache method. i.e. Outline cases where each is appropriate to use.

Specifically in order to avoid cases where someone may use the wrong method, like before. I think XML docs would be suitable for this. Say what a method does in short, and note `if you want to use for X purpose try Y API`.

src/Abstractions/NexusMods.Abstractions.Resources.Caching/InMemoryStore.cs
```diff
@@ -0,0 +1,243 @@
...
+                keyComparer: input.keyComparer,
+                shouldDeleteKey: input.shouldDeleteKey,
+                deleteDelay: input.deleteDelay,
+                inner: inner
+            )
+        );
+    }
+
+    /// <summary>
+    /// Store the results in-memory.
```
Sewer56, line 217: The code is good, but we still need to explain the relation between `StoreInMemory` (`InMemoryStore`), `UseCache` (`ResourceCache`) and `UseScopedCache` (`ScopedResourceCache`). Both are technically memory stores. 

Maybe link to https://github.com/bitfaster/BitFaster.Caching/wiki/IDisposable-and-Scoped-values#scoped for the scoped case. i.e. 'Prevent issues with disposal when there are multiple concurrent owners'/

erri120:
I'm probably gonna re-think the scoped cache in a separate PR.

Sewer56 (approved):

Al12rs:
I would like to discuss this a moment with @erri120 tomorrow morning before merging.

Sewer56:
Feel free. Whichever way we go, we eventually still need to avoid loading each entity one way or the other to determine if it should need evicting.

If it's through another strategy, or via MMDB PR, well, either works for me.

src/Games/NexusMods.Games.Larian/BaldursGate3/Pipelines.cs
```diff
@@ -49,10 +48,10 @@ public static IServiceCollection AddPipelines(this IServiceCollection serviceCol
                     }
                 }
             )
-            .UseCache(
-                keyGenerator: static hash => hash,
+            .StoreInMemory(
+                keySelector: static hash => hash,
                 keyComparer: EqualityComparer<Hash>.Default,
-                capacityPartition: new FavorWarmPartition(totalCapacity: 300)
+                shouldDeleteKey: async (tuple, _) => await fileStore.HaveFile(tuple.Item1)
```
Al12rs, line 54: I'm not sure I follow this.

I was thinking this should be something like:
```csharp
shouldDeleteKey: (tuple, _) => LoadoutFile.FindByHash(db, hash).Count == 0;
```

Remove if there are no loadout items that with this hash, we can have them in the library, but we don't care about those. So if the user removes a bunch of mods or deletes a loadout or something like that, all those entries can get evicted.

Al12rs (approved):

Issue #2778, opened 2025-03-05
Default sorts for views


Pickysaurus:
# Improvement

## Currently 
The default sort for various views is not well defined. 

## We want
@captainsandypants  to decide on the default sort order for:

- Collection Download page (#2777)
- Collection installed page 
- Library page
- All mods page
- My Mods collection

## Supporting Information
_(Include any supporting screenshots etc to support the above)_

## Design
_(Any designs to required for this feature)_

Issue #1229, opened 2024-04-17
Loadouts - Default sort order


LukeNexusMods:
The loadouts page currently does not have a default sort order. 

Until we receive user feedback about the default order we will use the installed date, most recently installed item appearing at the top.

A/C
Default sort order should be consistent on different pages of the app with the same content. 
When moving away from the app sort order does not change. (it currently forgets, then reorders)

Pickysaurus:
The default sort (to start) should be:
- **Mod Library** - Recently added, newest first
- **Loadout view** - Added to loadout, most recent first

erri120:
Done in https://github.com/Nexus-Mods/NexusMods.App/pull/2834.

PR #2834, opened 2025-03-11
Tree views sorting: Actions, Version, EnabledState and Collection columns

### LINKED ISSUE #2801: Actions do not sort updates in the library
https://github.com/user-attachments/assets/633072eb-c86a-4f14-b55e-5600ea5c5e7b

Updates should be sortable in the library.

Al12rs:
- Closes #2801 
- Part of #2778 
- Part of #1229

## Details:

- Tried implementing default sorting for TreeDataGrid views to mediocre results:
The sorting is applied while the tree is still being populated, and doesn't enforce sorting as new items come in, so it is only kinda sorted...
Edit: Removed since it was inconsistent
- Sorted Library Actions column as requested here https://github.com/Nexus-Mods/NexusMods.App/issues/2801#issuecomment-2710966531
- Sorted Library Version column as requested here: https://github.com/Nexus-Mods/NexusMods.App/issues/2801#issuecomment-2710977261
- Fixed bug in Collection and EnabledState column sorting, where same value was compared against itself

### DIFF
--- a/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
+++ b/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
@@ -125,16 +125,6 @@ public abstract class TreeDataGridAdapter<TModel, TKey> : ReactiveR3Object
                 .Subscribe()
                 .AddTo(disposables);
 
-            self.Source.AsObservable()
-                // ReSharper disable once ConditionIsAlwaysTrueOrFalseAccordingToNullableAPIContract
-                .Where(source => source is not null)
-                .Subscribe(static (source) =>
-                    {
-                        var sortColumn = source.Columns.FirstOrOptional(column => column.SortDirection is not null);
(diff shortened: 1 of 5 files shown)


src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
```diff
@@ -124,6 +125,16 @@ protected TreeDataGridAdapter()
...
+            self.Source.AsObservable()
+                // ReSharper disable once ConditionIsAlwaysTrueOrFalseAccordingToNullableAPIContract
+                .Where(source => source is not null)
+                .Subscribe(static (source) =>
+                    {
+                        var sortColumn = source.Columns.FirstOrOptional(column => column.SortDirection is not null);
+                        if (sortColumn.HasValue) source.SortBy(sortColumn.Value, sortColumn.Value.SortDirection!.Value);
+                    }
+                );
+
```
erri120, line 137: Let's remove this for now until we have something better.

erri120 (changes requested):

erri120 (approved):

PR #2836, opened 2025-03-11
Upgrade to the latest mnemonicdb and fix associated bugs

halgari:
This updates to the latest MnemonicDB. A few things had to change in this PR to remain compatible:

* Some `.Datoms` calls now return specific index types. For example `EntitySegment` is a much smaller version of `IndexSegment` that only stores the `A` and `V` parts of the datom resulting in a fairly substantial reduction in memory usage
* Some of our iterations over entities now use `.TryGetValue` which is much faster as internally it's setup to use .NET's SIMD operations over `Span<AttributeId>` 

Aside from the below changes, the rest of the API is hasn't changed, it uses less memory and is quite a bit faster

### DIFF
--- a/src/NexusMods.DataModel.SchemaVersions/Migrations/_0005_MD5Hashes.cs
+++ b/src/NexusMods.DataModel.SchemaVersions/Migrations/_0005_MD5Hashes.cs
@@ -40,7 +40,7 @@ internal class _0005_MD5Hashes : ITransactionalMigration
             .Select(entity =>
             {
                 // NOTE(erri120): need to use the IndexSegment because we don't want resolved datoms for removed attributes
-                foreach (var datom in entity.IndexSegment)
+                foreach (var datom in entity.EntitySegment)
                 {
                     if (attributesToRemove.Contains(datom.A)) return datom;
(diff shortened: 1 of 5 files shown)


Sewer56 (approved):

Sewer56 (commented):
Oops, wrong button.
Gimme a sec to check.

Sewer56 (approved):
Okay this matches the changes I had to make when I did the port prematurely.
I will presume this was tested beyond just booting App.

src/NexusMods.DataModel.SchemaVersions/Migrations/_0005_MD5Hashes.cs
```diff
@@ -40,7 +40,7 @@ public Task Prepare(IDb db)
             .Select(entity =>
             {
                 // NOTE(erri120): need to use the IndexSegment because we don't want resolved datoms for removed attributes
```
Sewer56, line 42: This comment should be renamed though.

halgari:
Reverted to a draft, I got an exception while testing just now.

PR #2850, opened 2025-03-13
Remove timeout and cancel exception from CliServer

erri120:
Before, the CliServer would create a cancellation token source that cancels after 1 second inside a while loop and use that token for `AcceptTcpClientAsync`. This meant that every second we're creating new token sources and every second we'd throw an `OperationCanceledException`.

This seemed very unusual, I couldn't figure out why we'd want that, so I updated the code to just use the main cancellation token.

### DIFF
--- a/src/NexusMods.SingleProcess/CliServer.cs
+++ b/src/NexusMods.SingleProcess/CliServer.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using System.Net;
 using System.Net.Sockets;
 using Microsoft.Extensions.Hosting;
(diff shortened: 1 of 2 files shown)


Al12rs:
I think the original reason may have been that this was taking too long to close and was keeping the application open. 
Maybe some issue with the cancellation of the `_cancellationTokenSource` ?

halgari:
One of the reasons here was to run the cleanup routines on a regular basis.
We can set the timeout longer, but we should make sure we don’t break
behavior with clients that are disconnecting

erri120:
> One of the reasons here was to run the cleanup routines on a regular basis. We can set the timeout longer, but we should make sure we don’t break behavior with clients that are disconnecting.

Ah you mean the `CleanClosedConnections` call in the while loop? All it does it remove completed `Task` objects from the `_runningClients` list. That could also be done by `HandleClientAsync` itself. I'll put up a commit to show what I mean.

erri120:
Updated the PR to remove finished tasks outside the while loop.

Al12rs (approved):

PR #2857, opened 2025-03-17
Reset login requests

### LINKED ISSUE #2062: Reset login requests
The login button is currently disabled while we wait for any login requests to be fulfilled or timed out after 5 minutes.

For a better UX, the button shouldn't be disabled. Consecutive login requests should cancel requests that came before them.

erri120:
Allows the user to re-try logging in without having to wait for the previous job to timeout.

Fixes #2062.


Sewer56 (approved):
Looks good to me.
At first porting all to R3 seemed overkill, but I see why it was done, for the await op.

PR #2858, opened 2025-03-17
"Remove Game"-overlay

### LINKED ISSUE #2662: When removing game give option to delete downloaded mods
As a user who is removing a game from the app
I want an option to delete downloaded mods for that game
So that I don't end up with files left behind on my drive

Currently if you remove a game it uninstalls all mods and collections. Any mods that you downloaded remain on your drive but cannot be seen in the app because you have removed the game. The files themselves have hashed names so you don't easily know which mod is which. 

@captainsandypants to provide designs

## Designs

[Figma link](https://www.figma.com/design/droQp7IUqRCq5AUGmXldvP/%F0%9F%93%B1-Add%2Fmanage-games?node-id=1087-5687&t=uHobRbnHR6dgv2tY-1)

Image

erri120:
Resolves #2662.


insomnious:
@captainsandypants added UI elements

image
image

captainsandypants:
@erri120 should be "Remove game and delete mods" not "Remove game, mods and delete downloads"

captainsandypants (approved):

src/NexusMods.App.UI/Pages/LocalFileDataProvider.cs
```diff
@@ -23,6 +24,8 @@ public LocalFileDataProvider(IServiceProvider serviceProvider)
         _connection = serviceProvider.GetRequiredService<IConnection>();
     }
 
+    public LibraryFile.ReadOnly[] GetAllFiles(GameId gameId, IDb? db = null) => [];
```
Al12rs, line 27: Add a comment explaining why this is empty (we don't have game information for local files)
```suggestion
    // TODO: update once we have game information on Local Files
    public LibraryFile.ReadOnly[] GetAllFiles(GameId gameId, IDb? db = null) => [];
```

src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
```diff
@@ -206,9 +232,17 @@ private OneOf<None, CreateLoadoutJob, UnmanageGameJob> GetJobRunningForGameInsta
...
     {
         await installation.GetGame().Synchronizer.UnManage(installation);
+
+        if (!shouldDeleteDownloads) return;
+        await _libraryService.RemoveItems(filesToDelete.Select(file => file.AsLibraryItem()));
+
+        foreach (var collection in collections)
+        {
+            await _collectionDownloader.DeleteCollection(collection);
+        }
```
Al12rs, line 245: I guess this is happening outside the unmanage Job then, should this be part of it instead?
Otherwise the app will say that unmanage has finished and the user might close the app while it is still deleting downloads and collections

Al12rs (commented):
I think we might want to move the removal of the items into the Unmanage job

src/NexusMods.App.UI/Pages/MyGames/MyGamesViewModel.cs
erri120: The widget will stay on the `RemovingGame` state until this task completes. I didn't want to put this stuff into the loadout synchronizer because our abstractions make this a complete mess.

Al12rs: I guess we would have the problem we had before of switching away and back and the state being reset.
I guess it would literally be simpler to show a modal with a spinner while the task is running, so the user can't navigate away.

Al12rs (approved):

PR #2871, opened 2025-03-18
UI styling refactor

insomnious:
A big ol PR for mostly behind the scene changes and to get us back in line with the current design system.

The point of this refactor is:

- Provide a consistent decision making workflow for how we use and style controls in Avalonia
- Clear up unused styles/files/classes now that we have been iterating on the UI
- Make my life easier and provide quicker iteration now we know what works and what doesn't
- Reduce tech debt (further identified debt are in issues)
  - #2873
  - #2876
  - #2875
  - #2902 
- Provide a consistent UI
- Futureproof the project by being able to teach others our system

Main changes for during review

- Full ControlTheme's have been copied from Avalonia so we can have real granular control that has been slowing us down until now
- `ControlResources.axaml` has been created to match how Avalonia works with it's ControlThemes and provides us a much needed abstraction layer between where Figma ends and Avalonia styling starts. Now we can share Button styles with anything that is like a Button... i.e. set `PrimaryStrongButtonBackground` as `PrimaryModerateBrush` and then use it for Primary ComboBox, Primary SplitButton, Primary ToggleButton etc. 
- Created new DebugControls page so can see controls, typography, colors and anything else we need all in one place.
- StandardButton default size was Medium but is now Small. Some files that assumed Medium needed to be specifically set. 
- TextBlock are now defaulted to `BodySMNormal`, a lot of TextBlocks were specifically set which no longer need to
- Font theme renamed from `Bold` to `Semi` so that it matches the font weight that was being used
- Added more Brand primitives such as Premium and Creator

Probably should really have been part of the PR

- Added start of styles for scrollbars
- Tweaked login and alpha overlay

This is far from an exhaustive "All UI issues and bugs are now fixed" but it's an awesome foundational change for future


src/NexusMods.App.UI/Pages/DebugControls/DebugControlsPageView.axaml
```diff
@@ -0,0 +1,773 @@
...
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:reactive="http://reactiveui.net"
+    xmlns:local="clr-namespace:NexusMods.App.UI.Pages.DebugControls"
+    xmlns:r3="clr-namespace:R3;assembly=R3"
+    xmlns:controls="clr-namespace:NexusMods.App.UI.Controls"
+    xmlns:icons="clr-namespace:NexusMods.Icons;assembly=NexusMods.Icons"
+    mc:Ignorable="d" d:DesignHeight="400"
+    x:Class="NexusMods.App.UI.Pages.DebugControls.DebugControlsPageView">
+
+    <Border Background="{StaticResource SurfaceLowBrush}">
```
erri120, line 15: I'd make this completely black, or at least make the color page background black. Otherwise you don't clearly see the colors.

src/NexusMods.App.UI/Controls/StandardButton.cs
erri120, line 1: The default values should be set to `0` so that they're actually the default.

docs/developers/development-guidelines/UINewSystem.md
```diff
@@ -0,0 +1,76 @@
...
+
+This document contains guidelines and tutorials for Stying and Templating the UI of the application.
+
+Depending on what the control will look like or what functionality is required will determine what method is used for styling. I considering this a type of flow chart for how complex we want to go.
+
+### Styling
+
+The most basic of visual changes when functionality stays the same. Styling is when the look of a control changes without changing any underlying markup and structure. Primarily used for color and font changes for various states but is used for layout changes such as margin, padding, alignment, spacing etc.
+
+For example, if we want to style an Avalonia `Seperator` control to match our system, we would create a new `SeperatorStyles.axaml` file inside of `src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Separator`. Any created style files also need referencing in the `src/Themes/NexusMods.Themes.NexusFluentDark/Styles/StylesIndex.axaml` file.
```
Al12rs, line 15: Put on separate line to make it more apparent. Forgetting this happened to me too many times
```suggestion
For example, if we want to style an Avalonia `Seperator` control to match our system, we would create a new `SeperatorStyles.axaml` file inside of `src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Separator`. 

Any created style files also need referencing in the `src/Themes/NexusMods.Themes.NexusFluentDark/Styles/StylesIndex.axaml` file.
```

docs/developers/development-guidelines/UINewSystem.md
```diff
@@ -0,0 +1,76 @@
...
+
+### ControlTheme
+
+If we need more elements to style than have been provided by Avalonia's default ControlTheme, we will need to provide a new one. 
+
+For example, if we want to change the structure of a `Button` control to have a `Border` around it, we would need to create a new `ControlTheme` for the `Button` control.
+
+Copy the default `Button` control theme from the Avalonia repository into `src/Themes/NexusMods.Themes.NexusFluentDark/Controls/Button.axaml` and modify it to suit our needs.
+
+Styles and classes can be added as part of the `ControlTheme` or alternatively as a separate `Style` file. This will be determined by the complexity of the control and how many controls will use the same styles. Any created `ControlTheme`'s also need referencing in the `src/Themes/NexusMods.Themes.NexusFluentDark/Resources/ResourceIndex.axaml` file.
```
Al12rs, line 30: Same as above
```suggestion
Styles and classes can be added as part of the `ControlTheme` or alternatively as a separate `Style` file. This will be determined by the complexity of the control and how many controls will use the same styles. 

Any created `ControlTheme`'s also need referencing in the `src/Themes/NexusMods.Themes.NexusFluentDark/Resources/ResourceIndex.axaml` file.
```

src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadView.axaml
```diff
@@ -220,17 +220,19 @@
...
+                    <StackPanel x:Name="OptionalModsPanel" Classes="TabItemHeaderStack">
                         <TextBlock>Optional</TextBlock>
                         <Border>
                             <TextBlock x:Name="OptionalDownloadsCount" />
                         </Border>
                     </StackPanel>
                 </TabItem.Header>
             </TabItem>
-            <TabItem x:Name="InstructionsTab" IsVisible="False">
+            <TabItem x:Name="InstructionsTab" IsVisible="True">
```
Al12rs, line 231: Should the default be true?

src/Themes/NexusMods.Themes.NexusFluentDark/NexusFluentDarkTheme.axaml
```diff
@@ -9,37 +10,38 @@
...
+        <!--                    ChromeGray="{StaticResource StrokeTranslucentStrong}" -->
+        <!--                    ChromeHigh="{StaticResource StrokeTranslucentStrong}" -->
+        <!--                    ChromeLow="{StaticResource StrokeTranslucentWeak}" -->
+        <!--                    ChromeMedium="{StaticResource SurfaceLow}" -->
+        <!--                    ChromeMediumLow="{StaticResource SurfaceMid}" -->
+        <!--                    ChromeWhite="{StaticResource SurfaceInverted}" -->
+        <!--                    ListLow="{StaticResource SurfaceLow}" -->
+        <!--                    ListMedium="{StaticResource StrokeTranslucentSubdued}" -->
+        <!--                    RegionColor="{StaticResource SurfaceBase}" /> -->
+        <!-- </FluentTheme.Palettes> -->
```
Al12rs, line 44: Are these no longer needed? Or are they set somewhere else?

insomnious: We aren't using the built in FluentTheme Palettes as we have all of our own. This was just a few less bindings and to remove the confusion in what is doing what. I'll remove comments.

src/NexusMods.App.UI/Pages/DebugControls/DebugControlsPageView.axaml
insomnious: This was a good suggestion. I've run this past @captainsandypants and we are going to leave it for now due to the fact that 80% of our controls\UI is based around SurfaceLow so it is a more accurate representation of how things look in context. We can reassess if it's proving difficult.

src/NexusMods.App.UI/Pages/CollectionDownload/CollectionDownloadView.axaml
insomnious: Good spot. I think this was just for testing as it is set in the code behind anyway, will put the default back to False

src/Themes/NexusMods.Themes.NexusFluentDark/Controls/Button.axaml
```diff
@@ -0,0 +1,96 @@
...
+    <!--     <Setter Property="BorderBrush" Value="{StaticResource AccentButtonBorderBrushPressed}" /> -->
+    <!--     <Setter Property="Foreground" Value="{StaticResource AccentButtonForegroundPressed}" /> -->
+    <!--   </Style> -->
+    <!-- -->
+    <!--   <Style Selector="^:disabled /template/ ContentPresenter#PART_ContentPresenter"> -->
+    <!--     <Setter Property="Background" Value="{StaticResource AccentButtonBackgroundDisabled}" /> -->
+    <!--     <Setter Property="BorderBrush" Value="{StaticResource AccentButtonBorderBrushDisabled}" /> -->
+    <!--     <Setter Property="Foreground" Value="{StaticResource AccentButtonForegroundDisabled}" /> -->
+    <!--   </Style> -->
+    <!-- </Style> -->
```
Al12rs, line 94: I'm guessing these are from Avalonia styles?
We would normally want to clear out dead code, but if this makes it easier to work with styles I'm fine with keeping it.

src/Themes/NexusMods.Themes.NexusFluentDark/Controls/TextBlock.axaml
```diff
@@ -0,0 +1,17 @@
...
+            <StackPanel>
+                <TextBlock Text="TextBlock" />
+                <TextBlock IsEnabled="False" Text="TextBlock" />
+                <TextBlock Text="Theme BodyLGNormalTheme" Theme="{StaticResource BodyLGNormalTheme}" />
+            </StackPanel>
+        </Border>
+    </Design.PreviewWith>
+
+    <!-- Defaults all TextBlock to BodySMNormalTheme -->
+    <ControlTheme x:Key="{x:Type TextBlock}" TargetType="TextBlock" BasedOn="{StaticResource BodySMNormalTheme}" />
```
Al12rs, line 15: I wasn't sure about setting a default for things, rather than using an explicit style every time.
The first one was actually choosing which style was supposed to be the default, since if the default isn't the most common one you lose most benefits of having one.

The main thing I worried about was granularity, if we ever need to change some of the default ones to something else, it is harder to know what is using that style and whether we want/need to update it or not. 

Explicit set of the style every time is obviously more bloat though.

There were also some concerns about style priority, e.g. the default unintentionally winning over the more granular replacements. I think this solution works for priority though.

insomnious: I had the same concerns - mainly with the priority of style selecting and shooting ourselves in the foot. We already have enough frustrations using native controls because of this. 

My tests when applying styles, classes and themes on top of this as a default have all worked fine and so the cleanliness is definitely nicer when 80%+ are going to be this default theme.

src/Themes/NexusMods.Themes.NexusFluentDark/NexusFluentDarkTheme.axaml
Al12rs: What if we set these all to pink or something, so we see where they are still used?

src/NexusMods.App.UI/Pages/DebugControls/DebugControlsPageView.axaml
```diff
@@ -0,0 +1,773 @@
...
+                    </TabItem>
+
+                    <TabItem Header="Tier 2 - Brand">
+                        <ScrollViewer>
+                            <StackPanel Margin="0,16,0,0">
+                                <StackPanel Classes="ElementSection">
+                                    <TextBlock Text="Brand" Theme="{StaticResource HeadingXSSemiTheme}" />
+                                    <StackPanel Classes="ColorRamp">
+                                        <TextBlock Text="TranslucentLight" />
+                                        <Grid ColumnDefinitions="*,*,*,*,*,*,*,*,*,*,*">
```
Al12rs, line 277: Why use Grids here? Seems long winded and hard to update

insomnious: This was so they just equally distributed to the width of the view. It was a debug page and so didn't spend much time looking into other options. We could just set a width and stack them however it's rarely, if ever, going to need to be changed.

src/NexusMods.App.UI/Pages/DebugControls/DebugControlsPageView.axaml
```diff
@@ -0,0 +1,773 @@
...
+                                    <ToggleSwitch IsChecked="True" IsEnabled="False" />
+                                </StackPanel>
+                                <StackPanel Spacing="8">
+                                    <TextBlock Text="CheckBox" />
+                                    <CheckBox>Enabled</CheckBox>
+                                    <CheckBox IsChecked="True">Checked</CheckBox>
+                                    <CheckBox IsChecked="{x:Null}" IsThreeState="True">3-State</CheckBox>
+                                    <CheckBox IsEnabled="False">Disabled</CheckBox>
+                                    <CheckBox IsChecked="True" IsEnabled="False">Checked</CheckBox>
+                                    <CheckBox IsChecked="{x:Null}" IsEnabled="False">3-State</CheckBox>
```
Al12rs, line 80: I'd want to keep the previews on this page in line with the design previews on the ControlThemes and styles. I see 6 here against 4 I think in the other places. 

I would want to keep the code the same so that if one needs updating it can just be copy pasted from the previews to here or vice versa.

insomnious: That's a good idea - I kind of kept the DebugControls simpler as an example of usage as opposed to showing every state and permutation, like on the themes and styles. I can definitely expand (if you think it'll be useful) on some controls on the debug page, maybe even give them their own tab\sub tab. The `StandardButton` design view, for instance, is rather big compared to what is in the DebugControls.
image

Al12rs: I mostly noticed the opposite, of some having less details in the styles files rather than the debug page, but Yeah I think it would make sense for maintainability to just keep the code the same.

I'm ok with making that a separate PR though

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Icons/IconsStyles.axaml
Al12rs, line 1: I think Check is still defined twice

src/Themes/NexusMods.Themes.NexusFluentDark/Styles/Controls/Icons/IconsStyles.axaml
```diff
@@ -64,6 +64,8 @@
                     <icons:UnifiedIcon Classes="DragVertical" />
                     <icons:UnifiedIcon Classes="Search" />
                     <icons:UnifiedIcon Classes="Home" />
+                    <icons:UnifiedIcon Classes="ToggleSwitchOn" />
+                    <icons:UnifiedIcon Classes="ToggleSwitchOff" />
```
Al12rs, line 68: These are already present in the next wrap panel preview

Al12rs (approved):

Issue #2908, opened 2025-03-20
Issues with changes config files not applying


Al12rs:
# User Report
Quoting a report of issues with config files not working correctly, needs investigation:

## Steps:
1. I install Lookup Anything. No config file present as they are generated when the game is run.
2. Run game
3. Open Generic Mod Config Menu (popular mod for editing config files in game)
4. Adjust Lookup Anything config options (in this case I set it so it only shows Hated gifts - not important what that is, it was just to have a clear setting - first screenshot)
5. Close game
6. Look in external changes, see that config.json is now present in Lookup Anything folder
7. Click Edit file, see that "Hated" is true while others are false which is correct
8. Go to My Mods, View mod files for Lookup Anything
9. As expected, no config.json file there (this was me just being thorough in my checking, I knew from your explanation it would be in external changes, not here)
10. Run game
11. Open GMCM menu
12. Change Lookup Anything config options to uncheck "Show hated gifts" and check "Show loved gifts" instead (second screenshot)
13. Close game
14. Look in external changes, see that config.json still has "Hated" set to true while the others are false
15. Double check there's no new config.json file in View mod files (again, not expecting it, just being thorough)
16. Run game
17. Open GMCM menu
18. Config settings are the same as they were the first time I set them and not the update values (third screenshot)

TL;DR It didn't allow SMAPI to update my config.json with the new settings meaning GMCM is unusable after the config file has been created

![Image](https://github.com/user-attachments/assets/986d4e6c-615d-4f00-91c8-65c7529f04a8)

![Image](https://github.com/user-attachments/assets/0211e67c-a47d-4481-9959-c8f19cdec8ed)

![Image](https://github.com/user-attachments/assets/6868b6e8-27ff-4d8d-b20e-cf94718be716)

### Details:
Link to discord conversation:
https://discord.com/channels/137344473976799233/1244072693054308362/1352236811073880136

Minimal mod setup:

![Image](https://github.com/user-attachments/assets/486b66e9-6327-4109-96f2-f7411b7133aa)

LukeNexusMods:
Needs testing

Al12rs:
Should be fixed with 
- #2919

Issue #2867, opened 2025-03-17
SMAPI May Fail to boot if managing pre-modded game due to StardewModdingAPI.deps.json generated file in overrides


LiamDawe:
(Sewer Edit)
- https://github.com/Nexus-Mods/NexusMods.App/issues/2867#issuecomment-2731389106

Cause and reproduction steps above.
This is an issue with unclean game install.

----------

# Bug Report

## Summary

Trying to run Stardew with SMAPI from the app, latest Nexus release from today, and it always gives the error of the deps file versioning not matching.

![Image](https://github.com/user-attachments/assets/dd47b362-9927-4b76-93af-463c91c34c83)

I can run it from Steam with this SMAPI just fine. So the issue is specific to the Nexus app.

## Steps to reproduce

Use Kubuntu, install Stardew, add in SMAPI via Nexus app and try to run it. Doesn't work.
I've tried the SMAPI download from Nexus and from the official SMAPI GitHub, it does it with both.

## What is the expected behaviour?

For it to work from Nexus.

## Other information

[nexusmods.app.slim.current.log](https://github.com/user-attachments/files/19292814/nexusmods.app.slim.current.log)

[nexusmods.app.main.current.log](https://github.com/user-attachments/files/19292816/nexusmods.app.main.current.log)

Sewer56:
[This happening on Linux is coincidental, this could happen on any OS hypothetically]

As I understand it, running SMAPI generates a file, but the App is not ingesting it into overrides, so the original file is restored when the Synchronizer is being ran.

```
00:10:39.375 [DEBUG] (NexusMods.Abstractions.Loadouts.Synchronizers.ALoadoutSynchronizer) Extracting 1 files to disk
```

Hitting 'play' triggers the synchronizer, resulting in an endless loop.

I'll try recreating this on my machine later today, see what I get.

Sewer56:
In any case, if it's there, you should be able to resolve the issue by deleting `StardewModdingAPI.deps.json` from your `External Changes` menu @LiamDawe

LiamDawe:
Removing that file caused other issues when then trying to mod it again, I forgot to get the logs (sorry).

I completely removed the game, removed the folders left by Nexus inside .local and then tried to re-do Stardew with SMAPI. So starting completely fresh.

Still doesn't work, looking again in External Changes, and that file was already there. Despite me removing everything, so I assume Nexus Mods adds it directly?

[...]

PR #2919, opened 2025-03-20
Don't re-insert files that change during ingestion

halgari:
Fixes issue-2908

We were not looking for and re-using existing external file records when ingesting changes. This would also cause the "duplicate key" exception on subsequent applies and in the "preview changes" dialogs


Sewer56:
Haven't confirmed but possibly also related/fixes #2867

Al12rs (approved):
I think I might have broken this when we changed the synchronizer to no longer replace the files inside mods. Good to have a test covering this now.

Al12rs:
Can confirm it fixed #2908

Issue #2885, opened 2025-03-18
Load Order: move TreeDataGrid to CompositeItemModel tech


Al12rs:
The current Load Order view is using the old TreeDataGridItemModel stuff instead of the composite item model.

We should update to it `CompositeItemModel` tech, same as all other trees we have to maintain everything using the same system and allow more complex UI styling and flexibility.

## Tasks:
- [x] Support `CompositeItemModel` based row styling #2893
    - [x] Find and implement a way to apply styling classes based on `CompositeItemModel` component contents
- [x] Add components and tree columns to represent the data to display in the LoadoutView #2893
- [x] Add a `LoadOrderDataProvider` class that takes in a `ISortableItemProvider` and returns `CompositeItemModels` #2893
    - [x] #2930
- [x] Implement TreeDataGrid column UIs based on components #2900
- [x] Integration in `LoadOrderViewModel` and `LoadOrderTreeDataGridAdaper` #2934
- [x] Rewire UI styling #2939
- [ ] Cleanup the old no longer used code

PR #2934, opened 2025-03-25
Load Order tree migration: integration into `LoadOrderViewModel` and `GridAdapter`

Al12rs:
- Part of #2885 

This actually moves the view over to using the new `CompositeItemModel` based tree items and updates all the functionality and bindings to work with the new system.

Some styling fixes work is going to be necessary, after which the leftover code from the previous approach can be cleaned up.

In particular, the `IsActive` styling has some issues due to `x:DataType` not being able to accept Types with type parameters correctly:
- https://github.com/AvaloniaUI/Avalonia/pull/18379
PR is merged but not released yet.


src/NexusMods.App.UI/Pages/Sorting/LoadOrder/LoadOrderItemModel.cs
erri120, line 1: Can this class be removed?

Al12rs: Yeah, just keeping it around until the styles have been fixed up since some reference it.

erri120 (approved):

PR #2939, opened 2025-03-26
Load Order UI fixes

insomnious:
A couple of quick fixes to bring load order back in line since the TreeDataGrid tech changes


Al12rs (approved):
Code looks good

Issue #2827, opened 2025-03-10
Implement compact UI elements


captainsandypants:
# Improvement

## Currently 
The app is designed to accommodate tap targets in an accessible way, this is due to us starting with the web design system as a base, as well as attempting to maintain accessibility on touch capable devices.

## We want
To test more compact components to make better use of space. The app design system is now mostly separate from the web design system meaning we are now able to refine it for desktop app use. The main touch capable device we currently want to support is the Steam Deck, we are finding that the hardware controls make navigating a more compact display viable without maintaining large tap targets. The following UI changes are for testing on Steam Deck and may not make the final cut if accessibility is not adequate.

- [x] Left hand menu
Font style change for menu items to Body/SM/Semi - this is for all states.
Font style change for menu headers to Body/SM/Normal (no longer all caps).  
[Figma link](https://www.figma.com/design/wgspwsneBneQn9ywd3dVhC/%F0%9F%93%B1%F0%9F%A7%AA-Design-test-area?node-id=2890-59010&t=zwDsJjvZrj9uNBSz-1)

- [x] Toolbar
"Floating" style toolbar, buttons are in a width hugging container with background set to surface/mid, this container then has a fullwidth parent container that is transparent (as in shows through the surface/low bg colour) and only has padding to align the button container.
[Figma link](https://www.figma.com/design/wgspwsneBneQn9ywd3dVhC/%F0%9F%93%B1%F0%9F%A7%AA-Design-test-area?node-id=2890-59010&t=zwDsJjvZrj9uNBSz-1)

- [x] XS Button and Split button variant
New XS button size, currently only used for toolbar buttons, height 24px (down from 28px on SM variant)
[Figma link](https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=1949-18019&t=BUzFHJnJOyAeEY49-1)

- [x] Small page tabs variant
A new variant for page tabs using font style Body/MD/Normal. Badges bg colour change to Surface/Translucent/Mid, font colours change to their translucent counterparts (see Figma designs).
[Figma link](https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=2756-28265&t=jrFybeRAEUAyZj8E-1)

- [x] Sub tabs redesign.
[...]

PR #2948, opened 2025-03-27
Compact UI - Part 1

insomnious:
Part of #2827 

This covers the first half of the Compact UI ticket

- Left hand menu
- Toolbar: I've also expanded this to start using a new `Toolbar` component to save the repetitive copy/paste of `StackPanel`, `Border` and controls. Only being used in `LibraryView` currently to see how we go.
- `ExtraSmall` button sizes are being used in above Toolbar
- TabControl and it's SubTabs class have been redesigned
- Removed Expanders and replaced with TabControl in Library


Al12rs (commented):
Left menu items size can be pushed up by 5px when there is an icon on the right side (collection downloads items and HealthCheck)

Al12rs (approved):

PR #2956, opened 2025-03-27
Add design verify test

erri120:
Since we now have a debug controls page, we can actually render it out in a test using the avalonia headless platform and verify if anything changed.

I couldn't get the controls and typography tabs to render properly, it wouldn't select the tab and just stay on the colors tab. Not sure about that.


erri120:
Closing for now.

PR #2960, opened 2025-03-31
Sort Mods Alphabetically By Name While Being Case-insensitive

### LINKED ISSUE #2824: Alphabetical sorting is not working as expected
# Bug Report

## Summary
When sorting the views in the app by Name, the way the mods are shown is not what users would expect.

We are sorting capital and lowercase letters separately, with capitals first. 

Image

## Steps to reproduce
Add mods to the app that have a mix of uppercase and lowercase names starting with the same letter. 

The sort order is case-insensitive.

Michael-Kowata:
Fixes #2824.

Sorts the mods alphabetically by the string name while being case-insensitive. 

NexusMods Issue #2824

### DIFF
--- a/src/NexusMods.App.UI/Controls/TreeDataGrid/SharedColumns.cs
+++ b/src/NexusMods.App.UI/Controls/TreeDataGrid/SharedColumns.cs
@@ -14,7 +14,10 @@ public static class SharedColumns
         {
             var aValue = a.GetOptional<StringComponent>(StringComponentKey);
             var bValue = b.GetOptional<StringComponent>(StringComponentKey);
-            return aValue.Compare(bValue);
+            var aName = aValue.HasValue ? aValue.Value.Value.ToString() : string.Empty;
+            var bName = bValue.HasValue ? bValue.Value.Value.ToString() : string.Empty;
+            
+            return string.Compare(aName, bName, StringComparison.OrdinalIgnoreCase);
         }
 
         public const string ColumnTemplateResourceKey = Prefix + "Name";


Michael-Kowata:
Rebased to clean up commit history, but also because one of the checks wasn't passing before. It should be good now.

src/NexusMods.App.UI/Controls/TreeDataGrid/SharedColumns.cs
```diff
@@ -14,7 +14,10 @@ public static int Compare<TKey>(CompositeItemModel<TKey> a, CompositeItemModel<T
         {
             var aValue = a.GetOptional<StringComponent>(StringComponentKey);
             var bValue = b.GetOptional<StringComponent>(StringComponentKey);
-            return aValue.Compare(bValue);
+            var aName = aValue.HasValue ? aValue.Value.Value.ToString() : string.Empty;
+            var bName = bValue.HasValue ? bValue.Value.Value.ToString() : string.Empty;
+            
+            return string.Compare(aName, bName, StringComparison.OrdinalIgnoreCase);
```
erri120, line 20:
```suggestion
            return aValue.Compare(bValue, (x, y) => StringComparer.OrdinalIgnoreCase.Compare(x.Value.Value, y.Value.Value));
```

erri120 (changes requested):

src/NexusMods.App.UI/Controls/TreeDataGrid/SharedColumns.cs
Michael-Kowata: I've committed the suggested changes.

Issue #2876, opened 2025-03-18
Overhaul overlays


insomnious:
A bit like the Advanced Installer UI, there is a lot of repetition happening when we are creating UIs for overlays/modals/messageboxes and if we don't nip it in the bud now, it'll become even more tech debt and be a bigger job to get in line with the design system.

This needs a bit of a deep dive into `src/NexusMods.App.UI/Overlays` and see what styles are reused, what styles aren't being used, where they are, what can be combined and some docs to support their usage.

Approximate time: 3 days

### Tasks

- Deep dive into what we are using and where
- Discussion with Design around what is likely to be defaults and what can be standardized
- Rebuild views using appropriate controls, layout elements and extract styles to seperate files
- Clearout unused classes
- Write supporting docs for when to use standard elements and when something more custom is required

erri120:
Possible duplicate of #1404.

erri120:
#1404 should be handled or at least discussed with this as well. #1404 is tech debt and has been around for a year and we never got around to doing it.

insomnious:
I'm going to go through what is on `main` and make sure they all have working UIs and no out of date buttons

insomnious:
The issues linked above (#1404 and #2942) will cover the further refactoring necessary. The linked PR contains enough to get back up to date and remove the urgency.

PR #2964, opened 2025-03-31
Overlay UI tweaks

insomnious:
Closes https://github.com/Nexus-Mods/NexusMods.App/issues/2876

A few tweaks to our current overlays to bring them in line before we do the bigger overhaul of the startup overlays and/or changing to use Avalonia Dialogs.

- Item delete confirmation button
- Upgrade to Premium no longer using markdown
- Metrics opt in no longer using markdown
- Strings in Language resource where possible


Al12rs (approved):
Conflict needs fixing otherwise looks good

Issue #2887, opened 2025-03-19
App deletes REDmod on apply


erri120:
# Bug Report

Repro:

1) Clean install Cyberpunk 2077 from Steam with REDmod
2) Add Cyberpunk 2077 to the app
3) Open Preview Changes

![Image](https://github.com/user-attachments/assets/00ca9a9b-ed53-4b26-9a9a-e45b1c44f4f4)

LukeNexusMods:
Could be to do with outside changes. Investigate this issue to determine size of work.

erri120:
When managing a game we got 3 locator IDs:

- `5840271637798999751` (REDmod)
- `6387618923890239308` (Phantom Liberty)
- `8420445566849588826` (Base Game)

Which we got from GameFinder after looking at the installed depots:

```
        "InstalledDepots"
        {
                "1091501"
                {
                        "manifest"              "8420445566849588826"
                        "size"          "66043436302"
                }
                "2060310"
                {
                        "manifest"              "5840271637798999751"
                        "size"          "96311717"
                        "dlcappid"              "2060310"
                }
                "2138330"
                {
                        "manifest"              "6387618923890239308"
                        "size"          "24745654120"
                        "dlcappid"              "2138330"
                }
        }
```

Then after running the actions we call `ReprocessGameUpdates` which returns the version definition that matches the most, given a list of disk entries (not including files we chose to ignore when indexing the game directory). This returns a version definition with just 1 locator ID which is `8420445566849588826`.

`BuildSyncTree` now calls `GetNormalGameState` which returns `154` files because it now only returns the base game files since we only have 1 locator ID. The REDmod files were previously marked as `DiskExists | LoadoutExists` because they were part of the "normal game state", now they aren't anymore because the locator IDs don't include the DLCs anymore which means that the REDmod files are now `DiskExists | PrevExists` which results in an action of `DeleteFromDisk`.

erri120:
Forcing the app to not have any hashes results in every file being marked as `DiskExists` and then being backed up and ingested. Note that this include files we ignored:

![Image](https://github.com/user-attachments/assets/0d78b00a-6710-46a1-be6c-b7ad35ed2622)

erri120:
Done with #2973, #2974, #2976.

LukeNexusMods:
@erri120 Would you be able to add steps of how QA would be able to test this please?

PR #2973, opened 2025-04-01
Replace `IsIgnoredPath` in Synchronizer

erri120:
Part of #2887.

`IsIgnoredPath` was added to reduce the time it takes to manage a game by skipping large files like archives. The issue is that skipping these files means they are never part of the sync tree which has the potential to cause a lot of issues and has already caused issues in #2887.

With the new minimal hashes and the file hashes DB, we can get the hashes for larger files very quickly and fallback to doing a full hash if we don't have the hashes already.

This PR removes `IsIgnoredPath` and adds `ShouldIgnorePathWhenIndexing` for BG3 specifically.


Al12rs (approved):

Al12rs (changes requested):
This removes the ability for us to Ignore Save Games changes for Baldur's Gate 3.

This is bad because it makes Saves be Loadout Locked (they should end up in Overrides). 

Meaning all user Save games will get nuked on Un-mange or uninstall of the App.

That sucks big time...

erri120:
Why are save games included in the first place?

Al12rs:
Because we need to manage the `modsettings.lsx` file (load order file) that is next to the saves folder.

Al12rs (approved):

PR #2974, opened 2025-04-01
Backup files in first synchronize pass instead of second pass

erri120:
Part of #2887.

Before: `DiskExists | LoadoutExists | DiskEqualsLoadout -> DoNothing`
After: `DiskExists | LoadoutExists | DiskEqualsLoadout -> BackupFile`

While debugging #2887 I've noticed that when managing a game, we don't backup files in the "first pass". We only backup in the "second pass" when we have previous state:

`DiskExists | PrevExists | LoadoutExists | DiskEqualsPrev | PrevEqualsLoadout | DiskEqualsLoadout -> BackupFile`

This PR updates the behavior of the first pass where we don't have previous state to match the behavior of the second pass.


Al12rs (approved):
I was thinking for how we could write a test for this, but it doesn't seem simple

PR #2982, opened 2025-04-02
Log falling back to doing a full hash

erri120:
Forgot to stage and commit this change in #2973. This logs files that the app doesn't have hash data for.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -1278,8 +1278,8 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
             if (hash.MinimalHash == diskMinimalHash)
                 return hash.XxHash3;
         }
-        
-        // If we didn't find a match, then we need to hash the file
+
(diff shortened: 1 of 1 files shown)


src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1278,8 +1278,8 @@ private async ValueTask<Hash> MaybeHashFile(IDb hashDb, GamePath gamePath, Absol
             if (hash.MinimalHash == diskMinimalHash)
                 return hash.XxHash3;
         }
-        
-        // If we didn't find a match, then we need to hash the file
+
+        _logger.LogDebug("Didn't find matching hash data for file `{Path}`, falling back to doing a full hash", file);
```
Al12rs, line 1282: This sounds like it would trigger a lot, for any non game file basically?

erri120: Every file we don't have hashes for, that's why I made this a debug log.

Al12rs: I wouldn't be surprised to see thousands of these, but sure

Al12rs (approved):

PR #2984, opened 2025-04-02
Update index code to be resilient against duplicate keys

### LINKED ISSUE #2869: Synchronizer error on game launch: Item with the same key has already been added
when installing new appimage for the upd i couldnt open the modmanager at all. With backups of my mods i didnt care about a reset so i deleted the Datafolder. I then went back into my account and added some mods back and pressed apply. Errors. I resetted the Datafolder again. When i try to launch the game with 1 mod im met with errors.

Image

[nexusmods.app.main.current.log](https://github.com/user-attachments/files/19304921/nexusmods.app.main.current.log)

[nexusmods.app.slim.current.log](https://github.com/user-attachments/files/19304935/nexusmods.app.slim.current.log)

erri120:
Fixes #2869.

This PR updates `ReindexState` in the Synchronizer to be more resilient against duplicate keys. I've added checks to make sure we don't index files we've already seen before which can lead to duplicate disk entry entities. I've also replaced the usage of `ToDictionary` which throws an exception on duplicate keys and we're now logging if we received duplicate entities.

I'm still not 100% how duplicate paths ended up in the disk state, my guess is that we previously had worse code that added duplicate entries and we fixed the places where we added duplicates but we never removed them, so users that upgraded to newer version still had duplicates in the DB.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -1183,81 +1183,96 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
     /// </summary>
     public async Task<bool> ReindexState(GameInstallation installation, IConnection connection, ITransaction tx)
     {
-        var seen = new HashSet<GamePath>();
-        var metadata = GameInstallMetadata.Load(connection.Db, installation.GameMetadataId);
-        var inState = metadata.DiskStateEntries.ToDictionary(e => (GamePath)e.Path);
-        var changes = false;
-        
-        
         var hashDb = await _fileHashService.GetFileHashesDb();
-        
-        foreach (var location in installation.LocationsRegister.GetTopLevelLocations())
+
+        var gameInstallMetadata = GameInstallMetadata.Load(connection.Db, installation.GameMetadataId);
+
+        var previousDiskStateEntities = gameInstallMetadata.DiskStateEntries;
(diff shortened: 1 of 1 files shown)


src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1183,81 +1183,96 @@ await Parallel.ForEachAsync(files, async (item, _) =>
...
+                var gamePath = installation.LocationsRegister.ToGamePath(file);
+                if (ShouldIgnorePathWhenIndexing(gamePath)) return;
+
+                bool isNewPath;
+                lock (seenPathsLock)
                 {
-                    {
-                        var gamePath = installation.LocationsRegister.ToGamePath(file);
-                        if (ShouldIgnorePathWhenIndexing(gamePath)) return;
+                    isNewPath = seenPaths.Add(gamePath);
```
Al12rs, line 1224: I guess in the case of a case-sensitive file system we could have the same path twice with different casing. 
Hopefully we don't run into that.

Al12rs (approved):

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
erri120: We noted that we wouldn't support such a case:

https://github.com/Nexus-Mods/NexusMods.App/blob/main/docs/developers/decisions/backend/0003-paths.md

Issue #2944, opened 2025-03-26
FileStore corruption after redownloading different versions of the same mod multiple times in different combinations


Al12rs:
Link to discord discussion: https://discord.com/channels/137344473976799233/1244072693054308362/1354397230110478336

## Outline:
User reported contents of a mod's `manifest.json` ending up spliced inside the `assets/data.json` file, resulting in a corrupted final output:

![Image](https://github.com/user-attachments/assets/5f87fd0c-81d3-4b86-adb5-754f8307a436)

![Image](https://github.com/user-attachments/assets/6283cec3-2d1b-44b4-a1c6-9105a952422c)

Discord message quote:
> NMA managed to insert the contents of the manifest into the assets/data.json file. The first screenshot is of my Lookup Anything data.json file from my ModDev Stardew install and the second is from the Stardew install that NMA is managing. I know that it was something in my install/uninstall/reinstall shenanigans that caused that, because I happened to look at that file after I uninstalled and deleted the older version of LA right before I then downloaded and installed the old version again (but didn't happen to take a screenshot of it, hence the screenshot from the other Stardew install).

## Steps:
Link to used mod page: 
https://www.nexusmods.com/stardewvalley/mods/541?tab=files

> Deleting from the library and redownloading does appear to be a necessary step. When I just uninstalled the older version, there were no changes applied. I had to:
- download 1.48.1, 
- install 1.48.1, check the contents of `data.json` with `View Mod Contents` and `Edit`
- download 1.49.0,
- install 1.49.0, check the contents of `data.json` should be the same
- delete from library version 1.48.1, 
- check the installed 1.49.0 `data.json` again and it will be corrupted

After doing that, the `data.json` file contained the `manifest.json` data (manifest for 1.48.1).

Al12rs:
Requires investigation, user seems to have reproduced twice, which is worrying.

LukeNexusMods:
What is the impact on the user?

Al12rs:
Corrupted file contents inside the game folder -> game/mods/everything not working as expected.

Al12rs:
This seems to be related to FileStore corrupting data when adding and deleting mods from the Library with shared files between them in different combinations.

Al12rs:
The seems to originate in the Nx library:

[...]

PR #3007, opened 2025-04-08
Avoid running the FileStore garbage collector in fire and forget 

Al12rs:
The GC behavior is very non deterministic and can be left around to be killed by an application close.

I changed most usage instances to prefer awaiting the results rather than fire and forgetting them.

I ran into a number of issues related to GC execution not being deterministic while debugging 
- #2944 

This Pr should make the behaviour more predictable and surface any exceptions or issues happening during GC.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.GC/GarbageCollectorRunMode.cs
+++ b/src/Abstractions/NexusMods.Abstractions.GC/GarbageCollectorRunMode.cs
@@ -21,4 +21,9 @@ public enum GarbageCollectorRunMode
     /// complete. The task is scheduled on the thread pool.
     /// </summary>
     RunAsyncInBackground,
+
+    /// <summary>
+    /// Runs the garbage collection in the background, waiting asynchronously for it to complete.
+    /// </summary>
+    RunAsynchronously,
 }
--- a/src/Abstractions/NexusMods.Abstractions.GC/IGarbageCollectorRunner.cs
+++ b/src/Abstractions/NexusMods.Abstractions.GC/IGarbageCollectorRunner.cs
@@ -20,5 +20,5 @@ public interface IGarbageCollectorRunner
     /// </summary>
     /// <param name="gcRunMode">The mode to run the GC in.</param>
     // ReSharper disable once InconsistentNaming
-    void RunWithMode(GarbageCollectorRunMode gcRunMode);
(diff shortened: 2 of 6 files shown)


Sewer56:
The most important question here is 'does this block the UI thread'.

The code originally fired in the background as to make deletions instant (which, at the time of implementation, they were).

If the user has a huge FileStore, and determining needed actions takes over a second, I'd rather avoid that, even if our ability to debug easier is impacted.

(Misc Note: You can always just have non fire-forget behaviour in debug, and optimized in release)

Sewer56 (approved):

Al12rs:
> As far as the GC and file streams are concerned, we should tie the lock's lifetime to the returned stream.
> 
> image
> 
> Otherwise the GC can start after the stream has been opened, causing potential issues. This is something that requires an issue and a patch, even if it may take a short time.

Could you open a ticket with the details please?

Sewer56:
Oh, I forgot `RunAsync` has a `Task.Run`, so we're actually good here :+1:

Sewer56:
> Could you open a ticket with the details please?

@Al12rs  Sure, though I'm not even sure if it's an 'issue' in the sense that anything would break per se.

The main risk is that `entry.OldFilePath.Delete();` throws during the GC process when:
- The user is on Windows
- A stream was opened before GC was ran
- A stream outlives the GC runtime
- So GC throws, trying to delete a file in use.

The risk is... an unhandled exception dialog and GC not running to completion.
On next GC everything would be cleaned up one way or another.
I'll just throw that on the ticket.

Sewer56:
> Technically, this could increase the chances of a deadlock happening between two tasks trying to access the FileStore lock, but I don't have a concrete example where that could happen.

Oh, also, there shouldn't be a difference.

The caller thread will be running some code, somewhere else in the App in either case.
Whether it be another Task, as the current one (caller) is awaited, or some other code following the GC call because it wasn't awaited.

The only possible source of deadlock, GC wise, is the GC waiting on itself; which is why it's critical GC does not call external non-GC code.

That's why the lock is released early. Because I have no guarantee what code the MnemonicDB callbacks will run once `tx.Commit()` is called. If I had an option to run callbacks on another thread within MMDB via API, the entirety of the GC, including file store updates would be under the lock; but given the current situation I cannot do that, as I have no API guarantee of what MMDB would do under the hood.

Issue #2276, opened 2024-11-20
Load Order - RedMod TreeDataGrid Drag&Drop support


Al12rs:
## Outline
Add Drag&Drop support to TreeDataGrid load order

Please investigate Drag and Drop load order. How does it work with the current tree data grid? Is there anything we need to overcome before being able to implement. Provide estimates for the work to be completed

## Tasks:
- [x] #2885 
- [x] Add `TreeeDataGridAdapter` option to enable Drag&Drop #3019
    -  Add option flag
    - Enable drag and drop on the TreeDataGrid
    - Add virtual method for `OnRowDrop` that has the dragged models and target location data extracted from the original event args.
- [x] Update `LoadOrderTreeDataGridAdapter` to use Drag&Drop and call some stubbed `MoveItems` function on `ISortableItemsProvider`  #3026
- [x] Implement `MoveItems` with tests in `RedModSortableItemsProvider` that supports moving multiple items to a specific index #3031 
- [x] Fix gaps between UI items preventing making some drop operations fail #3035

## Needs investigation:
- Prevent dragging of items that can't be moved: not implemented since not needed for RedMods
- Prevent dropping of items in invalid locations: not implemented since not needed for RedMods
- UI and styling changes for drag and drop operations:
This seems quite tricky to change. We would need to completely replace the drop adorners with our own, with likely access problems to private treeDataGrid stuff. This is the relevant code: https://github.com/AvaloniaUI/Avalonia.Controls.TreeDataGrid/blob/a684fd7d7f325f1134c7c43cc62af2f6ff727ff5/src/Avalonia.Controls.TreeDataGrid/TreeDataGrid.cs#L666

## DoD:
- User can drag and drop RedMod items in the Load Order view to change their sorting position.
- Sorting position is persisted across app restarts. 
- Starting the game has the redMods loaded correctly (harder to test, don't know which red mods work)

Load order is currently only available for Cyberpunk and only to sort RedMod type mods.
See comment below for a list of mods that would count.
Load order view can be accessed from the "New Tab" page, by selecting "Load Order WIP"

insomnious:
RedMods to test:

- https://www.nexusmods.com/cyberpunk2077/mods/16256?tab=files
- https://www.nexusmods.com/cyberpunk2077/mods/14245?tab=files
- https://www.nexusmods.com/cyberpunk2077/mods/19817?tab=files

Al12rs:
Finished with 
- #3031

PR #3019, opened 2025-04-09
TreeDataGrid Adapter support for drag/drop events

Al12rs:
- Part of #2276

This adds a `enableDragAndDrop` parameter to the setup of the `TreeDataGridAdapter`.
This enables drag and drop support on the tree and provides two subjects for `RowDragStarted` and `RowDrop` events.

Added code to extract the source and target models from the event data.
The original eventArgs are still passed to the subject since they can contain useful data like drop position, or can allow altering the state of the Drag indicator (can drop, can't drop etc).   

I checked with breakpoints that the extraction of the `sourceModels` and `targetModel` worked fine.

### DIFF
--- a/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
+++ b/src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
@@ -8,6 +8,7 @@ using NexusMods.App.UI.Extensions;
 using ObservableCollections;
 using R3;
 using System.Reactive.Linq;
+using Avalonia.Input;
 using DynamicData.Kernel;
 
 namespace NexusMods.App.UI.Controls;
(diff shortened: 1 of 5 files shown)


src/NexusMods.App.UI/Controls/TreeDataGrid/TreeDataGridAdapter.cs
```diff
@@ -20,6 +21,10 @@ public abstract class TreeDataGridAdapter<TModel, TKey> : ReactiveR3Object
     where TKey : notnull
 {
     public Subject<(TModel model, bool isActivating)> ModelActivationSubject { get; } = new();
+    
+    public Subject<(TModel[] sourceModels, TreeDataGridRowDragStartedEventArgs e)> RowDragStartedSubject { get; } = new();
+    
+    public Subject<(TModel[] sourceModels, TModel target, TreeDataGridRowDragEventArgs e)> RowDropSubject { get; } = new();
```
erri120, line 27: Is there anything other on the event args that we haven't extracted?

Al12rs:
Added OnDragOver event support

erri120 (approved):

PR #3026, opened 2025-04-14
Load Order: Drag and Drop Setup and prepare data 

Al12rs:
- Part of #2276 

## Details:
- Added stubbed `MoveIitems` method to `ILoadoutSortableItemProvider` that will contain the main moving logic for RedMods.
- Normalized the drag/drop event data in the `LoadOrderViewModel` depending on the `SortDirectionCurrent`.
- Made Drop position `Inside` result in either `Before` or `After` to avoid dropped items not moving due to drop position being too central.
- Enabled drag and drop in the UI for testing. Items won't move but it is possible to drag them and see the effects.

Next step is to implement the main logic inside `MoveItems`

### DIFF
--- a/src/NexusMods.App.UI/Pages/Sorting/LoadOrder/LoadOrderViewModel.cs
+++ b/src/NexusMods.App.UI/Pages/Sorting/LoadOrder/LoadOrderViewModel.cs
@@ -131,8 +131,7 @@ public class LoadOrderViewModel : AViewModel<ILoadOrderViewModel>, ILoadOrderVie
                             if (eventArgs.Position != TreeDataGridRowDropPosition.Inside) return;
                             
                             // Update the drop position for the inside case to be before or after
-                            var positionY = eventArgs.Inner.GetPosition(eventArgs.TargetRow).Y / eventArgs.TargetRow.Bounds.Height;
-                            eventArgs.Position = positionY < 0.5 ? TreeDataGridRowDropPosition.Before : TreeDataGridRowDropPosition.After;
+                            eventArgs.Position = PointerIsInVerticalTopHalf(eventArgs) ? TreeDataGridRowDropPosition.Before : TreeDataGridRowDropPosition.After;
                         }
                     );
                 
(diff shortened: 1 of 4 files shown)


src/NexusMods.App.UI/Pages/Sorting/LoadOrder/LoadOrderViewModel.cs
```diff
@@ -117,13 +117,72 @@ public LoadOrderViewModel(LoadoutId loadoutId, ISortableItemProviderFactory item
...
+                adapter.RowDragOverSubject
+                    .Subscribe(dragDropPayload =>
+                        {
+                            var (sourceModels, targetModel, eventArgs) = dragDropPayload;
+
+                            if (eventArgs.Position != TreeDataGridRowDropPosition.Inside) return;
+                            
+                            // Update the drop position for the inside case to be before or after
+                            var positionY = eventArgs.Inner.GetPosition(eventArgs.TargetRow).Y / eventArgs.TargetRow.Bounds.Height;
+                            eventArgs.Position = positionY < 0.5 ? TreeDataGridRowDropPosition.Before : TreeDataGridRowDropPosition.After;
```
erri120, line 135: Why are we changing the position when TreeDataGrid returns `TreeDataGridRowDropPosition.Inside`?

Al12rs: For UI purposes. The drop indicator line will show either Before or after rather than showing inside.

Since we don't support dropping inside for the moment, this is the approach that offers the best experience to users.
Otherwise trying to hit the Before or After becomes finnicky and annoying.

erri120: Since this code is repeated 3 times in this file, you should make that a method.

Al12rs: Made a method but I'm not sure how much clearer it is. 
If you have a better name for the method let me know.

erri120 (approved):

PR #3020, opened 2025-04-10
Updated Load Order UI

insomnious:
Part of https://github.com/Nexus-Mods/NexusMods.App/issues/2277

Updates the Load Order page UI to be closer to [Figma](https://www.figma.com/design/1sbIAJGr8URkY3wmisyaJg/%F0%9F%93%B1-Rules?node-id=1029-31403&m=dev).

image

Along with the load order page changes, the Alert control theme has been cleaned up by removing redundant styling and using native controls.

Ad-hoc changes from Figma (signed off by @captainsandypants) include:

- Show Alert button has been moved to above the trophy bar
- Hide Alert dismiss button
- Hide ModName and only show DisplayName (for now)

PR #3027, opened 2025-04-14
Adds Parent Mod column to Load Order UI

insomnious:
Not entirely happy with the naming, the column heading is "Parent mod" in Figma, but elsewhere in the codebase we just call it ModName, with the main Name column becoming DisplayName. I'd appreciate your thoughts.

Waiting on #3020 to be merged

### DIFF
--- a/src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProviderFactory.cs
+++ b/src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProviderFactory.cs
@@ -30,7 +30,8 @@ public class RedModSortableItemProviderFactory : ISortableItemProviderFactory
 
     public string IndexColumnHeader => "Load Order";
 
-    public string NameColumnHeader => "REDmod Name";
+    public string DisplayNameColumnHeader => "REDmod Name";
+    public string ParentModNameColumnHeader => "Parent mod";
 
     public string EmptyStateMessageTitle => "No REDmods detected";
     public string EmptyStateMessageContents => "Some mods contain REDmod items that alter core gameplay elements. When detected, they will appear here for load order configuration.";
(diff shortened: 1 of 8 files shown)


src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProviderFactory.cs
```diff
@@ -30,7 +30,8 @@ public class RedModSortableItemProviderFactory : ISortableItemProviderFactory
 
     public string IndexColumnHeader => "Load Order";
 
-    public string NameColumnHeader => "REDmod Name";
+    public string DisplayNameColumnHeader => "REDmod Name";
+    public string ParentModNameColumnHeader => "Parent mod";
```
Al12rs, line 34: This could probably just be hardcoded in the column definition instead of being here. I don't know hoe likely it is to change for different games.

src/NexusMods.App.UI/Pages/Sorting/LoadOrder/TreeDataGrid/LoadOrderComponents.cs
```diff
@@ -78,11 +78,28 @@ public static int Compare<TKey>(CompositeItemModel<TKey> a, CompositeItemModel<T
...
+
+        public static readonly ComponentKey DisplayNameComponentKey = ComponentKey.From(ColumnTemplateResourceKey + "_" + "DisplayNameComponent");
+
+        public static int Compare<TKey>(CompositeItemModel<TKey> a, CompositeItemModel<TKey> b) where TKey : notnull
+        {
+            throw new NotImplementedException();
+        }
+
+        // The header name should be set on column creation as it is game dependent
+        public static string GetColumnHeader() => throw new NotSupportedException();
```
Al12rs, line 93: We should set the column header here directly for ModNameColumn instead of passing it in later from the sortableItemProviderFactory, since it shouldn't be game dependent.

Al12rs (changes requested):
See comments above

insomnious:
I _think_ this is what you mean :)

Al12rs (approved):

PR #3031, opened 2025-04-14
Load Order: Implement move logic for Drag & Drop

Al12rs:
- Part of #2276 

This implements the main logic for drag moving items around with drag and drop.
It is now possible to use drag and drop to move items around.

This also implements some tests for the new `MoveItemsTo()` methods.

Things left to do are:
- Fix gap problem, where items dropped between item gaps in the list are not moved. #3035
- Investigate Styling options for Drag and drop decorations.


src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/SortOrder/RedMod/RedModSortableItemProvider.cs
```diff
@@ -151,10 +151,63 @@ public async Task SetRelativePosition(ISortableItem sortableItem, int delta, Can
     }
 
     /// <inheritdoc/>
-    public Task MoveItemsTo(ISortableItem[] sourceItems, ISortableItem targetItem, TargetRelativePosition relativePosition, CancellationToken token)
+    public async Task MoveItemsTo(ISortableItem[] sourceItems, ISortableItem targetItem, TargetRelativePosition relativePosition, CancellationToken token)
     {
-        // TODO: implement this method
-        return Task.CompletedTask;
+        await _semaphore.WaitAsync(token);
```
erri120, line 156: Use our custom `WaitDisposable` instead, then you don't need the `try/finally`:

```
using var _ = _semaphore.WaitDisposable(cancellationToken: token);
```

erri120 (changes requested):

erri120 (approved):

PR #3033, opened 2025-04-14
Add diagnostics for configuring Cyberpunk 2077 on Linux for Steam

### LINKED ISSUE #2171: Verify `WINEDLLOVERRIDES` and installed winetricks components for Cyberpunk 2077 on Linux
# Improvement
winmm.dll is needed for red4ext, however it's not found after installing the mod. it would be great if you could add this to the games wine config automatically.

Link to information on the issue:
https://wiki.redmodding.org/cyberpunk-2077-modding/for-mod-users/users-modding-cyberpunk-2077/modding-on-linux

## Currently 
_(Describe the current behaviour of the existing feature you want to improve)_

RED4Ext doesn't work on Steam Deck without adding a Launch Option: WINEDLLOVERRIDES="winmm,version=n,b" %command%

## We want
_(Describe the new behaviour you want to see for this feature)_

Really adding all the details listed on  the following site somewhere in your wiki or setup would be excellent so folks won't have to run around troubleshooting the issue like I did:

https://wiki.redmodding.org/cyberpunk-2077-modding/for-mod-users/users-modding-cyberpunk-2077/modding-on-linux

## Supporting Information
_(Include any supporting screenshots etc to support the above)_
PXL_20241015_141514228 MP

## Design

erri120:
Resolves #2171.

- Adds utilities for parsing `WINEDLLOVERRIDES` and `winetricks.log` files
- Adds two new critical diagnostics for Cyberpunk 2077

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.GameLocators/Stores/Steam/SteamLocatorResultMetadata.cs
+++ b/src/Abstractions/NexusMods.Abstractions.GameLocators/Stores/Steam/SteamLocatorResultMetadata.cs
@@ -26,4 +26,6 @@ public record SteamLocatorResultMetadata : IGameLocatorResultMetadata
     
     /// <inheritdoc />
     public IEnumerable<LocatorId> ToLocatorIds() => ManifestIds.Select(m => LocatorId.From(m.ToString()));
+
+    public Func<string?>? GetLaunchOptions { get; init; }
 }
--- a/src/Abstractions/NexusMods.Abstractions.GameLocators/Stores/Steam/SteamLocatorResultMetadata.cs
+++ b/src/Abstractions/NexusMods.Abstractions.GameLocators/Stores/Steam/SteamLocatorResultMetadata.cs
@@ -27,5 +27,7 @@ public record SteamLocatorResultMetadata : IGameLocatorResultMetadata
     /// <inheritdoc />
     public IEnumerable<LocatorId> ToLocatorIds() => ManifestIds.Select(m => LocatorId.From(m.ToString()));
 
+    public AbsolutePath? ProtonPrefixDirectory { get; init; }
+
     public Func<string?>? GetLaunchOptions { get; init; }
 }
--- a/src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
+++ b/src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
@@ -1,4 +1,6 @@
+using System.Text;
 using JetBrains.Annotations;
+using NexusMods.Paths;
 
 namespace NexusMods.Abstractions.GameLocators;
 
@@ -8,6 +10,37 @@ namespace NexusMods.Abstractions.GameLocators;
 public static class WineParser
 {
     public const string EnvironmentVariableName = "WINEDLLOVERRIDES";
+
+    /// <summary>
+    /// Parses the given `winetricks.log` file and returns all installed packages.
+    /// </summary>
+    public static IReadOnlySet<string> ParseWinetricksLogFile(AbsolutePath filePath)
+    {
+        if (!filePath.FileExists) return new HashSet<string>();
+
+        using var stream = filePath.Read();
+        return ParseWinetricksLogFile(stream);
+    }
+
+    /// <summary>
+    /// Parses the given `winetricks.log` file and returns all installed packages.
+    /// </summary>
+    public static IReadOnlySet<string> ParseWinetricksLogFile(Stream stream)
+    {
+        using var sr = new StreamReader(stream, Encoding.UTF8);
+
+        var result = new HashSet<string>();
+
+        while (sr.ReadLine() is { } line)
+        {
+            var trimmed = line.Trim();
+            if (trimmed.Length == 0) continue;
+
+            result.Add(line);
+        }
+
+        return result;
+    }
 
     /// <summary>
     /// Parses the <see cref="EnvironmentVariableName"/>.
@@ -28,6 +61,8 @@ public static class WineParser
             var section = span[splitRange];
 
             var index = section.LastIndexOf('=');
+            if (index == -1) continue;
+
             var namesSpan = section[..index];
 
             var dllNamesEnumerator = namesSpan.Split(',');
--- a/src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
+++ b/src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
@@ -1,0 +1,123 @@
+using JetBrains.Annotations;
+
+namespace NexusMods.Abstractions.GameLocators;
+
+/// <summary>
+/// Parser for WINE.
+/// </summary>
+public static class WineParser
+{
+    public const string EnvironmentVariableName = "WINEDLLOVERRIDES";
+
+    /// <summary>
+    /// Parses the <see cref="EnvironmentVariableName"/>.
+    /// </summary>
+    public static IReadOnlyList<WineDllOverride> ParseEnvironmentVariable(ReadOnlySpan<char> environmentVariableValue)
+    {
+        var results = new List<WineDllOverride>();
+
+        // https://gitlab.winehq.org/wine/wine/-/wikis/Wine-User's-Guide#winedlloverrides-dll-overrides
+        var span = GetWineDllOverridesSection(environmentVariableValue);
+
+        // NOTE(erri120): DLLs are separated with a semicolon
+        var splitEnumerator = span.Split(';');
+        foreach (var splitRange in splitEnumerator)
+        {
+            var section = span[splitRange];
+
+            var index = section.LastIndexOf('=');
+            var namesSpan = section[..index];
+
+            var dllNamesEnumerator = namesSpan.Split(',');
+            var dllOverrideTypes = GetOverrideTypes(section);
+
+            foreach (var range in dllNamesEnumerator)
+            {
+                var name = FixDllName(namesSpan[range]).ToString();
+                results.Add(new WineDllOverride(name, dllOverrideTypes));
+            }
+        }
+
+        return results;
+    }
+
+    private static ReadOnlySpan<char> FixDllName(ReadOnlySpan<char> input)
+    {
+        return input.TrimEnd(".dll");
+    }
+
+    private static WineDllOverrideType[] GetOverrideTypes(ReadOnlySpan<char> section)
+    {
+        var index = section.LastIndexOf('=');
+        if (index == section.Length - 1) return WineDllOverride.Disabled;
+
+        var typesSpan = section[(index + 1)..];
+
+        var numTypes = typesSpan.Count(',') + 1;
+        Span<WineDllOverrideType> overrideTypesSpan = stackalloc WineDllOverrideType[numTypes];
+
+        var typesIndex = 0;
+        var typesEnumerator = typesSpan.Split(',');
+        foreach (var splitRange in typesEnumerator)
+        {
+            var typeSpan = typesSpan[splitRange];
+            if (typeSpan.Length != 1) continue;
+
+            var c = typeSpan[0];
+            if (c is 'n') overrideTypesSpan[typesIndex++] = WineDllOverrideType.Native;
+            if (c is 'b') overrideTypesSpan[typesIndex++] = WineDllOverrideType.BuiltIn;
+        }
+
+        return overrideTypesSpan[..typesIndex].ToArray();
+    }
+
+    private static ReadOnlySpan<char> GetWineDllOverridesSection(ReadOnlySpan<char> input)
+    {
+        const string prefix = "WINEDLLOVERRIDES=";
+
+        var span = input;
+        if (!span.StartsWith(prefix, StringComparison.OrdinalIgnoreCase)) return span;
+
+        span = span[prefix.Length..];
+        if (!span.StartsWith('"')) return span;
+
+        span = span[1..];
+        var index = span.IndexOf('"');
+        span = span[..index];
+
+        return span;
+    }
+}
+
+public record WineDllOverride(string DllName, WineDllOverrideType[] OverrideTypes)
+{
+    /// <summary>
+    /// Gets whether the DLL is disabled.
+    /// </summary>
+    public bool IsDisabled => OverrideTypes is [WineDllOverrideType.Disabled];
+
+    internal static readonly WineDllOverrideType[] Disabled = [WineDllOverrideType.Disabled];
+}
+
+/// <summary>
+/// DLL override types for Wine.
+/// </summary>
+[PublicAPI]
+public enum WineDllOverrideType
+{
+    /// <summary>
+    /// The dll is disabled.
+    /// </summary>
+    Disabled = 0,
+
+    /// <summary>
+    /// The built-in version should be loaded.
+    /// </summary>
+    BuiltIn = 1,
+
+    /// <summary>
+    /// The native version should be loaded.
+    /// </summary>
+    Native = 2,
+}
+
--- a/src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
+++ b/src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
@@ -14,6 +14,8 @@ public static class WineParser
     /// </summary>
     public static IReadOnlyList<WineDllOverride> ParseEnvironmentVariable(ReadOnlySpan<char> environmentVariableValue)
     {
+        if (environmentVariableValue.Length == 0) return [];
+
         var results = new List<WineDllOverride>();
 
         // https://gitlab.winehq.org/wine/wine/-/wikis/Wine-User's-Guide#winedlloverrides-dll-overrides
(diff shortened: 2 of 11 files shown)


MattSturgeon:
Parsing the dll overrides explicitly from steam's launch command feels a little conceptually flawed, given there are several ways users (and/or a future "fix diagnostic" script) could configure dll overrides. It would cover _most_ cases though, I guess, so it's a fair starting point.

1. `WINEDLLOVERRIDES` env var configured in steam
2. `WINEDLLOVERRIDES` env var configured elsewhere (e.g. in a shell profile or systemwide)
3. `[DllOverrides]` in `wine.conf` or `.winerc`
4. `Software\Wine\DllOverrides` in the registry (e.g. `user.reg` as used by `winecfg`)

I wonder if there is a way to ask wine what the _effective_ dll override setting is for a particular dll name? If so, maybe a script could be run through steam (so it gets run by wine in the correct wineprefix) that prints the dll's override setting to stdout or something.

At the very least, I think this only _partially_ resolves https://github.com/Nexus-Mods/NexusMods.App/issues/2171, so it shouldn't close that issue.

src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
```diff
@@ -0,0 +1,174 @@
...
+            if (c is 'n') overrideTypesSpan[typesIndex++] = WineDllOverrideType.Native;
+            if (c is 'b') overrideTypesSpan[typesIndex++] = WineDllOverrideType.BuiltIn;
+        }
+
+        return overrideTypesSpan[..typesIndex].ToArray();
+    }
+
+    private static ReadOnlySpan<char> GetWineDllOverridesSection(ReadOnlySpan<char> input)
+    {
+        const string prefix = "WINEDLLOVERRIDES=";
```
MattSturgeon, line 113: I think `EnvironmentVariableName` on L12 is unused. Was it intended to be used here as `prefix = EnvironmentVariableName + "="`?

Sewer56:
I had a peek at this yesterday, but will only comment until I look at this more properly later today.

Sewer56:
Will check next morning, was stuck unexpectedly earlier.

src/Abstractions/NexusMods.Abstractions.GameLocators/Stores/Steam/SteamLocatorResultMetadata.cs
```diff
@@ -26,4 +26,8 @@ public record SteamLocatorResultMetadata : IGameLocatorResultMetadata
     
     /// <inheritdoc />
     public IEnumerable<LocatorId> ToLocatorIds() => ManifestIds.Select(m => LocatorId.From(m.ToString()));
+
+    public AbsolutePath? ProtonPrefixDirectory { get; init; }
```
Sewer56, line 30: This should be documented as the path to the 'WINEPREFIX' used by the Proton distribution from Steam. If it was me, I'd call it `WinePrefixDirectory` as it's technically the most correct; and just that it's used by proton in the docs, but it works either way.

Mostly keeping that in mind for the non-Windows users; not everyone reading this code may know exactly what's going on.

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/Emitters/WinePrefixRequirementsEmitter.cs
```diff
@@ -0,0 +1,36 @@
...
+    private static readonly WineDllOverride[] RequiredOverrides =
+    [
+        new("winmm", [WineDllOverrideType.Native, WineDllOverrideType.BuiltIn]),
+        new("version", [WineDllOverrideType.Native, WineDllOverrideType.BuiltIn]),
+    ];
+
+    private static readonly ImmutableHashSet<string> RequiredWinetricksPackages =
+    [
+        "d3dcompiler_47",
+        "vcrun2022",
```
Sewer56, line 23: This is kinda hardcoded, but we can get away with it for now, mostly on the basis that binaries built with VisualC++ are forwards compatible, so we should always point this to the newest vcrun if possible.

But worth noting, they're not forwards compatible (newer binary on older runtime). [A great example of where that can go wrong](<https://dolphin-emu.org/blog/2024/09/04/dolphin-progress-report-release-2407-2409/#visual-studio-twenty-twenty-woes>).

So we really should strive to ensure it's up to date, as we'll never know when a component like red4ext or CET update their compiler.

src/Games/NexusMods.Games.Generic/WineDiagnosticHelper.cs
```diff
@@ -0,0 +1,86 @@
...
+        // TODO: support more than Steam
+        if (gameInstallation.LocatorResultMetadata is not SteamLocatorResultMetadata steamLocatorResultMetadata) return null;
+
+        var protonPrefixDirectory = steamLocatorResultMetadata.ProtonPrefixDirectory;
+        if (protonPrefixDirectory is null) return null;
+
+        var winePrefixDirectory = protonPrefixDirectory.Value.Combine("pfx");
+
+        // https://github.com/Winetricks/winetricks/blob/e73c4d8f71801fe842c0276b603d9c8024d6d957/src/winetricks#L4216-L4225
+        var winetricksFilePath = winePrefixDirectory.Combine("winetricks.log");
```
Sewer56, line 22: https://github.com/Winetricks/winetricks/blob/e73c4d8f71801fe842c0276b603d9c8024d6d957/src/winetricks#L12811-L12818

https://github.com/Winetricks/winetricks/blob/e73c4d8f71801fe842c0276b603d9c8024d6d957/src/winetricks#L12894-L12901

https://github.com/Winetricks/winetricks/blob/e73c4d8f71801fe842c0276b603d9c8024d6d957/src/winetricks#L12951-L12958

https://github.com/Winetricks/winetricks/blob/e73c4d8f71801fe842c0276b603d9c8024d6d957/src/winetricks#L13068-L13075

In case it doesn't embed:

```bash
w_metadata vcrun2015 dlls \
    title="Visual C++ 2015 libraries (concrt140.dll,mfc140.dll,mfc140u.dll,mfcm140.dll,mfcm140u.dll,msvcp140.dll,msvcp140_1.dll,msvcp140_atomic_wait.dll,vcamp140.dll,vccorlib140.dll,vcomp140.dll,vcruntime140.dll,vcruntime140_1.dll)" \
    publisher="Microsoft" \
    year="2015" \
    media="download" \
    conflicts="vcrun2017 vcrun2019 ucrtbase2019 vcrun2022" \
    file1="vc_redist.x86.exe" \
    installed_file1="${W_SYSTEM32_DLLS_WIN}/mfc140.dll"
```

```bash
w_metadata vcrun2017 dlls \
    title="Visual C++ 2017 libraries (concrt140.dll,mfc140.dll,mfc140u.dll,mfcm140.dll,mfcm140u.dll,msvcp140.dll,msvcp140_1.dll,msvcp140_2.dll,msvcp140_atomic_wait.dll,vcamp140.dll,vccorlib140.dll,vcomp140.dll,vcruntime140.dll,vcruntime140_1.dll)" \
    publisher="Microsoft" \
    year="2017" \
    media="download" \
    conflicts="vcrun2015 vcrun2019 ucrtbase2019 vcrun2022" \
    file1="vc_redist.x86.exe" \
    installed_file1="${W_SYSTEM32_DLLS_WIN}/mfc140.dll"
```

```bash
w_metadata vcrun2019 dlls \
    title="Visual C++ 2015-2019 libraries (concrt140.dll,mfc140.dll,mfc140u.dll,mfcm140.dll,mfcm140u.dll,msvcp140.dll,msvcp140_1.dll,msvcp140_2.dll,msvcp140_atomic_wait.dll,msvcp140_codecvt_ids.dll,vcamp140.dll,vccorlib140.dll,vcomp140.dll,vcruntime140.dll,vcruntime140_1.dll" \
    publisher="Microsoft" \
    year="2019" \
    media="download" \
    conflicts="vcrun2015 vcrun2017 vcrun2022" \
    file1="vc_redist.x86.exe" \
    installed_file1="${W_SYSTEM32_DLLS_WIN}/mfc140.dll"
```

```bash
w_metadata vcrun2022 dlls \
    title="Visual C++ 2015-2022 libraries (concrt140.dll,mfc140.dll,mfc140chs.dll,mfc140cht.dll,mfc140deu.dll,mfc140enu.dll,mfc140esn.dll,mfc140fra.dll,mfc140ita.dll,mfc140jpn.dll,mfc140kor.dll,mfc140rus.dll,mfc140u.dll,mfcm140.dll,mfcm140u.dll,msvcp140.dll,msvcp140_1.dll,msvcp140_2.dll,msvcp140_atomic_wait.dll,msvcp140_codecvt_ids.dll,vcamp140.dll,vccorlib140.dll,vcomp140.dll,vcruntime140.dll,vcruntime140_1.dll)" \
    publisher="Microsoft" \
    year="2022" \
    media="download" \
    conflicts="vcrun2015 vcrun2017 vcrun2019" \
    file1="vc_redist.x86.exe" \
    installed_file1="${W_SYSTEM32_DLLS_WIN}/vcruntime140.dll"
```

The `winetricks` project checks the existence of the file `installed_file1` or `installed_exe1` to denote the install state of a given component. BUT as I noted in:
- https://github.com/Nexus-Mods/NexusMods.App/pull/3033/files#r2047736618

> But worth noting, they're not backwards compatible (newer binary on older runtime).

What this means is that if the 2015-2017-2019 trend continues, and the same DLL name is reused, a hypothetical `vcrun2025` etc. package may be detected as 'installed'; when it is not.

While this is not blocking, we will need *more robust* runtime detection; one way or another, and I would *really* like to see that to be surfaced to a new GitHub issue once this is merged.

src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        span = span[1..];
+        var index = span.IndexOf('"');
+        span = span[..index];
+
+        return span;
+    }
+}
+
+public record WineDllOverride(string DllName, WineDllOverrideType[] OverrideTypes)
```
Sewer56, line 129: This should be a `record struct`, the item does not need boxing as it's composed of 2 reference types at current, meaning that it's cheaper to copy around as a value type.

As for the load orders, there's a fixed set of these,

- `b` Builtin
- `n` Native
- `b,n` Builtin then Native
- `n,b` Native then Builtin

This should ideally be represented as an enum, with each of the 4 orders represented as an enum variant.

`winecfg` UI for reference: 

image

Not sure if I'm being too picky here or not though; but I thought it's worth pointing out, as the number of possibilities here is finite.

Do at least change to `record struct`.

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/Diagnostics.cs
```diff
@@ -172,4 +172,40 @@ Please install the {RedmodLink} and start the game at least once.
             .AddValue<NamedLink>("RedmodLink")
         )
         .Finish();
+
+    [DiagnosticTemplate]
+    [UsedImplicitly]
+    internal static IDiagnosticTemplate RequiredWineDllOverrides = DiagnosticTemplateBuilder
+        .Start()
+        .WithId(new DiagnosticId(Source, number: 5))
+        .WithTitle("Update WINE dll overrides to mod on Linux")
```
Sewer56, line 181: `dll` should be in capitals, i.e. `DLL`

src/NexusMods.StandardGameLocators/SteamLocator.cs
```diff
@@ -46,6 +48,21 @@ protected override IGameLocatorResultMetadata CreateMetadata(SteamGame game)
...
+            {
+                var localConfigPath = SteamLocationFinder.GetUserDataDirectoryPath(game.SteamPath, game.AppManifest.LastOwner).Combine("config").Combine("localconfig.vdf");
+                var parserResult = LocalUserConfigParser.ParseConfigFile(game.AppManifest.LastOwner, localConfigPath);
+                if (parserResult.IsFailed)
+                {
+                    Logger.LogWarning("Error while parsing local user config at `{Path}`: `{Error}`", localConfigPath, parserResult.Errors);
+                    return null;
+                }
+
+                if (!parserResult.Value.LocalAppData.TryGetValue(game.AppId, out var localAppData)) return null;
```
Sewer56, line 62: Worth noting this is not available until the game is booted at least once.
If we don't do it, someone else will; not a big issue as App would issue diagnostic after the user closes the game, but it's there.

src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
```diff
@@ -0,0 +1,174 @@
...
+                results.Add(new WineDllOverride(name, dllOverrideTypes));
+            }
+        }
+
+        return results;
+    }
+
+    private static ReadOnlySpan<char> FixDllName(ReadOnlySpan<char> input)
+    {
+        return input.TrimEnd(".dll");
```
Sewer56, line 83: We need to ignore the case here!
This could be a `.DLL`, with DLL name in uppercase, have seen it before.

In other words, an IndexOf and Slice!

src/Abstractions/NexusMods.Abstractions.GameLocators/WineParser.cs
Sewer56: +1

Sewer56 (approved):
Anyway, the core logic's good, and there's nothing out of the ordinary to stop merging, so I'll just approve it to avoid any unnecessary waiting tomorrow morning.

Do look through the comments and caveats above, there are a lot, and at least one should be promoted to an issue. That said, good enough for first iteration/release.

src/Games/NexusMods.Games.Generic/WineDiagnosticHelper.cs
erri120: I don't really understand what this has to do with the referenced line of code. We're using the `winetricks.log` file to check which component is installed.

src/Games/NexusMods.Games.RedEngine/Cyberpunk2077/Emitters/WinePrefixRequirementsEmitter.cs
Sewer56: To be fair, we shouldn't be blindly following the wiki. Components like CET could change their wrapper names any time they like (even id they likely wouldn't), or someone could come up with a completely new loader.

We should ideally be detecting the overrides the user needs dynamically and then emitting diagnostics as needed. That code could then also be reused for all native games, and would not need adjustments on a per game basis. (Unless in the super super super rare case a game has multiple exe files and you are forced to start a wrapper due to DRM)

To be specific, we'd be parsing the PE header of the exe file, getting the list of module imports and checking if the synchronizer will deploy a file to disk that matches the name of any of the modules. If that's not in the overrides, emit a diagnostic.

Can just copy the parser from one of my projects if needed.

I'm just fine with this for now as the functionality around here is still in development. We can do better long term though.

src/Games/NexusMods.Games.Generic/WineDiagnosticHelper.cs
erri120: Why would that not be robust enough? The user will be using winetricks to install the components, so we'll check `winetricks.log` if they did that.

Sewer56: (I didn't get notified about reply here via email, strange, was having a busy morning, so only got back now, sorry about that)

Anyway, I think a step by step would be better here.

So, `winetricks` checks the existence of a single file, that is `installed_file1` as per above to determine if the component is installed.

The problem is, `vcrun2015`, `vcrun2017` and `vcrun2019` all use the same file `"${W_SYSTEM32_DLLS_WIN}/mfc140.dll"`.
So if you install `vcrun2015`, then `vcrun2017` and `vcrun2019` will also be assumed as 'installed' by winetricks. This is not good.

So what can happen in this situation is that the user installs a game via Steam; and Steam would install (for instance) `vcrun2015` into the WINEPREFIX of the game as part of the game's setup. If the user later installs a mod that was built with a newer toolchain (e.g. `vcrun2019`); that may fail, because Visual C++ is not forward compatible (e.g. [in the real life example above](<https://dolphin-emu.org/blog/2024/09/04/dolphin-progress-report-release-2407-2409/#visual-studio-twenty-twenty-woes>)); only backwards compatible.

--------------

An example of a game where something like this could happen is [Skyrim SE](<https://steamdb.info/app/489830/depots/>). It installs `vcrun2015` during game setup. 
image
If the user then installs something like an ENB, DLL mod, or otherwise that was built on the 2019 toolchain then:

- winetricks says `vcrun2019` is installed because `"${W_SYSTEM32_DLLS_WIN}/mfc140.dll"` exists.
- App wouldn't emit a diagnostic.
- `vcrun2019` (or newer) is not installed.
- There is a risk of the mod failing to run.

Most mod authors are unfortunately not aware of forward compatibility issues with Visual C++ either.
But, on Windows they can kinda sorta get away with it because new games will install a newer Visual C++, so there's a high chance the end user might already have whatever a mod was built with.

However, because we start with a clean slate on Linux per game (fresh WINEPREFIX), that may not be the case. And this could lead to random hard to diagnose runtime issues for users with specific mods down the road.

This is why we will need something more robust. It's important to know what we will have behind the scenes.

--------------

We are very lucky (for now) that, for Cyberpunk, CET is built with `vcrun2022`. And, it just so happens that `winetricks` project chose to check a different file name in `"${W_SYSTEM32_DLLS_WIN}/vcruntime140.dll"` for 2022 specifically. 

However, that is pure luck, and we shouldn't rely on that to last for long.
There can be forward compatibility breakages even within the same Visual C++ 'year', as seen in the above example (Dolphin).

I would argue the best course of action down the road is just have our own check to ensure that the ***latest*** Visual C++ Redist is installed in a given WINEPREFIX. Since all releases 2015 and above are backwards compatible, just checking for `latest` and asking the user to update would be handy. We could also use that code on Windows, too.

Sewer56 (approved):

src/Games/NexusMods.Games.Generic/WineDiagnosticHelper.cs
erri120: Again, what does any of this have to do with the `winetricks.log` file?

PR #3046, opened 2025-04-15
Added: 'Synchronizer Backed Up Files as Roots'

Sewer56:
Just the thing I worked on late yesterday and today morning.

It boils down to the following:

- We used to have 'Game Files' as a hidden mod.
- We no longer have 'Game Files' as a hidden mod.
- Thus game files were not rooted by GC.

This would lead to their deletion, every mod deletion.
Consequently, files from the game folders would be 're-backed-up as a side effect'.

This change happened around February, when we transitioned away from using 'Game Files' mod and onto the Game Hashes DB.

-------------

With the new behaviour:

- On synchronize, files backed up by the synchronizer are tagged `SynchronizerBackedUpFile` (inherits from `BackedUpFile`).
    - These store the install instance, for future migration work, and unmanage.
- Instances of `BackedUpFile` are considered rooted (not removable) by the GC.

More intricate details in the actual code.

As for cleaning up, there's improvements to be made there but I've kept it very simple for now.
When unmanaging a game, all 'rooted' backups associated with the game installation are removed.

This isn't ideal behaviour (yet), as we will want to be more flexible down the road, e.g. 'what if the user chose the full backup option and now wants to get rid of it', or 'what if the game updates and we got ancient backups sitting there'?

And of course, 'why back up when we could pull files from game stores?'
But that's further work, I'll leave that part to @halgari ; as this part of the code/design is their domain.


Sewer56:
I forgot to update the verify schema yesterday, fixed just now before standup.

NexusMods.Abstractions.GC.DataModel/BackedUpFile.cs
```diff
@@ -0,0 +1,21 @@
...
+[PublicAPI]
+public partial class BackedUpFile : IModelDefinition
+{
+    private const string Namespace = "NexusMods.GC.BackedUpFile";
+
+    /// <summary>
+    /// Hash of the file.
+    /// </summary>
+    public static readonly HashAttribute Hash = new(Namespace, nameof(Hash)) { IsIndexed = true };
+}
```
Al12rs, line 21: I think this should go in `Abstractions.Loadout` and we can avoid the `Abstractions.GC.DataModel` project.

src/NexusMods.App.GarbageCollection.DataModel/DataStoreReferenceMarker.cs
```diff
@@ -31,6 +32,17 @@ public static void MarkUsedFiles<TParsedHeaderState, TFileEntryWrapper>(IConnect
...
+        // Mark non Loadout/Library items that are backups (e.g. Backups of original game files)
+        MarkItemsMarkedAsBackups(archiveGc, db);
+    }
+
+    private static void MarkItemsMarkedAsBackups<TParsedHeaderState, TFileEntryWrapper>(ArchiveGarbageCollector<TParsedHeaderState, TFileEntryWrapper> archiveGc, IDb db) where TParsedHeaderState : ICanProvideFileHashes<TFileEntryWrapper> where TFileEntryWrapper : IHaveFileHash
+    {
+        // All files explicitly marked as roots should be preserved.
+        var libraryFiles = BackedUpFile.All(db);
+        foreach (var file in libraryFiles)
+            archiveGc.AddReferencedFile(file.Hash);
```
Al12rs, line 45: Rename names here for clarity

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1168,6 +1169,21 @@ await Parallel.ForEachAsync(files, async (item, _) =>
...
+            _ = new SynchronizerBackedUpFile.New(tx, out var id)
+            {
+                GameInstallId = installation.GameMetadataId,
+                BackedUpFile = new BackedUpFile.New(tx, id)
+                {
+                    Hash = item.Hash,
+                }
+            };
+        }
+        await tx.Commit();
```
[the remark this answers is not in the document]
Al12rs: We also need to update `ReprocessGameUpdates` that deletes game file entries from Override in case we get game hashes for them. 

Because we didn't identify those files to be game files the first time around and we need to add GameFileBackups markers for them at the moment they are no longer rooted in Override.

NexusMods.Abstractions.GC.DataModel/BackedUpFile.cs
Al12rs: Use single level without inheritance, with both hash and gameMetadata

src/NexusMods.App/Services.cs
```diff
@@ -63,6 +64,7 @@ public static IServiceCollection AddApp(this IServiceCollection services,
                 .AddLibrary()
                 .AddLibraryModels()
                 .AddJobMonitor()
+                .AddLoadoutsSynchronizers()
```
Al12rs, line 67: Is this still needed? I think it might not since you moved the models out of the Loadout.Synchronizer project

Al12rs:
```suggestion
```

tests/NexusMods.App.GarbageCollection.DataModel.Tests/FindUsedFiles/MarkGameFilesTest.cs
```diff
@@ -0,0 +1,129 @@
...
+        // https://github.com/pengweiqhca/Xunit.DependencyInjection?tab=readme-ov-file#3-closest-startup
+        // A trick for parallelizing tests with Xunit.DependencyInjection
+        public void ConfigureServices(IServiceCollection services) => DIHelpers.ConfigureServices(services);
+    }
+}
+
+/// <summary>
+/// A stubbed game for GC testing.
+/// </summary>
+public class GCStubbedGame : AGameTest<StubbedGame>
```
Al12rs, line 80: Is it possible to use `AIsolatedGameTest` here instead? Not sure on the differences in detail, but we have been using the isolated version for new tests since it was redone with the purpose of avoiding parallel execution issues.

src/NexusMods.App/Services.cs
```diff
@@ -3,6 +3,7 @@
 using NexusMods.Abstractions.Games;
 using NexusMods.Abstractions.Library.Models;
 using NexusMods.Abstractions.Loadouts;
+using NexusMods.Abstractions.Loadouts.Synchronizers;
```
Al12rs, line 6:
```suggestion
```

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1143,6 +1157,7 @@ public virtual async Task ActionBackupNewFiles(GameInstallation installation, Di
         // TODO: This may be slow for very large games when other games/mods already exist.
         // Backup the files that are new or changed
         var archivedFiles = new ConcurrentBag<ArchivedFileEntry>();
+        var pinnedFiles = new ConcurrentBag<ArchivedFileEntry>();
```
Al12rs, line 1160: You can make this a bag of hashes instead of a bag of `ArchivedFileEntry`

src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1475,6 +1504,16 @@ await _jobMonitor.Begin(new UnmanageGameJob(installation), async ctx =>
                     await ctx.YieldAsync();
                     await DeleteLoadout(loadout, GarbageCollectorRunMode.DoNotRun);
                 }
+                
+                // Retract all roots / synchronizer backed up files for this installation.
```
Al12rs, line 1508:
```suggestion
                // Retract all `GameBakedUpFile` entries to allow for game file backups to be cleaned up from the FileStore
```

tests/NexusMods.App.GarbageCollection.DataModel.Tests/FindUsedFiles/MarkGameFilesTest.cs
Sewer56: I can change it, but it wouldn't make a difference, I still need a `Startup` object for XUnit DI.

Sewer56: That said, I'd need to write some additional boilerplate somewhere it seems, to handle getting the stubbed game into the GameRegistry, it's not showing up there.

Sewer56: @Al12rs I had to add some missing functionality, notably the 'AIsolatedGameTest` did not support the `GenerateGameFiles` API. But I did it just to make you happy.

It seems all of the code there was duplicated and slightly tweaked in `AIsolatedGameTest`. I wonder why this couldn't just have been a bool flag into the constructor of `AGameTest`, we shouldn't really be duplicating code like this :sweat_smile: . Not sure how that wound up all the way in the codebase.

There isn't a difference here at runtime though, I had to add a tiny bit of extra boilerplate I didn't have before actually to make this work now.

Al12rs: I think the plan to to delete AGameTest when we don't have any uses of it anymore

tests/Games/NexusMods.Games.TestFramework/AIsolatedGameTest.cs
```diff
@@ -336,6 +336,11 @@ private void ResetGameFolders()
     /// </summary>
     protected async Task<Loadout.ReadOnly> CreateLoadout()
     {
+        if (!_gameFilesWritten)
+        {
+            await GenerateGameFiles();
+            _gameFilesWritten = true;
+        }
```
Al12rs, line 343: And this has no chance of creating parallel execution problems right? The files are generated in a different folder for each test?

Sewer56: The Game Installation and DataModel location are randomised per run, and those places are obtained from DI, which is unique per run.

So I believe it's unique per run, given that the sources are.

Al12rs: I'm not sure if Isolated tests were supposed to use inMemoryFS, could that be the case or did I dream that up?

Al12rs (approved):

PR #3063, opened 2025-04-17
Fix app crashing on disposed bitmaps in the guided installer

### LINKED ISSUE #3056: Fatal Exception
03:18:18.517 [FATAL] (NexusMods.App.Startup) Exception crashed the application!|System.NullReferenceException: Object reference not set to an instance of an object.

[nexusmods.app.main.14.log](https://github.com/user-attachments/files/19779581/nexusmods.app.main.14.log)

Found when installing FOMODS as part of the following collection:
https://www.nexusmods.com/games/cyberpunk2077/collections/ayfbwl

erri120:
Fixes #3056.

The issue came from our use of the scoped cache and the interaction between disposed lifetimes and bitmaps. Removing the scoped cache and having better handling of disposing the image fixes the issue.

### DIFF
--- a/src/NexusMods.App/Startup.cs
+++ b/src/NexusMods.App/Startup.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics;
 using Avalonia;
 using Avalonia.Controls;
 using Avalonia.ReactiveUI;
@@ -59,6 +60,7 @@ public class Startup
         }
         catch (Exception e)
         {
+            Debugger.Break();
             logger.LogCritical(e, "Exception crashed the application!");
         }
     }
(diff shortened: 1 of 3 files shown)


src/NexusMods.App/Startup.cs
```diff
@@ -59,6 +60,7 @@ public static void Main(IServiceProvider provider, string[] args)
         }
         catch (Exception e)
         {
+            Debugger.Break();
```
Al12rs, line 63: Leftover from debugging I think

Al12rs:
```suggestion
```

src/NexusMods.App/Startup.cs
```diff
@@ -1,3 +1,4 @@
+using System.Diagnostics;
```
Al12rs, line 1:
```suggestion
```

src/NexusMods.App/Startup.cs
erri120: Nope, this is in here on purpose. Otherwise the debugger doesn't break.

Al12rs:
Closing and reopening to rebase the CI tests

PR #3067, opened 2025-04-17
WIP - New dialog system

insomnious:
This is very early days. Creating PR to share and discuss.


src/NexusMods.App.UI/MessageBox/Enums/ButtonEnum.cs
```diff
@@ -0,0 +1,14 @@
...
+/// </summary>
+public enum ButtonEnum
+{
+    Ok,
+    YesNo,
+    OkCancel,
+    OkAbort,
+    YesNoCancel,
+    YesNoAbort,
+}
```
Al12rs, line 14: We have three separate enums here, I'd like to reduce that and instead have an extensible system for defining which buttons should be present in a dialog.

I used copilot to write something up:

A flexible approach is to define a record or class storing essential button properties (like text, icons, etc.), together with a static class of common defaults. That allows for on-the-fly creation of new buttons without passing actual UI controls. Below is a simple example:

```csharp
public record MessageBoxButtonDefinition(
    string Text,
    IconValue? LeftIcon,
    IconValue? RightIcon,
    StandardButton.ShowIconOptions ShowIcon
);
```

```csharp
public static class MessageBoxButtons
{
    public static readonly MessageBoxButtonDefinition Ok = new(
        "OK",
        null,
        null,
        StandardButton.ShowIconOptions.None
    );

    public static readonly MessageBoxButtonDefinition Cancel = new(
        "Cancel",
        null,
        null,
        StandardButton.ShowIconOptions.None
    );
    
    // Add more as needed...
}
```

You can then pass an array of `MessageBoxButtonDefinition` objects to the `ShowMessageBox()` method and return the pressed button (or pressed button's Id). This keeps the system extensible by letting you define new button definitions at runtime.

src/NexusMods.App.UI/MessageBox/MessageBoxViewModel.cs
```diff
@@ -0,0 +1,189 @@
...
+    private IMessageBoxView<ButtonResult>? _view;
+
+    public string ContentTitle { get; }
+    public string ContentMessage { get; set; }
+
+    public bool IsOkShowed { get; private set; }
+    public bool IsYesShowed { get; private set; }
+    public bool IsNoShowed { get; private set; }
+    public bool IsAbortShowed { get; private set; }
+    public bool IsCancelShowed { get; private set; }
```
Al12rs, line 22: A list of buttons should be generated and shown based on the inputs rather than having a limited set ready. 

I'd expect button creation and handling to happen in the codebehind. 
VM would have SetResult(buttonId) to be called when a button in pressed.

src/NexusMods.App.UI/MessageBox/MessageBox.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        _view.SetCloseAction(() =>
+        {
+            tcs.TrySetResult(_view.GetButtonResult());
+            window.Close();
+        });
+        
+        window.ShowDialog(owner);
+        return tcs.Task;
+    }
```
Al12rs, line 74: `window.ShowDialog(owner);` is an async task itself that should be awaited.
I don't think you need to use the `TaskCompletionSource` in this case as you can instead await `ShowDialog` directly.

insomnious:
Added a Factory to create the MessageBox that is then shown using `WindowManager.ShowMessageBox()`. Added seperate button ID and Text so it can be properly used in localisation

image

```csharp
// create new MessageBox
var messageBox = MessageBoxFactory.Create(
  "Delete this mod?", 
  "Deleting this mod will remove it from all collections. This action cannot be undone.", 
  [
  MessageBoxStandardButtons.Cancel,
    new MessageBoxButtonDefinition(
        "Yes, delete",
        ButtonDefinitionId.From("yes-delete"),
        null,
        null,
        ButtonRole.AcceptRole | ButtonRole.DestructiveRole
    )
  ],
  MessageBoxSize.Small
);

// Use WindowManager to show it
var result = await ViewModel.WindowManager.ShowMessageBox(messageBox, MessageBoxWindowType.Modal);

Console.WriteLine($@"Result: {result}");

// Output when 'Yes, Delete' is clicked or Enter is pressed
Result: yes-delete
```

src/NexusMods.App.UI/Dialog/ICloseable.cs
```diff
@@ -0,0 +1,8 @@
+using System.Reactive;
+using ReactiveUI;
+
+namespace NexusMods.App.UI.Dialog;
+
+public interface ICloseable
```
Al12rs, line 6: Closable sounds better than Closeable
```suggestion
public interface IClosable
```

Al12rs (commented):
- [ ] IDialogViewModel should use TResult rather than `ButtonDefinitionId`
- [ ] Merge MessageBox and Dialog container into one thing
	- [ ] Similar to StandardButton we add all the things and hide them
	- [ ] With ViewModelViewHost at the middle, completely generic ViewModel
	- [ ] If we need to close the dialog from the inner VM, the caller of the dialog needs to pass the container manually for now, custom solution.
- [ ] Cleanup the button roles:
	- [ ] Remove unused ones
	- [ ] Split up into Styling focused ones and Shortcut/Action related ones (consider whether they need to be flags or not)

insomnious:
Known issues:

- Large vertical modals don't spawn exactly in the center due to I think the height being calculated afterward positioning. Only when using ViewModelViewHost. We do, however, set a max window height of parent height minus 50.
- Custom Content, when used, can't currently close the window. (We can add this later if it becomes it's an issue, until then we always have the titlebar).
- Titlebar of spawned window can't be turned off (related to the above, we can add later but this keeps it within scope).


https://github.com/user-attachments/assets/a8936b92-21fa-4a1c-b59a-08b148015b02

insomnious:
- [ ] Rename MessageBox to Dialog (merge into the empty interfaces)
- [ ] Keep message box for the relevant factory methods
- [ ] Investigate dialogs opening in wrong monitor
- [ ] Investigate weird DPI
- [ ] Show icon even if you have custom VMVH
- [ ] Investigate screen relative avalonia size rather than using pixels for starting sizes
- [ ] Rename Close button ID to DefaultClose

src/NexusMods.App.UI/Dialog/CustomContentView.axaml
Al12rs, line 1: Should we rename these to `CustomContentExampleView` ? To clarify that this isn't some to use to create custom dialogs for example.

insomnious: Yeah sure, makes sense

PR #3070, opened 2025-04-17
Remove LFS and migrate data

halgari:
This removes the LFS entries in the `.gitattributes` file and adds the .zip files directly into the repo


halgari:
Merged this as everything is blocked on LFS no longer working :/

Issue #658, opened 2023-09-20
Define policy for GamePaths referencing Nested locations


Al12rs:
Games define a list of base locations (`GameFolderType`) where all GamePaths are relative to (`AGame.GetLocations()`).

These locations could be nested to each other (e.g. `Data` is under `Game` in case of `Skyrim/Data`).

This poses the problem of how to internally store `GamePath`s that are relative to a nested location like `Data`.

The entities that can generate new GamePaths are Mod Installers, Games, and the LoadoutSynchronizer. 
Generic Mod installers could have a hard time knowing where to properly map GamePaths.

The Synchronizer that detects new files in a folder doesn't have any information regarding where these should be placed and so can only pick either the closest parent Location (most nested), or the highest level one (TopLevelLocation). 
Currently the Synchronizer uses to most nested location.

An important note to make is the fact that, wile for the Steam install two game locations might be nested, that might no longer be the case for the GOG installation for example. 

To maintain this portability across Game installations we might need to accept arbitrary GamePaths, TopLevel, nested, inbetween etc.

This unfortunately means that there could be two different `GamePath`s that actually point to the same disk location, making detection of conflicts potentially awkward. 

Things to do:
* Decide on general approach (only top level, only most nested, arbitrary)
* Adapt the code and make sure everything works with the chosen policy.

Al12rs:
This will be a problem for bethesda games

LukeNexusMods:
Every component that needs a nested folder should normalize it before talking to the rest of the system.

PR #3077, opened 2025-04-22
Update suggested directories for the advanced installer

### LINKED ISSUE #2463: Advanced Installer - Better suggested folders
# Improvement

## Currently 
The only suggested location provided by the advanced installer is the game root.

## We want
To provide some sensible, per-game suggestions. 

## Supporting Information

Image

## Design

erri120:
Cleaning up the `AGame` class a bit and adding `Mods` as a suggested directory for Stardew Valley.

image

Fixes #2463.


src/Games/NexusMods.Games.StardewValley/StardewValley.cs
```diff
@@ -101,7 +92,9 @@ protected override IReadOnlyDictionary<LocationId, AbsolutePath> GetLocations(IF
         var result = new Dictionary<LocationId, AbsolutePath>()
         {
             { LocationId.Game, installation.Path },
+            { LocationId.From("Mods"), installation.Path.Combine(Constants.ModsFolder) },
```
Al12rs, line 95: I think this might be a problem:
- #658

We can have two ways to indicate the same `GamePath`: 
- `{LocationId.Game}, mods/modA/manifest.json`
- `{LocationId.From("Mods")}, /modA/manifest.json`

Figuring out that they are the same file required resolving the absolute path, something we only do late in the synchronizer.

A potential solution would be to have the installer normalize the `GamePaths` to the top level one, but the relationship between Game and Mods locations needs to be expressed somewhere.

erri120: I'm not sure I fully understand the issue. The synchronizer only looks at top level locations. Currently, the advanced installer is the only place where we look at nested locations as suggested destinations.

Al12rs: So if the user were to choose a nested location in the Advanced installer as the install location, the files under the `GamePaths` of `{LocationId.From("Mods")} would never get deployed.

I guess that would be the current issue

erri120: It does mess with preview changes but it applies just fine:

image

image

Al12rs: I don't think you had any conflicts. 
`ShouldWin` uses `GamePaths` not absolute paths for example, so it would miss the conflicts completely.
Other parts of the app, like the Preview Changes view, or the Health Checks would not be equipped to resolve a `LocationId.Mods` to be inside `LocationId.Game`.

I don't think that we want to have `LoadoutItems` pointing to nested locations, like the mods one for SDV, we want all LoadoutItems that resolve to the same `AbsolutePath` to have the same `GamePath`, otherwise we would just introduce complexity everywhere we want to use `GamePaths`, just to resolve whether two paths are actually the same.

PR #3088, opened 2025-04-23
Remove automatic protocol registration on Linux

### LINKED ISSUE #3068: Update protocol registration on Linux
Protocol registration continues to be a pain for us on Linux. Diagnosing the issue is very difficult due to all the different system components in play. Instead of trying to make sense of the mess, we can just hand over protocol registration for someone else to deal with.

Flatpak handles protocol registrations for us, so we don't need anything special in a flatpak build. For AppImage builds we can redirect users to use something like [gearlever](https://github.com/mijorus/gearlever).

What would also help is a debug page with a button that launches `nxm://registration-test/{id}`. If the app doesn't receive this link after a timeout then we can conclude that the protocol registration is likely broken.

List of tasks I propose we should do:
- Add debug page to verify we can receive `nxm` links
- Add dialog when the app receives an `nxm` registration test link with an unknown ID
- Remove custom protocol registration handling from AppImage builds
- Redirect AppImage users to something like [gearlever](https://github.com/mijorus/gearlever) for protocol registration

erri120:
Resolves #3068.

### DIFF
--- a/src/NexusMods.App.UI/Pages/ProtocolRegistration/ProtocolRegistrationTestPage.cs
+++ b/src/NexusMods.App.UI/Pages/ProtocolRegistration/ProtocolRegistrationTestPage.cs
@@ -1,0 +1,39 @@
+using NexusMods.Abstractions.Serialization.Attributes;
+using NexusMods.App.UI.WorkspaceSystem;
+using NexusMods.Icons;
+
+namespace NexusMods.App.UI.Pages;
+
+[JsonName(nameof(ProtocolRegistrationTestPageContext))]
+public record ProtocolRegistrationTestPageContext : IPageFactoryContext;
+
+public class ProtocolRegistrationTestPageFactory : APageFactory<IProtocolRegistrationTestPageViewModel, ProtocolRegistrationTestPageContext>
+{
+    public ProtocolRegistrationTestPageFactory(IServiceProvider serviceProvider) : base(serviceProvider) { }
+
+    public override PageFactoryId Id => StaticId;
+    public static readonly PageFactoryId StaticId = PageFactoryId.From(Guid.Parse("3C1B0632-8407-4904-8F30-5B86067E9715"));
+
+    public override IProtocolRegistrationTestPageViewModel CreateViewModel(ProtocolRegistrationTestPageContext context)
+    {
+        return new ProtocolRegistrationTestPageViewModel(ServiceProvider, WindowManager);
+    }
+
+    public override IEnumerable<PageDiscoveryDetails?> GetDiscoveryDetails(IWorkspaceContext workspaceContext)
+    {
+        return
+        [
+            new PageDiscoveryDetails
+            {
+                Icon = IconValues.Cog,
+                ItemName = "Protocol Registration Test",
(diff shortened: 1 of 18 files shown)


src/Abstractions/NexusMods.Abstractions.NexusWebApi/Types/NXMProtocolRegistrationTest.cs
```diff
@@ -0,0 +1,28 @@
+using JetBrains.Annotations;
+
+namespace NexusMods.Abstractions.NexusWebApi.Types;
+
+/// <summary>
+/// NXM protocol registration test.
+/// </summary>
+[PublicAPI]
+public class NXMProtocolRegistrationTest : NXMUrl
```
Al12rs, line 9: Should this be called "check" rather than "test" to avoid overload with unit tests?

src/NexusMods.App.UI/Pages/ProtocolRegistration/ProtocolRegistrationTestPage.cs
```diff
@@ -0,0 +1,39 @@
...
+                ItemName = "Protocol Registration Test",
+                SectionName = "Utilities",
+                PageData = new PageData
+                {
+                    Context = new ProtocolRegistrationTestPageContext(),
+                    FactoryId = StaticId,
+                },
+            },
+        ];
+    }
```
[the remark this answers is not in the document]
erri120: It should be visible to all users for them to debug any issues they have with protocol registration.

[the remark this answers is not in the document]
Al12rs: I think Design will want this hidden behind an option or something.
I think having this button in the settings directly would be the long term plan, but I don't think we have the setup for that at the moment.

src/NexusMods.CrossPlatform/ProtocolRegistration/ProtocolRegistrationLinux.cs
```diff
@@ -13,124 +13,18 @@ namespace NexusMods.CrossPlatform.ProtocolRegistration;
...
     }
 
-    private async Task UpdateMIMECacheDatabase(AbsolutePath applicationsDirectory, CancellationToken cancellationToken = default)
+    public Task RegisterHandler(string uriScheme, bool setAsDefaultHandler = true, CancellationToken cancellationToken = default)
     {
-        _logger.LogInformation("Updating MIME cache database");
-
-        var command = _updateDesktopDatabaseDependency.BuildUpdateCommand(EscapeWhitespaceForCli(applicationsDirectory));
-        await _processFactory.ExecuteAsync(command, cancellationToken: cancellationToken);
+        return Task.CompletedTask;
```
Al12rs, line 28: So none of the linux install methods will have protocol registration? They will all rely on package managers/custom user setup?

erri120: Yup.

Al12rs (approved):

PR #3089, opened 2025-04-23
Don't validate exit code by default

### LINKED ISSUE #3087: Remove exit code validation when running games

erri120:
Exit code validation is questionable at best and unreasonable to have as a default set to true. This PR changes the default behavior to not have exit code validation. Any consumer of the API that cares about non-zero exit codes can set the option to have validation.

Resolves #3087.


Al12rs (approved):

Sewer56:
I'm not sure if 'remove validation on all binaries' is a correct response to not wanting it on some binaries.

Wouldn't you want this the other way around?
i.e. opt-out instead of opt-in.

With opt-in there's always a chance we'll be missing errors because we chose not to check for them. Which could lead to unintended side effects not as explicitly logged or presented to the user as desired.

Issue #2273, opened 2024-11-20
Load Order - Backend Abstraction


Al12rs:
## Outline
Abstract Load Order backend to support more than one game.

## Details
Implement necessary abstractions to facilitate Load Order implementation for new games, and supporting behaviour differences between them.

Objectives:
- Support behavior differences without requiring big changes to the system- 
- Avoid code duplication
- Reduce boilerplate code
- Make adding support for new games easy

## Definition of done
- Developers are satisfied with the abstractions
- Supported at least 3 types of load orders (e.g. REDmods, CP77 .archives, BG3 pak files)

PR #3090, opened 2025-04-23
Load Order: Backend abstraction

Al12rs:
- Part of #2273 

Initial effort to make parts of the RedMod load order code more generic and reusable for other load orders.

- Introduced `ASortableItemProviderFactory` abstraction which simplifies the definition of new provider factories
- Introduced `ASortableItemProvider` abstraction which implements some virtual method implementations
- Reworked a lot of the code in the provider to make it more structured and easier to abstract and emulate
- Introduced `ISortItemKey` as way to define a game-specific id to deduplicate and identify sortable items instead of using temporary Guids.

Credits to @IDCs for a lot of the starting code used for this branch.

Tests seem to fail still, will try to investigate


src/Abstractions/NexusMods.Abstractions.Games/ISortItemKey.cs
erri120, line 1: Why do we want an opaque key type instead of using generics? The benefit of using opaque types is that you're able to mix them without issues, but for load order a Skyrim ESP plugin won't have the key type of a Cyberpunk redmod:

```csharp
class FooKey : ISortableItemKey;
class BarKey : ISortableItemKey;

class FooSortableItem : ISortableItem
{
  FooSortableItem()
  {
    Key = new BarKey(); // no safeguards
  }
}

class FooSortableItemProvider : ILoadoutSortableItemProvider
{
  void DoSomething(ISortableItem sortableItem)
  {
    // unnecessary checks
    if (sortableItem is not FooSortableItem foo) throw new NotSupportedException();
    if (foo.Key is not FooKey fooKey) throw new NotSupportedException();
  }
}
```

I believe the API would be easier to use if we have generics on the key and item type:

```csharp
interface ISortableItem<TKey> where TKey : IEquatable<TKey>
{
  TKey Key { get; }
}
```

```csharp
interface ILoadoutSortableItemProvider<TItem, TKey>
  where TKey : IEquatable<TKey>
  where TItem : ISortableItem<TKey>;
```

Similar to how our custom TreeDataGrid APIs work.

erri120 (changes requested):

src/Abstractions/NexusMods.Abstractions.Games/ISortItemKey.cs
Al12rs: I tried something like this but I ran into issues with the bindings XAML.

But I can try again. I also think that generics would be more suitable than the current approach.

erri120: The core types (item and provider) should definitely be generic. If you run into a situation where the generic types are unknown you can always have a non-generic interface that's essentially empty:

```csharp
interface IFoo;

interface IFoo<T1, T2, T3> : IFoo;
```

src/Abstractions/NexusMods.Abstractions.Games/ASortableItemProviderFactory.cs
```diff
@@ -0,0 +1,143 @@
...
+    public virtual string IndexColumnHeader => "LOAD ORDER";
+
+    /// <inheritdoc />
+    public virtual string DisplayNameColumnHeader => "NAME";
+
+    /// <inheritdoc />
+    public virtual string EmptyStateMessageTitle => "No Sortable Mods detected";
+
+    /// <inheritdoc />
+    public virtual string EmptyStateMessageContents => "Some mods may modify the same game assets. When detected, they will be sortable via this interface.";
```
erri120, line 45: Can we group these UI messages into a type? `LoadOrderUiMessages` or something. It seems odd to have this in here.

Al12rs: Done

erri120 (approved):

PR #3099, opened 2025-04-24
Add Russian language

MjKey:
I like what you're doing and would really like to have Russian in the app, thank you!


MjKey:
I am prepared to offer translation services into Russian for the application during its development and for subsequent updates.

PR #3104, opened 2025-04-28
Generate new Loadout name as late as possible

### LINKED ISSUE #2977: Loadouts with same name
If you click copy loadout on load outs with lots of mods in the loadouts they create all have the same letter. 

https://github.com/user-attachments/assets/02778fff-6436-45d7-99a9-e6d54bb9be8e

erri120:
Should prevent duplicates when copying the same Loadout multiple times.

Fixes #2977.

### DIFF
--- a/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -1354,18 +1354,8 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
             {
                 // Prime the hash database to make sure it's loaded
                 await _fileHashService.GetFileHashesDb();
-                
-                var existingLoadoutNames = Loadout.All(Connection.Db)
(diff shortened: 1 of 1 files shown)


src/Abstractions/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1610,9 +1592,27 @@ await _jobMonitor.Begin(new UnmanageGameJob(installation), async ctx =>
...
+
+        {
+            using var tx2 = Connection.BeginTransaction();
+            tx2.Add(newLoadoutId, Loadout.Name, newName);
+            tx2.Add(newLoadoutId, Loadout.ShortName, newShortName);
+
+            await tx2.Commit();
+        }
+
+        return Loadout.Load(Connection.Db, newLoadoutId);
```
Al12rs, line 1614: This would first create the copy with the wrong name, then update the name.

I would very much prefer to have a single commit.

What is stopping us from changing the names right before the commit for the first transaction?

Al12rs (changes requested):

Al12rs (approved):

PR #3107, opened 2025-04-28
Use recent collection ratings

### LINKED ISSUE #2370: Change revision percentage on collection download page
On new revisions the number on the collection download page for the positive ratings is blank. 

On the main website we use success ratings from the last 30 days or all time. 

Please change the success rating on the collection download page to 'All Time'.  So there is a correlation between the app and the main page.

erri120:
Resolves #2370


Al12rs:
It seems we have some projects leftover that we should remove:
image

PR #3115, opened 2025-04-29
Cyberpunk 2077 health check

Pickysaurus:
- Updated structure to match SDV/Bannerlord diagnostics
- REDmod DLC diagnostic is now a warning (as there are no ill effects of it being missing, other than REDmods not loading) 
- REDmod DLC diagnostic now only shows if the user has at least 1 REDMod installed
- REDmod DLC diagnostic now gives a count of how many mods require it
- ArchiveXL diagnostic now checks `/mods/` and `/archives/` rather than the game root as it's base folder.
- Linux WINE DLL and Winetricks diagnostics now use correct bullet point styles
- Only show the Protontricks Health Check if REDmod DLC is installed.


src/NexusMods.Games.RedEngine/Cyberpunk2077/Diagnostics.cs
```diff
@@ -142,13 +175,22 @@ the following usecase of {DependencyName}
...
         .Start()
         .WithId(new DiagnosticId(Source, number: 3))
-        .WithTitle("Missing 'Protontricks' dependency")
+        .WithTitle("Protontricks is not installed")
         .WithSeverity(DiagnosticSeverity.Critical)
-        .WithSummary("Protontricks is required to install REDmods but is not present.")
+        .WithSummary("Protontricks is required to use REDmods but is not present.")
         .WithDetails("""
-Protontricks is required to install REDmods but is not present.
+Protontricks is not installed by is required to use the REDmod DLC when playing Cyberpunk 2077 on Linux.
```
Al12rs, line 182:
```suggestion
Protontricks is not installed but is required to use the REDmod DLC when playing Cyberpunk 2077 on Linux.
```

src/NexusMods.Games.RedEngine/Cyberpunk2077/Diagnostics.cs
```diff
@@ -142,13 +175,22 @@ the following usecase of {DependencyName}
...
+
+### How to Resolve
+1. Install Protontricks by following the instructions in the {ProtontricksUri}.
+2. Apply a loadout with Protontricks installed.
```
Al12rs, line 186: I think we might need the user to restart the app for this to detect the presence of protontricks, but I'm not 100% on that

src/NexusMods.Games.RedEngine/Cyberpunk2077/Diagnostics.cs
```diff
@@ -178,13 +237,17 @@ Please install the {RedmodLink} and start the game at least once.
...
         .Start()
         .WithId(new DiagnosticId(Source, number: 5))
-        .WithTitle("Update WINE DLL overrides to mod on Linux")
+        .WithTitle("Incorrect WINE configuration")
         .WithSeverity(DiagnosticSeverity.Critical)
-        .WithSummary("These changes are required for Linux")
+        .WithSummary("Update the WINE DLL overrides to mod on Linux")
         .WithDetails("""
-Modding Cyberpunk 2077 on Linux requires you to configure WINE.
+Modding Cyberpunk 2077 on Linux requires you to configure WINE or the game will not launch correctly.
```
Al12rs, line 244: This could be slightly confusing, since the user has to change command line options of the game executable, and not some setting of WINE.

Under technical details we could add the explanation for why this change is necessary.

Something like this (from Gemini):

> What is a WINE DLL OVERRIDE?

> A WINE DLL override is a configuration setting that tells WINE not to use its own built-in version of a specific Windows DLL, but instead to use a native (real) Windows DLL file that you provide.

> Normally, when a Windows program running on WINE asks for a function from d3d11.dll (a DirectX graphics library), WINE provides its own version of d3d11.dll.
With an override for d3d11.dll, you put a real d3d11.dll file (copied from a Windows installation, for example) into the WINE environment, and you configure WINE to use that file instead of its internal version.
You typically configure these overrides using the winecfg tool, where you can set the loading order for specific DLLs (e.g., "native, builtin" meaning try the native file first, then fall back to WINE's if the native one isn't found, or "native" meaning only use the native file).

> Why is it Needed to Mod Certain Games on WINE?

> Mod Injectors/Loaders: Many mods use custom loaders or injectors to hook into the game process that are implemented as wrapper or replacements for system DLLs. These mod-provided DLLs would normally get loaded on Windows when they are found inside the game folder, but on WINE, the default behaviour is different and the internal WINE versions are used by default instead. Configuring WINE DLL OVERRIDEs allows changing that behaviour to instead load the mod provided DLLs.

src/NexusMods.Games.RedEngine/Cyberpunk2077/Diagnostics.cs
```diff
@@ -196,13 +259,17 @@ Modding Cyberpunk 2077 on Linux requires you to configure WINE.
...
-        .WithSummary("These packages are required or Linux")
+        .WithSummary("Install required packages to mod on Linux")
         .WithDetails("""
 Modding Cyberpunk 2077 on Linux requires packages to be installed in your WINE prefix.
 
+### How to Resolve
 {Instructions}  
+
+### Technical Details
+The current WINE prefix does not have the required packages installed for modding on Linux.
```
Al12rs, line 272: We could add a more detailed explanation of what Winetricks is and why it would be helpful (ask gemini or chatGPT for starting point)

Al12rs (commented):
Some suggestions above

src/NexusMods.Games.RedEngine/Cyberpunk2077/Diagnostics.cs
Pickysaurus: @Sewer56 do you happen to know? The command to check for PT seems to run every time the Health Check is evaluated.

Pickysaurus:
Added some more information to the Linux messages.

src/NexusMods.Games.RedEngine/Cyberpunk2077/Diagnostics.cs
```diff
@@ -263,7 +267,8 @@ Modding Cyberpunk 2077 on Linux requires you to configure WINE or the game will
         .WithSeverity(DiagnosticSeverity.Critical)
         .WithSummary("Install required packages to mod on Linux")
         .WithDetails("""
-Modding Cyberpunk 2077 on Linux requires packages to be installed in your WINE prefix.
+Modding Cyberpunk 2077 on Linux requires certain packages to be installed in your WINE prefix. Winetricks packages are often required for modding because they provide essential components or libraries that the game or its mods depend on to function correctly.
```
Al12rs, line 270: It's the app that requires Winetricks to be able to manage the game installation correctly, rather than the game or the mods.

Sewer56 (approved):

PR #3123, opened 2025-04-30
Close tabs on middle click

### LINKED ISSUE #1755: Support closing tabs with Middle Click
# User story

**As a** user

**I want to** close workspace panel tabs with middle click

**So that** tabs are easier to manage and behave like browser tabs.

# Details
Using mouse Middle Click anywhere in the body of the tab header should close the tab as if pressing the close button, same as how browsers work.

Al12rs:
Repost of #3122 to add changes

- Closes #1755.

This only closes tabs if the pointer is inside the bounds of the tab header at the time of both pressing and releasing, similar to how it works in popular web browsers.

Credits to @JonathanFeenstra

### DIFF
--- a/src/NexusMods.App.UI/WorkspaceSystem/PanelTabHeader/PanelTabHeaderView.axaml.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/PanelTabHeader/PanelTabHeaderView.axaml.cs
@@ -4,7 +4,6 @@ using System.Reactive.Linq;
 using Avalonia.Controls;
 using Avalonia.Input;
 using Avalonia.ReactiveUI;
-using DynamicData.Kernel;
 using JetBrains.Annotations;
 using NexusMods.App.UI.Extensions;
 using ReactiveUI;
--- a/src/NexusMods.App.UI/WorkspaceSystem/PanelTabHeader/PanelTabHeaderView.axaml.cs
+++ b/src/NexusMods.App.UI/WorkspaceSystem/PanelTabHeader/PanelTabHeaderView.axaml.cs
@@ -4,6 +4,7 @@ using System.Reactive.Linq;
 using Avalonia.Controls;
 using Avalonia.Input;
 using Avalonia.ReactiveUI;
+using DynamicData.Kernel;
 using JetBrains.Annotations;
 using NexusMods.App.UI.Extensions;
 using ReactiveUI;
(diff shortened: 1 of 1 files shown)


src/NexusMods.App.UI/WorkspaceSystem/PanelTabHeader/PanelTabHeaderView.axaml.cs
```diff
@@ -55,6 +58,19 @@ public PanelTabHeaderView()
...
+                removeHandler => Container.PointerReleased -= removeHandler
+                ).Where(eventPattern => eventPattern.EventArgs.InitialPressMouseButton == MouseButton.Middle && 
+                                        Bounds.Contains(eventPattern.EventArgs.GetPosition(this)) && 
+                                        eventPattern.EventArgs.GetIntermediatePoints(this).FirstOrOptional(_ => true)
+                                            .Convert(point => Bounds.Contains(point.Position)) 
+                                            .ValueOr(false)
+                )
+                .Select(_ => Unit.Default)
+                .InvokeReactiveCommand(this, view => view.ViewModel!.CloseTabCommand)
+                .DisposeWith(disposables);
```
erri120, line 73: The check can be simplified. I'd almost wager that we don't have to check the position at all and we can get by with only checking the mouse button since the release event would only be send to the control that is underneath the pointer when a button gets released.

`GetIntermediatePoints` will always return one item for the pointer released event which is the same as returned by `GetPosition`/`GetCurrentPoint`. Those points are only set for the generic mouse move event.

```suggestion
            Observable
                .FromEventPattern<PointerReleasedEventArgs>(
                    addHandler => Container.PointerReleased += addHandler,
                    removeHandler => Container.PointerReleased -= removeHandler
                )
                .Select(static eventPattern => eventPattern.EventArgs)
                .Where(eventArgs =>
                {
                    if (eventArgs.InitialPressMouseButton != MouseButton.Middle) return false;

                    var point = eventArgs.GetPosition(this);
                    return Bounds.Contains(point);
                })
                .Select(_ => Unit.Default)
                .InvokeReactiveCommand(this, view => view.ViewModel!.CloseTabCommand)
                .DisposeWith(disposables);
```

erri120 (changes requested):

src/NexusMods.App.UI/WorkspaceSystem/PanelTabHeader/PanelTabHeaderView.axaml.cs
Al12rs: Good catch, this seems to work as desired

erri120 (approved):

PR #3131, opened 2025-04-30
Simple collection creation from CLI

erri120:
Only available in release, no tests yet or anything special. This is just for testing the limitations and an initial implementation.

This PR updates a bunch of models but only for marking attributes as optional.

### DIFF
--- a/src/NexusMods.Collections/CollectionCreator.cs
+++ b/src/NexusMods.Collections/CollectionCreator.cs
@@ -15,11 +15,11 @@ namespace NexusMods.Collections;
 
 public static class CollectionCreator
 {
-    public static CollectionRoot CreateCollectionManifest(
+    public static (CollectionRoot, CollectionMetadata.New, CollectionRevisionMetadata.New) CreateCollection(
         ITransaction tx,
         LoadoutItemGroup.ReadOnly group,
         IGameDomainToGameIdMappingCache mappingCache,
-        User.ReadOnly creator,
+        User.ReadOnly author,
         CollectionSlug collectionSlug)
     {
         var gameId = group.AsLoadoutItem().Loadout.Installation.GameId;
@@ -30,7 +30,7 @@ public static class CollectionCreator
             Name = group.AsLoadoutItem().Name,
             GameId = gameId,
             Slug = collectionSlug,
-            AuthorId = creator.UserId,
+            AuthorId = author,
         };
 
         var collectionRevisionMetadata = new CollectionRevisionMetadata.New(tx)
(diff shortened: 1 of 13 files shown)


src/NexusMods.Collections/Verbs.cs
```diff
@@ -107,4 +119,91 @@ private static async Task<int> GatherCollectionDefinitions([Injected] IRenderer
 
         return 0;
     }
+
+    [Verb("create-collection", "Creates a collection")]
+    private static async Task<int> CreateCollection(
```
Al12rs, line 124: Please better define what the verb does. Does it create a writable collection? A read-only one? Does it create a collection on Nexus?

src/NexusMods.Collections/Verbs.cs
```diff
@@ -107,4 +119,91 @@ private static async Task<int> GatherCollectionDefinitions([Injected] IRenderer
...
+            collectionSlug: CollectionSlug.From(slug)
+        );
+
+        await AddManifest(libraryService, temporaryFileManager, jsonSerializerOptions, tx, collectionManifest, collectionMetadata, collectionRevisionManifest, cancellationToken);
+
+        await tx.Commit();
+        return 0;
+    }
+
+    private static async ValueTask AddManifest(
```
Al12rs, line 180: Add manifest to what? To Nexus? I would suggest using a more descriptive name for the function or adding a comment to define its scope

src/NexusMods.Collections/CollectionCreator.cs
```diff
@@ -0,0 +1,159 @@
...
+using NexusMods.Abstractions.NexusModsLibrary.Models;
+using NexusMods.Abstractions.NexusWebApi;
+using NexusMods.Abstractions.NexusWebApi.Types;
+using NexusMods.MnemonicDB.Abstractions;
+using NexusMods.Paths;
+using CollectionMod = NexusMods.Abstractions.Collections.Json.Mod;
+
+namespace NexusMods.Collections;
+
+public static class CollectionCreator
```
Al12rs, line 16: It isn't clear what the CollectionCreator is meant to do. Does it add new collections to the app? Does it add new collections to Nexus? Does it just add new revisions to Nexus? 

Please use comments to define the responsibility scope of the class. Can also consider alternative names if they reduce ambiguousness.

src/NexusMods.Collections/CollectionCreator.cs
```diff
@@ -0,0 +1,159 @@
...
+{
+    public static (CollectionRoot, CollectionMetadata.New, CollectionRevisionMetadata.New) CreateCollection(
```
Al12rs, line 18: Similar to class, please add comments to explain what this supposed to do, as CreateCollection is ambiguous and doesn't define the scope of the operation well.

Al12rs (changes requested):

Al12rs (approved):

PR #3144, opened 2025-05-03
Support adding/removing games with CLI without installation required

### LINKED ISSUE #3143: Add and remove supportable games with CLI without installation required
## Summary of Current Behavior:
Currently, there is no feature where you could add/remove supported games without having the game in your local machine. Future developers may want to contribute to the app such as improving UI features or testing current features but do not want to have/buy a supported game.

## Details:
A CLI command exists where you can add a game with `NexusMods.App add-game -g <game> -v <version> - p <path>` However, this command only lets you add games if there exists an installation from your PC. The auto-detector would handle this usually. 

For example, if the developer has Stardew Valley and runs `NexusMods.App add-game -g "Stardew Valley" -v 1.6.15 -p "/foo/faz/bar"`, the CLI works and vice versa. 
Future developers that want to contribute to the app but do not have a supported game cannot access most features. 

As a developer, I want to be able to register a game into the database without having the game to access other features for future development/contributions. Although, it is expected that some features may not function due to requiring the game files. This may include installing mods into the game's loadouts and playing the game.

What if the developer wants to only remove the manually-added game from the database? The only way to doing so is to use the following CLI command: `NexusMods.App uninstall-app`, which not only does this but also wipes everything to a new state.

As a developer, I want to remove a manually-added game without necessarily clearing out all app settings.

### Steps to Reproduce Bug
1. __Without a supported game installed__, run the CLI `NexusMods.App as-main add-game -g <game> -v <version> - p <path>`
2. The app should error out by not finding the game in your PC.

### Expected behaviour
1. Run the CLI `NexusMods.App as-main add-game -g <game> -v <version> - p <path>`
2. Opening the app should have the game appear in "My Games." and will have access to most features such as Collections, downloading mods, accessing Loadouts, etc. Some features are expected to not work as well.
3. If the developer does not want the test game anymore in the app's database, use a CLI command to remove the game, including the loadouts associated with the game. (i.e. `NexusMods.App as-main remove-game -g <game> -id <EntityID>`)
4. When reopening the app, the game and its loadouts should not be listed as installed, and the app should not clean reset.

## Development:
See PR #3144 

### Design
N/A

### Definition of Done
I do not believe users should have this CLI feature due to its purposes for development, but I am not sure how it would be restricted to something like "Developer Mode" only. Users are free to have it however if allowed.

### Test Data:
_(Any test data required to test the feature, e.g. links to mods, required setups, collections etc.)_
Saving the `EntityID` generated for the manually added game should be required when removing a game due to the `locator.Remove()` function, where an EntityID is needed to verify the game as "Manually Added"

#### Operating system:

- [ ]  Windows 10
- [x] Windows 11
- [ ] Linux (distro details)

**App version: 0.10.2 dev**

#### NexusMods account type:
- [x] Premium
- [x] Free

### QA Checklist:
- [ ] Add QA label if relevant
- [x] Add repro steps
- [x] Add expected behaviour
- [x] Add known limitations with the fix or issues
- [ ] QA test Windows
- [ ] QA test Linux

axmanalad:
Fixes #3143 

This PR refactors the game parsing behavior for `add-game` CLI to allow access for future developers/contributors to add supported games without having the game installed in their PC. The output is updated to provide the EntityID for deletion purposes. 

This also adds a new CLI command `NexusMods.App remove-game -g <game> -id <EntityID>` which will remove the game from the app’s database and the loadouts associated with the game. The game parameter is required to remove loadouts while the `EntityID` is used to remove the game from the app’s database.

I included more documentation in the Contributing docs to emphasize the requirement of a supported game for full functionality and steps to contribute without a supported game installed.

_Note: I am not sure how to return the boolean value of True with warning message to the developer in at the end of `GameParser.cs` so I just left it there as a note._

### DIFF
--- a/src/NexusMods.App.Cli/OptionParsers/GameParser.cs
+++ b/src/NexusMods.App.Cli/OptionParsers/GameParser.cs
@@ -11,11 +11,33 @@ internal class GameParser(IGameRegistry gameRegistry) : IOptionParser<IGame>
 {
     public bool TryParse(string toParse, out IGame value, out string error)
     {
-        var install = gameRegistry.Installations.Values.FirstOrDefault(g => g.Game.GameId.ToString() == toParse) ??
-                    gameRegistry.Installations.Values.FirstOrDefault(g => g.Game.Name.Equals(toParse, StringComparison.CurrentCultureIgnoreCase))!;
-
-        value = (IGame)install.Game;
-        error = string.Empty;
-        return true;
+        try
+        {
+            var install = gameRegistry.Installations.Values.FirstOrDefault(g => g.Game.GameId.ToString() == toParse) ??
+                          gameRegistry.Installations.Values.FirstOrDefault(g => g.Game.Name.Equals(toParse, StringComparison.CurrentCultureIgnoreCase)) ??
+                          GameInstallation.Empty;
+            if (install.Equals(GameInstallation.Empty))
+            {
+                throw new NullReferenceException();
+            }
+            value = (IGame)install.Game;
+            error = string.Empty;
+            return true;
+        }
+        catch (NullReferenceException)
+        {
+            // Recheck if current game parse is supported
+            var install = gameRegistry.SupportedGames.FirstOrDefault(g => g.GameId.ToString() == toParse) ??
+                       gameRegistry.SupportedGames.FirstOrDefault(g => g.Name.Equals(toParse, StringComparison.CurrentCultureIgnoreCase))!;
+            if (install == null!)
+            {
+                value = null!;
+                error = $"Game '{toParse}' is not found or supported.";
+                return false;
+            }
+            value = (IGame)install;
+            error = $"Warning: Game '{toParse}' is supported but not installed. Please install it properly for full functionality.";
+            return true;
+        }
     }
 }
--- a/docs/developers/Contributing.md
+++ b/docs/developers/Contributing.md
@@ -21,6 +21,27 @@ helping you make things like new 'Windows' and 'Controls' easier.
 
 Make sure to follow our [Development Guidelines](./development-guidelines/UICodingGuidelines.md).
 
+### Debugging the App
+This app is designed to be used with [games](../users/games/index.md) that are currently supported. It is recommended to
+have a supported game installed on your system to debug the app with full functionality.
+
+If you want to contribute to the app without a supported game installed, you can continue
+doing so by manually adding a supported game. However, supported games that are manually added and not installed properly for 
+testing purposes may not be playable. You can still use the app to test the UI and other features, 
+but some features may not work as expected. To add a supported game manually:
+
+1. Have a directory ready and copy its path for the game you want to add to store its mods.
+2. After building the project, go to your terminal and change the directory to the `src/NexusMods.App/bin/Debug/net9.0` folder.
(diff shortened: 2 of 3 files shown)


src/NexusMods.App.Cli/OptionParsers/GameParser.cs
```diff
@@ -11,11 +11,33 @@ internal class GameParser(IGameRegistry gameRegistry) : IOptionParser<IGame>
...
+            if (install == null!)
+            {
+                value = null!;
+                error = $"Game '{toParse}' is not found or supported.";
+                return false;
+            }
+            value = (IGame)install;
+            error = $"Warning: Game '{toParse}' is supported but not installed. Please install it properly for full functionality.";
+            return true;
+        }
```
erri120, line 41: Don't use exceptions for control flow, this can be simplified:

```suggestion
        var game = gameRegistry.SupportedGames.FirstOrDefault(g => g.Name.Equals(toParse, StringComparison.OrdinalIgnoreCase));
        if (game is null && uint.TryParse(toParse, out var parsedGameId))
        {
            game = gameRegistry.SupportedGames.FirstOrDefault(g => g.GameId.Equals(parsedGameId));
        }

        if (game is null)
        {
            value = null!;
            error = $"Unknown game: `{toParse}`";
            return false;
        }

        value = (IGame)game;
        error = string.Empty;
        return true;
```

erri120 (changes requested):

axmanalad:
Thank you @erri120 for reviewing my PR. I noticed that adding a game manually shows version "Unknown" in the front-end even if the add-game CLI passes in the version; I went forward to fix that issue as well.

erri120:
> Thank you @erri120 for reviewing my PR. I noticed that adding a game manually shows version "Unknown" in the front-end even if the add-game CLI passes in the version; I went forward to fix that issue as well.

The vanity version we show in the UI comes from our [game hashes](https://github.com/Nexus-Mods/game-hashes) database. The `ManuallyAddedGame.Version` field is kinda a left-over as it does nothing anymore similar to how `GameLocatorResult.Version` does nothing anymore. No need to change any code related to the versions.

axmanalad:
> The vanity version we show in the UI comes from our [game hashes](https://github.com/Nexus-Mods/game-hashes) database. The `ManuallyAddedGame.Version` field is kinda a left-over as it does nothing anymore similar to how `GameLocatorResult.Version` does nothing anymore. No need to change any code related to the versions.

Understandable, I did not know that the versions fields for `ManuallyAddedGame` and `GameLocatorResult` were outdated. I guess for the time being, the version will be another part of the UI that does not work as intended when adding a game manually. Thanks for letting me know. 👍

Pickysaurus:
What's the use case for this? I would say we actually don't want contributions where the developer hasn't even installed the game as they haven't verified it actually works.

Al12rs:
> I would say we actually don't want contributions where the developer hasn't even installed the game as they haven't verified it actually works.

I don't think this would be a problem. Having a fake game or a real one doesn't matter for a lot of code changes (anything with the UI for example). Our CI pipeline uses the same idea here of stubbing the game, but it is still useful.

erri120 (changes requested):

erri120 (approved):

PR #2985, opened 2025-04-02
[Draft] Added: TreeFolderGenerator for Dynamically Generating Observable Folders given a List of Files

Sewer56:
More details coming soon. Need to resolve conflicts with `main`, and do a second take over the code.

#2862.

PR #3158, opened 2025-05-06
Added: Support for Folders in the View Loadout Group Files

Sewer56:
Adds the following:

- A Folder Generator (`TreeFolderGenerator.cs`) that (originally #2985) which can be used to build a dynamic, reactive tree structure based off of Receive/Remove file events.
- An adapter `TreeFolderGeneratorCompositeItemModelAdapter` which integrates the observable stream of files into the `TreeFolderGenerator`.
- Support for hierarchical view in `ViewLoadoutGroupFilesTreeDataGridAdapter` via the new `Folder Generator`.

- Feature: View Loadout Group files now support showing multiple categories, e.g. `GAME`, `SAVES`, `APPDATA` at the same time.
    - Before it was hardcoded to show only `GAME` files.
    - Now it will show files from only 1 category if there's only 1 category (e.g. if we only have `GAME` files, behaviour is the same).
    - But if we have multiple roots, `GAME`, `SAVES`, `APPDATA`, we will now show all the roots. 

There are also a lot of (basic) tests. Since the code was fairly complex, it helped me catch and prevent buggy behaviour a lot.

### DIFF
--- a/tests/NexusMods.UI.Tests/Helpers/TreeDataGrid/FolderGenerator/TreeFolderGeneratorTests.cs
+++ b/tests/NexusMods.UI.Tests/Helpers/TreeDataGrid/FolderGenerator/TreeFolderGeneratorTests.cs
@@ -241,7 +241,7 @@ public class TreeFolderGeneratorTests
         generator.OnReceiveFile(testItem2, fileModel2);
         
         // Assert
-        addedItems.Should().HaveCount(1, "one root item should be added");
+        addedItems.Should().HaveCount(2, "two root items should be added");
     }
     
     [Fact]


Sewer56:
Fails tests? Weird, I think I have a patch around this code on my other branch, lemme backport for a sec.

src/NexusMods.App.UI/Helpers/TreeDataGrid/New/FolderGenerator/Helpers/TreeFolderGeneratorCompositeItemModelAdapter.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// <param name="changes">The changeset containing changes to process.</param>
+    private void Adapt(IChangeSet<CompositeItemModel<EntityId>, EntityId> changes)
+    {
+        foreach (var change in changes)
+        {
+            switch (change.Reason)
+            {
+                case ChangeReason.Add:
+                case ChangeReason.Update:
+                    FolderGenerator.OnReceiveFile(_factory.CreateItem(change.Current.Key), change.Current);
```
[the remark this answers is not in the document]
Sewer56: This goes in a future PR.
Please open a ticket.

It's not always possible to do all removals/additions in batch per-se, due to the fact that new additions would go into multiple SourceCache(s). However, doing it in batch per-folder is possible.

That would yield approximately a 7x reduction in notifications, as that's the average ratio of files to folders when I tested on 100 random games back in 2023.

src/NexusMods.App.UI/Pages/ItemContentsFileTree/New/ViewLoadoutGroupFiles/ViewLoadoutGroupFilesTreeDataGridAdapter.cs
```diff
@@ -12,11 +19,22 @@ namespace NexusMods.App.UI.Pages.ItemContentsFileTree.New.ViewLoadoutGroupFiles;
...
+        if (!viewHierarchical)
+            return filesObservable;
+
+        // Note(sewer): This is 'hardcoded' to display only mod files for the time being,
+        // as to retain the existing behaviour/UI we had beforehand. But we can now display
+        // things such as saves.
+        
+        // Make a folder generator and adapt
+        var connection = serviceProvider.GetRequiredService<IConnection>();
+        var adapter = new TreeFolderGeneratorLoadoutTreeItemAdapter<LoadoutGroupFilesTreeFolderModelInitializer>(connection, filesObservable);
```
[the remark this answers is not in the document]
Sewer56: Handled in commit c646db083d8a9c72419a1c32a5f536487c13b7a4 , will backport if no merge conflicts.

src/NexusMods.App.UI/Helpers/TreeDataGrid/New/FolderGenerator/TreeFolderGenerator.cs
```diff
@@ -0,0 +1,99 @@
...
+    internal readonly SourceCache<CompositeItemModel<EntityId>, EntityId> RootCache = new(model => model.Key);
+    private IncrementingNumberGenerator _incrementingNumberGenerator = new();
+    
+    /// <summary>
+    /// Returns an observable changeset of root items, suitable for binding to a TreeDataGrid.
+    /// </summary>
+    public IObservable<IChangeSet<CompositeItemModel<EntityId>, EntityId>> ObservableRoots()
+    {
+        return RootCache.Connect();
+    }
```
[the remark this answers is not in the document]
Sewer56: Handled in commit e1d53ab9e54db2acf6a2e510813d46102984e9ab , will backport.

src/NexusMods.App.UI/Helpers/TreeDataGrid/New/FolderGenerator/TreeFolderGenerator.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// In simpler words, don't show the 'GAME' folder if we only have files in 'GAME'.
+    /// But if we have 'GAME' and 'SAVES', show both!
+    /// </summary>
+    public IObservable<IChangeSet<CompositeItemModel<EntityId>, EntityId>> SimplifiedObservableRoots()
+    {
+        return ObservableRoots()
+            .Select(_ => LocationIdToTree.Count) // tied 1:1 with root count
+            .Select(GetAdaptedChangeSet) // get either changeset with 1 root, or with all roots.
+            .Switch();
+    }
```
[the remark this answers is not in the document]
Sewer56: (I lost the original comment because of browser crash on unsuspend -.-)

Anyway, the refcount here isn't strictly necessary. There is only 1 instance of `SimplifiedObservableRoots` / `ObservableRoots` used at any given time; and that's created on activation of the `TreeDataGridAdapter`, by the `TreeDataGridAdapter` base class.

When the adapter is disposed, that single connection is dropped. When `ViewHierarchical` is toggled, causing another call, the previous observable will be disposed.

Where it's useful to have refcount is to reduce the latency between `ViewHierarchical` switches (avoid memory pressure); so I applied it to just the core `ObservableRoots()` to achieve this.

src/NexusMods.App.UI/Helpers/TreeDataGrid/New/FolderGenerator/Helpers/TreeFolderGeneratorCompositeItemModelAdapter.cs
```diff
@@ -0,0 +1,79 @@
...
+    private readonly TTreeItemWithPathFactory _factory;
+
+    /// <summary>
+    /// Creates a new instance of <see cref="TreeFolderGeneratorCompositeItemModelAdapter{TTreeItemWithPath,TTreeItemWithPathFactory,TKey,TFolderModelInitializer}"/>.
+    /// </summary>
+    /// <param name="factory">Factory that creates <typeparamref name="TTreeItemWithPath"/> instances.</param>
+    /// <param name="changes">Observable of changes to automatically process.</param>
+    public TreeFolderGeneratorCompositeItemModelAdapter(
+        TTreeItemWithPathFactory factory,
+        IObservable<IChangeSet<CompositeItemModel<EntityId>, EntityId>> changes)
```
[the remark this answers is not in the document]
Sewer56: Saves memory, since we have no use for the key otherwise; just that DynamicData wants one.
But ya, it's doable.

src/NexusMods.App.UI/Helpers/TreeDataGrid/New/FolderGenerator/TreeFolderGenerator.cs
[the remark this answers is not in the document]
Sewer56: I improved memory further in the other branch:
- 8eeddd2c3f32f008bf9c06b66a224a597cc1b2d9 , immediately disposes the observable when `ViewHierarchical` is toggled.
- f2663a5c4298fd4f7b3d72c235042f0a75e0bc64 ,  ensures data is immediately disposed in case the `Context` is ever switched out; and on view activation/deactivation (useful for tab switching).
- e1d53ab9e54db2acf6a2e510813d46102984e9ab has the requested change.

Sewer56:
@Al12rs I moved the code/items as requested in 9bc6da932d3e4a69ef557d8546a0702cde858411

src/NexusMods.App.UI/Helpers/TreeDataGrid/New/FolderGenerator/Helpers/TreeFolderGeneratorCompositeItemModelAdapter.cs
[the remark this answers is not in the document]
Sewer56: Getting this sorted on the other branch, it's a bit of an expensive change, but I'm mostly there.

tests/NexusMods.UI.Tests/Helpers/TreeDataGrid/FolderGenerator/FolderModelInitializer.Example.FileSize.cs
Al12rs, line 1: Same for as FileCount, once the we have this in the actual view, this should be removed

tests/NexusMods.UI.Tests/Helpers/TreeDataGrid/FolderGenerator/TreeFolderGeneratorForLocationIdTests.cs
Al12rs, line 1: There are a lot of tests here, for something that is technically an implementation detail (`TreeFolderGeneratorForLocationId`).

I'm sure they helped a lot during development, but I'm not sure about the cost/benefit they offer now.

You would normally want to test the higher layer, like the DataProvider, but in this case this generator was extracted from the provider and attached as a step after it in the Adapter. 

For example, some of these tests will have to be rewritten or changed for 
- #3166 
since they rely on the implementation detail of  `OnReceiveFile` rather than using the DB state as input state like the data provider would.

tests/NexusMods.UI.Tests/Helpers/TreeDataGrid/FolderGenerator/TreeFolderGeneratorTests.cs
```diff
@@ -0,0 +1,594 @@
...
+                        addedItems.Add(change.Current);
+                    }
+                }
+            });
+            
+        // Act
+        generator.OnReceiveFile(testItem, fileModel);
+        
+        // Assert
+        addedItems.Should().HaveCount(1, "one root item should be added");
```
Al12rs, line 164: All tests that are using a subscription here are assuming that the the data in the subscription is updated immediately as part of `OnReceiveFile`.

This is actually not guaranteed as soon as there is an `ObserveOn` somewhere in the observable chain, in which case the OnNext() returns early after scheduling the rest of the chain on a dispatcher (even if it stays on the same thread).

That would mean that `generator.OnReceiveFile(testItem, fileModel);` could return without `addedItems` having been updated yet, which will fail the test.

Al12rs: Not sure what to do here, you could add a tascCompletionSource and add a timeout, similar to what I tried to do in the Load Order tests, but I've had to disable those because they seemed to keep stalling the CI (though we are still stalling on other tests now).

Alternative is to just scrap the tests since they can fail even if the code continues to works perfectly fine.

Al12rs (changes requested):
See comments for Observing changes in the tests

src/NexusMods.App.UI/Helpers/TreeDataGrid/New/FolderGenerator/TreeFolderGeneratorForLocationId.cs
Al12rs, line 1: I would maybe have considered just passing a bool flag or something to define whether we want the locationId entry or not. This seems like a lot of boilerplate.

Didn't like the stuff with the EntityId in here lol

tests/NexusMods.UI.Tests/Helpers/TreeDataGrid/FolderGenerator/TreeFolderGeneratorForLocationIdTests.cs
Sewer56: The purpose of tests is threefold:

- Ensure existing code does not break with future changes
- Prototype during development / aid development
- As an example of how to use an API (filecount / filesize).

In this case, the FileCount and FileSize are done in the actual view, on the other branch; and the 
are already done in the actual view; and this is synced with that. 

These tests are useful, do not touch them. I just did a pretty expensive migration to using `GamePath` as a key (as requested) that took hours. The tests took minutes to fix, and they identified issues immediately; issues which I otherwise would have not noticed without having to manually debug through the code while in the app and try every single operation. Am still patching the minor details, mostly done on that end though.

src/NexusMods.App.UI/Helpers/TreeDataGrid/New/FolderGenerator/Helpers/TreeFolderGeneratorCompositeItemModelAdapter.cs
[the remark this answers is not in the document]
Sewer56: eb53647462ec48fd52416b403df8bbb87731c137

Took like half a day, but there we are.

src/NexusMods.App.UI/Helpers/TreeDataGrid/New/FolderGenerator/TreeFolderGeneratorForLocationId.cs
Sewer56: Been moved to GamePath last night, in either case.

tests/NexusMods.UI.Tests/Helpers/TreeDataGrid/FolderGenerator/TreeFolderGeneratorTests.cs
Sewer56: The data here *is* updated immediately.

There is no `ObserveOn` in the chain here, as the `TreeFolderGenerator` does not enforce a default observer on the caller (and it shouldn't).

For UI code, i.e. ViewModels, it's the ViewModels' responsibility to ensure the updates are received on the correct thread. 

In this case, for the purposes of the UI it's reached from the `TreeDataGridAdapter`

```csharp
return self
    .GetRootsObservable(viewHierarchical)
    .OnUI() // ObserveOn
/// ...
```

But when in test code, there is not (and there wouldn't ever be) an ObserveOn in the chain, since it's not the responsibility of the source to specify how messages are received. So the test code runs on the default scheduler, which queues the work on the same thread.

If you run under a debugger and look at the `Subscribe` call, then you'll notice the inner subscribe block is ran on the same thread (and stack trace) as the call to `OnReceiveFile`. That isn't going to change.

Sewer56: I'm aware some tests in the past have explicitly bound themselves to the UI thread, because the underlying code was written to be bound to the UI thread; which enforced that limitation.

That limitation here is not present; I'm not even working with any UI code in these tests.

PR #3163, opened 2025-05-07
Fixes thumbnails being resized incorrectly

### LINKED ISSUE #2975: Thumbnails get stretched with differing widths
> Image
> Doesn't handle differing width formats very well 

 _Originally posted by @RedRanger14 in [#1974](https://github.com/Nexus-Mods/NexusMods.App/issues/1974#issuecomment-2769611303)_

insomnious:
Fixes #2975

I discovered it was an issue in the ImagePipeline code (and not just a UI/Styling stretch option) as all thumbnails downloaded from Nexus were being resized to `90 x 46` before being stored locally.

I've added an optional `maintainAspectRatio` parameter to the ImagePipeline resize function and so we still pass in our thumbnail size we want, but it will resize while maintaining aspect ratio, with the larger of the width/height values being the dominate one.

Also removed the original UI-based placeholder icon for when no images where there to display. The ImagePipeline adds a transparent image as a fallback and so previous way of detecting if null no longer works. I've added a fallback `.png' like the collection one now and so it fixes the problem of the icon showing behind a transparent png.

image

NOTE: This will only fix newly downloaded mods and doesn't go back and resize/redownload thumbnails of items already in the Library. @halgari agreed this is fine.


Al12rs:
@insomnious

insomnious:
> @insomnious `NexusMods.UI.Tests.ImageLoaderTests.Test` seems to be failing

Oh thanks, I'll look

src/NexusMods.Media/ImageResizer.cs
```diff
@@ -7,38 +7,69 @@ public sealed class ImageResizer<TResourceIdentifier> : ANestedResourceLoader<TR
...
+            if (_newSize.Width > _newSize.Height)
+            {
+                var newHeight = (int)(_newSize.Width * aspectRatio);
+                targetSize = new SKSizeI(_newSize.Width, newHeight);
+            }
+            else
+            {
+                var newWidth = (int)(_newSize.Height / aspectRatio);
+                targetSize = new SKSizeI(newWidth, _newSize.Height);
+            }
```
Al12rs, line 47: This currently doesn't take into account the case where the original major axis doesn't match the target major axis, potentially resulting in images that have greater dimensions than the targets.

This alternative should handle that as well:
```suggestion
            var originalAspectRatio = (float)original.Width / original.Height;
            var targetAspectRatio = (float)_newSize.Width / _newSize.Height;
    
            if (originalAspectRatio > targetAspectRatio)
            {
                // Original is wider compared to target - width becomes the limiting factor
                targetSize = new SKSizeI(_newSize.Width, (int)(_newSize.Width / originalAspectRatio));
            }
            else
            {
                // Original is taller compared to target - height becomes the limiting factor
                targetSize = new SKSizeI((int)(_newSize.Height * originalAspectRatio), _newSize.Height);
            }
```

Al12rs (approved):

PR #3167, opened 2025-05-08
Revert parts of #3088

erri120:
Re-adds automatic protocol registration on Linux until we don't need it anymore.


Al12rs (approved):
I didn't test this since I'm on windows, trust you did on your end.

PR #3187, opened 2025-05-14
Fix collection rating parsing

### LINKED ISSUE #2370: Change revision percentage on collection download page

erri120:
Some API endpoints return values between 0 and 100, others between 0 and 1. We previously used the latter which didn't work when switching to the former that resulted in always showing 100%.

Resolves #2370.


Sewer56 (approved):
Am not qualified to comment on these endpoint but I'll trust your word

PR #3190, opened 2025-05-14
Update AppStream metadata file

erri120:
Part of #2436. Also adds a script to generate AppStream release files for us.

### DIFF
--- a/scripts/python/releases-to-appstream.py
+++ b/scripts/python/releases-to-appstream.py
@@ -1,0 +1,72 @@
+#!/usr/bin/python3
+import sys
+import requests
+import xml.etree.ElementTree as ET
+import xml.dom.minidom
+from datetime import datetime
+
+OWNER = "Nexus-Mods"
+REPO = "NexusMods.App"
+GITHUB_API = f"https://api.github.com/repos/{OWNER}/{REPO}/releases"
+
+def fetch_releases():
+    releases = []
+    page = 1
+
+    while True:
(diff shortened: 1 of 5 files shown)


Al12rs:
@Sewer56 Could you have a look over this?

src/NexusMods.App/com.nexusmods.app.metainfo.xml
```diff
@@ -29,7 +34,7 @@
     <url type="contribute">https://nexus-mods.github.io/NexusMods.App/developers/Contributing/</url>
 
     <content_rating type="oars-1.0">
-        <content_attribute id="social-info">moderate</content_attribute>
+        <content_attribute id="social-info">intense</content_attribute>
```
Sewer56, line 37: Our optional, opt-in diagnostics don't (or are not meant to) share anything user identifiable; I think the previous rating of 'moderate' was actually correct.

scripts/python/releases-to-appstream.py
```diff
@@ -0,0 +1,72 @@
...
+        page_releases = response.json()
+        if not page_releases:
+            break
+
+        releases.extend(page_releases)
+        page += 1
+
+    return releases
+
+def generate_appstream_xml(releases):
```
Sewer56, line 30: This function name might confuse someone thinking it's doing substitution into the `metainfo` file at first glance, when it's the `releases` file.

I don't think it's a huge deal, as this is a trivial script, thought I'd just point it out. 
Quite honestly a 1 liner comment at the top of the file would be sufficient.

PR #3194, opened 2025-05-14
Fix for no apply button

halgari:
If a loadout has never been applied (and neither has the game), then there's no "previously applied state" and the loadout/tx state is set to `default` we later on in the synchronizer filter out any default states. 

TL;DR - in some rare cases where a user has *only* ever installed a collection, they may not see an apply button, this fixes that issue.

--

This also does the following: 

* Handles a case in Cyberpunk where there's two files that collide on the minimal hash. If we detect one or more possible hash files for a given minimal hash we bail out and hash the slow way
* Changes a broken mapping that was backing up files for no reason (leftover from when we had ignore paths)
* Adds a verb for manually reindexing a loadout's disk state, for testing

### DIFF
--- a/src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
+++ b/src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
@@ -198,8 +198,7 @@ public class SynchronizerService : ISynchronizerService
         var isBusy = loadoutState.ObservePropertyChanged(l => l.Busy);
 
         var lastApplied = LastAppliedRevisionFor(loadout.InstallationInstance)
-            .ToObservable()
-            .Where(last => last != default(LoadoutWithTxId));
+            .ToObservable();
 
         var revisions = Loadout.RevisionsWithChildUpdates(_conn, loadoutId)
             .ToObservable()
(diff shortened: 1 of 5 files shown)


src/NexusMods.DataModel/Synchronizer/SynchronizerService.cs
```diff
@@ -227,7 +226,7 @@ private Observable<LoadoutSynchronizerState> CreateStatusObservable(LoadoutId lo
                     if (busy)
                         return LoadoutSynchronizerState.Pending;
 
-                    if (last.Id != loadoutId)
+                    if (last.Id != loadoutId && last != default(LoadoutWithTxId))
```
Al12rs, line 229: If the loadout is not the last one applied though, we do need to Apply. 

The problem is that we are in a state with no loadout applied.

This should currently happen while switching loadouts, and only very briefly on loadout creation. But it is somehow currently possible to get past loadout creation and still have no Last Applied loadout

halgari: in that case it's handled further down b the "needs sync"

PR #3199, opened 2025-05-15
Added: Replace Library Items Atomically in Place & Other Library Refactorings

Sewer56:
This PR adds a few things:

- The ability to 'replace' instances of a `LibraryItem` with another `LibraryItem` across all loadouts.
    - Including filter for read only collections.
- A small cleanup of API usage; removing duplicate operations. 
    - e.g. Removing library linked loadout items is now in `LibraryService`, rather than duplicated in multiple classes.
- Fix: Installing an item uses latest DB state rather than one from the item.
    - Fixes a potential issue where a user:
        - Opens Library View
        - Adds a collection from the browser
        - Installs a Library Item to that collection
    - And the operation fails because collection didn't exist at the time the library item was read from the DB.
- Some other library APIs and refactorings, such as `LoadoutsWithLibraryItem` API.

Replace operation does not preserve user ingested files such as configs.

Some of the new APIs for the LibraryService also query or interact with loadouts.
These can live in either a `Loadout` service, a `Library` service (currently), or another service. 
For now I just added them to existing service. Thoughts?

-----------

The ability to properly 'replace' here is dependent on the ability for installers to not leave junk should they suddenly decide that a mod may be incompatible mid way through.

Therefore, I've gone through the effort of reviewing every single installer to ensure that skipping/termination happens before anything is added to the current transaction:

| Installer                           | No Junk | Comment                                                                                                                |
| ----------------------------------- | ------- | ---------------------------------------------------------------------------------------------------------------------- |
| AdvancedManualInstallerUI           | [X]     | Adds to tx after user made choices, if cancelled, tx is not touched.                                                   |
| AdvancedManualInstaller             | [X]     |                                                                                                                        |
| FomodXmlInstaller                   | [X]     | Bails out before adding to tx                                                                                          |
| GenericPatternMatchInstaller        | [X]     | Succeeds if *any* file was deployed. Any file means `tx` was used at least once. No early terminya~tion after `tx`.    |
| BG3SEInstaller                      | [X]     | Bails out early if DLL Hijacking stub DLL is missing.                                                                  |
| BLSEInstaller                       | [X]     | Bails if no BLSE launcher present.                                                                                     |
| BannerlordModInstaller              | [X]     | Bails out if no module info, else proceeds till end.                                                                   |
| AppearancePresetInstaller           | [x]     | Only creates items and uses `tx` if an `".preset"` is found. Success/failure depends if any item was found.            |
| FolderlessModInstaller              | [x]     | Matches `.archive` and docs files. Only uses `tx` if any found, and finding indicates success.                         |
| RedModInstaller                     | [x]     | Bails out if no `info.json` found, `redModInfoFile == null` can terminate but that throws, failing all, so we're good. |
| SimpleOverlayModInstaller           | [x]     | Exits if no roots, else exits only if 1 file.                                                                          |
| GenericInstaller                    | [x]     | All checks done beforehand.                                                                                            |
| SMAPIInstaller                      | [x]     | Bails out early. May encounter errors later, but this does not return unsupported.                                     |
| FallbackCollectionDownloadInstaller | [x]     | Always succeeds.                                                                                                       |

There are cases where installers throw exceptions if something 'exceptional' happens; but that terminates the whole operation rather than delegating to next, which is considered ok in this scenario.


src/NexusMods.Abstractions.Library/ILibraryService.cs
```diff
@@ -28,30 +29,166 @@ public interface ILibraryService
...
+    /// <summary>
+    /// Ignores items in ReadOnly collections such as collections from Nexus Mods.
+    /// </summary>
+    public bool IgnoreReadOnlyCollections { get; set; }
+    
+    /// <summary>
+    /// Gets the <see cref="ReplaceLibraryItemOptions"/> for this <see cref="ReplaceLibraryItemsOptions"/>
+    /// </summary>
+    public ReplaceLibraryItemOptions ToReplaceLibraryItemOptions() => new() { IgnoreReadOnlyCollections = IgnoreReadOnlyCollections };
 }
```
erri120, line 194: Why do we need all of these methods?

Al12rs:
@Sewer56 Build is failing

Sewer56:
> @Sewer56 Build is failing

I guess something must have changed on main since, hmmmmm.

Sewer56:
CC @Al12rs

src/NexusMods.Abstractions.Loadouts/Models/LibraryLinkedLoadoutItem.cs
```diff
@@ -21,4 +21,13 @@ public partial class LibraryLinkedLoadoutItem : IModelDefinition
...
     /// </summary>
     public static readonly ReferenceAttribute<LibraryItem> LibraryItem = new(Namespace, nameof(LibraryItem)) { IsIndexed = true };
+
+    public readonly partial struct ReadOnly
+    {
+        /// <summary>
+        /// Tries converting this entity to a <see cref="LoadoutItem"/> entity,
+        /// if the entity is not a <see cref="LoadoutItem"/> entity, it returns false.
+        /// </summary>
+        public LoadoutItem.ReadOnly AsLoadoutItem() => new(Db, Id);
```
Al12rs, line 31: Comment doesn't seem correct

src/NexusMods.Abstractions.Library/Jobs/IInstallLoadoutItemJob.cs
```diff
@@ -30,3 +31,15 @@ public interface IInstallLoadoutItemJob : IJobDefinition<LoadoutItemGroup.ReadOn
...
+/// The result of installing a loadout item via the <see cref="IInstallLoadoutItemJob"/>.
+/// This struct holds a <see cref="LoadoutItemGroup"/> for the item which was just installed.
+///
+/// If the value is 'null' then the job was attached to an existing, external transaction
+/// to be part of a larger atomic operation.
+/// (Done by passing an <see cref="ITransaction"/> transaction to the job.)
+/// This is because the value is not yet available; as the transaction
+/// needs to be externally committed by the caller.
+/// </summary>
+public record struct InstallLoadoutItemJobResult(LoadoutItemGroup.ReadOnly? LoadoutItemGroup);
```
Al12rs, line 45: What is the advantage of passing a struct record of a nullable value rather than just passing the nullable value?

Sewer56: I wanted to:

- Add documentation to the return type in an explicit manner.
- Avoid excessive refactors if there's ever a need to add more values in the return structure/type.

Mainly the first one. This is actually called `Nominal Typing`, https://nexus-mods.github.io/NexusMods.App/developers/decisions/project/0004-use-nominal-typing/

Al12rs (approved):

Sewer56:
:eyes: :ok_hand:

PR #3202, opened 2025-05-16
Fix for the app never exiting

halgari:
Updates to the newest version of MnemonicDB which reworks how the application exits. Hard-exits the app after 15 seconds if it doesn't properly shutdown. Fix a but with crashing if the last scanned transaction datom wasn't set, we don't need it anyways, so we just use `.DiskEntries` on the game object instead


halgari:
Merging this so I can back-port and get the changes into 0.11.1 for QA testing

PR #3193, opened 2025-05-14
More http download fixes

erri120:
- Use default HTTP client config
- Add retry on network exceptions
- Add setting to control max parallel download count for collections

PR #3203, opened 2025-05-16
Backport fixes from main

halgari:
I'm pretty sure most of these fixes need backporting and/or we want these in 0.11.1, will verify after the PR builds


Al12rs:
This is missing 
- #3193

Al12rs (approved):

PR #3215, opened 2025-05-20
Added: Accurate protocol registration compliant with the Freedesktop .desktop Entry Spec 1.5.

Sewer56:
This implements accurate protocol registration that is compliant with the [Freedesktop .desktop Entry Spec 1.5](<https://specifications.freedesktop.org/desktop-entry-spec/latest/index.html#introduction>).

This has been tested on several machines around the office:

- SteamOS + KDE
- Unknown Distro+ KDE
- PopOS + GNOME
- Mint + Cinnamon

I'm at the office and don't have Linux on my laptop currently, so I haven't tested on more DEs; instead I went around and tested on different people's machines.

On each machine, I've tried the following (minimum) for file paths:

- `\` in folder name
- ` ` in folder name
- `$` in folder name

This was derived from the `escape characters` in [`exec variables`](https://specifications.freedesktop.org/desktop-entry-spec/latest-single/#exec-variables) and [`string types`](https://specifications.freedesktop.org/desktop-entry-spec/latest-single/#value-types) rules.

Namely...

-----

Characters in both sets:

- Backslash (`\`) - Both methods escape it with \\

Characters unique to [`exec variables`](https://specifications.freedesktop.org/desktop-entry-spec/latest-single/#exec-variables):

- Double quotes `"`
- Backticks ```
- Dollar signs `$`

Characters unique to [`string types`](https://specifications.freedesktop.org/desktop-entry-spec/latest-single/#value-types) :

- Spaces (` `)
- Newlines (`\n`)
- Tabs (`\t`)
- Carriage returns (`\r`)

The created paths are consistent with the spec, including edge cases listed in the docs.

For instance, although unintuitive, `\` in the `Exec` field is represented as `\\\\` (both sets of rules escape `\` by prepending a slash), and `$` is represented as `\\$`.

----------

The other thing to note is that the `Exec` and `TryExec` fields have different rules.
`Exec` follows both sets of quoting rules ( [`exec variable`](https://specifications.freedesktop.org/desktop-entry-spec/latest-single/#exec-variables) and [`string types`](https://specifications.freedesktop.org/desktop-entry-spec/latest-single/#value-types)), while `TryExec` follows only [`string types`](https://specifications.freedesktop.org/desktop-entry-spec/latest-single/#value-types).

The specific quoting rules for `Exec` are applied *first* and the standard quoting rules for all `string` types are applied after for the `Exec` field. I have verified this is the correct order via trial and error on different DEs, as it is not (sufficiently) explicit enough in the spec.

The exact details, quotes, notes etc. are all in the source.

----------

Other Notes:

- Icon does not point to right location/file path yet. I handed the Deck I was developing on back to QA for the time being; so will fix in separate PR later.

Anyway, hopefully this is the end of our Protocol Handler problemos.


Sewer56:
https://github.com/Nexus-Mods/NexusMods.App/actions/runs/15141873612/artifacts/3161365244

src/NexusMods.CrossPlatform/ProtocolRegistration/ProtocolRegistrationLinux.cs
```diff
@@ -125,15 +131,101 @@ private async Task UpdateMIMECacheDatabase(AbsolutePath applicationsDirectory, C
...
+    // Note that 'Exec' is a string https://specifications.freedesktop.org/desktop-entry-spec/1.0/recognized-keys.html
+    private string EscapeDesktopFilePath(string path)
+    {
+        return path
+            .Replace(@"\", @"\\") // Escape backslashes first to avoid double escaping
+            .Replace(" ", @"\s")
+            .Replace("\n", @"\n")
+            .Replace("\t", @"\t")
+            .Replace("\r", @"\r");
+    }
```
erri120, line 230: Could you make these `internal static` and add tests?

Sewer56: I added tests for every single escape character mentioned in the relevant spec sections.
And notes regarding some others like `%` and `;` which don't apply in our case, including why they don't apply; all within spec.

erri120 (approved):

PR #3224, opened 2025-05-20
Fix the redmod deploy order

halgari:
This does a few things:

* adds `-modded` to Cyberpunk when launching (always adds it)
* makes sure we provide the proper CLI commands to `redmod.exe` when we run it
* use `-modlist` instead of `--modlist` when calling redmod
* use `-force` so it always redeploys
* update the redmod `.bat` file so we can pass in arguments to the script 

NOTE: this also removes the `BulkCache` call from the app during the synchronizer runs. This was an optimization from the past that was breaking some things in the app. It's a long story, but with the recent bloom filter changes to MnemonicDB the bulk cache function would sometimes load partial entities, resulting in the `Missing attribute` errors we've been seeing. The loading features of MnemonicDB are so much faster now (and will get faster in the future) that we no longer need this functionality.

### DIFF
--- a/src/NexusMods.Abstractions.Games/RunGameTool.cs
+++ b/src/NexusMods.Abstractions.Games/RunGameTool.cs
@@ -103,7 +103,7 @@ public class RunGameTool<T> : IRunGameTool
         }
         else
         {
-            _ = await RunCommand(cancellationToken, program);
+            _ = await RunCommand(cancellationToken, commandLineArgs, program);
         }
 
         // Check if the process has spawned any new processes that we need to wait for (e.g. Launcher -> Game)
@@ -127,9 +127,10 @@ public class RunGameTool<T> : IRunGameTool
         
     }
 
-    private async Task<CommandResult> RunCommand(CancellationToken cancellationToken, AbsolutePath program)
+    private async Task<CommandResult> RunCommand(CancellationToken cancellationToken, string[] arguments, AbsolutePath program)
     {
         var command = new Command(program.ToString())
+            .WithArguments(arguments)
             .WithWorkingDirectory(program.Parent.ToString());
 
         var result = await _processFactory.ExecuteAsync(command, cancellationToken: cancellationToken);
--- a/src/NexusMods.Abstractions.Loadouts/Models/DiskStateExtensions.cs
+++ b/src/NexusMods.Abstractions.Loadouts/Models/DiskStateExtensions.cs
@@ -45,8 +45,6 @@ public static class DiskStateExtensions
         // Get the attributes for the entries in the disk state
 
         var ret = DiskStateEntry.FindByGame(asOfDb, metadata.Id);
-        //asOfDb.BulkCache(ret.EntityIds);
-        
         return ret;
     }
     
--- a/src/NexusMods.Abstractions.Loadouts/Models/DiskStateExtensions.cs
+++ b/src/NexusMods.Abstractions.Loadouts/Models/DiskStateExtensions.cs
@@ -45,7 +45,7 @@ public static class DiskStateExtensions
         // Get the attributes for the entries in the disk state
 
         var ret = DiskStateEntry.FindByGame(asOfDb, metadata.Id);
-        asOfDb.BulkCache(ret.EntityIds);
+        //asOfDb.BulkCache(ret.EntityIds);
         
         return ret;
     }
--- a/src/NexusMods.Games.RedEngine/RedModDeployTool.cs
+++ b/src/NexusMods.Games.RedEngine/RedModDeployTool.cs
@@ -48,9 +48,9 @@ public class RedModDeployTool : ITool
         if (!deployFolder.DirectoryExists())
             deployFolder.CreateDirectory();
 
-        // Note (halgari): When we change the redmod order, we need to delete the cache file so that redmod will realize it needs
-        // to redeploy. This is due to a bug in redmod. If that bug ever gets fixed (does an order check) then  
-
+        // Note (halgari): When we change the redmod order, we need to force redmod to re-run. This is due to a bug in redmod
+        // not recognizing order as a change in the mod configuration. If that bug ever gets fixed (does an order check) then  
+        // we can resolve this by removing the -force flag.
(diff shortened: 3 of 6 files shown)


src/NexusMods.Games.RedEngine/RedModDeployTool.cs
```diff
@@ -48,11 +48,14 @@ public async Task Execute(Loadout.ReadOnly loadout, CancellationToken cancellati
         if (!deployFolder.DirectoryExists())
             deployFolder.CreateDirectory();
 
+        // Note (halgari): When we change the redmod order, we need to delete the cache file so that redmod will realize it needs
+        // to redeploy. This is due to a bug in redmod. If that bug ever gets fixed (does an order check) then
```
Al12rs, line 52: Comment is out of date

src/NexusMods.Abstractions.Loadouts/Models/DiskStateExtensions.cs
```diff
@@ -45,7 +45,7 @@ public static GameInstallMetadata.ReadOnly GetMetadata(this GameInstallation ins
         // Get the attributes for the entries in the disk state
 
         var ret = DiskStateEntry.FindByGame(asOfDb, metadata.Id);
-        asOfDb.BulkCache(ret.EntityIds);
+        //asOfDb.BulkCache(ret.EntityIds);
```
Al12rs, line 48: Should be removed or a comment be added to explain why it is disabled

src/NexusMods.Abstractions.Games/RunGameTool.cs
```diff
@@ -127,9 +127,10 @@ public virtual async Task Execute(Loadout.ReadOnly loadout, CancellationToken ca
         
     }
 
-    private async Task<CommandResult> RunCommand(CancellationToken cancellationToken, AbsolutePath program)
+    private async Task<CommandResult> RunCommand(CancellationToken cancellationToken, string[] arguments, AbsolutePath program)
```
Al12rs, line 130: We should also update `RunWithShell()` to accept command line parameters, similar to `RunCommand()`

Al12rs (changes requested):
Some cleanup and we should update the `RunGameTool.RunWithShell()` code path to correctly use parameters as well

src/NexusMods.Games.RedEngine/RedModDeployTool.cs
```diff
@@ -61,7 +64,9 @@ public async Task Execute(Loadout.ReadOnly loadout, CancellationToken cancellati
             if (loadout.InstallationInstance.LocatorResultMetadata is SteamLocatorResultMetadata)
             {
                 await using var batchPath = await ExtractTemporaryDeployScript();
-                await _toolRunner.ExecuteAsync(loadout, Cli.Wrap(batchPath.ToString()), true, cancellationToken);
+                var command = Cli.Wrap(batchPath.ToString())
+                    .WithArguments(["deploy", "-force", "-modlist=" + loadorderFile.Path], true);
```
erri120, line 68: We need to update the path since the script runs inside the WINE prefix. Currently, the command looks something like this:

```
protontricks-launch --appid 1091500 "/home/USER/.local/state/NexusMods.App/Temp/6ed5b0ed-64a8-4009-b323-035226c17336.bat" deploy -force -modlist=/home/USER/.local/state/NexusMods.App/Temp/9efcdabe-e29f-4b94-bdf0-b6c7e53a9ecf.tmp
```

This part is the issue:

```
-modlist=/home/USER/.local/state/NexusMods.App/Temp/9efcdabe-e29f-4b94-bdf0-b6c7e53a9ecf.tmp
```

redmod will run inside the WINE prefix of the game so it won't be able to resolve this Linux path.

We can do two things:
1) Use the WINE mapped path `Z:`
2) Put the file in the WINE prefix

With the first approach we can make use of the fact that WINE maps `/` to `Z:` by default, so `/home/USER/foo/bar/baz` will be accessible inside the WINE prefix at `Z:\home\USER\foo\bar\baz`. This is default behavior so it might not be super reliable if the behavior changes or the user changes the prefix settings.

The second approach is more reliable, we can put the modlist file inside the prefix directly. You can get the wine directory from the `SteamLocatorResultMetadata` like this:

```csharp
if (steamLocatorResultMetadata.LinuxCompatibilityDataProvider is null) return;
var winePrefixDirectory = steamLocatorResultMetadata.LinuxCompatibilityDataProvider.WinePrefixDirectoryPath;
var cDriveDirectory = winePrefixDirectory.Combine("drive_c");
```

then we can just put the file somewhere inside the `C:` directory and call redmod with `-modlist="C:\\something`.

erri120 (changes requested):

src/NexusMods.Games.RedEngine/RedModDeployTool.cs
erri120:
```csharp
if (loadout.InstallationInstance.LocatorResultMetadata is SteamLocatorResultMetadata steamLocatorResultMetadata)
{
    if (steamLocatorResultMetadata.LinuxCompatibilityDataProvider is null) return;
    var wineDirectory = steamLocatorResultMetadata.LinuxCompatibilityDataProvider.WinePrefixDirectoryPath;
    var cDriveDirectory = wineDirectory.Combine("drive_c");

    if (!cDriveDirectory.DirectoryExists()) cDriveDirectory.CreateDirectory();
    var tmpFile = cDriveDirectory.Combine("modlist.txt");
    await using (var input = loadorderFile.Path.Read())
    await using (var output = tmpFile.Open(FileMode.Create, FileAccess.ReadWrite, FileShare.None))
    {
        await input.CopyToAsync(output, cancellationToken: cancellationToken);
    }

    await using var batchPath = await ExtractTemporaryDeployScript();
    var command = Cli.Wrap(batchPath.ToString()).WithArguments(["deploy", "-force", "-modlist=C:\\modlist.txt"], escape: true);
    await _toolRunner.ExecuteAsync(loadout, command, true, cancellationToken);
}
```

Al12rs (approved):

Issue #2271, opened 2024-11-20
Epic: Load Order


Al12rs:
# Epic

## Outline
Add Load Order management support to the app

PR #3269, opened 2025-05-27
Add Load Order Management docs

Al12rs:
- Part of #2271 

This adds a document that aims to cover all the potentially relevant points when dealing with load orders. 

This is likely not complete, we should maintain this by adding anything relevant.

The purpose of the doc is to be able to reference it when making decisions about UX or code design.

A way to check if a particular approach is compatible with other elements of load order management.


erri120 (approved):

Al12rs:
Updated

PR #3277, opened 2025-05-27
Added: Ability to ignore individual files in update results in the backend.

Sewer56:
Notes:

- Filters are optional, so you can choose to view a non-filtered view of the updates where necessary in the UI.
- I added an observer to `ModUpdateService` that will automatically refresh all filtered views (observables) when a filter is added.
- Filters currently are limited to disabling single files in update results, regardless of cause.
    - I discussed with @captainsandypants last week at the office, topics like 'should we specifically ignore `ver a -> ver b`' update; we settled on ignoring specific files for the time being.
- Unit tests included, and technically, an integration test.


Al12rs:
Tests are failing, @Sewer56 can you check what is up there? Might need a rebase onto latest main perhaps

Sewer56:
Merged main and updated schema, all good.

Al12rs:
Could you include QA steps? Or should that wait for future PRs?

Al12rs (approved):

PR #3288, opened 2025-05-29
Fixes for the loadout group page

halgari:
The recent updates to the LoadoutGroup page didn't take deleted files into account. This change fixes this issue by changing `LoadoutFile` to `LoadoutItemWithTargetedPath` and conditionally loading `Size`. It also updates the `Pending Changes` page to show `WarnOfUnableToExtract`, so we actually see deleted files even if we don't have a source for them. 

## To test:

* Create a game loadout
* Add a mod
* Go to the game folder and delete a game file
* Apply changes
* Verify that there is a `0` sized entry in the `External Changes` pane
* Delete the `0` sized file
* Verify that a warning appears in Health Check about a file not being found for the base game
* Verify that the file deleted shows up as a green entry in the `Pending Changes` dialog

We can improve this UX in the future, but at least for now we can see these issuse instead of the problems being invisible.


Al12rs:
Added the StrikeThrough styling to deleted entries as well. 

Only for file entries, not for parent folders that only have deleted items as children for example, but I don't think that is super important.

Issue #3147, opened 2025-05-05
C# project cleanup


erri120:
We currently have 121 C# projects in our solution, 85 projects in `src`, 35 projects in `tests` and one benchmark project. Our abstractions projects especially are a total mess for no good reason. I personally find it very hard to justify having this many projects around, they only make it harder to add new features since our dependency graph is very convoluted and it's very easy to create circular dependencies. We've wanted a UI project with just controls but creating one would be a massive undertaking due to how the dependency graph is set up.

Instead of making things worse and even more confusing, I propose reducing the amount of projects we have to the bare minimum:

```mermaid
flowchart TD
    NexusMods.Sdk
    NexusMods.UI.Sdk
    NexusMods.UI.Sdk --> NexusMods.Sdk
    NexusMods.Cli --> NexusMods.Sdk
    NexusMods.UI --> NexusMods.UI.Sdk
    NexusMods.Themes.FluentDark --> NexusMods.UI
    NexusMods.Games.Foo --> NexusMods.UI.Sdk
    NexusMods.Games.Bar --> NexusMods.UI.Sdk
    NexusMods.DataModel --> NexusMods.Sdk
    NexusMods.App --> NexusMods.DataModel
    NexusMods.App --> NexusMods.Cli
    NexusMods.App --> NexusMods.Themes.FluentDark
    NexusMods.App --> NexusMods.Games.Foo
    NexusMods.App --> NexusMods.Games.Bar
```

- `NexusMods.App`: executable project
- `NexusMods.UI`: contains the main window and all pages
- `NexusMods.UI.Sdk`: contains all reusable controls and our UI abstractions
- `NexusMods.Themes.FluentDark`: nothing changes here, still the same theme project
- `NexusMods.Sdk`: contains **all** of our abstractions and models
- `NexusMods.Games.*`: games are still separate projects
- `NexusMods.Cli`: contains **all** of our CLI commands
- `NexusMods.DataModel`: contains **all** implementations (could find a better name for this project`

As an example, the `ILibraryService` interface would be in `NexusMods.Sdk`, all the library database models would also be `NexusMods.Sdk` and the implementation of `ILibraryService` would be in `NexusMods.DataModel`.

This architecture is very simple and makes it very easy to add new features or do any refactors without dealing with dependency issues.

halgari:
[...]

PR #3290, opened 2025-05-30
Epic game index

halgari:
Adds support for indexing Epic games. This code is a bit different from the other two iterations in that it doesn't do network indexing. Instead it is broken into two phases:

* Metadata - we query `egdata.app` and pull the JSON data to give us the build information and the Sha1 hashes
* Hashing - we hash the files on-disk to generate the minimal hashes and hash relations. 

The outcome of this approach is that we can pull build data for any build that `egdata.app` has, and we do that during indexing, so we're not scraping their site or making a call per user. It's just a call when we run the CLI command to grab and process the data. When a new version releases we have to download it somewhere using EGS, index it, and then grab the latest builds. 

This PR also puts the new EGS code into the `.Backend` project, a future PR will move the other store libraries into that project as well.

Note: There aren't any functional changes with this PR for users. This is only the indexing + hash db building so we can a new DB out. A future PR will wire up the user facing code.

### DIFF
--- a/src/NexusMods.Backend/NexusMods.Backend.csproj
+++ b/src/NexusMods.Backend/NexusMods.Backend.csproj
@@ -1,6 +1,4 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-    <!-- NuGet Package Shared Details -->
-    <Import Project="$([MSBuild]::GetPathOfFileAbove('NuGet.Build.props', '$(MSBuildThisFileDirectory)../'))" />
     <ItemGroup>
         <ProjectReference Include="..\NexusMods.Abstractions.GameLocators\NexusMods.Abstractions.GameLocators.csproj" />
         <ProjectReference Include="..\NexusMods.Abstractions.Games\NexusMods.Abstractions.Games.csproj" />


erri120 (changes requested):
https://github.com/Nexus-Mods/NexusMods.App/blob/e2b48a2a35abc412d5658d74f88d0549be5ae728/src/NexusMods.Games.FileHashes/NexusMods.Games.FileHashes.csproj#L19

With the new project structure from #3147 this reference is not allowed. The `NexusMods.Backend` project will only contain implementations, the interfaces and any shared types would go into `NexusMods.Sdk`. The only project that is "allowed" to reference `NexusMods.Backend` is `NexusMods.App` for DI registration.

https://github.com/Nexus-Mods/NexusMods.App/blob/e2b48a2a35abc412d5658d74f88d0549be5ae728/src/NexusMods.Backend/NexusMods.Backend.csproj#L2-L3

The `NexusMods.Backend` is also not allowed to be published to NuGet since it'll only contain implementations.

I recommend not putting stuff into `NexusMods.Backend` just yet, I'm still in the process of merging projects into `NexusMods.Sdk`, once the shared types are in place I'll begin with merging implementations into `NexusMods.Backend`.

For this PR you can make `NexusMods.Networking.Epic` or similar and put everything in there for now.

src/NexusMods.Backend/NexusMods.Backend.csproj
```diff
@@ -1,5 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+    <!-- NuGet Package Shared Details -->
+    <Import Project="$([MSBuild]::GetPathOfFileAbove('NuGet.Build.props', '$(MSBuildThisFileDirectory)../'))" />
```
erri120, line 3:
```suggestion
```

src/NexusMods.Games.FileHashes/NexusMods.Games.FileHashes.csproj
```diff
@@ -16,6 +16,8 @@
       <ProjectReference Include="..\NexusMods.Abstractions.Games\NexusMods.Abstractions.Games.csproj" />
       <ProjectReference Include="..\NexusMods.Abstractions.Settings\NexusMods.Abstractions.Settings.csproj" />
       <ProjectReference Include="..\NexusMods.App.Generators.Diagnostics\NexusMods.App.Generators.Diagnostics.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false"  />
+      <ProjectReference Include="..\NexusMods.Backend\NexusMods.Backend.csproj" />
+      <ProjectReference Include="..\NexusMods.Networking.EpicGameStore\NexusMods.Networking.EpicGameStore.csproj" />
```
erri120, line 20:
```suggestion
      <ProjectReference Include="..\NexusMods.Networking.EpicGameStore\NexusMods.Networking.EpicGameStore.csproj" />
```

Issue #3293, opened 2025-05-30
xdg-utils is non-conformant to the .desktop entry specification, leading to possible login failures


Sewer56:
# Bug Report

This is an upstream bug, I reported it to:
- https://gitlab.freedesktop.org/xdg/xdg-utils/-/issues/279

So you can find the details.
In the meantime, I'll make a workaround.

PR #3294, opened 2025-05-30
Added: Workaround around xdg-utils 'generic' fallback in protocol handlers

Sewer56:
Workaround for 
-  #3293

Need urgent merge before next release.


Al12rs (commented):
I can't really test this, from reading the comments and looking at the code it appears ok.
I'm assuming you already tested this on your end

erri120 (approved):

PR #3309, opened 2025-06-03
Add collection ID to DB

### LINKED ISSUE #3304: Store collection IDs in DB
GraphQL returns `id` values that we're currently not saving in the DB. We're using the collection slug as identifiers instead of the actual id.

erri120:
Resolves #3304.

Similar to `RevisionId`, `CollectionId` is a globally unique identifier used by some GraphQL endpoints. This PR adds the attribute as a required unique attribute to `CollectionMetadata`.

This PR adds migration 8, which is a migration that does a GraphQL API call to get the ID of the collection based on the slug.


halgari:
Makes sense, if this doesn't require a login I'm okay with this

Al12rs:
There is a conflict on Schema.verified.md

PR #3315, opened 2025-06-03
Added: Kebab 🍢 Menu to Library Items

Sewer56:
Adds a three dot (kebab) menu to the Library:

image

## Mods from Nexus

For mods from Nexus, the following are available:

- [x] View Changelog: Takes the user to mod page (see notes in code)
- [x] Open Mod Page: Takes the user to mod page

This should work on both mod page rows and mod rows.

## External Mods

For local mods where metadata is not available, items are disabled:

image

Needs a styling pass from @insomnious before review.

## Other Info

All new strings are localizable.


src/NexusMods.App.UI/Pages/LibraryPage/TreeDataGridLibraryResources.axaml
```diff
@@ -136,21 +137,81 @@
...
+           
+            <!-- more options -->
+            <controls:StandardButton
+                ToolTip.Tip="More options"
+                x:Name="HamburgerMenuButton"
+                LeftIcon="{x:Static icons:IconValues.MoreVertical}"
+                ShowIcon="Left"
+                ShowLabel="False"
+                Fill="None">
+                <controls:StandardButton.Flyout>
```
Al12rs, line 162: This can be empty if dealing with a local file mod (not from nexus).
What is the expected behaviour of the button in that case?

Al12rs (approved):

PR #3328, opened 2025-06-04
Configurable file paths

halgari:
Add UI support for changing the archives path. This PR does a few things

* Updates `ConfigurablePath` so that the `KnownFolder` is optional. *This was the cause of the config file changes not taking effect*, there was no way to point to an absolute path. Now if `ConfigurablePath` has a `KnownPath` of null, the path is considered an absolute path
* Adds UI elements and settings builders for a list of configurable paths
* Wires up the UI for these elements in a very simple add/remove operation


captainsandypants:
> If you move the location when you already have downloads, does it move the files that are already there?

No, they'll need to move them manually for now.

captainsandypants:
Design:

[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=3716-8577&t=Z3IZs1fyoCRVuihN-1)

image

insomnious:
`src\NexusMods.App.UI\Controls\Settings\SettingEntries\SettingEntryViewModel.cs:45` main part that we know isn't quite right

erri120:
Updated the PR with fixes for the path selector to prevent exceptions from an invalid thread.

emwrn:
Works fine on Ubuntu 25.04 ✅

emwrn:
v0.13.4.0 works fine on Ubuntu 25.04 ✅

Issue #3118, opened 2025-04-29
Phase 1 - Collection Create


LukeNexusMods:
- [x] #3304 (#3309)
- [x] Create a new collection from the left menu (#3302)
- [x] Library: drop-down to select the collection the item installs to (#3320)
- [x] Local collections: one-click upload draft (#3373)

### Left menu - Create collection
[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=3562-101811&t=6ImPjmnFqIFVrVMU-1)

![Image](https://github.com/user-attachments/assets/56d417a7-26a5-4a6d-8cad-da685f1ed5d0)

### Library - Collection tab - Create collection
[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=3562-101811&t=6ImPjmnFqIFVrVMU-1)

![Image](https://github.com/user-attachments/assets/207b446b-9541-40bc-9078-ef55b5f7c8be)

### Library - Select collection install location
[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=3354-40656&t=6ImPjmnFqIFVrVMU-1)

![Image](https://github.com/user-attachments/assets/418ccd4d-c267-4c6d-b8d9-b5d0011c60f2)

### Clone collection
[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=3607-25720&t=6ImPjmnFqIFVrVMU-1)

![Image](https://github.com/user-attachments/assets/65a43d62-40be-4198-8a24-b7141c231230)

erri120:
@captainsandypants what's the name generation logic for new collections that the user can create?

erri120:
> > [@captainsandypants](https://github.com/captainsandypants) what's the name generation logic for new collections that the user can create?
> 
> In terms of what name to give it by default if they don't rename it?

The default name is `My new collection` so if you create two new collections you have `My new collection` and `My new collection`.

captainsandypants:
> > > [@captainsandypants](https://github.com/captainsandypants) what's the name generation logic for new collections that the user can create?
> > 
> > 
> > In terms of what name to give it by default if they don't rename it?
> 
[...]

PR #3373, opened 2025-06-12
Upload collection revisions

erri120:
Part of #3118.

Adds a simple upload button for now that creates the collection and uploads the revision. If the collection already exists, the revision gets updated instead. Also opens Nexus Mods after the collection has been uploaded, note that if the collection has just been created it takes a few seconds to show up on nexus mods.

### DIFF
--- a/src/NexusMods.Abstractions.NexusModsLibrary/Models/CollectionRevisionMetadata.cs
+++ b/src/NexusMods.Abstractions.NexusModsLibrary/Models/CollectionRevisionMetadata.cs
@@ -53,7 +53,7 @@ public partial class CollectionRevisionMetadata : IModelDefinition
     public static readonly UInt64Attribute TotalRatings = new(Namespace, nameof(TotalRatings)) { IsOptional = true };
 
     /// <summary>
-    /// Status.
+    /// Publication status.
     /// </summary>
     public static readonly EnumAttribute<RevisionStatus> Status = new(Namespace, nameof(Status)) { IsOptional = true };
 }
--- a/src/NexusMods.Abstractions.Collections/ManagedCollectionLoadoutGroup.cs
+++ b/src/NexusMods.Abstractions.Collections/ManagedCollectionLoadoutGroup.cs
@@ -5,6 +5,9 @@ using NexusMods.MnemonicDB.Abstractions.Models;
 
 namespace NexusMods.Abstractions.Collections;
 
+/// <summary>
+/// Represents an uploaded collection managed by the app.
+/// </summary>
 [Include<CollectionGroup>]
 public partial class ManagedCollectionLoadoutGroup : IModelDefinition
 {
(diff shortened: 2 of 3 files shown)


src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutView.axaml
```diff
@@ -28,11 +28,15 @@
                                    Title="All"
                                    Description="A combined view of all your installed mods from all collections." />
         </Panel>
+
+        <StackPanel Grid.Row="1">
+            <controls:StandardButton x:Name="ButtonUploadCollectionRevision" Text="Upload" />
```
Al12rs, line 33: This should probably be hidden behind a feature flag. Either only available in debug mode or only available behind a setting

src/NexusMods.Abstractions.NexusModsLibrary/Models/CollectionRevisionMetadata.cs
```diff
@@ -51,4 +51,9 @@ public partial class CollectionRevisionMetadata : IModelDefinition
     /// The total number of ratings this revision has.
     /// </summary>
     public static readonly UInt64Attribute TotalRatings = new(Namespace, nameof(TotalRatings)) { IsOptional = true };
+
+    /// <summary>
+    /// Status.
```
Al12rs, line 56:
```suggestion
    /// Revision published status.
```

Al12rs: Since the attribute is optional, please include what default (no value) means

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutView.axaml
Al12rs: This is done in the codebehind 👍

src/NexusMods.Collections/CollectionCreator.cs
```diff
@@ -59,6 +76,92 @@ private static string GenerateNewCollectionName(string[] allNames)
...
 
+    public static async ValueTask<CollectionRevisionMetadata.ReadOnly> UploadCollectionRevision(IServiceProvider serviceProvider, LoadoutItemGroupId groupId, CancellationToken cancellationToken)
+    {
+        var connection = serviceProvider.GetRequiredService<IConnection>();
+        var loginManager = serviceProvider.GetRequiredService<ILoginManager>();
+        var libraryService = serviceProvider.GetRequiredService<ILibraryService>();
+        var nexusModsLibrary = serviceProvider.GetRequiredService<NexusModsLibrary>();
+        var temporaryFileManager = serviceProvider.GetRequiredService<TemporaryFileManager>();
+        var jsonSerializerOptions = serviceProvider.GetRequiredService<JsonSerializerOptions>();
+        var mappingCache = serviceProvider.GetRequiredService<IGameDomainToGameIdMappingCache>();
```
Al12rs, line 87: Should the creator just be a singleton and these services be readonly state? I think this would make sense if these services needed to be transient, but it seems to me they are all singletons.

PR #3377, opened 2025-06-12
Added dialog to uninstall mod

insomnious:
First implementation test of the new dialog system

Would like Linux users (@Sewer56 @erri120) to find 5 mins just to run this and see what happens regarding sizing and positioning of the modal.

Screenshot 2025-06-12 142313

My Mods -> select a mod -> click uninstall.

Other dialogs and tests can be found in Debug Controls -> Dialogs

image


src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutViewModel.cs
```diff
@@ -199,6 +201,31 @@ public LoadoutViewModel(
                         .ToArray();
 
                     if (ids.Length == 0) return;
+
+                    var dialog = DialogFactory.CreateMessageBox(
```
Sewer56, line 205: This should be localized. 
@Al12rs @insomnious 

There's a lot of PRs that make it in with hardcoded English; I find myself fixing those down the road.

PR #3405, opened 2025-06-17
Fix duplicate collection downloads on upload

erri120:
When uploading a collection revision, we can either create a new revision or update an existing revision. The latter causes issues in our data model because we previously assumed that revisions are mutable.

This "fix" is more of a hack and will remove all the previous downloads before adding the new ones. It will throw an exception if any of the previous download entities are referenced by anything.

Part of #3118.


Al12rs (approved):
I'd like to talk about this a bit after standup, for now it's fine

PR #3406, opened 2025-06-17
Fix R3 error logging

erri120:
We were actually never logging unhandled exceptions in R3 observables because we set the handler and then it gets overwritten by `UseR3()`.


Al12rs:
You can go bump this PR to fix it on their end: 
https://github.com/Cysharp/R3/pull/329

PR #3411, opened 2025-06-17
Added: Ability to Hide/Show Updates in the Library View`

Sewer56:
This has 1/2 visual edge cases; but they are *subjective*; around counts/numbers. One person may argue it's correct, another may argue it's not. They're for things that have not been defined in design. We can refine those later.

Also disables the `view mod page/changelog` buttons for local mods.


emwrn:
Looks good on Ubuntu 25.04 ✅

PR #3412, opened 2025-06-17
Update com.nexusmods.app.desktop

god-bypass:
for some reason this line makes the .desktop file not valid and it doesnt update it

### DIFF
--- a/src/NexusMods.App/com.nexusmods.app.desktop
+++ b/src/NexusMods.App/com.nexusmods.app.desktop
@@ -11,6 +11,7 @@ StartupWMClass=NexusMods.App
 MimeType=x-scheme-handler/nxm
 Icon=com.nexusmods.app
 Exec=${INSTALL_EXEC} %u
+TryExec=${INSTALL_TRYEXEC}
 X-AppImage-Integrate=true
 X-AppImage-Name=NexusMods.App
 X-AppImage-Arch=x86_64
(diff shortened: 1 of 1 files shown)


god-bypass:
my .desktop was originally this
```[Desktop Entry]
Type=Application
# Version=1.5
# Commented out because PupNet needs updating to recognise 1.5, so letting it default to latest is ideal.
Name=Nexus Mods App
GenericName=Mod Manager
Comment=Mod Manager for your games
Categories=Game
Terminal=false
StartupWMClass=NexusMods.App
MimeType=x-scheme-handler/nxm
Icon=com.nexusmods.app
Exec=/opt/nexusmods-app/NexusMods.App %u
TryExec=${INSTALL_TRYEXEC}
X-AppImage-Integrate=true
X-AppImage-Name=NexusMods.App
X-AppImage-Arch=x86_64 i still cant see```

tho i tried the one from the source it also didn't work without the line removal

**Comment by Sewer56** [comment]
This is probably related to:
- https://github.com/Nexus-Mods/NexusMods.App/issues/3385#issue-3144974172

```
public async Task RegisterHandler(string uriScheme, bool setAsDefaultHandler = true, CancellationToken cancellationToken = default)
```

In this method we're not properly checking the build types, and skipping some steps when we shouldn't be on certain build types.
I'm guessing this method is not ran as a result:

```
private async Task CreateDesktopFile(AbsolutePath applicationsDirectory, CancellationToken cancellationToken = default)
```

Sewer56:
I'd fix that right away normally, but my week is stupid busy.

god-bypass:
yea i might understand more that it doesnt rename which it should

PR #3416, opened 2025-06-18
Fix exception when writing empty RedMod loadorder file

Al12rs:
Replaces Aggregate with string.Join to handle empty collections gracefully.

Previously, calling Aggregate on an empty collection would throw an exception, causing the Synchronizer to fail when no redmods were enabled.

I wasn't able to test this, so if someone wants to check please do.

## Steps for testing:
- Ensure you have RedMod DLC installed
- Install a redmod mod but disable it (you should have no other enabled remods)
- Apply
- No exceptions should happen


erri120 (approved):

chnexusmods:
Windows 11

PR #3446, opened 2025-06-23
Add user info to the DB after logged in

erri120:


src/NexusMods.Networking.NexusWebApi/LoginManager.cs
```diff
@@ -94,9 +97,37 @@ public LoginManager(
...
+            entityId = existingDatom.E;
+        }
+        else
+        {
+            entityId = tx.TempId();
+            tx.Add(entityId, UserInDb.NexusId, userInfo.UserId.Value);
+        }
+
+        tx.Add(entityId, UserInDb.Name, userInfo.Name);
+        tx.Add(entityId, UserInDb.NexusId, userInfo.UserId.Value);
```
Al12rs, line 123: `UserInDb.NexusId` added twice in the else case

Al12rs (changes requested):
Small thing above

Al12rs (approved):

PR #3452, opened 2025-06-24
Epic support phase2

halgari:
Finish up support for Epic games. 

* Switches Epic LocatorIds over to ManifestHash which is the official "build id" for a given build
* Saves ManfiestIds into the Game files hash DB

Testing this is a bit odd at the moment, but it shouldn't affect existing code. To test this (before it hits main) you'd need to export the file hashes DB and copy it into the app's AppData folder.


src/NexusMods.Abstractions.Games.FileHashes/Models/VersionDefinition.cs
```diff
@@ -38,6 +39,11 @@ public partial class VersionDefinition : IModelDefinition
     /// </summary>
     public static readonly StringsAttribute Steam = new(Namespace, nameof(Steam)) { IsIndexed = true };
     
+    /// <summary>
+    /// The associated Steam ManifestIDs for this version
```
Al12rs, line 43: I think it's supposed to be
```suggestion
    /// The associated Epic ManifestIDs for this version
```

Al12rs (approved):
A couple of docs issues note above.

It would be good to have documentation or links to documentation on what we use to track each game store and a description of the hierarchy of manifests and ids that they use.

Also basic overview of how the app gest the information, files on disk and/or APIs used.

Issue #3493, opened 2025-06-30
Add collection and revision status for created collections


erri120:
Part of collection creation.

```mermaid
---
config:
    gitGraph:
        mainBranchName: "App"
---

gitGraph
    commit id: "Create Collection (local)"
    commit id: "Install mod A"
    commit id: "Install mod B"
    branch "NexusMods"
    commit id: "fake commit"
    branch "NexusMods_Revision"
    switch "NexusMods"
    merge "App" id: "Upload Collection"
    switch "NexusMods_Revision"
    merge "App" id: "Upload draft Revision 1"
    switch "App"
    commit id: "Install mod C"
    switch "NexusMods_Revision"
    merge "App" id: "Update draft Revision 1 (a)"
    switch "App"
    commit id: "Remove mod A"
    switch "NexusMods_Revision"
    merge "App" id: "Update draft Revision 1 (b)"
    switch "NexusMods"
    merge "NexusMods_Revision" id: "Publish Revision 1"
    switch "App"
    commit id: "Install mod D"
    switch "NexusMods_Revision"
    merge "App" id: "Upload draft Revision 2"
    switch "NexusMods"
    merge "NexusMods_Revision" id: "Publish Revision 2"
```

erri120:
Done with #3509.

PR #3509, opened 2025-07-02
Fetch collection status on page load and use R3 bindings

erri120:
Requires #3502, part of #3493.

Cleans up `ILoadoutViewModel` to move to R3 bindings instead of ReactiveUI.

Fetches the current collection status from Nexus Mods on page loads and exposes the data in the VM. UI updates to follow soon :tm:.

### DIFF
--- a/src/NexusMods.Networking.NexusWebApi/NexusModsLibrary.Collections.cs
+++ b/src/NexusMods.Networking.NexusWebApi/NexusModsLibrary.Collections.cs
@@ -80,7 +80,7 @@ public partial class NexusModsLibrary
     /// <summary>
     /// Uploads a collection revision, either creating a new revision or updating an existing draft revision.
     /// </summary>
-    public async ValueTask UploadDraftRevision(
+    public async ValueTask<RevisionNumber> UploadDraftRevision(
         CollectionMetadata.ReadOnly collection,
         IStreamFactory archiveStreamFactory,
         CollectionRoot collectionManifest,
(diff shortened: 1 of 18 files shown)


src/NexusMods.Networking.NexusWebApi/NexusModsLibrary.Collections.cs
```diff
@@ -262,28 +264,22 @@ static UpdatePolicy ConvertUpdatePolicy(Abstractions.Collections.Json.UpdatePoli
...
-        return newer;
+        var revisionNumbers = await GetAllRevisionNumbers(collection, cancellationToken);
+        return revisionNumbers.FirstOrOptional(static tuple => tuple.Status == RevisionStatus.Published).Convert(static tuple => tuple.Number);
     }
 
     /// <summary>
     /// Gets all revision numbers for the given collection in descending order.
     /// </summary>
-    public async ValueTask<RevisionNumber[]> GetAllRevisionNumbers(
+    public async ValueTask<(RevisionNumber Number, RevisionStatus Status)[]> GetAllRevisionNumbers(
```
Al12rs, line 282: Users can delete collections from Nexus if they only uploaded a draft and never published the first revision.

This means that this query can fail.

erri120: I'll deal with API errors in a separate PR.

Issue #3518, opened 2025-07-03
Publish created collections


erri120:
- [ ] Dialog when creating the collection using the "Share"-button. Allows setting the initial visibility of the collection (unlisted, listed)
- [ ] Dialog after successfully publishing
- [ ] Add new operation to upload a revision and publish it immediately

## Designs

[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Library---Collections?node-id=3562-101810&t=W440upf4MnoLRwP4-1)

<img width="1176" height="876" alt="Image" src="https://github.com/user-attachments/assets/ee745bdd-57dd-4f56-b888-e0bb41c66fb5" />

<img width="1176" height="704" alt="Image" src="https://github.com/user-attachments/assets/359e591e-3a8a-4a1f-9c00-647d1496800c" />

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/174c4c71-6fa4-4db4-a2b0-40cbd961ca37" />

PR #3534, opened 2025-07-08
Update API endpoints and commands for collection creation

erri120:
Part of #3518.

### DIFF
--- a/src/NexusMods.Collections/CollectionCreator.cs
+++ b/src/NexusMods.Collections/CollectionCreator.cs
@@ -91,7 +91,7 @@ public static class CollectionCreator
         return result.Remap(group);
     }
 
-    private static async ValueTask<(IStreamFactory, CollectionRoot)> PrepareForUpload(
+    private static async ValueTask<(IStreamFactory, TemporaryPath, CollectionRoot)> PrepareForUpload(
         IServiceProvider serviceProvider,
         CollectionGroupId groupId,
         CancellationToken cancellationToken)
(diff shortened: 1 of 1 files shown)


src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutViewModel.cs
```diff
@@ -135,24 +137,28 @@ public LoadoutViewModel(
 
             RulesSectionViewModel = new SortingSelectionViewModel(serviceProvider, windowManager, loadoutId, canEditObservable: isSingleCollectionObservable);
 
-            CommandUploadRevision = new ReactiveCommand<Unit>(async (unit, cancellationToken) =>
+            CommandShareCollection = IsCollectionUploaded.Select(static b => !b).ToReactiveCommand<Unit>(async (_, cancellationToken) =>
```
Al12rs, line 140: When should these commands be enabled? Only if you already have the collection uploaded, but that seems wrong?

src/NexusMods.Networking.NexusWebApi/GraphQL/CreateCollection.graphql
```diff
@@ -58,8 +58,21 @@ mutation EditCollectionSummary($collectionId: Int!, $summary: String!) {
...
-mutation PublishRevision($revisionId: ID!, $collectionStatus: CollectionStatus!) {
-    publishRevision(revisionId: $revisionId, collectionStatus: $collectionStatus) {
+mutation PublishRevision($revisionId: ID!) {
+    publishRevision(revisionId: $revisionId) {
         success
     }
 }
+
+mutation ListCollection($collectionId: Int!) {
+    listCollection(collectionId: $collectionId) {
```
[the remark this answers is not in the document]
erri120: Not much we can do about that.

src/NexusMods.App.UI/Pages/LoadoutPage/LoadoutViewModel.cs
Al12rs: Missed the negation in the select, was confused because of that.

PR #3559, opened 2025-07-15
Abstract GraphQL API calls

erri120:
This PR moves all GraphQL calls to a new abstraction layer called `IGraphQlClient`. Every GraphQL call is wrapped with strong types for both inputs and results.


Sewer56:
Will check as requested

PR #3563, opened 2025-07-16
Enable collection creation and add events

erri120:
Enables collection creation for everyone, but collection sharing is now locked behind a setting. Also adds tracking events so we know how many people are using this feature.


Sewer56 (approved):
LGTM.

Reviewed from the laptop.

PR #3596, opened 2025-07-21
Implement group deletion

erri120:
Uses custom deletion code for the entities as `tx.Delete(recursive: true)` is too dangerous.


erri120:
> I think this is fine.
> 
> Question: Does this affect `LibraryLinkedLoadoutFile`(s)? (Has this been considered)
> 
> Technically we're removing the LoadoutFile(s) here, including their children, but the links may still hang around, risking inconsistent state. I feel like you'd know better than me here, so I haven't dug in too deep to find out myself.

`LibraryLinkedLoadoutFile` goes from the Loadout to the Library, not the other way around.

Issue #3561, opened 2025-07-15
Bundle game hashes DB on build


erri120:
Similar to how we bundle `games.json` file with #3476:

https://github.com/Nexus-Mods/NexusMods.App/blob/aa5839b8d61f1b5c325a6054bdabdf3be25cd962/src/NexusMods.Networking.NexusWebApi/NexusMods.Networking.NexusWebApi.csproj#L27-L41

PR #3597, opened 2025-07-21
Add build date

erri120:
Required for #3561.


Sewer56 (approved):
Looks good to me.
I will assume you tested on Windows (I haven't).

Technically speaking, you could use `pwsh` to unify the code inside the `.csproj` file; that would work for CI- but would create an extra dependency for regular user building.

PR #3599, opened 2025-07-21
Show delete button by default

erri120:
Shows the delete button for collections by default. Also prevents deleting the last group in a Loadout.


Sewer56 (approved):
(Not sure why the comment there posted- don't mind it)

PR #3650, opened 2025-07-28
Force kill unresponsive process

### LINKED ISSUE #3633: Force kill unresponsive previous app processes
Although we haven't identified and fixed the issue of the app not shutting down properly, we can at least force kill those unresponsive processes.

On startup, the `CLIServer` uses a`MultiProcessSharedArray` to write the process identifier and port of the main process into a sync file. If this shared array exists we can extract the process identifier of the previous main process. If the process identifier points to a still open process we can try and ping the process using our `CLIServer`. If the process refuses a connection then it's very likely that the `CLIServer` of that process has shut down already.

I believe with this approach we can run a quick test and startup to see if the a previous process is unresponsive and then force kill it. This doesn't fix the underlying issue but would at least allow users to continue using the app without issue.

erri120:
Resolves #3633.

How to test:

1) Enable multiple run instances in Rider.
2) Set a breakpoint in the `DebuggingHost.Dispose` method and start the first instance.
3) Close the first instance normally, this should hit the breakpoint in the dispose method. The first instance is now in an unresponsive state because of the debugger.
4) Set a breakpoint in the new `CleanupOldProcess` method and start the second instance.
5) Observe how the new method tries to connect to the first instance and fails. The first instance should be killed by the second instance.

### DIFF
--- a/src/NexusMods.App/Program.cs
+++ b/src/NexusMods.App/Program.cs
@@ -71,7 +71,8 @@ public class Program
         var services = host.Services;
         _logger = services.GetRequiredService<ILogger<Program>>();
 
-        CleanupOldProcesses(services).Wait(timeout: TimeSpan.FromSeconds(10));
+        // NOTE(erri120): has to come before host startup
+        CleanupUnresponsiveProcesses(services).Wait(timeout: TimeSpan.FromSeconds(10));
 
         // Okay to do wait here, as we are in the main process thread.
         host.StartAsync().Wait(timeout: TimeSpan.FromMinutes(5));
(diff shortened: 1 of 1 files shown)


src/NexusMods.App/Program.cs
```diff
@@ -148,6 +152,48 @@ public static int Main(string[] args)
         }
     }
 
+    private static async Task CleanupOldProcesses(IServiceProvider serviceProvider)
```
Al12rs, line 155: Function should be renamed to something more accurate like `CheckForExistingProcess`, since if another process is alive and responding it wouldn't get cleaned up.

```suggestion
    private static async Task CheckForExistingProcess(IServiceProvider serviceProvider)
```

erri120: `CheckForExistingProcess` is even less accurate since it doesn't indicate that the existing process will be killed, which is pretty important. I can rename to `CleanupUnresponsiveProcesses`.

Al12rs (approved):

PR #3669, opened 2025-07-30
Added: Support for Pause and Resume

### LINKED ISSUE #3604: Add pause and resume support to the job system
Pausing is just cancellation that can be recovered from. The behavior is very similar, pausing a job means requesting cancellation on the cancellation token. Any code using the cancellation token will throw an `OperationCanceledException` as usual. The difference comes in the handling of the exception, instead of transitioning the job status to `Cancelled` we check a boolean `isRequestingPause` to either transition to the `Cancelled` or `Paused` state. This boolean will be set when someone is calling the `Pause` method on the job.

Additionally, a `Resume` method is required to transition from the `Paused` state back to a `Running` state. Note that this requires resetting internal state.

Sewer56:
# Job Pause/Resume Implementation

Spent a bit thinking about the problem; including how to best handle
requested upcoming features such as downloads pausing/resuming.

I think the solution here is decent; it's a bit different than originally proposed, which was a cancel/restart.
This is more of a true pause/resume, by stalling existing jobs without hogging the threadpool.

## Summary

Adds pause/resume functionality to the job system using a cooperative model, with both standard and pre-emptive pause modes.

Jobs can be paused using existing `YieldAsync()` yield points and resumed later,
with cancellation taking precedence over pause.

## Main Points

- **`JobCancellationToken`**: New token combining cancellation + pause/resume using `AsyncManualResetEvent`
- **`YieldAsync()`**: Now handles both pause detection and waiting for resume
    - Many of our existing jobs are now pausable without requiring code changes.
- **API**: Added `Pause()`/`Resume()` methods to `IJob`, `IJobGroup`, `IJobMonitor`
- **Force Pause**: For jobs with long-running nested operations a 'pre-emptive' pause mode is available:
    - Set `SupportsForcePause => true` and handle `OperationCanceledException` with `HandlePauseExceptionAsync()`
    - See example.

## Tests & Examples

- Example (co-operative) usage in `Examples/BestPractices/PauseResumeExample.cs`
- Force pause (pre-emptive) example in `Examples/BestPractices/ForcePauseExample.cs`
- Test for new `JobCancellationToken` (e.g. Pausing, Waiting, etc.) in `NexusMods.Jobs.Tests/Unit/JobCancellationTokenTests.cs`
- Test for pausing in jobs in `NexusMods.Jobs.Tests/Unit/JobPauseResumeTests.cs`
- Test for force pause functionality in `NexusMods.Jobs.Tests/Unit/JobForcePauseResumeTests.cs`

## Misc

- Added [AsyncEx](https://github.com/StephenCleary/AsyncEx) package rather than implementing own `AsyncManualResetEvent` by hand.

Tests were written with extra care to prevent flakiness in CI.
As we know, the Windows GitHub Actions test runner can be unreliable and stall several seconds at a time. So I've been very careful with how we wait.

fixes #3604


erri120 (changes requested):
This doesn't follow the requirements set out by #3604.

If you think a different approach to #3604 is more appropriate, you should've commented on #3604, explained your approach and allowed for others to comment on it as well.

Al12rs:
From looking at the code, the main reason this alternative approach was implemented was to:
- Support cancellation of a paused job

But the pausing approach proposed doesn't work well for our purposes because it 
- requires cooperative `YieldAsync` calls rather than Token cancellation, which means that we can't pause external code that takes tokens. 
- We want to be able to pause arbitrary async code that supports token cancellation.

erri120 (changes requested):
The job implementation shouldn't deal explicitly with pausing. `SupportsForcePause` should be default `true` and `HandlePauseException` should be used by the job runner directly. Job implementation shouldn't have to call it explicitly. There's also no point in having two different ways pausing, it just adds more complexity for now real benefit.

Sewer56:
(This was refined as requested and can be reviewed for feedback)
(But please don't merge, I got a tiny bit of test flakiness over here. I think it's jobs transitioning to cancelled before hitting running, a race condition which matters for tests, but not real code- but I still need to confirm.)

PR #3692, opened 2025-08-05
Update exception message being send

erri120:
Changes the telemetry data for exceptions to use the first location of the stacktrace as the message. Matomo groups by exception message for some ungodly reason, so this will help with organizing exceptions.

### DIFF
--- a/src/NexusMods.Telemetry/ExceptionData.cs
+++ b/src/NexusMods.Telemetry/ExceptionData.cs
@@ -24,41 +24,62 @@ internal record struct ExceptionData(string Type, string Message, string? StackT
     private static ExceptionData From(Exception exception)
     {
         var type = exception.GetType().ToString();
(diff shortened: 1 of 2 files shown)


Sewer56 (approved):

src/NexusMods.Telemetry/ExceptionData.cs
```diff
@@ -15,13 +15,50 @@ public static List<ExceptionData> Create(Exception exception, List<ExceptionData
...
-                Message: RedactUtils.Redact(exception.Message),
-                StackTrace: exception.StackTrace
-            ));
+            result.Add(From(exception));
         }
 
         return result;
     }
+
+    private static ExceptionData From(Exception exception)
```
Al12rs, line 24: Want to add a comment explaining what From is meant to do and why for future reference?

PR #3697, opened 2025-08-06
Added: Documentation for API Key in contributing, and skipping of tests if no API key is set.

Sewer56:
I mentioned this on Monday, but forgot to PR it seems.
Basically what it says on the can. It's just some of the the basic stuff to make contributing to the project easier.

I made this patch on the side, while experimenting with Nix in between downloads stuff.

### DIFF
--- a/docs/developers/Contributing.md
+++ b/docs/developers/Contributing.md
@@ -42,6 +42,20 @@ but some features may not work as expected. To add a supported game manually:
    NexusMods.App as-main remove-game -g "Game Name" -id "EntityID"
    ```
 
+### Testing with Nexus API Integration
+
+Some tests in the project require a valid Nexus Mods API key to run properly.
+These tests are marked with the `RequiresApiKey` trait and will be skipped if the API key is not set.
+
+#### Obtaining a Nexus API Key
+
+1. Go to your [API Key management page](https://www.nexusmods.com/users/myaccount?tab=api)
+2. Generate a new API key for development/testing purposes
+
+#### Setting Up the Environment Variable
+
+To run networking tests locally, set the `NEXUS_API_KEY` environment variable.
+
 ## Translations
 
 Translations are currently handled via the IDE. See [this issue](https://github.com/Nexus-Mods/NexusMods.App/issues/598) for more details.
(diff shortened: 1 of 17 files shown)


Al12rs (changes requested):
See above

Sewer56:
When the user runs the test ***they are informed an API key is required to proceed***.

The only difference is that now:
- They see a gray icon beside the test as opposed to a red failed cross
- The user gets a consistent error message (not a message that's different in each test, or a random exception)

The intent is merely to make it so a dev without a fully set up development environment would not be swarmed with errors.
I am aware of the extra cases, e.g. requires premium vs not requires premium.
I have no plans to address that at the moment.

This was merely a quick fix I did on the laptop (new development machine) while building for the first time.
I just figured committing the fix would be more worthwhile than throwing it away.

--

(To be more specific, it was useful to me to see what system libraries I didn't have configured in my Linux Nix FHS environment)

Al12rs:
The "RequiresNetworking" tag was removed from the test classes, assuming it was working correctly, should it not remain there?

Sewer56:
That seems to be a mistake, yeah.
Will revert there.

Al12rs (changes requested):
I tested this without an API key and the tests are skipped silently, very easy to miss them.

I would prefer for the tests not to be skipped automatically if the API key is missing, but rather it should be a conscious decision of the developer to configure the runner to skip all the API Key requiring tests.

So if the API key is missing or invalid, the tests should fail with a clear error unless they are configured to be skipped.

docs/developers/Contributing.md
```diff
@@ -42,6 +42,20 @@ but some features may not work as expected. To add a supported game manually:
...
+
+#### Obtaining a Nexus API Key
+
+1. Go to your [API Key management page](https://www.nexusmods.com/users/myaccount?tab=api)
+2. Generate a new API key for development/testing purposes
+
+#### Setting Up the Environment Variable
+
+To run networking tests locally, set the `NEXUS_API_KEY` environment variable.
+
```
Al12rs, line 58: Needs updating to match new behaviour

PR #3701, opened 2025-08-06
Switch to new Query engine

halgari:


erri120:
Could you update our docs to include a short section about setting up Rider to use the proper SQL dialect and how to create and run queries at runtime?

Al12rs:
There seems to be problem with the tests on linux

Al12rs (changes requested):
Tested with the most recent changes but it still seems to be hanging on close on windows.
Will investigate

Al12rs:
Opened a PR on the DB repo to fix the hanging issues:
- https://github.com/Nexus-Mods/NexusMods.MnemonicDB/pull/157

I tested it using project references, and it closes correctly with that. 

Please let me know if there are any issues with the approach.

Al12rs:
> > Could you update our docs to include a short section about setting up Rider to use the proper SQL dialect and how to create and run queries at runtime?
> 
> It did it automatically for me in rider, it's called "Language Injection" and there's a setting for it

Are there ways to open the app DB like it was demoed in Exeter? 
So that we could write and run queries live?

It would simplify testing for new queries, and would also make for a powerful debugging tool if we can access the DB that way.

halgari:
Updated to the latest version of MDB, and included the missing types discussed at standup. 

Yes @Al12rs there is now a way to see the UI: 

`NexusMods.App datamodel ui` will start the UI backend and open the link in a browser window

Issue #3714, opened 2025-08-12
Hash Linux native GOG builds


erri120:
We're using `https://content-system.gog.com/products/{productId}/os/{os}/builds?generation=2` as the endpoint to get per-os builds however native Linux builds don't appear here.

Test with Stardew Valley (`1453375253`) and OS set to `linux` to see the endpoint return `404 Not Found`.

See [Heroic-gogdl](https://github.com/Heroic-Games-Launcher/heroic-gogdl) for reference, this is used by Heroic to install native Linux builds: https://github.com/Heroic-Games-Launcher/heroic-gogdl/blob/a126d4309b41155faf634c9e26fa7928725a8ad0/gogdl/dl/managers/linux.py#L90-L136

PR #3721, opened 2025-08-13
Fetch linux installer archives

erri120:
Part of #3714. Getting this code in first so we can start fetching and extracting installers. Indexing and putting it in the file hashes DB will be done later.

### DIFF
--- a/src/NexusMods.Networking.GOG/Client.cs
+++ b/src/NexusMods.Networking.GOG/Client.cs
@@ -320,7 +320,7 @@ internal class Client : IClient
             if (numRead <= 0) break;
 
             var span = buffer.Span.Slice(start: 0, length: numRead);
-            if (!IsArchive(span, out var index)) continue;
+            if (!ContainsZipArchiveHeader(span, out var index)) continue;
 
             found = true;
             await output.WriteAsync(buffer[index..], cancellationToken: cancellationToken);
@@ -331,8 +331,9 @@ internal class Client : IClient
         await input.CopyToAsync(output, cancellationToken: cancellationToken);
     }
 
-    private static bool IsArchive(ReadOnlySpan<byte> span, out int index)
+    private static bool ContainsZipArchiveHeader(ReadOnlySpan<byte> span, out int index)
     {
+        // ZIP archive magic (50 4B 03 04)
         Span<byte> magic = stackalloc byte[4];
         magic[0] = 0x50;
         magic[1] = 0x4B;


src/NexusMods.Networking.GOG/Client.cs
```diff
@@ -241,6 +242,106 @@ public bool TryGetAuthInfo(out AuthInfo.ReadOnly authInfo)
...
+            found = true;
+            await output.WriteAsync(buffer[index..], cancellationToken: cancellationToken);
+            break;
+        }
+
+        if (!found) throw new NotSupportedException("The input Stream doesn't appear to contain an archive");
+        await input.CopyToAsync(output, cancellationToken: cancellationToken);
+    }
+
+    private static bool IsArchive(ReadOnlySpan<byte> span, out int index)
```
Sewer56, line 334: This is verifying a zip magic header.
Can you just make a quick comment noting that; it may not be immediately obvious.

src/NexusMods.Networking.GOG/Client.cs
```diff
@@ -241,6 +242,106 @@ public bool TryGetAuthInfo(out AuthInfo.ReadOnly authInfo)
...
+                await using var stream = await response.Content.ReadAsStreamAsync(cancellationToken: cancellationToken);
+                await CopyArchive(stream, output, cancellationToken);
+            }
+        }, cancellationToken: token);
+    }
+
+    private static async ValueTask CopyArchive(Stream input, Stream output, CancellationToken cancellationToken)
+    {
+        var bufferSize = Size.MB;
+        var buffer = new Memory<byte>(array: new byte[bufferSize.Value]);
```
Sewer56, line 313: We could probably skip zero initialization here of `new byte[`, but it's not too important in the grand scheme of things.

Sewer56 (approved):
I downloaded one or two Linux games from GOG, and checked the archives; just as a sanity test. Seems good to me.

It's technically possible for `CopyArchive` method to pull data that isn't part of the ZIP archive; since we slice to end.
Though I consider it unlikely enough we shouldn't care for now- same way we wouldn't care about whole installer format changing.
(Misc note: There being more than 1 archive after the script certainly caught me offguard, I would have expected just one archive with both installer and data lol)

That said, down the road, it could technically be better to pass the input stream directly into `ZipArchive` constructor, that means we could download as we extract, which speeds things up; and avoids pulling data beyond the archive.
[Also keeps it in memory, but only what we need]

But stuff here's clearly WIP, per
`// TODO: use something else for production, only using MemoryStream for testing`, so it's not a blocker as the code's not in use yet.

Issue #3159, opened 2025-05-06
Load Order Backend: Support Collection Load Orders


Al12rs:
## Summary:
Ticket for updating the Load Order backend to support dealing with Collection load orders.

## Details:
Depending on the game, collections may come with load order information. 
The app needs to be able to parse this information and use it when installing collections, and be able to write it when creating them.

### Collection data:
Different games store load order information in different ways:
- Stardew Valley: No load order
- Cyberpunk: Apparently collections don't include the RedMod load order
- Awoved (most UE5 games will be the same): flat load order with indexes for UE pak mods:
<details>

```json
  "loadOrder": {
    "Optimized Tweaks EXPT-4-1-0-1739502191": {
      "enabled": true,
      "locked": false,
      "pos": 0
    },
    "Abilities Unlocked-9-1-0-1739537266": {
      "enabled": true,
      "locked": false,
      "pos": 1
    },
    "Better Attributes-16-1-0-1739635856": {
      "enabled": true,
      "locked": false,
      "pos": 2
    },
    "Better Companions-10-1-0-1739548461": {
      "enabled": true,
      "locked": false,
      "pos": 3
    },
    "Less Punishing Encumbrance (Jog)-14-1-0-1739618015": {
      "enabled": true,
      "locked": false,
      "pos": 4
    },
    "More Attributes-16-1-0-1739635911": {
      "enabled": true,
      "locked": false,
      "pos": 5
    },
    "Shorter Death Screen-8-1-0-1739534336": {
      "enabled": true,
      "locked": false,
      "pos": 6
    },
    "Easy Breathing-21-1-0-1739646446": {
      "enabled": true,
      "locked": false,
      "pos": 7
    }
  },
```

[collection.json](https://github.com/user-attachments/files/20065641/collection.json)
</details>
- BG3: custom load order for bg3 pak files:
<details>

```json
{
      "id": "ModFixer.pak",
      "modId": "Mod Fixer-141-1-0-1692715645",
      "enabled": true,
      "name": "ModFixer",
      "data": {
        "author": "Unknown",
        "description": "Missing",
        "folder": "ModFixer",
        "md5": "",
        "name": "ModFixer",
        "type": "Adventure",
        "uuid": "2adde66c-3c56-4fb9-a37b-3806b885b557",
        "version": "1",
        "publishHandle": "0",
        "isListed": true
      },
      "locked": true
    },
    {
[...]

PR #3728, opened 2025-08-14
DuckDB Enabled State Queries

Al12rs:
- This adds a number of queries for getting the enabled state of loadout items using the new DuckDB query system.
- Added a number of tests to verify the correct behaviour of the new query system, and everything seems to be correct so far.
- This doesn't do any query composition at the moment, while we still decide how to approach that, so there is code duplication. 

### Details:

This served as an exploration of the new system and will be used as base for:
- #3159 

Any feedback on the query structure itself is welcome and appreciated, in case there are more performant alternatives.
I did some experimenting with performance but didn't notice significant differences while testing some variations.


Al12rs:
@halgari Removed the single point queries as discussed, which also reduced a lot of the SQL code duplication as well.

halgari (approved):

Issue #3517, opened 2025-07-03
Clone installed collection to local collection


erri120:
Add clone functionality to installed read-only collections from Nexus Mods that will clone the entire collection to a new local mutable collection.

Estimate: at least 2 weeks

Difficulties: only cloning the data that you actually want, not including data like collection references. Would need much testing to make sure this works as intended. Ideally with verify tests.

[Figma link](https://www.figma.com/design/5LmVL8YpgDHFSFA76fgCZA/%F0%9F%93%B1-Mod-library?node-id=3607-25720&t=6ImPjmnFqIFVrVMU-1)

![Image](https://github.com/user-attachments/assets/65a43d62-40be-4198-8a24-b7141c231230)

dkNykita:
So much needed to fix collection issues with "broken" downloads / installs

PR #3729, opened 2025-08-14
Cloning Collections

halgari:
Adds support for #3517, adds a button on the menu of Nexus Collections allowing the user to provide a name (via a dialog box). 

The resulting collection is a clone of the first, with the name changed, the `readonly` flag removed, and references to the Nexus Collection purged. 

Known Issue: the UI glitches slightly and is still orange for "nexus collection" even though the collection is now editable.

### DIFF
--- a/src/NexusMods.Abstractions.Collections/NexusCollectionLoadoutGroup.cs
+++ b/src/NexusMods.Abstractions.Collections/NexusCollectionLoadoutGroup.cs
@@ -1,6 +1,7 @@
 using NexusMods.Abstractions.Loadouts;
 using NexusMods.Abstractions.NexusModsLibrary;
 using NexusMods.Abstractions.NexusModsLibrary.Models;
+using NexusMods.MnemonicDB.Abstractions;
 using NexusMods.MnemonicDB.Abstractions.Attributes;
 using NexusMods.MnemonicDB.Abstractions.Models;
 
@@ -38,5 +39,23 @@ public partial class NexusCollectionLoadoutGroup : IModelDefinition
                 .AsModels<NexusCollectionItemLoadoutGroup.ReadOnly>(Db)
                 .Where(static model => model.IsValid());
         }
+    }
+
+    public static async Task<EntityId> MakeEditableLocalCollection(IConnection conn, EntityId collId, string newName)
+    {
+        var cloneId = await CollectionGroup.Clone(conn, collId);
+        var cloneEnt = Load(conn.Db, cloneId);
+                
+        using var tx = conn.BeginTransaction();
+        // Remap the name
+        tx.Add(cloneId, LoadoutItem.Name, newName);
+        // Make it editable
+        tx.Add(cloneId, CollectionGroup.IsReadOnly, false);
+        // Retract the Nexus references as this is no longer associated with the official collection
+        tx.Retract(cloneId, RevisionId, RevisionId.Get(cloneEnt));
+        tx.Retract(cloneId, CollectionId, CollectionId.Get(cloneEnt));
+        tx.Retract(cloneId, LibraryFileId, LibraryFileId.Get(cloneEnt));
+        await tx.Commit();
+        return cloneId;
     }
 }
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -18,8 +18,8 @@
     <PackageVersion Include="Microsoft.Extensions.TimeProvider.Testing" Version="9.7.0" />
     <PackageVersion Include="Nerdbank.FullDuplexStream" Version="1.1.12" />
     <PackageVersion Include="Nerdbank.Streams" Version="2.11.79" />
-    <PackageVersion Include="NexusMods.MnemonicDB" Version="0.20.0" />
-    <PackageVersion Include="NexusMods.MnemonicDB.Abstractions" Version="0.20.0" />
+    <PackageVersion Include="NexusMods.MnemonicDB" Version="0.21.0" />
+    <PackageVersion Include="NexusMods.MnemonicDB.Abstractions" Version="0.21.0" />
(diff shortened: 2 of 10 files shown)


halgari:
@insomnious next week it would be nice to get your eyes on this, I think I'm missing something on how I'm navigating to the new collection page.

src/NexusMods.Abstractions.Collections/NexusCollectionLoadoutGroup.cs
```diff
@@ -39,4 +40,32 @@ public partial struct ReadOnly
...
+        tx.Retract(cloneId, LibraryFileId, LibraryFileId.Get(cloneEnt));
+
+        // Retract the Nexus references in items so that this is no longer associated with the official collection
+        var db = conn.Db;
+        foreach (var item in conn.Query<EntityId>("SELECT Id FROM mdb_NexusCollectionItemLoadoutGroup(Db=>$Db) WHERE Parent = $Collection", new { Db = db, Collection = cloneId }))
+        {
+            var ent = NexusCollectionItemLoadoutGroup.Load(db, item);
+            tx.Retract(item, NexusCollectionItemLoadoutGroup.Download, NexusCollectionItemLoadoutGroup.Download.Get(ent));
+            tx.Retract(item, NexusCollectionItemLoadoutGroup.IsRequired, NexusCollectionItemLoadoutGroup.IsRequired.Get(ent));
+        }
```
Al12rs, line 66: ## Leftover "BundledGroup" and "ReplicatedGroup: Attributes

This isn't removing the attributes from the extensions of `NexusCollectionItemLoadoutGroup`
- `NexusCollectionBundledLoadoutGroup`
- `NexusCollectionReplicatedLoadoutGroup`

This leaves those entities in a precarious position where they have the attributes of `NexusCollectionBundledLoadoutGroup` for example, but don't have the attributes of `NexusCollectionItemLoadoutGroup` which they are supposed to extend. 

I was able to obtain those half cleaned up `NexusCollectionBundledLoadoutGroup` using a query and they had garbage data in the other attributes.

I tested using SDV Witchcore collection, which has some bundled mods: https://www.nexusmods.com/games/stardewvalley/collections/g14kxi/mods

I think keeping the `NexusCollectionBundledLoadoutGroup` attributes while removing the base `NexusCollectionItemLoadoutGroup` is too error prone. We expect entities to have all the attributes from up in the inheritance chain. 

I think the correct approach in this case would be to strip the attributes from `NexusCollectionBundledLoadoutGroup` and `NexusCollectionReplicatedLoadoutGroup`.

## Missing data provider for stripped Bundled groups
We only have three `ILoadoutDataProvider` implementations:
- `LocalFileDataProvider`
- `NexusModsDataProvider`
- `BundledDataProvider`

If we properly remove `NexusCollectionBundledLoadoutGroup` it results in those mod groups disappearing from the collection page, since there is no provider covering them.

We would need to either convert the bundledGroup to one of the other groups (ensuring the corresponding data is kept even if the collection is removed) or add a new provider that covers entries that don't have any Library correspondence. 

In the future, if we want to be able to create mods out of External Changes, we will likely need a new Data provider for those mods anyways, so I'm leaning towards that being the correct solution.

## Error prone way of stripping attributes
The current approach removes each attribute individually. If a new attribute was to be added, it would be easy to forget to add the corresponding cleanup line in this code.

Do you think there could be a way to get the list of attributes programmatically, so that new attributes would be caught automatically?

Ideally we would be able to catch extensions of `NexusCollectionBundledLoadoutGroup` and their attributes as well, so that if a new extension is added, it would still be cleaned up correctly here, but I think that might be harder still.

Al12rs (changes requested):

halgari:
@Al12rs @erri120 this should handle the changes we discussed, we create a new "ManuallyCreatedArchive" type and link items to that. Also now has a provider for this type that shows up in the collections/library.

Al12rs (changes requested):
I tested cloning the same collection from yesterday on a clean setup
- https://www.nexusmods.com/games/stardewvalley/collections/tckf0m/mods

It seems My Mods was somehow also cloned?
I had no mods installed in My Mods, don't know if it matters at the moment.

Image

src/NexusMods.Abstractions.Collections/NexusCollectionLoadoutGroup.cs
```diff
@@ -39,4 +43,88 @@ public partial struct ReadOnly
...
+        // Make it editable
+        tx.Add(cloneId, CollectionGroup.IsReadOnly, false);
+        // Retract the Nexus references as this is no longer associated with the official collection
+        tx.Retract(cloneId, RevisionId, RevisionId.Get(cloneEnt));
+        tx.Retract(cloneId, CollectionId, CollectionId.Get(cloneEnt));
+        tx.Retract(cloneId, LibraryFileId, LibraryFileId.Get(cloneEnt));
+
+        // Retract the Nexus references in items so that this is no longer associated with the official collection
+        var db = conn.Db;
+        foreach (var item in conn.Query<EntityId>("SELECT Id FROM mdb_NexusCollectionItemLoadoutGroup(Db=>$Db) WHERE Parent = $Collection AND IsRequired = True", new { Db = db, Collection = cloneId }))
```
Al12rs, line 64: Why the `AND IsRequred = True` check?

Isn't that just going to miss optional mods that also need to be stripped of attributes?
For example Bundled Optional mods?

halgari:
I disagree with the idea of not having these show up in the Library, they're a source of files and they aren't tied to any other method of re-installation so they have to show up somewhere for re-installation. It's valid about the duplication I'll see what I can do about that, and remove the check of `IsReadOnly` I forgot about. 

At that point I'm good with the status of this feature

halgari:
Why wouldn't you want those files to be re-installable? If they aren't re-installable how would they be managed? What would the lifetime of those files be?

PR #3733, opened 2025-08-19
Turn Archives into Only Source of Truth for Available Files [And Hopefully Fix Some Missing Archive Errors]

halgari:
```
The student asked, “Master, the code falters and I see no cause. How shall I proceed?”

The master replied,
“When the forest is tangled, cut away the undergrowth.
When the stream runs clear, what silt remains?
When the path is open, what traveler is lost?”
```

---

There's a lot that is mildly wrong with the old code. Essentially we had file entries in MnemonicDB that mapped to archive entries, and those had a unique ID that could exist in multiple archive paths on disk. This means several things could break to give us the "hash not found" error. There was no clear method in the code that maintained this relationship. The GC would orphan files (not retract all their datoms), and would occasionally remap entries into new archives. The `HaveFile` would only ever make sure that a given file entry existed with the given hash, not if the whole chain of files existed.

On top of all this we have the "two sources of truth" problem. We were using MDB for fast access to data, and queries, but the real source of truth is what was in the archives themselves. 

This PR removes a lot of this jank by making the .nx archives the ultimate source of truth for what files exist. When the `ReloadCaches` method is called, all the archive headers are parsed and indexed into memory structures. Those structures are then used for all quick lookups.

I can see how the old code *could* cause issues, even if I can't reproduce these errors. But whatever is happening here, this shouldn't be worse than the old code, and it's a much cleaner implementation. So hopefully this resolves the issues. 

Note: this won't fix cases where the GC killed files it shouldn't have, but there's not much we can do about those cases.


Sewer56:
A few things come to mind, here.

First of all, I'm glad this is being simplified down to using the archives as the source of truth.
I've brought up this issue before; but it was never considered large enough to warrant a fix as it
wasn't considered broken; so I've never had the opportunity. This aligns with what I had in mind in
the past for the FileStore in the past. 

Outside of that; there's one major blocker right now here.

## Resetting Cache in `BackupFiles` is a no-no.

`ReloadCaches` is now called as part of `BackupFiles`, which will lead to a notable performance regression.
This is because `ReloadCaches` accesses every archive on disk, after searching for them.

Consider the following situation:

- You have 1000 mods installed (>= 1000 archives)
- You are about to install a 1000 mod collection.

Every mod download means backing up files during repacking; and thus a `ReloadCaches` call.
This unfortunately means `1000 + 1001 + 1002 ... + 1999` == `1,499,500` times an archive would be opened;
and its headers read into memory. Unfortunately, that would be very slow, especially on Windows
where most of the users would be.

-----

PS. I did some cleanup on this PR, mainly cleaning up a few loose ends:

- Fixing a bug where NxFileStore could throw an exception on next boot refresh if the process dies in the middle of GC.
    - Either due to an explicit process kill, power outage or due to a [Main Battle Tank Shooting your Computer](https://www.youtube.com/watch?v=rSvBFm_MuXw).
    - Basically, there can be a duplicate entry if the old archive is not deleted, but we used `.Add` which throws on duplicates, which is not desirable.
- Dead code/variables.
- Leftover documentation that was no longer indicative of what the code was doing.
- Design document for GC

Outside of that, I got a slightly improved approach to this; I'll post it in the next comment; standup's
about to start. 😅

Sewer56 (approved):
I wouldn't block merging, but I must remind.

Our current goal is to optimise the App for collection downloads in the next week + sprint; so we can push/advertise that.

I'm quite worried about shipping it this way. Although not profiled/tested, I suspect calling `ReloadCaches` for every mod download would likely slow down large collection downloads far more than I'm comfortable with.

https://github.com/Nexus-Mods/NexusMods.App/pull/3733#issuecomment-3200768119

I'll leave it to your discretion.
Fix in place, or open a new ticket to pick up.
Use the comment above or a variation thereof for the ticket.

halgari:
Updated the PR to re-use previous archive states when scanning. Should improve performance quite a bit.

Sewer56:
(I saw the commit earlier when it was put up. Bear in mind, this still keeps the O(N^2) nature of filesystem accesses, in which case 1.5 million filesystem accesses for the example of 1000 existing mods + 1000 new mods. The problem wasn't during GC, but much rather calling it repeatedly during new backups.)

The simple fix is to just append to the existing dictionary. i.e. Add only new data to existing frozen dict. It should be a single method to do so; shorter than the existing one. You have the guarantee the backed up data is always new.

---

Not sure how much this amounts to. Overhead of opening and reading first 4K page for 1.5 million files on Windows is 1m 20s on an NVMe SSD (I know this from a benchmark I did last year). But now we're only requesting metadata, in exchange for not reading 4K pages. Not sure how much that amounts to, not worth writing a benchmark here; easier to just fix.

PR #3760, opened 2025-08-25
Add progress reporting to loadout Synchronization

halgari:
I tried adding progress bars, but there's just too much variance to the timing of the code. So this is an alternative: updating the text on the spinner during apply. This way we can at least see where stalls are happening (deleting files, extracting files, backups, etc). The updates can be fine-grained like "extracting file X of Y" or more course "gathering information".

### DIFF
--- a/src/NexusMods.App.UI/LeftMenu/Items/ApplyControl/ApplyControlViewModel.cs
+++ b/src/NexusMods.App.UI/LeftMenu/Items/ApplyControl/ApplyControlViewModel.cs
@@ -1,4 +1,3 @@
-using System.Reactive;
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using DynamicData;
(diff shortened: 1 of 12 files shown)


src/NexusMods.App.UI/LeftMenu/Items/ApplyControl/ApplyControlViewModel.cs
```diff
@@ -126,6 +130,18 @@ public ApplyControlViewModel(LoadoutId loadoutId, IServiceProvider serviceProvid
                     .Subscribe(isApplying => IsApplying = isApplying)
                     .DisposeWith(disposables);
                 
+                _jobMonitor.ObserveActiveJobs<SynchronizeLoadoutJob>()
+                    .Prepend(ChangeSet<IJob, JobId>.Empty)
+                    .QueryWhenChanged(jobs =>
+                        {
+                            if (jobs.Items.FirstOrDefault()?.Definition is SynchronizeLoadoutJob sJob && sJob.LoadoutId == loadoutId)
+                                return sJob.StatusMessage.AsSystemObservable();;
+                            return new BindableReactiveProperty<string>(value: "NONE").AsSystemObservable();
```
Al12rs, line 139:
```suggestion
                            return new BindableReactiveProperty<string>(value: "").AsSystemObservable();
```

Al12rs, line 139:
```suggestion
                                return sJob.StatusMessage.AsSystemObservable();
                            return new BindableReactiveProperty<string>(value: "").AsSystemObservable();
```

PR #3767, opened 2025-08-28
Avoid accessing all archive headers during NxFileStore Backup

Al12rs:
Avoid calling `ReloadCaches()` and accessing all archives during backup operation, instead only add entries from the newly created archive to improve concurrency.

As outlined by @Sewer56 this should be offer an improvement by reducing the filesystem access and thus the possibility of file access collision on windows.


Sewer56:
Discussed, made a change and we're okay.

PR #3770, opened 2025-08-30
Add initial support for Fallout 4 with Creation Engine service refactor

halgari:
Adds FO4 game stubs so we can start indexing FO4 game files


halgari:
Very simple addition, I'm merging this because it doesn't do anything but add a stub for FO4 that is disabled by default.

Sewer56:
Alright, no worries.

PR #3794, opened 2025-09-04
Upgrade most deps to latest versions

halgari:
Updates most non-transient dependencies to latest versions. Does not update fluent assertions or other pinned libraries


erri120:
I recommend updating all the Avalonia packages in a separate PR and adapt our code to the new release if needed. It's the most important dependency we have after all and the only one worth reading the release notes for.

Al12rs:
Closing in favour of 
- #3808

PR #3819, opened 2025-09-09
Added: Boilerplate/Skeleton of the Downloads Page

Sewer56:
## Summary

Adds Downloads page boilerplate with navigation and left menu structure, the foundation for future downloads page functionality.

## Changes

- Added Downloads page boilerplate with views, view models, and page factory
- Implemented spine downloads button navigation to open the new Downloads page (click download icon)
- Created left menu structure with sections for All Downloads, per-game downloads, and completed downloads
- Added game icons to download menu items with async loading and a fallback
- Displays empty state placeholder indicating download management will be available in next update

## Checklist

- [x] Style Matches Design
- [x] New Pages can be revisited after close+reopen
- [x] All permanent strings are localizable

## Modified Files

- `src/NexusMods.App.UI/Controls/Spine/SpineViewModel.cs` - Added navigation to Downloads page
- `src/NexusMods.App.UI/LeftMenu/Downloads/` - New left menu infrastructure for downloads
- `src/NexusMods.App.UI/Pages/Downloads/` - Complete downloads page implementation
- `src/NexusMods.App.UI/Helpers/ImageHelper.cs` - Shared utility for loading game icons
- `src/NexusMods.App.UI/ImageSizes.cs` - Added LeftMenu icon size constant

PR #3832, opened 2025-09-12
Downloads Page Backend: First Bigger Half

Sewer56:
This comes after:
- https://github.com/Nexus-Mods/NexusMods.App/pull/3819

Go review that first.

## Summary

Implements the core data layer for the Downloads page TreeDataGrid, including filtering, data transformation, and reactive UI components.

This is the first, larger half of the downloads page backend.

## Changes

- **DownloadsFilter**: Filter for grabbing downloads with scopes (All, Active, Completed, GameSpecific).
- **IDownloadsDataProvider & DownloadsDataProvider**: Transforms `IDownloadsService` observables into `CompositeItemModel` collections for `TreeDataGrid`
- **TreeDataGrid Download Components**: i.e. GameComponent, SizeProgressComponent, SpeedComponent, StatusComponent. For custom columns specific for downloads page. 
- **Column Definitions**: Game, Size, Speed, Status column mappings with sorting capabilities

## Modified Files

- `src/NexusMods.App.UI/Pages/Downloads/DownloadsFilter.cs` - Filter types and scopes
- `src/NexusMods.App.UI/Pages/Downloads/IDownloadsDataProvider.cs` - Data provider interface
- `src/NexusMods.App.UI/Pages/Downloads/DownloadsDataProvider.cs` - Data transformation implementation
- `src/NexusMods.App.UI/Pages/Downloads/DownloadComponents.cs` - Reactive UI components
- `src/NexusMods.App.UI/Pages/Downloads/DownloadColumns.cs` - Column definitions
- `src/NexusMods.App.UI/Resources/Language.resx` - UI string resources


src/NexusMods.App.UI/Pages/Downloads/DownloadColumns.cs
```diff
@@ -0,0 +1,139 @@
...
+ *    - Maps to: DownloadComponents.SpeedComponent (from DownloadComponents.cs)  
+ *    - Shows: Transfer rate "5.2 MB/s" or "--"
+ *    - Sortable by current transfer rate
+ * 
+ * 5. STATUS COLUMN (defined here as DownloadColumns.Status):
+ *    - Maps to: DownloadComponents.StatusComponent (from DownloadComponents.cs)
+ *    - Contains: Progress bar + Pause/Resume + Cancel + Kebab menu
+ *    - Consolidated all download actions into single column
+ *    - Sortable by completion percentage (0% to 100%)
+ */
```
Al12rs, line 39: Put the comments next to the actual columns definitions, otherwise they will become outdated too easily.

Al12rs: I just added the Collections column to the Library.
Plan is to allow users to toggle which columns they want to see in the future (no idea how yet), but that means there will be opportunity to add a bunch of columns.

Sewer56: Ah. My original intent at the time was to have a listing; since some columns were reused from library and other views.
However, in the end, I had to duplicate some of those columns instead in the final branch/PR, due to the usage of generics.

I can inline it as comments to the items in this case; yeah; that works.

I'll add that to my notes for the final PR to avoid conflicts.

Al12rs (approved):

PR #3837, opened 2025-09-15
first attempt at gimmicky PluginDelegate

killzoms:
I am not all that familiar with this particular code base, but this allows Fomod to locate files on Linux and auto toggle fomod options.


Sewer56:
What's an example file I could test with?
I'm not sure when this would be reviewed, but having a unit test that covers this would be handy.

Sewer56:
:ok_hand:

erri120:
The FOMOD format is nonsensical and doesn't match modern use cases as it expects a definitive installation order for these "plugin delegates". 

We're currently in the process of adding Fallout 4 and Skyrim SE support and might take a look at this again after we have load order support and after we decide whether we want to implement these methods at all or just keep them as is.

PR #3868, opened 2025-09-17
Refactor hash DB formats and protect against "full game backup"

halgari:
This code does the following: 

* Adds depot and manifest info to GOG file hashes
* Reworks Steam hashes to be based mostly on manifests (instead of AppIds), so that old manifests are correctly emitted even when the appIds point to new manifests
* Correctly indexes DLC for GOG. These productIds don't have builds, so they didn't show up before
* Refuses to back up game files in batches larger than 2GB
* Provides SQL routines to figure out LocatorIds based on the indexed file state
* Lots of small changes to the generation of the hashes DB

TL;DR - I was able to replicate the full set of issues users have seen in the past. I got an error as the app tried to back up 80GB of Cyberpunk data (due to an outdated hash database). I updated the database and the backup attempt went away. I also verified that PL is now correctly detected and not put into external changes. 

Notes for reviewers: 
* I had to disable one of the checks on the `SDV.2_5_2025` test. This is because the hashdb purged a old version of SDV and that test now fails because we can't get the locator Ids.


src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1314,6 +1325,10 @@ await Parallel.ForEachAsync(files, async (item, _) =>
             }
         );
 
+        var totalSize = archivedFiles.Sum(static x => x.Size);
+        if (totalSize > MaximumBackupSize)
+            throw new Exception($"Cannot backup files, total size is {totalSize}, which is larger than the maximum of {MaximumBackupSize}");
```
halgari, line 1330: This is the backup size check

tests/NexusMods.DataModel.SchemaVersions.Tests/MigrationSpecificTests/TestsFor_0004_RemoveGameFiles.cs
```diff
@@ -21,7 +21,8 @@ public async Task Test(string databaseName)
         var db = tempConnection.Connection.Db;
         foreach (var loadout in Loadout.All(db))
         {
-            loadout.Contains(Loadout.LocatorIds).Should().BeTrue("Loadout should contain LocatorIds");
+            if (databaseName != "SDV.2_5_2025.rocksdb.zip")
```
erri120, line 24: What's up with this?

Issue #3286, opened 2025-05-29
Bring back global download page


captainsandypants:
## Summary:
We need a place to view global downloads. Users have confirmed this is something they want.

## Details:
For now we only need a temporary page that shows everything that is downloading.
The implementation should be as simple as possible for Phase 1, it's a placeholder for something better.
This also means users will know that clicking the download button in the spine is intended to give them more details, which will improve over time, where as currently it does nothing.

### Expected behaviour
When I click on the download spine button, I'm taken to a page that shows me everything that is downloading.
Ideally this would be a list of rows with what's currently downloading and what's queueing but if it's easier it can just be one progress bar with the total.

### Design
[Figma Link](https://www.figma.com/design/Ntr9PttR5VrqpTRwwZHzLu/%F0%9F%93%B1-Mod-list---adding-removing-mods?node-id=5642-20566&p=f&t=u8jtgyaIowRqeqHR-0)

### Definition of Done
- [x] Clicking the download button at the bottom of the spine menu opens a download page
- [x] On the download page I can see, in some form, everything that is currently downloading.

### Test Data:
Download something.

### QA Checklist:
- [ ] Add QA label if relevant
- [ ] Add DODs
- [ ] Add test data/scenarios
- [ ] QA test Windows
- [ ] QA test Linux

Sewer56:
Make the tickets, close this.

Al12rs:
👍 QA will close this after doing their pass as part of release testing.

PR #3870, opened 2025-09-18
First pass at download page UI

insomnious:
Not sure much more is worth doing until more functionality in place.

This PR is part of #3286 and targets the `downloads-page-ui` branch for @Sewer56 

- Now using StandardButton component instead of a random Button
- Uses a single Toolbar component instead of 3
- ProgressBar now uses the 'DownloadBar' class
 
image


Sewer56:
(Merged into my branch, not main)

PR #3875, opened 2025-09-18
Rework loadout queries

### LINKED ISSUE #3874: Update loadout queries to pass enabled state along
We can use a macro parameter `onlyEnabled` to specify whether a macro should only include enabled items or not. This would allow us to better combine macros and make our queries simpler.

erri120:
Closes #3874.

- Renamed the macros to be consistent
- Added `onlyEnabled` parameter to macros
- Removed duplicate queries
- Simplified where possible


halgari (approved):
Good cleanup over all

PR #3878, opened 2025-09-18
Fix hashed service startup issues

halgari:
Fixes a few issues in the existing file hashes service

* Makes the service a IHostedService so we can get async startup routines. This is required because if there's a connection error
in an existing DB there's no way to force the app startup to halt until we get a fixed DB. This means that something else in the app could call `.Current` causing an error. Now, if the service starts, we'll have an up-to-date db.
* Catches errors connecting to the existing DB, downloads a new DB if the most recent one fails. 

This code helps in the cases where DBs are not forwards compatible. That is a new version of the app, can't work with a old version of the hash db.


Al12rs:
Will backport to 0.17.1

Issue #3902, opened 2025-09-25
Add system to query and access Game Files as well as LoadoutItems


Al12rs:
## Problem Statement:
Currently items in the Loadout don't include game files, meaning that various systems interested in what files are part of a loadout, can't see them.

Ideally there would be a homogeneous interface to query and access files of a loadout, regardless of whether they are from mods, game files, generated etc.

#### Examples of systems requiring this:
- File Conflicts (conflicts with game files)
- Load Order (DLCs that need to be sorted)
- Health Checks (checks on required DLCs, parse game data)

These systems are interested in all the files, not just the conflict winners, but knowing which of them are winners would be very helpful as well.

#### Use cases:
- Find all files matching a particular path or path pattern
- Open/read the contents of a file, regardless of it being a LoadoutFile, Game file, generated file etc.

## Implementation approach:
TBD

PR #3914, opened 2025-09-29
Synchronizer in query

halgari:
- Part of #3902

Implements much of the guts of the Synchronizer "layers" in SQL. 

`synchronizer.WinningFiles(db)` will list all the winning files for all loadouts (filter on the loadout column to constrain the results), these can then be used in any number of queries. 

The results will include

* Game files
* Loadout Files
* Overrides
* Intrinsics


Al12rs:
Please keep in mind that for #3902 we need to have access to all the files in the loadout, not just the winning files.

For certain use cases, knowing the which files are winning is also useful, but for many cases we also need to know all the disabled/non winning files.

erri120 (approved):

Issue #3912, opened 2025-09-29
Failed to register table function with duplicate name


Al12rs:
## Problem:
When starting the app, it sometimes crashes with the following exception:

```
System.AggregateException: One or more errors occurred. (Failed to register table function.)
 ---> System.InvalidOperationException: Failed to register table function.
   at NexusMods.HyperDuck.ATableFunction.Register(Connection connection)
   at NexusMods.HyperDuck.DuckDB.Register(ATableFunction aTableFunction)
   at NexusMods.MnemonicDB.Connection.RegisterWithEngine(DuckDB duckDb)
   at NexusMods.MnemonicDB.Connection..ctor(ILogger`1 logger, IDatomStore store, IServiceProvider provider, IEnumerable`1 analyzers, IQueryEngine queryEngine, Boolean readOnlyMode, String prefix)
   at NexusMods.Games.FileHashes.FileHashesService.OpenDb(DatabaseInfo databaseInfo) in G:\Development\Repos\NexusMods.App\src\NexusMods.Games.FileHashes\FileHashesService.cs:line 86
   at NexusMods.Games.FileHashes.FileHashesService.CheckForUpdateCore(Boolean forceUpdate, CancellationToken cancellationToken) in G:\Development\Repos\NexusMods.App\src\NexusMods.Games.FileHashes\FileHashesService.cs:line 207
   at NexusMods.Games.FileHashes.FileHashesService.StartAsync(CancellationToken cancellationToken) in G:\Development\Repos\NexusMods.App\src\NexusMods.Games.FileHashes\FileHashesService.cs:line 693
   at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>b__14_1(IHostedService service, CancellationToken token)
   at Microsoft.Extensions.Hosting.Internal.Host.ForeachService[T](IEnumerable`1 services, CancellationToken token, Boolean concurrent, Boolean abortOnFirstException, List`1 exceptions, Func`3 operation)
   at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>g__LogAndRethrow|14_3(<>c__DisplayClass14_0&)
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
[...]

PR #3928, opened 2025-09-30
Update to MnemonicDB 0.28

halgari:
As discussed, we discussed in #3912 MnemonicDB was upgraded so that all connections get a global ID and all models are read via `mdb_{name}()`.  Now to access the file hashes you can call a model function via `mdb_gogbuilds(DbName=>'hashes')` (case insensitive). This connection is resolved at query execution time, so new hash databases will be read as soon as the new connection is spun up (and the old one shut down? Need to look into that). If no name is provided, `default` is used instead, meaning we only need to use DbName when not providing the DB as a parameter or when using the non-primary database.

This won't crash as before, may still be some issues around db upgrades.


Al12rs (approved):
Tested and seems to be working fine

PR #3940, opened 2025-10-01
Update LocatorIds before synchronization

### LINKED ISSUE #3455: Synchronizer is using old LocatorIds in case of Game updates/DLC changes during `BuildSynchTree`
# Bug Report

## Summary of Current behaviour
If the game updates, or installed DLCs are changed, even if we have the hashes for the new LocatorIds, the app will still ingest and backup all the game files, just to remove them at the end of Synchronize (inside `ReprocessGameUpdates`).

`Synchronize` uses the old `LocatorIds` stored on the the `Loadout` to determine what the game state should look like, rather than looking at the new `LocatorIds` advertised by the game store.

This leads to all updates to fist be ingested into Override, just to then get removed a step later.

This can lead to long Apply times, as game updates can be of multiple GBs.

## Details:
App currently only checks if the `LocatorIds` for the game have changed (e.g. Steam locatorIds) during `ReprocessUpdates` step which happens after Synchronization has already backed up and ingest new files and deletions into External changes.

### Steps to reproduce
Adding a DLC will backup the DLC files and then discard them:

- Manage a Cyberpunk install without RedMod
- Close the app
- Add RedMod DLC from Steam or GOG
- Open the app and Apply (add a mod or do some other change to allow Applying)
- App will backup the REDmod DLC files as observable by the log line
- `[DEBUG] Backing up 4721 files of 90.123 MB in size`

Removing a DLC will put a bunch of `DeletedFile` entries in External Changes
- Manage a Cyberpunk install with RedMod
- Close the app
- Remove RedMod from Steam or GOG
- Open the app and Apply
- All the RedMod files will be in Override as Deletion entries.

### What is the expected behaviour?
App should realize that LocatorIds have changed early and avoid backing up or Ingesting game changes and instead use the correct locatorId data from the start.

### QA Checklist:
- [ ] Add QA label if relevant
- [ ] Add repro steps
- [ ] Add expected behaviour
- [ ] Add known limitations with the fix or issues
- [ ] QA test Windows
- [ ] QA test Linux

halgari:
fixes #3455 

This pull request refactors the loadout synchronization logic to improve the handling of locator IDs and game version updates when ingesting from disk. The main changes involve separating the update of locator IDs from the override reprocessing logic, ensuring locator IDs are updated before building the sync tree.

### DIFF
--- a/src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -639,11 +639,9 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
             CleanDirectories(foldersWithDeletedFiles, newState, loadout.InstallationInstance);
         }
 
-        loadout = await ReprocessOverrides(loadout);
-
         job?.SetStatus("Archive Cleanup");
         await _garbageCollectorRunner.RunAsync();
-        
+
 
         return loadout;
     }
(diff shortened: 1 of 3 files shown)


src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1148,10 +1146,17 @@ private bool ActionIngestFromDisk(Dictionary<GamePath, SyncNode> syncTree, Loado
             }
         }
 
+        // Update locator IDs before building the sync tree
+        loadout = await UpdateLocatorIds(loadout);
```
Al12rs, line 1150: I think `UpdateLocatorIds(loadout)` might need to go before the `await ActivateLoadout(loadout);`  (L1144) during the loadout switching code.

Otherwise if there was a game update, and the user tried to switch to a different loadout we would:
- Correctly deactivate the previous Loadout, updating its `LocatorIds`
- Reset the game to the new `LocatorIds` vanilla state
- Try to Activate the other loadout against a vanilla game state that doesn't match the loadout locatorIds (still on the old game version).
- This breaks the Activate code which assumes we are doing a clean Apply of a loadout against its matching vanilla state.

Al12rs:
Moved the `UpdateLocatorIds()` call at the start of Synchronize to cover the loadout switch case. Doing this now to get a 0.18.1 build to send to QA and have them test this.

PR #3942, opened 2025-10-02
Added: Right Click Functionality to Loadout and Library Pages

### LINKED ISSUE #3863: Support right click for kebab menu on library tree data grid
## Description

Implement right-click context menu functionality (kebab menu) for the library tree data grid component. This will provide users with contextual actions when right-clicking on library items.

## Background

This builds upon the design concepts established in issue #3746, specifically focusing on the library tree data grid implementation rather than collections management.

## Requirements

- [ ] Implement right-click event handling for library tree data grid rows
- [ ] Display contextual kebab menu on right-click
- [ ] Reuse existing menu component for consistency
- [ ] Ensure proper menu positioning and accessibility
- [ ] Include relevant library-specific actions in the menu (e.g., install, uninstall, remove from library, view details)

## Design Reference

The kebab menu design should follow the established patterns from the [Figma component library](https://www.figma.com/design/RGRSmIC4KoVlIosQB5YmQY/%F0%9F%93%B1%F0%9F%A7%B1-App-components?m=auto&node-id=1682-7579&t=exL8s0b5P4Et22wt-1) referenced in #3746.

## Implementation Notes

- Focus specifically on the library tree data grid component
- Consider which actions are most relevant for library items
- Ensure the menu appears at the cursor position
- Handle edge cases for menu positioning near screen boundaries
- Maintain consistency with other right-click menus in the application

## Related

- Related to #3746 (design reference)

### LINKED ISSUE #3864: Implement right-click context menu functionality for TreeDataGrid in Loadout page
## Description

Implement right-click context menu functionality for the TreeDataGrid component in the Loadout page. This will provide users with quick access to common actions for loadout items normally present in the 'kebab menu'. (3 dots)

## Acceptance Criteria

- [ ] Right-clicking on items in the Loadout page TreeDataGrid shows a context menu
- [ ] Context menu is reused from the 'Kebab' menu at the end of each row.

Sewer56:
fixes #3864
fixes #3863
- #3864
- #3863


Sewer56:
I figured out the underlying issue that was preventing this from working with XAML, actually.
Took a bit, but I got it.

There's a couple of facets as to why we couldn't get anything to work.
In the most simplest level it narrows down to this:

```
Updating `DataContext` for `TreeDataGridRow` does not update the `DataContext` of `ComponentControl` `ComponentTemplate`(s).
```

To illustrate:

```xaml
<!-- LoadoutItemMenuFlyout - Reusable MenuFlyout for loadout items -->
<MenuFlyout x:Key="LoadoutItemMenuFlyout" FlyoutPresenterClasses="TreeDataGridMenu"
            x:DataType="{x:Type controls:CompositeItemModel, x:TypeArguments=abstractions:EntityId}">
    
    <!-- View Mod Page -->                                                   <!-- 👇 ✅ Updated Correctly -->
    <controls:ComponentControl x:TypeArguments="abstractions:EntityId" Content="{CompiledBinding}"> 
        <!-- 👇 ❌ Not Updated -->
        <controls:ComponentControl.ComponentTemplate>
            <controls:ComponentTemplate x:TypeArguments="controls:SharedComponents+ViewModPageAction"
                                        ComponentKey="{x:Static local:LoadoutColumns+EnabledState.ViewModPageComponentKey}">
                <controls:ComponentTemplate.DataTemplate>
                    <DataTemplate DataType="{x:Type controls:SharedComponents+ViewModPageAction}">
                        <MenuItem Header="{x:Static resources:Language.Library_ViewModPage_Text}"
                                    Command="{CompiledBinding CommandViewModPage}"
                                    IsEnabled="{CompiledBinding IsEnabled.Value}">
                            <MenuItem.Icon>
                                <icons:UnifiedIcon Value="{x:Static icons:IconValues.OpenInNew}" />
                            </MenuItem.Icon>
                        </MenuItem>
                    </DataTemplate>
                </controls:ComponentTemplate.DataTemplate>
            </controls:ComponentTemplate>
        </controls:ComponentControl.ComponentTemplate>
         <!-- 👆 ❌ Not Updated -->
    </controls:ComponentControl>

    <!-- ... -->
```

The way our current system works, is when you assign a `CompositeItemModel<TKey>` to `Content`; we actually do a key lookup, and extract the individual `IItemModelComponent`(s), before assigning them to the inner `ComponentTemplate`(s).

```csharp
protected override Control? BuildContentControl(CompositeItemModel<TKey> itemModel, out Optional<string> contentPresenterClass)
{
    contentPresenterClass = Optional<string>.None;

    if (ComponentTemplate is null) throw new InvalidOperationException();
    if (!itemModel.TryGet(ComponentTemplate.ComponentKey, ComponentTemplate.ComponentType, out var component)) return Fallback;

    // NOTE(erri120): DataTemplate.Build doesn't use the
    // data you give it, need to manually set the DataContext.
    // Otherwise. the new control will inherit the parent context,
    // which is CompositeItemModel<TKey>.
    var control = ComponentTemplate.DataTemplate.Build(data: null);
    if (control is null) return Fallback;

    control.DataContext = component;
    contentPresenterClass = ComponentTemplate.ComponentKey.Value;
    return control;
}
```

Unfortunately, this was not working quite right for `MenuFlyout`(s), and all other controls/UI components which may be reused.

------

This is a multi layered issue:

1. In cases like opening the `MenuFlyout`, `ContextMenu`, and other reuses of UI controls behind the scenes by Avalonia; the `DataContext` is changed before the control is rendered.
    - This means `Content` is changed before the item is loaded.
    - But this was being dropped because of `if (IsLoaded) _serialDisposable.Disposable = Subscribe(content);`.
    - We didn't factor in the `DataContext` can change before the item is loaded.
        - Since with items like `MenuFlyout` we would do unload + reload, rather than creating new each time.

2. Even if this check is removed and `Subscribe` is called, content wouldn't refresh.
    - We only refresh if the number of components (i.e. usually columns since 1 per column), changes.
    - But the number of columns was not changing since last usage, as all rows have same column count, so `GetObservable(content)` returned no values.

Also; in normal usage `_serialDisposable.Disposable = Subscribe(content);` was actually ran across twice on each control load; because the previous code had it inside both `OnPropertyChanged` and `OnLoaded`. But- `OnPropertyChanged` would always be called when setting the `DataContext` (`Content`) in the first place, making `OnLoaded` redundant.

So I removed `OnLoaded`, and made `OnPropertyChanged` always update the subscription if `Content` is changed; meaning the subscription is only attempted once. In addition, I made it so calling `Subscribe` will always update the inner `ComponentTemplate`(s) if this control was previously rendered- as that indicates control reuse.

Al12rs (commented):
Nice work finding a fix for that issue!

The code looks good and it seems to be working great from testing.

I noticed that the `CollectionLoadoutView` (installed nexus collection page) is missing the context menu though.

It should the same context menu as the `LoadoutView`, so it should just be a matter of adding the following to `CollectionLoadoutView.axaml` I think

```xml
<TreeDataGrid.Styles>
    <Style Selector="TreeDataGridRow">
        <Setter Property="ContextFlyout" Value="{StaticResource LoadoutItemMenuFlyout}" />
    </Style>
</TreeDataGrid.Styles>
```

Issue #3903, opened 2025-09-25
Handle INI files in Fallout 4


halgari:

Al12rs:
```ini
[Archive]
bInvalidateOlderFiles=1
sResourceDataDirsFinal=
```

Standard practice is to put these into `Fallout4Custom.ini` rather than `Fallout4.ini`.

A simple fist step could be to to update these values manually to the  `Fallout4Custom.ini` on disk (creating it if necessary) without actually managing the ini files themselves for now.

PR #3943, opened 2025-10-02
Add preferences paths to Skyrim/Fallout4 

halgari:
Adds a mapping for `LocationId.Preferences` for Fallout4/SkyrimSE and ignores saves in those folders. I'll add some diagnostics for this in the future, but at least for now we can track/change inis for the games. I was planning to go deeper with this by parsing/ingesting the contents of the INIs themselves, but that code ended up way more complex than I had hoped. This is a simpler approach and works for now.

In a future PR we'll add a diagnostic to inform the user about FO4 changes they need to make

### DIFF
--- a/src/NexusMods.Games.CreationEngine/ACreationEngineSynchronizer.cs
+++ b/src/NexusMods.Games.CreationEngine/ACreationEngineSynchronizer.cs
@@ -21,7 +21,11 @@ public abstract class ACreationEngineSynchronizer : ALoadoutSynchronizer
             {pluginsFile.Path, pluginsFile},
         };
     }
- 
+
+    
+    private static readonly GamePath SavesPath = new GamePath(LocationId.Preferences, "Saves");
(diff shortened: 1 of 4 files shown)


Al12rs (commented):
The main issue we currently have is that many mods will not work without adding the archive invalidation tweaks to the ini files.

- #3903 

This could be achieved by adding a `PrepareGame()` step to game management, where games can define custom one time setups to do. 

This PR would add the game `MyGames` folder and all contents except for saves to the managed files.

This has the downside of changes to ini files appearing in external changes. 
We discussed that we likely don't wish to have users manage ini files like other normal loadoutFile entries long term, with direct conflicts and what not.

With that in mind I'm not sure we should do this change, as reverting this would likely require a migration to remove all Loadout items with the matching locationId. 

I'd rather not do anything with INI files inside the loadout (and by extension anything else in `MyGames` folder) until we have a clearer idea of how we want to handle them long term.

Al12rs (approved):
We can merge this for now but we should consider whether we want to keep this approach before Skyrim/FO4 general support release

PR #3948, opened 2025-10-05
Add shared semaphore lock for un-managing game from loadout synchronization

### LINKED ISSUE #3939: Prevent users from un-managing game while a loadout is synchronizing
## Problem statement:
If a loadout is Synchronizing (applying) and a user tries to unmanage the game, we can get in an inconsistent state and get exceptions.

We should prevent users from being able to unamange a game if one of its loadouts are in the process of Applying.

## Desired behaviour:
TBD by design

axmanalad:
Fixes #3939 

This PR retrieves the same semaphore instance from `SynchronizerService.cs`, and apply this semaphore in `UnManage()` from `ALoadoutSynchronizer`. Based on the semaphore reference, this will follow the locking mechanism where `UnManage()` will resume its request if there is no other loadouts synchronizing Users will have to wait for the corresponding loadout to complete its synchronization before the un-manage job starts.

There is another bug vice versa, where a user can synchronize a loadout during the un-manage job; users should not be able to synchronize a loadout while the game is un-managing. The PR also adds a fix to the UI where the "Apply" button will not appear when the game is un-managing.  In `Synchronize()`, the latest loadout synchronize job is cancelled if there is an un-manage job. However, I am not sure what to return for this case: whether it should be thrown an error or it should still return the default value.

### DIFF
--- a/src/NexusMods.App.UI/LeftMenu/Items/ApplyControl/ApplyControlViewModel.cs
+++ b/src/NexusMods.App.UI/LeftMenu/Items/ApplyControl/ApplyControlViewModel.cs
@@ -111,13 +111,14 @@ public class ApplyControlViewModel : AViewModel<IApplyControlViewModel>, IApplyC
                     .Subscribe(status =>
                     {
                         var (ldStatus, isProcessing,  gameStatus, running) = status;
-                        
+                        var hasUnmanageJob = _jobMonitor.Jobs.Any(job => job.Definition is UnmanageGameJob);
                         IsProcessing = isProcessing;
                         CanApply = !isProcessing
                                    && !running
                                    && gameStatus != GameSynchronizerState.Busy
                                    && ldStatus != LoadoutSynchronizerState.Pending
-                                   && ldStatus != LoadoutSynchronizerState.Current;
+                                   && ldStatus != LoadoutSynchronizerState.Current
+                                   && !hasUnmanageJob;
                         IsLaunchButtonEnabled = !isProcessing 
                                                 && !running
                                                 && gameStatus != GameSynchronizerState.Busy
--- a/src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -1788,65 +1788,76 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
     /// <inheritdoc />
     public async Task UnManage(GameInstallation installation, bool runGc = true, bool cleanGameFolder = true)
     {
-        await _jobMonitor.Begin(new UnmanageGameJob(installation), async ctx =>
-            {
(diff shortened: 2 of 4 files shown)


Al12rs (changes requested):
Thanks for the contribution again!

I'm thinking that instead of passing around the semaphore we could add an Unmanage method to the `SynchronizerService` and have the locking happen in there, to remain consistent with how the synchronize works.

We also shouldn't use the presence of the job to determine whether we are unmanaging or not during Synchronize. The lock should prevent the events from happening concurrently. What we can do is to check that the loadout is still valid after entering the lock and return early if it isn't (has been deleted/unmanaged).

Good thinking of using the presence of the job in the ApplyControlViewModel to determine the state of the button. I'm not sure if the current approach is correct though or if we should instead subscribe to an observable of the job. We might not get updated correctly on status changes otherwise.

src/NexusMods.App.UI/LeftMenu/Items/ApplyControl/ApplyControlViewModel.cs
Al12rs, line 128:
```suggestion
var hasUnmanagingJob = _jobMonitor.HasActiveJob<UnmanageGameJob>(job => job.Installation.GameMetadataId == _gameMetadataId.Value)
    .Prepend(false);

// Note(sewer):
// Fire an initial value with StartWith because CombineLatest requires all stuff to have latest values.
// In any case, we should prevent Apply from being available while a file is in use.
// A file may be in use because:
// - The user launched the game externally (e.g. through Steam).
//     - Approximate this by seeing if any EXE in any of the game folders are running.
//     - This is done in 'Synchronize' method.
// - They're running a tool from within the App.
//     - Check running jobs.
loadoutStatuses.CombineLatest(isProcessingObservable, gameStatuses, gameRunningTracker.GetWithCurrentStateAsStarting(), hasUnmanagingJob)
    .OnUI()
    .Subscribe(status =>
    {
        var (ldStatus, isProcessing,  gameStatus, running, isUnmanaging) = status;
        IsProcessing = isProcessing;
        CanApply = !isProcessing
                   && !running
                   && gameStatus != GameSynchronizerState.Busy
                   && ldStatus != LoadoutSynchronizerState.Pending
                   && ldStatus != LoadoutSynchronizerState.Current
                   && !isUnmanaging;
        IsLaunchButtonEnabled = !isProcessing 
                                && !running
                                && gameStatus != GameSynchronizerState.Busy
                                && ldStatus == LoadoutSynchronizerState.Current
                                && !isUnmanaging;
        
    })
    .DisposeWith(disposables);
```

Al12rs: This should make sure we that both the Apply and Launch buttons are disabled while we are unmanaging the game.

Al12rs:
In the future we will likely want to switch from a semaphore to a Queue of operations. 
With the queue we can have much more granular control of which operations can run at the same time and do things like condensing or skipping operations.

PR #3955, opened 2025-10-07
Implement initial set of Mixpanel events

### LINKED ISSUE #3937: Add Mixpanel events
See internal spreadsheet for events. This ticket will only mention issues with some of the properties:

Super properties with issues:

- `is_verified_mod_author`: don't have this data
- `is_staff`: don't have this data

Events with issues:

- `mods_download_failed`, `mods_installation_failed`, `collections_download_failed`, `collections_installation_failed`, `app_crashed`:
  - `error_code`: app doesn't have error codes
  - `error_message`: unusable due to 255 character limit (see #3936)
- `app_downloaded`, `app_installed`, `app_uninstalled`, `app_updated`, `app_crashed`, `app_upsell_clicked`: what and how?

erri120:
Resolves #3937.

Implemented everything that makes sense for the app and was straightforward to do.


src/NexusMods.Sdk/Tracking/Events.cs
```diff
@@ -252,6 +252,10 @@ public readonly record struct DurationValue(long Milliseconds)
     {
         public static implicit operator DurationValue(long value) => new(value);
         public static implicit operator DurationValue(TimeSpan value) => new((long)value.TotalMilliseconds);
-        public static implicit operator DurationValue(Stopwatch value) => new(value.ElapsedMilliseconds);
+        public static implicit operator DurationValue(Stopwatch value)
+        {
+            value.Stop();
+            return new DurationValue(value.ElapsedMilliseconds);
+        }
```
Al12rs, line 259: Shouldn't the stop happen on the caller side? This seems like it could easily be delayed, making the communicated elapsed time wrong

erri120: `Stop` is a no-op when called on an already stopped Stopwatch. The caller could stop the watch themselves but since the precision is only on a millisecond basis, the time loss is often `0` or negligible, especially since this implicit operator gets called first before the actual method gets called.

PR #3958, opened 2025-10-07
Refresh `LocatorId` game store data during Synchronize to detect game version changes

### LINKED ISSUE #3957: App does not detect Game `LocatorId` (versions) changes until restart
## Problem statement:
App doesn't detect changes in the games `LocatorIds` until it is restarted. 

This can lead to Synchronizer using the wrong locator ids when applying and result in App labelling Game Files as external changes to backup.

## Details:
We rely on `GameFinder` to get us the installed games information. This information is then cached and reused.

`GameFinder` does not currently expose a way to refresh information for a specific game, but instead only exposes method to get the information for all the games installed to a specific game store.

This makes refreshing the `LocatorId` information rather inefficient.

## Solution approaches:
- A. Refresh the information from all the games of the current game store each time we run the synchronizer.
- B. Update Game Finder and game store handler code in it to expose a method to return only the information for a single game, that is hopefully more efficient than getting the info for all the games.
- C. Other?

Al12rs:
- Fixes #3957

### DIFF
--- a/src/NexusMods.Abstractions.GameLocators/IGameLocator.cs
+++ b/src/NexusMods.Abstractions.GameLocators/IGameLocator.cs
@@ -10,5 +10,5 @@ public interface IGameLocator
     /// </summary>
     /// <param name="game">The game to find the location of.</param>
     /// <returns>Location of the game.</returns>
-    public IEnumerable<GameLocatorResult> Find(ILocatableGame game);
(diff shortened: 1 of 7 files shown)


Al12rs:
> This is going to spam the logs.
Good point, I could skip the logging if the force refresh is set perhaps

Issue #3852, opened 2025-09-16
Resolving file conflicts


erri120:
## User Story

**As a** user

**I want to** resolve file conflicts

**So that** I can decide what files are deployed over others.

## Requirements

- Resolve file conflicts by changing the conflict priority of mods
- If both `Mod A` and `Mod B` want to deploy file `foo.txt`, the user should be able to resolve the conflict by changing the priority of either `Mod A` or `Mod B` to take precedence.
- Conflict priority is defined as greater number wins
- Game files should always have the lowest priority
- Overrides should always have the highest priority

## Designs

https://www.figma.com/design/1sbIAJGr8URkY3wmisyaJg/%F0%9F%93%B1-Rules?node-id=1603-27074&t=KhFTy0OFErKxWCnT-1

TreeDataGrid columns:

- "Conflict priority": priority number and up/down arrows
- "Name": name of the loadout group with the conflicts
- "Conflicts": number of "wins" and "loses" using icons
- "Collection": parent collection of the group
- "Actions": button to view conflicts

TreeDataGrid should support drag & drop to re-arrange the order, similar to load order.

PR #3991, opened 2025-10-14
Assign loadout item group priorities

### LINKED ISSUE #3966: Add loadout manager API
## Problem statement:
Currently the code that affects Loadouts is scattered throughout the code base, with some of the code even being in the UI project. This makes it impossible to modify behaviour for things such as File Conflict state management.

These operations should go through a centralized API instead, where it will be possible to add side effects to the transactions as needed.

The synchronizer has become a loadout manager of sorts but it should really just deal with reading and writing files to/from disk. The new loadout manager API will deal with creating, removing and duplicating loadouts as well as adding and removing mods to loadouts.

erri120:
Resolves #3966. Part of #3852.

Assigns new priorities and rebalances existing priorities on loadout actions.

### DIFF
--- a/src/NexusMods.DataModel/LoadoutManager.cs
+++ b/src/NexusMods.DataModel/LoadoutManager.cs
@@ -383,7 +383,6 @@ internal partial class LoadoutManager : ILoadoutManager
     public void RemoveItems(ITransaction tx, LoadoutItemGroupId[] groupIds)
     {
         var db = _connection.Db;
-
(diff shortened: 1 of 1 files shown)


src/NexusMods.Library/LibraryService.cs
Al12rs: Maybe in the future we will want to move Mod updates to the LoadoutManager, but not sure yet.

erri120: Forgot the library had this, updated the code so now everything goes back to the loadout manager.

Issue #3978, opened 2025-10-13
Cleanup loadout and synchronizer management


erri120:
Types:
- `ILoadoutManager` from #3966
- `ILoadoutSynchronizer`
- `ISynchronizerService`

The synchronizer and synchronizer service have an identity crisis with very vague responsibilities, inconsistent usages, and terrible ad-hoc extension points. The synchronizer especially has become a huge tech debt as it keeps growing in size.

We should revise the structure, responsibilities, and relationships of the various components in an effort to reduce tech debt, bring clarity, reduce bugs, improve testability, and provide better extension points and APIs for consumers.

Things to consider:
- We want to regulate concurrent access to game folder for synchronize/mange operations
- We want to allow games to override/add modify behaviours of certain synchronize/loadout operations
- We want to have oversight on transactions affecting a loadout to potentially add side effects (file conflicts sorting data, load order computation).

PR #3997, opened 2025-10-15
Add `IGameLocationsService`

erri120:
Part of #3978.

This new service will handle operations on game locations/installations like indexing the files, removing empty directories, writing, deleting, and moving files.

### DIFF
--- a/src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
+++ b/src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
@@ -1163,7 +1163,7 @@ public class ALoadoutSynchronizer : ILoadoutSynchronizer
         return loadout;
     }
 
-    public async Task<GameInstallMetadata.ReadOnly> RescanFiles(GameInstallation gameInstallation, bool ignoreModifiedDates)
+    public async Task<GameInstallMetadata.ReadOnly> RescanFiles(GameInstallation gameInstallation)
     {
         // Make sure the file hashes are up to date
         await _fileHashService.GetFileHashesDb();
--- a/src/NexusMods.Abstractions.Loadouts.Synchronizers/ILoadoutSynchronizer.cs
+++ b/src/NexusMods.Abstractions.Loadouts.Synchronizers/ILoadoutSynchronizer.cs
@@ -54,11 +54,7 @@ public interface ILoadoutSynchronizer
     /// whatever is on disk.
     /// </summary>
     /// <param name="gameInstallation">The game installation to rescan.</param>
-    /// <param name="ignoreModifiedDate">
-    /// If false, files that have unchanged modified date since the last scan will be skipped.
-    /// If true, all files will be rehashed.
-    /// </param>
-    Task<GameInstallMetadata.ReadOnly> RescanFiles(GameInstallation gameInstallation, bool ignoreModifiedDate = false);
+    Task<GameInstallMetadata.ReadOnly> RescanFiles(GameInstallation gameInstallation);
 
     /// <summary>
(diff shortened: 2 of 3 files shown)


src/NexusMods.Abstractions.Loadouts.Synchronizers/IGameLocationsService.cs
```diff
@@ -0,0 +1,68 @@
...
+
+[PublicAPI]
+public record IndexGameResult(
+    FrozenDictionary<GamePath, IndexFileResult> NewFiles,
+    FrozenDictionary<GamePath, IndexFileResult> ModifiedFiles,
+    FrozenSet<GamePath> RemovedFiles
+);
+
+[PublicAPI]
+public interface IGameLocationsService
```
Al12rs: We should add good documentation to these new interfaces as we go along, to ensure that the scope is well defined and they don't something they are not meant to be.

src/NexusMods.Abstractions.Loadouts.Synchronizers/IGameLocationsService.cs
```diff
@@ -0,0 +1,68 @@
...
+        FrozenDictionary<GamePath, DiskStateEntry.ReadOnly> previousDiskState,
+        ILocationsFilter filter,
+        CancellationToken cancellationToken = default
+    );
+
+    // Task RemoveEmptyDirectories(GameInstallation installation, CancellationToken cancellationToken = default);
+}
+
+[PublicAPI]
+public interface ILocationsFilter
```
Al12rs: This is more a `GamePathFilter` rather than a `LocationsFilter`

src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1165,7 +1167,7 @@ private bool ActionIngestFromDisk(Dictionary<GamePath, SyncNode> syncTree, Loado
     {
         // Make sure the file hashes are up to date
         await _fileHashService.GetFileHashesDb();
-        return await ReindexState(gameInstallation, ignoreModifiedDates, Connection);
+        return await ReindexState(gameInstallation);
```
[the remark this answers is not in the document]
erri120: If there was such a case then it should've been documented. As it stands right now, nothing was using that flag or the CLI command so I removed it.

src/NexusMods.Abstractions.Loadouts.Synchronizers/ALoadoutSynchronizer.cs
```diff
@@ -1464,196 +1466,110 @@ await Parallel.ForEachAsync(files, async (item, _) =>
...
-    {
-        var hashDb = await _fileHashService.GetFileHashesDb();
 
-        var gameInstallMetadata = GameInstallMetadata.Load(connection.Db, installation.GameMetadataId);
+        return GameInstallMetadata.Load(Connection.Db, installation.GameMetadataId);
+    }
 
-        var previousDiskStateEntities = gameInstallMetadata.DiskStateEntries;
-        var previousDiskState = new Dictionary<GamePath, DiskStateEntry.ReadOnly>(capacity: previousDiskStateEntities.Count);
+    private FrozenDictionary<GamePath, DiskStateEntry.ReadOnly> GetDiskState(GameInstallMetadata.ReadOnly gameInstallMetadata)
```
Al12rs: This code seems to be doing some of the same things as `GetDiskStateForGame`, but the ForGame variant has some high performance optimization and returns `PathPartPair`s.

We should probably encapsulate the fast iteration part and share it. Currently there are two ways to get the DiskState with different outputs but presumably one version is faster than the other.

PR #4009, opened 2025-10-18
Fix download progress reset conditionally

### LINKED ISSUE #3927: Rapid pause/unpause cycling causes downloads to restart from 0 bytes
## What happened

When rapidly cycling between pause and unpause states on individual downloads in the Download Menu, downloads can sometimes restart from 0 bytes instead of resuming from their previous position. 

This occurs specifically when the user quickly toggles pause/unpause multiple times in succession.

## Context

The problem is probably some sort of race condition that occurs when download jobs don't have sufficient time to properly transition into running state before they're paused again- or something of the kind.

## Steps to Reproduce

1. Start a download (any size)
2. Wait for download to progress (at least a second)
3. Rapidly click pause, then unpause, then pause again in quick succession (multiple cycles)
4. Observe that the download may restart from 0 bytes instead of resuming from previous position

## Expected Behaviour

Downloads should maintain their downloaded bytes count and resume from the previous position regardless of how quickly pause/unpause commands are issued.

## Actual Behaviour

During rapid pause/unpause cycling, downloads sometimes lose their progress and restart from 0 bytes, requiring the entire download to begin again.

axmanalad:
Fixes #3927 

This PR adds conditional checks to all instances that `_state.TotalBytesDownloaded` is modified. When a mod is unpaused, `bytesWritten` initially resumes at 0 bytes. Rapidly calling `StartAsyncImpl()` causes the Pause job's finally block to read the Resume job's initial bytes and save that result. The addition of these conditions would prevent overwriting the progress if no byte is written. The finally block also checks for the output stream to make sure the download progress stays up to date.

Currently, `JobMonitor.cs` does not atomically check the job status for `Resume()` and `Pause()`, which may cause these functions to run on the wrong job status. The PR also adds both semaphores and a new atomic state operation `TryTransitionStatus()` to ensure the correct state transition happens without race conditions to `Resume()` and `Pause()` (i.e. `Pause()` will not execute if the job status is "Pause" at the exact moment of calling the actual operation).

### DIFF
--- a/src/NexusMods.Networking.HttpDownloader/HttpDownloadJob.cs
+++ b/src/NexusMods.Networking.HttpDownloader/HttpDownloadJob.cs
@@ -139,8 +139,6 @@ public record HttpDownloadJob : IJobDefinitionWithStart<HttpDownloadJob, Absolut
             if (outputStream.Length != contentLength) outputStream.SetLength(contentLength);
         }
 
-        outputStream.Position = (long)_state.TotalBytesDownloaded.Value;
-
         await context.YieldAsync();
         using var request = PrepareRequest(out var isRangeRequest);
         using var response = await Client.SendAsync(


Al12rs:
Hi @axmanalad, I looked at the code and I think the issue might be with `StreamProgressWrapper` that isn't properly initialized for resume (`_currentBytesWritten`, `_lastBytesWritten` are 0 and not set to the current stream position in the constructor).

I'm not sure we otherwise need any of the CAS and locks. We want to avoid locks as much as possible in general.

Al12rs (changes requested):
See comment above

Al12rs (approved):
Nice, I tested with a 1.0 GB file, paused and resumed a bunch of times. Stayed consistent and in the end installed correctly.

Nice work!

PR #4075, opened 2025-10-29
Scoped settings improvements

Al12rs:
- Part of #1043

## Changes:
- Added a `TryGet` method to allow keyed settings to test for presence of persisted value without getting back a default.
This allows keyed settings to adopt custom default values if needed (different SortOrders have different default sorting), since otherwise defaults are defined at the type level.
- Made settings override scope aware. It is now possible to override specific scoped settings.

### DIFF
--- a/src/NexusMods.Backend/Settings/SettingsManager.cs
+++ b/src/NexusMods.Backend/Settings/SettingsManager.cs
@@ -136,13 +136,13 @@ internal class SettingsManager : ISettingsManager
             return true;
         }
         
-        var overrideVal = GetDefault<T>();
-        overrideVal = GetOverride(settingKey, overrideVal, out var didOverride);
+        var defaultValue = GetDefault<T>();
+        defaultValue = GetOverride(settingKey, defaultValue, out var didOverride);
 
         if (!didOverride) return false;
         
-        CoreSet(overrideVal, settingKey, notify: !didOverride);
-        value = overrideVal;
+        CoreSet(defaultValue, settingKey, notify: !didOverride);
+        value = defaultValue;
         return true;
     }
     


src/NexusMods.Backend/Settings/SettingsManager.cs
```diff
@@ -102,29 +103,59 @@ public SettingsManager(IServiceProvider serviceProvider)
...
+        if (savedValue is not null)
+        {
+            savedValue = GetOverride(settingKey, savedValue, out _);
+            CoreSet(savedValue, settingKey, notify: false);
+            value = savedValue;
+            return true;
+        }
+        
+        var overrideVal = GetDefault<T>();
+        overrideVal = GetOverride(settingKey, overrideVal, out var didOverride);
```
erri120:
```suggestion
        var defaultValue = GetDefault<T>();
        defaultValue = GetOverride(settingKey, defaultValue, out var didOverride);
```

src/NexusMods.Backend/Settings/SettingsManager.cs
```diff
@@ -102,29 +103,59 @@ public SettingsManager(IServiceProvider serviceProvider)
...
 
-            var overriden = overrideMethod.Invoke(value);
-            Debug.Assert(overriden.GetType() == typeof(T));
+        if (!didOverride) return false;
+        
+        CoreSet(overrideVal, settingKey, notify: !didOverride);
+        value = overrideVal;
```
erri120:
```suggestion
        CoreSet(defaultValue, settingKey, notify: !didOverride);
        value = defaultValue;
```

erri120 (approved):

PR #4094, opened 2025-11-02
Use Heroic's new protocol format to support passing arguments to the game

CommandMC:


Sewer56 (commented):
I'll just trust it was tested.

Sewer56 (approved):

CommandMC:
I've checked that the protocol URLs in general work, although I don't have Cyberpunk 2077 installed to test the exact use case that lead to this PR. We may want to wait until the workflows finish and have the OP test it on their system

PR #4096, opened 2025-11-03
Add download queue state to prevent new download jobs from starting

### LINKED ISSUE #3923: Pause All doesn't prevent new downloads from being queued
## What happened?

When hitting "Pause All", it doesn't prevent additional downloads from being queued. The system uses parallel downloading with `Math.Max(Environment.ProcessorCount, 16)` concurrent slots, but "Pause All" only pauses currently downloading jobs, not the entire queue system.

## Context

The issue occurs because "Pause All" only affects downloads that are actively downloading, not those in other phases like extracting or finishing up. When download jobs are in the process of transitioning past running, pause is ineffective in preventing the download from transitioning to `Complete`. Afterwards, new download(s) can still be queued to maintain the parallel slot count.

**Example scenario:**
- 16 downloads are active (parallel limit)
- 15 are actively downloading, 1 is extracting/finishing
- User hits "Pause All"
- The 15 downloading jobs pause, but the finishing/extracting job continues
- When the extracting job completes its current phase, a new download is queued to fill the slot
- User needs to press "Pause All" multiple times on fast connections

This creates a whack-a-mole situation where pausing downloads doesn't actually stop the queue from advancing.

**Original report**: https://github.com/Nexus-Mods/NexusMods.App/pull/3898#issuecomment-3347589970

## Steps to Reproduce

1. Start multiple downloads using a large collection with many mods
2. Click "Pause All" button
3. Observe that only actively downloading jobs pause
4. When download jobs are in the process of transitioning past running, new downloads are queued
5. Need to click "Pause All" multiple times in quick succession to actually pause all downloads

## Expected Behavior

Clicking "Pause All" should immediately pause the entire download queue system, preventing any new downloads from being queued until explicitly resumed, regardless of the current state of existing jobs.

## Actual Behavior

"Pause All" only pauses actively downloading jobs. Downloads in other states (extracting, finishing) continue, and when they complete, new downloads are automatically queued to maintain the parallel slot count, requiring multiple button presses.

axmanalad:
Fixes #3923 

This PR introduces a global pause/resume conditional mechanism for the download job queue via `JobMonitor`, allowing the system to prevent new download jobs from starting while the queue is paused. 

- The `volatile bool _isDownloadQueuePaused` field controls and queries the paused state in `JobMonitor` when using `PauseDownloadQueue()`, `ResumeDownloadQueue()`, and `IsDownloadQueuePaused()`.

- In `DownloadsService`, the methods `PauseAll`, `PauseAllForGame`, `ResumeAll`, and `ResumeAllForGame` now pause or resume the download queue globally.

- `DownloadCollectionJob` and `HttpDownloadJob` now check the global paused state before starting new downloads.
  
  - `HttpDownloadJob` pauses the job context of new/future downloads if `IsDownloadQueuePaused` is active.
  
  - `DownloadCollectionJob` waits until `IsDownloadQueuePaused` is false, resuming to add the current download awaiting as a new download job.
  
  **NOTE:** Need of testing with premium; I have not tested `DownloadCollectionJob` since I am not a premium user (i.e. `DownloadItems` does not work for a free user).

A few things I wanted to consider for the solution are if `ResumeDownloadQueue` should also get called if a user resumes one download, and if  job creation should be atomically handled (e.g. a new download job starts immediately after `PauseAll` but before the flag is checked, leading to briefly start before pausing)?

### DIFF
--- a/src/NexusMods.Collections/DownloadCollectionJob.cs
+++ b/src/NexusMods.Collections/DownloadCollectionJob.cs
@@ -27,7 +27,11 @@ public class DownloadCollectionJob : IJobDefinitionWithStart<DownloadCollectionJ
             var download = downloads[index];
             if (!CollectionDownloader.DownloadMatchesItemType(download, ItemType)) return;
             if (CollectionDownloader.GetStatus(download, Db).IsDownloaded()) return;
-
+            
+            // Waits until the download queue is unpaused
+            while (context.Monitor.IsDownloadQueuePaused())
+                await Task.Delay(500, token); // Checks every 500ms
+            
             try
             {
                 if (download.TryGetAsCollectionDownloadNexusMods(out var nexusModsDownload))


axmanalad:
Hi @Al12rs, thanks again for reviewing my PR,

While I am currently brainstorming a solution with your approach, I wanted to share my thoughts about your design questions (maybe even initiate suggestions):
> * Do we pause or cancel the `DownloadCollectionJob`?

I believe pausing `DownloadCollectionJob` is the better approach than cancelling. Cancelling the job would require users to start their downloads again from scratch, whereas pausing allows to resume normally without losing much progress (except for finished downloads, which would remain).

> * If we pause it, we need to add ability for users to resume it from the Collection Download page (do we add new buttons or do we just show the start download button?)

We could reuse the Start button (maybe even change the text of the UI "Resume Download" when it is an existing job) and find the matching job of the collection download to resume all downloads. If there is none, we would create a new one. Adding extra buttons similar to the Downloads page might be redundant but a quality of life feature. 
This question relates to something that my current solution does not consider (see the last paragraph).

> * There is some weirdness when dealing with nested jobs. If we pause the parent job, do we pause or cancel the ongoing children? If we pause them how do we resume them when we resume the parent job.

If we pause the parent job, I believe it should also pause the ongoing children. Resuming the parent should signal children to resume as well, which requires some form of IPC implementation.
I noticed that there are a few unused job monitor methods, mainly those that use `JobGroup` (such as `ResumeGroup` and `PauseGroup`). However, I have not figured out how the groups are organized, nor do I know how will it work as of right now. I believe these would be useful for any group-related tasks like Collections.

As a **note**, I have no idea how downloading Collections work for a premium user. Most of the current solution heavily relates to downloading mods individually since I only tested this as a free user. Could you please provide a short video demonstrating how downloading, pausing, and resuming a collection works as a premium user. Sharing it on Discord or in the PR would be very helpful for me to understand the expected functionality. Thank you! 👍

